PR #66, opened 2018-05-29
Fix address defaultValue/value, hide testing accounts

jacogr:


packages/ui-app/src/InputAddress/index.js
```diff
@@ -44,36 +51,112 @@ const transform = (value: string): Uint8Array => {
...
+    }
+  }
+
+  componentDidMount () {
+    const { api } = this.props;
+
+    this.setState({
+      subscriptions:
+        [
+          () => api.system.chain().subscribe((apiChain?: string) => {
```
jacogr: Should be set globally, not filtered here.

PR #82, opened 2018-06-09
fix: Fixes #81 by converting returned color to string and passes failing test

### LINKED ISSUE #81: Unable to Create account due to colors package error
When running the latest version of polkadot-js, if I click "Create account" on the Accounts page it returns the following error in Chrome Inspector Console `Uncaught TypeError: all.splice(...)[0].alpha(...).string is not a function`

ltfschoen:


packages/ui-identicon/src/colors.js
```diff
@@ -23,8 +23,6 @@ module.exports = function colors (seeder: Seeder): ColorGen {
   return (alpha: number = 0.9): string => {
     const index = Math.floor(all.length * seeder());
 
-    return all.splice(index, 1)[0]
-      .alpha(alpha)
-      .string();
+    return String(all.splice(index, 1)[0].alpha(alpha));
```
jacogr: Looking at the color docs, I believe the correct approach should be something like -

```
// string hsl, i.e. will be something like 'hsl(320, 50%, 100%)'
return all.splice(index, 1)[0].alpha(alpha).hsl().string();
```

https://www.npmjs.com/package/color

ltfschoen:
I've summarised what I did this morning below. I haven't yet had time to look at whether it was caused by the index out of bound as you pointed out 

When I go to http://localhost:3000/#/accounts, and click the Accounts tab, and then click "Create account", it loads packages/app-accounts/src/Creator.js, which is dependent on packages/app-accounts/src/Address.js, which is in turn dependent on @polkadot/ui-react/IdentityIcon, which calls packages/ui-react/src/IdentityIcon/appendIcon.js, which calls the `identicon` function of @polkadot/ui-identicon in packages/ui-identicon/src/index.js with a seed argument being the Polkadot Address shown in the UI. It then calls colors.js with a seeder argument.

I tried replicating the functionality in Node.js v9.8.0 REPL by running `node`:

```
const Color = require('color');

const COLORS = [
  '#01888C', // teal
  '#FC7500', // bright orange
  '#034F5D', // dark teal
  '#F73F01', // orangered
  '#FC1960', // magenta
  '#C7144C', // raspberry
  '#F3C100', // goldenrod
  '#1598F2', // lightning blue
  '#2465E1', // sail blue
  '#F19E02' // gold
];

const SHAPE_COUNT = 5;
const WOBBLE = 30;
const seeder = 0.3528900146484375;
const amount = (seeder * WOBBLE) - (WOBBLE / 2);
const all = COLORS.map((hex) => Color(hex).rotate(amount));
const alpha = 0.9;
const index = Math.floor(all.length * seeder);
let color = all.splice(index, 1)[0].alpha(alpha);

color.string() 
color.string(3) 
color.toString() 
String(color)
```

The last four commands return the following now.

'hsla(10.699999999999989, 99.2%, 48.6%, 0.9)'
'hsla(10.709000000000003, 99.194%, 48.627%, 0.9)'
'hsla(10.699999999999989, 99.2%, 48.6%, 0.9)'
'hsla(10.699999999999989, 99.2%, 48.6%, 0.9)'

Earlier this morning I was still getting errors when I tried calling `.string()` on a Color, It started working at around the time that I uninstalled https://www.npmjs.com/package/color-string and reinstalled https://www.npmjs.com/package/color, so perhaps there was a dependency conflict between those packages, since the color package is dependent on color-string package, see https://github.com/Qix-/color/blob/master/index.js#L128

jacogr:
I tried to re-produce the issue here with a test-case, sadly couldn't get it to break https://github.com/polkadot-js/apps/pull/87/commits/111b0c4f9ea676ecaba405fbc9a12f726cbc74b9

PR #83, opened 2018-06-09
feat: Fixes #75. Add edit icon from semantic-ui to account name input field

### LINKED ISSUE #75: Account edit currently is a bit confusing...
This is just something small... however it is not very obvious that if you change the name the save becomes active. Most will expect to "click before doing". The functionality without extra clicks is not bad, however it is completely non-obvious.

ltfschoen:
Adding an edit icon that is embedded in the account name input field makes it more obvious to the user that the input field isn't read-only, and that there is a way to "click before doing" (i.e. editing the input field that is identified as being editable before the ability to save changes is granted). Some users may click the edit icon itself thinking it toggles the ability to edit the input field, whilst others may just recognise that the edit icon means the input field is editable. In either case just having the edit icon should suffice as being the initial "call to action" for a user wishing to edit the input field, and since the Reset and Save buttons illuminate when the account name is modified, they'll recognise that a further step is required to save their changes.
Screenshot of text field now with edit icon is shown below:
screen shot 2018-06-09 at 7 27 53 pm

### DIFF
--- a/packages/app-accounts/src/Editor.js
+++ b/packages/app-accounts/src/Editor.js
@@ -135,6 +135,7 @@ class Editor extends React.PureComponent<Props, State> {
               })}
               onChange={this.onChangeName}
               value={editedName}
+              isEditable={true}
             />
           </div>
         </div>
--- a/packages/app-accounts/src/Editor.js
+++ b/packages/app-accounts/src/Editor.js
@@ -130,12 +130,12 @@ class Editor extends React.PureComponent<Props, State> {
           <div className='ui--row'>
             <Input
               className='full'
+              isEditable
               label={t('editor.name', {
                 defaultValue: 'identified by the name'
               })}
               onChange={this.onChangeName}
               value={editedName}
-              isEditable
             />
           </div>
         </div>
--- a/packages/app-accounts/src/Editor.js
+++ b/packages/app-accounts/src/Editor.js
@@ -135,7 +135,7 @@ class Editor extends React.PureComponent<Props, State> {
               })}
               onChange={this.onChangeName}
               value={editedName}
-              isEditable={true}
+              isEditable
             />
           </div>
         </div>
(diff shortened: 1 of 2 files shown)


packages/ui-app/src/Input.js
```diff
@@ -78,6 +79,7 @@ export default class Input extends React.PureComponent<Props, State> {
           placeholder={placeholder}
           type={type}
           value={value}
+          className={isEditable ? 'ui icon input edit' : ''}
```
jacogr: Could we move this up? (Alphabetical?)

packages/ui-app/src/Input.js
```diff
@@ -86,6 +88,7 @@ export default class Input extends React.PureComponent<Props, State> {
                 : 'off'
             }
           />
+          {isEditable ? <i className='edit icon' /> : null}
```
jacogr: We are specifying the icon in 2 places, i.e. classname and then as an actual icon. Shouldn't 1 place be sufficient? (And yes, have seen some weird things in this regard before, just feels like we are missing something and it should only be needed once, possibly as this version since we are manually including the `input`)

packages/app-accounts/src/Editor.js
```diff
@@ -135,6 +135,7 @@ class Editor extends React.PureComponent<Props, State> {
               })}
               onChange={this.onChangeName}
               value={editedName}
+              isEditable
```
jacogr: Move it up, alphabetical. (Just makes it a bit easier to spot if you know that things are always sorted)

packages/ui-app/src/Input.js
ltfschoen: Previously I'd tried loads of different combinations trying to get the edit icon to be contained within the input field. Unfortunately the way I've done it is the only way I could get it to work. I just tried removing some of the duplicate classes as you suggested and was impressed to find that it still works.
Another approach that I tried previously was creating an `EditIcon` component and passing it down to the `icon` prop of the `Input` component, but I couldn't get that to work.

jacogr: Yes, I've seen funnies with icons + inputs before as well. (I think there is something basic I'm missing.)

PR #84, opened 2018-06-09
config: Enable source maps in development for front-end debugging

ltfschoen:


ltfschoen:
Note that this PR is not necessary, as source map debugging may be performed in development if the server is run with the production environment variable enabled:
```
NODE_ENV=production yarn run start
```

jacogr:
I think it is a good idea to have it enabled, just making the dev experience better - probably need to go with something "cheaper" for dev though, since `source-map` takes quite a time to build. https://webpack.js.org/configuration/devtool/

`cheap-eval-source-map` may do the trick for dev mode.

packages/apps/webpack.config.js
```diff
@@ -23,7 +23,7 @@ function createWebpack ({ alias = {}, context, name = 'index' }) {
 
   return {
     context,
-    devtool: isProd ? 'source-map' : false,
+    devtool: 'source-map',
```
jacogr: Use `cheap-eval-source-map` for dev-mode (faster compiles), `source-map` for prod.

Alternatively (have not looked into pros/cons it is probably better to just use what react-create-app uses, https://github.com/facebook/create-react-app/blob/next/packages/react-scripts/config/webpack.config.dev.js#L80

(That has been battle-tested)

packages/ui-identicon/webpack.config.js
```diff
@@ -22,5 +22,6 @@ module.exports = {
       }
     ]
   },
+  devtool: 'source-map',
```
jacogr: As above.

packages/ui-react-rx/webpack.config.js
```diff
@@ -25,5 +25,6 @@ module.exports = {
       }
     ]
   },
+  devtool: 'source-map',
```
jacogr: As above.

packages/ui-react/webpack.config.js
```diff
@@ -43,5 +43,6 @@ module.exports = {
       }
     ]
   },
+  devtool: 'source-map',
```
jacogr: As above.

packages/apps/webpack.config.js
```diff
@@ -23,9 +23,9 @@ function createWebpack ({ alias = {}, context, name = 'index' }) {
 
   return {
     context,
-    devtool: isProd ? 'source-map' : false,
+    devtool: isProd ? 'source-map' : 'cheap-eval-source-map',
     entry: `./src/${name}.js`,
-    mode: isProd ? 'production' : 'development',
+    mode: ENV,
```
jacogr: Nice.

PR #85, opened 2018-06-09
fix: Fixes #70 by Rename Submit Extrinsic to Submit Transaction

### LINKED ISSUE #70: Rename "submit extrinsic" to "submit transaction"

ltfschoen:


jacogr:
I think it makes sense to have the modal pop-up also updated, i.e. `ui-signer/src/Extrinsic.js` - at least in terms of the title. (i.e. user clicks "submit tx" and it shows "submit tx")

ltfschoen:
Should we also change any of the following?

* Extrinsics tab view
  * change `from extrinsic section` to `from transaction section`
  * change `submit the following extrinsic` to `submit the following transaction`
* Storage tab view (with `system` value selected for the `query storage area` label, and when selecting a value from selection box of the `with storage key` label)
  * change `The data associated with an extrinsic` to `The data associated with a transaction` when `extrinsicData(extrinsic)` is selected
  * change `The extrinsics index` to `The transactions index` when `extrinsicIndex()` is selected
  * change `The extrinsicsRoot for the current block` to `The transactions root for the current block` when `extrinsicsRoot()` is selected
* Raw RPC tab view
  * change `Submit a fully formatted extrinsic for block inclusion` to `Submit a fully formatted transaction for block inclusion` when `submitExtrinsic(extrinsic)` is selected for the `with method name` label and when `author` value is selected from the `rpc area` selection box
  * change the label `extrinsic: Bytes` to `transaction: Bytes` when the `author` value is selected from the `rpc area` selection box

jacogr:
Very good question - the naming is a bit of a conumdrum. We specifically went with "extrinsic" in Polkadot since it is closer to the truth than "transaction". at the same time it is what users are used to. So the general rule here would be that what the user submits is a transaction (i.e. the blob of data he/she signs), I'm copying the original comment from Gavin here -

"re: terminology, i'd like to see "extrinsic" used as a general term for pieces of data that get referenced by a header and bundled with a block, just as we do right now pretty much across the board. but i'd like to use "transaction" specifically for those extrinsics that are signed by a user. so you can say "submit a transaction" which is the same thing as saying "submit an extrinsic that is signed".

this doesn't make much difference in the codebase

but does mean the UI and general surrounding terminology can continue to use "transaction" for the most part

(for those that don't know, the extrinsics that are not transactions (i.e. not signed) are things like the timestamp and parachain header hashes, which are not generally verified (that's the job of fishermen) - their mere inclusion in a validated/finalised block means that >2/3 validators have testified to their correctness)
"

Things like the "extrinsic: Bytes" is actually pulled in from the types repo, and `extrinsicsRoot` is what is actually contained in a substrate block, so those need to stay. So for the fields/etc, keep to "extrinsic", where we have specific `t('...', {...})` labels, it would make sense to change.

So the "Extrinsics" can be renamed to "Transactions", however things like rawRPC needs to stay to match up with the actual exposed APIs.

ltfschoen:
In that case the proposed changes would include:

* **Extrinsics tab view**
  * Changing `submit the following extrinsic` to show `submit the following transaction (signed extrinsic)` in app-extrinsics/src/Selection.js, unless the user chooses either the `timestamp` or `parachains` value from the options in the previous `from extrinsic section` selection box.

**If the user chooses any of `consensus, council, councilVoting, democracy, parachains, session, or staking` from the previous `from extrinsic section` selection box, would any of those constitute a transaction (signed extrinsic)?**

* **Storage tab view**
  * No change

* **Raw RPC tab view**
  * No change

jacogr:
Timestamp & parachains _should_ not be visible atm - since they each only include internal methods. (There may be "user submitted" in the future)

For least confusion, it probably makes sense to change the type labels to "transaction" as well. Should just give it a go and see how it "feels like" from a UX perspective when actually using.

PR #90, opened 2018-06-11
fix: Fixes #88 buttonbar style cornering on Accounts and Addresses pages

### LINKED ISSUE #88: Format buttonbar + spacer with round corners
This setup is just "off"

polkadot apps portal 2018-06-11 17-24-10

The "Forget" should have rounding right, the "Reset" rounding left. Pure CSS in the case where spacers are used in Button Groups.

ltfschoen:
- Add right corners to Forget
- Add left corners to Reset

screen shot 2018-06-11 at 10 29 01 pm


jacogr:
Nice! I would suggest putting the overrides in here, so it applies everywhere (instead of just the 2 apps) - 

https://github.com/polkadot-js/apps/blob/master/packages/ui-app/src/styles/semantic.css

Apart from that, looks absolutely perfect.

ltfschoen:
I'm hesitant to remove `.accounts--App .accounts--Editor` and `.addresses--App .addresses--Editor` incase other different buttons groups (perhaps even button groups where we want them to have square corners instead) with CSS styles `.ui--Button-Group .ui.buttons` are introduced later on other pages, as it would cause the 1st button to have right corners and 3rd button to have left corners

I was also hoping to leverage PostCSS variables and nesting plugins. I read the following blogpost https://ashleynolan.co.uk/blog/postcss-a-review.
I tried installing those dependencies plus a few of my Sass favourites with `yarn add precss autoprefixer postcss-simple-vars postcss-nested postcss-import postcss-mixins --dev`.
I then tried adding them to packages/apps/webpack.config.js
```
{
  loader: require.resolve('postcss-loader'),
  options: {
    ident: 'postcss',
    plugins: () => [
      require('postcss-clean')(),
      require('postcss-flexbugs-fixes'),
      require('precss'),
      require('autoprefixer'),
      require('postcss-simple-vars'),
      require('postcss-nested'),
      require('postcss-import'),
      require('postcss-mixins')
    ]
  }
}
```
I then added the line of code to try using PostCSS variables `$small-corner: .28571429rem;`.
I then restarted the server, but it gave me error `clean: Invalid selector '$small-corner: .28571429rem;` 

Note that I initially I tried adding the plugins directly into the dependency @polkadot/dev-react/config/postcss.js, since it's loaded by postcss.config.js to see if that would work (with the intention of creating a PR to that library if it worked), but it didn't work. I then also tried just adding the plugins directly into postcss.config.js, but that didn't work either.

I was hoping to change the code to use PostCSS variables and nesting like so:
```
$small-corner: .28571429rem;

.accounts--Editor .addresses--Editor {
  .ui--Button-Group {
    .ui.buttons .button:nth-child(1) {
      border-top-right-radius: $small-corner;
      border-bottom-right-radius: $small-corner;
    }
    .ui.buttons .button:nth-child(3) {
      border-top-left-radius: $small-corner;
      border-bottom-left-radius: $small-corner;
    }
  }
}
```

jacogr:
Yes, PostCss is great, well worth pulling-in. (The complexity has not being that high thus far, but just the ability to nest is worth it's weight in gold - have used it elsewhere extensively and it just cleans up things so much.) 

I understand applying the current solution everywhere has issues, but it is the right thing in this case. It just means it needs to be a bit more generic. So we need to detect the spacer and act accordingly, as opposed to just taking the order and applying the rounding there. (Which could have some issues if new buttons are added to the group)

So logic (obviously not that simple in CSS is something like) - if button is followed by spacer, round the right, if a button is preceded by a spacer, round the left.

ltfschoen:
Oh, I just realised that the plugins are loaded sequentially, and the error I encountered was probably being caused because postcss-clean which minifies the CSS was being it had loaded postcss-variables so it knew how to interpret. I changed packages/apps/webpack.config.js and it works now

I've used `.ui.button.compact.mini.basic + .button:nth-child(3)` to conditionally apply the corners to the 3rd button only if the preceding sibling elements class matches that of the spacer.

Although I haven't figured out how to conditionally apply the corners to the 1st button only if next sibling element class matches that of the spacer.

ltfschoen:
I've pushed another commit with a solution that works.

**Proposed Solution**: The work-around that I came up with uses the existing generic ui-app/src/Button/Divider.js component, but I've passed custom styling with no padding as props to it so it has no width. The proposal is to add this "widthless" divider to the start of each usage of the "Forget, Reset, Save" button group, such that the custom cornering CSS overrides only apply to the "Forget, Reset, Save" button groups, which contain 1st "widthless" divider element and a 3rd "normal" divider element.

**Challenges & Alternatives:** 

I did some research and found that there is no CSS selector combinator that allows conditional styling of an element conditional upon "subsequent" sibling elements existing (i.e. only apply cornering to the first button iff the next sibling element is a divider). The only options for conditional styling using pure CSS that I found that are compatible across browsers according to caniuse.com work when there's a match with any of the "preceding" sibling elements or just the immediate "preceding" sibling element, they include the 'Adjacent sibling combinator' by use of `+` and the 'General sibling combinator' by use of `~`. See this link for details under the heading "Combinators" https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Selectors.

Another alternative solution that was considered by not deemed generic enough was to conditionally apply the custom cornering only to button groups that are contained within a classname that contains the text "--Editor", on the assumption that the Forget, Reset, Save button group would only be used within a nesting containing such a wildcard value (as is currently the case)
```
div [class*="--Editor"] {
  .ui--Button-Group {
    .ui.buttons {
      .button:nth-child(1) {
        border-top-right-radius: $small-corner;
        border-bottom-right-radius: $small-corner;
      }
      .ui.button.compact.mini.basic + .button:nth-child(3) {
        border-top-left-radius: $small-corner;
        border-bottom-left-radius: $small-corner;
      }
    }
  }
}
```

packages/app-accounts/src/Editor.js
```diff
@@ -67,6 +67,7 @@ class Editor extends React.PureComponent<Props, State> {
 
     return (
       <Button.Group>
+        <Button.Group.Divider style={{ padding: '0em' }} />
```
jacogr: Is it possible that we can move this to the base component- feels a bit wrong having to do this everywhere?

packages/app-addresses/src/Editor.js
```diff
@@ -67,6 +67,7 @@ class Editor extends React.PureComponent<Props, State> {
 
     return (
       <Button.Group>
+        <Button.Group.Divider style={{ padding: '0em' }} />
```
jacogr: As above.

packages/ui-app/src/styles/semantic.css
```diff
@@ -77,3 +79,17 @@
   left: -7px;
   opacity: 1;
 }
+
+.ui--Button-Group {
```
jacogr: Since this is nested under `.ui--` (i.e. our namespace), probably makes more sense (as-is) in `components.css`?

PR #78, opened 2018-06-09
config: Save dependencies in Node.js configuration

ltfschoen:
Currently when a user clones the repository and runs `yarn run start` it returns a load configuration error or module not found error for each of the required dependencies if they do not already have them installed globally, such that the user is required to install them all manually (i.e. `yarn add ...` or `npm install ...`. This change allows new users to easily install all relevant dependencies by just running `yarn install` before they run `yarn run start`

PR #99, opened 2018-06-18
fix: Fixes #98 by running existing Node.js CLI script

### LINKED ISSUE #98: polkadot-dev-yarn-only: command not found
I pulled the latest changes from the master branch and then switched to the latest Node.js version using NVM with `nvm use v10.4.0`. I then ran `yarn install`, but it gave the following error `sh: polkadot-dev-yarn-only: command not found`:

```
$ yarn
yarn install v0.24.6
(node:12387) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
[1/4] 🔍  Resolving packages...
[2/4] 🚚  Fetching packages...
[3/4] 🔗  Linking dependencies...
warning "react-hot-loader@4.2.0" has unmet peer dependency "react@^15.0.0 || ^16.0.0".
[4/4] 📃  Building fresh packages...
$ polkadot-dev-yarn-only
sh: polkadot-dev-yarn-only: command not found
error Command failed with exit code 127.
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
``` 

It's trying to run the line `"postinstall": "polkadot-dev-yarn-only",` in package.json.
I understand that the intention is that after installing the dependencies into the node_modules/ directory, it'd run indirectly via a shell script the Node.js CLI script that would then be available in ./node_modules/@polkadot/dev/scripts/polkadot-dev-yarn-only.js. When using `"postinstall": "polkadot-dev-yarn-only",` it tries to calling a binary executable file in node_modules/.bin named `polkadot-dev-yarn-only` but it doesn't exist and the shell script /node_modules/@polkadot/dev/scripts/polkadot-dev-yarn-only doesn't exist either.

ltfschoen:
With the existing approach it is expecting to call a binary file in node_modules/.bin/polkadot-dev-yarn-only that is associated with a shell script would be in node_modules/dev/scripts/polkadot-dev-yarn-only.sh. This PR simply calls the Node.js CLI script directly

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "demo:rx": "webpack-serve --config packages/ui-react-rx/webpack.config.js --content packages/ui-react-rx --port 3000",
     "demo:ui": "webpack-serve --config packages/ui-react/webpack.config.js --content packages/ui-react --port 3000",
     "deploy:ghpages": "gh-pages -d packages/apps/build",
-    "postinstall": "polkadot-dev-yarn-only",
+    "postinstall": "./node_modules/@polkadot/dev/scripts/polkadot-dev-yarn-only.js",
     "vanitygen": "polkadot-dev-build-babel && node packages/app-vanitygen/build/generator/cli.js",
     "start": "cd packages/apps && webpack-serve --config webpack.config.js --port 3000",
     "test": "jest --coverage"


ltfschoen:
This error was occurring because I was using an old version of Yarn. After I followed the recommendations to update to the latest version of Yarn that were suggested [here](https://github.com/polkadot-js/apps/pull/78), it resolved the issue.

PR #100, opened 2018-06-18
Update Readme with Yarn version requirement

ltfschoen:
To avoid issues with installation it is necessary to install a compatible version of Yarn.
It was highlighted [here](https://github.com/polkadot-js/apps/pull/78) that Yarn 1.0 included workspaces, and we are successfully using Yarn 1.3.2 for Travis CI, so propose to request users to ensure they are using a version later than that to avoid and problems.
For instance, I was using Yarn 0.24.6 and it [wouldn't install the sub-package dependencies](https://github.com/polkadot-js/apps/pull/78) nor would it [run a necessary binary](https://github.com/polkadot-js/apps/issues/98), but after upgrading to Yarn 1.7.0 those issues were resolved.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,10 +45,11 @@ To start off, this repo (along with others in the [@polkadot](https://github.com
 To get started -
 
 1. Clone the repo locally, via `git clone https://github.com/polkadot-js/apps <optional local path>`
-2. Ensure that you have a recent version of `node.js`, for development purposes [Node 10](https://nodejs.org/en/) is recommended.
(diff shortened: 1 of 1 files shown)


jacogr:
Thanks. All repos now use (this script)[https://github.com/polkadot-js/dev/blob/master/packages/dev/scripts/polkadot-dev-yarn-only.js] now as part of install as well, so it will warn you at least when using npm and (engines will do a >= 1.3.2)[https://github.com/polkadot-js/apps/blob/master/package.json#L6] check as well

PR #107, opened 2018-06-27
Add nvmrc

chevdor:
This allows lazy devs using `nvm` to read less doc :)


jacogr:
We are all a bit lazy - well, not lazy, but when the tools exist and you can just get on with it, why not ;)

PR #109, opened 2018-06-28
Add missing locale resources

chevdor:
Add missing files


jacogr:
It is actually a harmless and expected error, as horrible as it looks...

The issue here however is that, even with these added, you may still get the errors. i.e. It will actually try to locate the locale based on the browser language settings. (So it your case it may look for de first, before falling back to the default.)

Having said that, at least the English-browser users won't see errors anymore with this.

chevdor:
Ok I will add the missing ones, no problem.
I am testing using `en` for now.

PR #133, opened 2018-07-10
fix: Fixes #130 latest rust reports chains differently

### LINKED ISSUE #130: Latest Rust reports chain differently
https://github.com/polkadot-js/apps/blob/master/packages/apps/src/NodeInfo.tsx#L20

This needs to change now to check for -

- `dev`
- `Development` (as reported by latest master)
- `local` (previously missing, but also a `dev` chain)
- `Local`

As it stands master does not make the dev accounts available due to this check.

ltfschoen:
### Learnings
When the developer starts a Polkadot Node backend, the value of the CLI chain specification option is processed in https://github.com/paritytech/polkadot/blob/master/polkadot/cli/src/chain_spec.rs, which accepts any of the following values `"dev"`, `"local"`, `"poc-1"`, `"poc-2"`. Note however that the `poc-2` value is not shown as an available option when the user views the help menu after building the "master" branch with `polkadot --help`. 
When the developer runs the Polkadot Apps frontend it queries the Rust backend in the file ui-react-rx/src/Chain.tsx#L9, which contains the method used for the JSON RPC call `rpcs.system.public.chain` that is used to retrieve the chain specification from the backend Polkadot Node. It returns a value that is pre-formatted ready for presentation in frontend UI in the file https://github.com/paritytech/polkadot/blob/master/polkadot/service/src/chain_spec.rs (i.e. `"Development", "Local Testnet", "PoC-1 Testnet", "PoC-2 Testnet")
For instance, when a Polkadot Node is run using the `dev` Chain with `cargo run -- --chain=dev --validator --key Alice -d /tmp/alice`, it sets the `chain` variable value to "Development" in Polkadot Apps file ui-react-rx/src/Api/index.tsx#L28. 

Note the Development chain spec only requires one local Polkadot Node to be running to show the test accounts (i.e. Alice, Bob, Ferdie, etc) in the drop-down of the Extrinsics tab of Polkadot Apps, whereas the Local chain spec requires two.

### DIFF
--- a/packages/apps/src/NodeInfo.tsx
+++ b/packages/apps/src/NodeInfo.tsx
@@ -19,7 +19,7 @@ type Props = I18nProps & {};
 export function isTestChain(chain?: string): boolean {
   if (typeof chain === 'undefined') return false;
   const re = new RegExp("(dev|loc)", "i");
-  const match = re.test(chain.toLowerCase());
+  const match = re.test(chain.toString().toLowerCase());
   return match ? true : false;
 }
 
--- a/packages/apps/src/NodeInfo.tsx
+++ b/packages/apps/src/NodeInfo.tsx
@@ -16,11 +16,16 @@ import translate from './translate';
 
 type Props = I18nProps & {};
 
-function updateTestInfo (chain?: string) {
-  keyring.setTestMode(chain === 'dev');
+export function updateTestInfo (chain?: string): Boolean {
+  if (typeof chain === 'undefined') return false;
+  const re = new RegExp("(dev|loc)", "i");
+  const match = re.test(chain.toLowerCase());
+  keyring.setTestMode(match);
+  return match ? true : false;
 }
 
 function NodeInfo ({ className, style, t }: Props) {
+  console.log('t', t)
   return (
     <div className={classes('apps--NodeInfo', className)}>
       <Chain
--- a/packages/apps/src/NodeInfo.tsx
+++ b/packages/apps/src/NodeInfo.tsx
@@ -17,12 +17,13 @@ import translate from './translate';
 
 type Props = I18nProps & {};
 
-export function isTestChain(chain?: string): boolean {
+const re = new RegExp('(dev|loc)', 'i');
+
+export function isTestChain (chain?: string): boolean {
   if (isUndefined(chain)) {
     return false;
   }
 
-  const re = new RegExp("(dev|loc)", "i");
   const match = re.test(chain.toString().toLowerCase());
 
   return match ? true : false;
--- a/packages/apps/src/NodeInfo.tsx
+++ b/packages/apps/src/NodeInfo.tsx
@@ -28,7 +28,6 @@ function updateTestInfo (chain?: string) {
 }
 
 function NodeInfo ({ className, style, t }: Props) {
-  console.log('t', t)
   return (
     <div className={classes('apps--NodeInfo', className)}>
       <Chain
--- a/packages/apps/src/NodeInfo.tsx
+++ b/packages/apps/src/NodeInfo.tsx
@@ -6,7 +6,7 @@ import { I18nProps } from '@polkadot/ui-app/types';
 
 import React from 'react';
 
-import isUndefined from '@polkadot/util/is/undefined';
+import isTestChain from '@polkadot/ui-react-rx/util/isTestChain';
 import classes from '@polkadot/ui-app/util/classes';
 import keyring from '@polkadot/ui-keyring/index';
 import Chain from '@polkadot/ui-react-rx/Chain';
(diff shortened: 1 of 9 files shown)


packages/apps/src/NodeInfo.tsx
```diff
@@ -16,8 +16,15 @@ import translate from './translate';
 
 type Props = I18nProps & {};
 
+export function isTestChain(chain?: string): boolean {
+  if (typeof chain === 'undefined') return false;
+  const re = new RegExp("(dev|loc)", "i");
+  const match = re.test(chain.toLowerCase());
+  return match ? true : false;
```
jacogr: Space before `const` as well as after - can be grouped, but just leave some visual indication that the declaration block is different.

packages/apps/src/NodeInfo.tsx
```diff
@@ -16,8 +16,15 @@ import translate from './translate';
 
 type Props = I18nProps & {};
 
+export function isTestChain(chain?: string): boolean {
```
jacogr: Generally prefer that re-usable functions gets split - put it in a separate file. (We probably need to think about moving this into common/api at some point, it is a general utility after all - especially now that is is cleaner)

packages/ui-react-rx/src/Api/index.spec.js
```diff
@@ -0,0 +1,25 @@
...
+
+  it('enables Latest encoding when chain specification is NOT PoC-1 or undefined', () => {
+    expect(shouldUseLatestChain('poc-2')).toEqual(true);
+    expect(shouldUseLatestChain('PoC-2 Testnet')).toEqual(true);
+    expect(shouldUseLatestChain('Local')).toEqual(true);
+    expect(shouldUseLatestChain('Development')).toEqual(true);
+    expect(shouldUseLatestChain(' development')).toEqual(true);
+    expect(shouldUseLatestChain('dev')).toEqual(true);
+    expect(shouldUseLatestChain('Local')).toEqual(true);
+    expect(shouldUseLatestChain(' local ')).toEqual(true);
```
jacogr: Same comment as with the previous tests, i.e. location.

packages/ui-react-rx/src/Api/index.tsx
```diff
@@ -25,10 +25,16 @@ type State = ApiProps & {
   subscriptions: Array<any> // rxjs$ISubscription | null>;
 };
 
+export function shouldUseLatestChain(chain?: string): boolean {
+  if (typeof chain === 'undefined') return false;
+  const re = new RegExp("(poc-1)", "i");
+  const match = re.test(chain.toLowerCase());
+  if (match) return false;
+  return true;
+}
```
jacogr: As per above, give this function some breathing room, it is very condensed for no real benefit.

packages/ui-react-rx/src/Api/index.tsx
```diff
@@ -25,10 +25,16 @@ type State = ApiProps & {
   subscriptions: Array<any> // rxjs$ISubscription | null>;
 };
 
+export function shouldUseLatestChain(chain?: string): boolean {
+  if (typeof chain === 'undefined') return false;
+  const re = new RegExp("(poc-1)", "i");
```
jacogr: ' as opposed to "

packages/apps/src/NodeInfo.tsx
```diff
@@ -16,8 +16,15 @@ import translate from './translate';
...
+  if (typeof chain === 'undefined') return false;
```
jacogr: Use 

```
import isUndefined from '@polkadot/util/is/undefined
```

packages/apps/src/NodeInfo.tsx
```diff
@@ -16,8 +16,15 @@ import translate from './translate';
...
+  const re = new RegExp("(dev|loc)", "i");
```
jacogr: ' as opposed to "

jacogr: Does the RegEx need to be instantiated each time or can it be defined higher up?

packages/ui-react-rx/src/Api/index.tsx
```diff
@@ -25,10 +25,16 @@ type State = ApiProps & {
   subscriptions: Array<any> // rxjs$ISubscription | null>;
 };
 
+export function shouldUseLatestChain(chain?: string): boolean {
+  if (typeof chain === 'undefined') return false;
```
jacogr: `isUndefined` as above

packages/ui-react-rx/src/Api/index.tsx
ltfschoen: i've run `yarn run check` and fixed all linting problems. could we add a pre-push git hook to run linting and/or tests?

jacogr: Feel free. 

(I guess it never bothered me since I've always had the editor just run the stuff - tests are an issue since some repos are big testing wise, this one doesn't have that much. Linting would make sense)

packages/apps/src/NodeInfo.tsx
ltfschoen: agreed. moved to ui-react-rx/src/util as discussed

ltfschoen:
i believe i've addressed all review comments now

ltfschoen:
i'll have to make further updates since `poc-2` has been renamed to `staging` in this recent commit https://github.com/paritytech/polkadot/commit/3426552c763a157fa872ab11c1029ed4834f7a42

Issue #114, opened 2018-06-30
Stricter typing for children props


jacogr:
Currently children are defined as `any` across all of the base, this should be `React.ReactNode`, example -

https://github.com/polkadot-js/apps/blob/master/packages/ui-app/src/Input.tsx#L17

jacogr:
Closed in https://github.com/polkadot-js/apps/pull/134

PR #134, opened 2018-07-11
#114 Stricter typing for children props

yjkimjunior:
@jacogr just like this? #114


jacogr:
Thank you! Much appreciated. This was one of those one that I let slip through the cracks when converting earlier - any is a bit of a smell having any all over the place.

jacogr:
Closes #114

Issue #144, opened 2018-07-12
Extrinsics tab with Balance entered greater than ~9007100000000000 restarts the webpage


ltfschoen:
If you go to the and enter a value such as 9007200000000000, then the webpage will refresh, which may prevent users like Alice with 1152921504606846976 DOTs from entering their entire account balance.

![screen shot 2018-07-12 at 5 46 33 pm](https://user-images.githubusercontent.com/6226175/42644361-b609c9ec-85fb-11e8-94a9-b6e53c872b48.png)

jacogr:
So the input handles it as a number - would probably need to make an input (text values) for Balance then.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER

ltfschoen:
Note that with the changes so far it no longer pre-populates the input field as it did previously with the initial value specified in https://github.com/polkadot-js/apps/blob/master/packages/ui-app/src/Params/initValue.ts#L17

jacogr:
Closed in master as well as "next" (along with denomination input there), will be fully live with "next" branch due for poc-3.

PR #146, opened 2018-07-12
fix: Fixes Issue #144. Balance to submit with transaction changed to string

ltfschoen:
- Placeholder indicates acceptable range of values to user in the interim until form validation incorporated.
- Max value must be the balance of the account minus one, otherwise the Polkadot Node will give error `Runtime: proposer's balance too low`.
- Min value must be above 0.000000000000001, otherwise the Polkadot Node will give error `Runtime: proposer's balance too low`

screen shot 2018-07-12 at 9 53 03 pm

### DIFF
--- a/packages/ui-app/src/Params/Param/StringBalance.tsx
+++ b/packages/ui-app/src/Params/Param/StringBalance.tsx
@@ -1,0 +1,45 @@
+// Copyright 2017-2018 @polkadot/ui-app authors & contributors
+// This software may be modified and distributed under the terms
+// of the ISC license. See the LICENSE file for details.
+
+import { Props } from '../types';
+
+import React from 'react';
+
+import Input from '../../Input';
+import Bare from './Bare';
+
+export default class StringBalance extends React.PureComponent<Props> {
+  render () {
+    const { className, isDisabled, isError, label, style, withLabel } = this.props;
+
+    return (
+      <Bare
+        className={className}
+        style={style}
+      >
+        <Input
+          className='large'
+          isDisabled={isDisabled}
+          isError={isError}
+          label={label}
+          onChange={this.onChange}
+          placeholder='<any number between [0.000000000000001] and [available balance minus 1])>'
+          type='text'
+          withLabel={withLabel}
+        />
+      </Bare>
+    );
+  }
+
+  onChange = (value: string): void => {
+    const { onChange } = this.props;
+
+    const isValid = value.length !== 0;
+
+    onChange({
+      isValid,
+      value
+    });
+  }
+}
--- a/packages/ui-app/src/Params/Param/findComponent.ts
+++ b/packages/ui-app/src/Params/Param/findComponent.ts
@@ -14,12 +14,13 @@ import Hash from './Hash';
 import KeyValue from './KeyValue';
 import KeyValueStorageArray from './KeyValueStorageArray';
 import StringParam from './String';
+import StringBalance from './StringBalance';
 import Unknown from './Unknown';
 import VoteThreshold from './VoteThreshold';
 
 const components: ComponentMap = {
   'AccountId': Account,
-  'Balance': Amount,
+  'Balance': StringBalance,
   'BlockNumber': Amount,
   'bool': Bool,
   'Bytes': Bytes,
--- a/packages/ui-app/src/Params/Param/findComponent.ts
+++ b/packages/ui-app/src/Params/Param/findComponent.ts
@@ -14,13 +14,13 @@ import Hash from './Hash';
 import KeyValue from './KeyValue';
 import KeyValueStorageArray from './KeyValueStorageArray';
 import StringParam from './String';
-import StringBalance from './StringBalance';
+import Balance from './Balance';
 import Unknown from './Unknown';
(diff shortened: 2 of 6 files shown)


packages/ui-app/src/Params/Param/StringBalance.tsx
```diff
@@ -0,0 +1,45 @@
...
+        className={className}
+        style={style}
+      >
+        <Input
+          className='large'
+          isDisabled={isDisabled}
+          isError={isError}
+          label={label}
+          onChange={this.onChange}
+          placeholder='<any number between [0.000000000000001] and [available balance minus 1])>'
```
jacogr: This needs to be translatable.

packages/ui-app/src/Params/Param/StringBalance.tsx
```diff
@@ -0,0 +1,45 @@
...
+          withLabel={withLabel}
+        />
+      </Bare>
+    );
+  }
+
+  onChange = (value: string): void => {
+    const { onChange } = this.props;
+
+    const isValid = value.length !== 0;
```
jacogr: Currently stuff like `123,ghf` would be valid, it is not the intent - when `isValid` is set, the valis is supposed to be valid for submission.

packages/ui-app/src/Params/Param/StringBalance.tsx
jacogr: This needs to be translatable.

packages/ui-app/src/Params/Param/findComponent.ts
```diff
@@ -14,12 +14,13 @@ import Hash from './Hash';
...
 import KeyValueStorageArray from './KeyValueStorageArray';
 import StringParam from './String';
+import StringBalance from './StringBalance';
 import Unknown from './Unknown';
 import VoteThreshold from './VoteThreshold';
 
 const components: ComponentMap = {
   'AccountId': Account,
-  'Balance': Amount,
+  'Balance': StringBalance,
```
jacogr: `BalanceInput`

packages/ui-app/src/Params/Param/StringBalance.tsx
```diff
@@ -0,0 +1,45 @@
...
+export default class StringBalance extends React.PureComponent<Props> {
+  render () {
+    const { className, isDisabled, isError, label, style, withLabel } = this.props;
+
+    return (
+      <Bare
+        className={className}
+        style={style}
+      >
+        <Input
```
jacogr: It needs a pattern to strip invalid chars at the very least. We can add a pattern prop to the base input and have it attached to the raw HTML input - https://www.w3schools.com/tags/att_input_pattern.asp

ltfschoen:
I've also noticed that with these changes so far that it no longer pre-populates the input field with a visible value of 1 as it did previously with the initial value specified in https://github.com/polkadot-js/apps/blob/master/packages/ui-app/src/Params/initValue.ts#L17. It allows Submit Transaction when the Balance input field appears to be empty, but the Polkadot Node still processes a value of 1, unless the user changes the value in the input field

packages/ui-app/src/Params/Param/StringBalance.tsx
ltfschoen: As we discovered, that approach doesn't work, however I've implemented the `maxLength` as you recommended with success

packages/ui-app/src/Params/Param/StringBalance.tsx
ltfschoen: i've now wrapped the component with `translate`

packages/ui-app/src/Params/Param/StringBalance.tsx
ltfschoen: i've now added an input field validation function and associated unit tests

packages/ui-app/src/Params/Param/findComponent.ts
ltfschoen: i've renamed it to `Balance` as discussed

ltfschoen:
i've added additional commits. It now pre-populates the input field correctly.

**Updates** include:
- add internationalisation
- add maxLength to input field depending on whether poc-1 (64bit) or latest (128bit) chain used
- passed down api chain data via props
- add unit tests for balance input field
- add balance input field validation to only accept numbers

**Problems** i'm having include:
- a couple of the new unit tests that I added are not all passing but don't understand why. it's to do with comparing the maximum unsigned 64bit or 128bit value of the balance depending on what chain they are using (i.e. poc-1 or latest)
- when i add `&& maxBN64Bit.gte(inputBN)` or `&& maxBN128Bit.gte(inputBN)` it doesn't appear
to work correctly, as in the UI it doesn't allow the length of the balance to be as long as I would expect. I tried breakpoint debugging but couldn't figure it out. I've just commented their usage out in the meantime

jacogr:
Could we please move the branch, so it is possible to just switch to it? Just easier without having to clone to take a look and fix the issues.

ltfschoen:
Closing as I created a PR from a branch of this repo here https://github.com/polkadot-js/apps/pull/168 instead of using the PR from my fork of it

PR #154, opened 2018-07-18
fix: Fixes #94 by adding missing filter for extrinsics timestamp

### LINKED ISSUE #94: Assertion error in the "Extrinsic tab page" when user selects "timestamp" from the selection box labelled "from extrinsic section"
When select the "Extrinsics tab" and then choose "timestamp" from the selection box under the "from extrinsic section" label it crashes with `Error: Assertion failed`. See screenshots below:

screen shot 2018-06-13 at 10 42 46 am

screen shot 2018-06-13 at 10 43 18 am

ltfschoen:


packages/ui-app/src/InputExtrinsic/options/section.test.js
```diff
@@ -0,0 +1,67 @@
+// Copyright 2017-2018 @polkadot/ui-app authors & contributors
```
jacogr: .spec.js

packages/ui-app/src/InputExtrinsic/options/section.ts
```diff
@@ -8,6 +8,21 @@ import { DropdownOptions } from '../types';
 
 import map from '@polkadot/extrinsics';
 
+export function anyMethodToDisplay (methods: any) {
```
jacogr: Since these are re-usable (e.g. same logic applies to eg. storage), put them in seperate files

packages/ui-app/src/InputStorage/index.tsx
```diff
@@ -19,7 +19,7 @@ import classes from '../util/classes';
 import translate from '../translate';
 import SelectKey from './SelectKey';
 import SelectSection from './SelectSection';
-import keyOptions from './options/key';
+import methodOptions from './options/method';
```
jacogr: Not sure why this is done - in storage the term "method" doesn't make much sense since it actually refers to a key in the state storage. (For extrinsics methods are fine since that is what they are)

ltfschoen:
@jacogr I've made updates to address the review comments

packages/ui-app/src/InputExtrinsic/options/section.ts
```diff
@@ -2,27 +2,21 @@
...
-
+import { Section } from '@polkadot/params/types';
+import { Extrinsic$Sections, Extrinsics } from '@polkadot/extrinsics/types';
 import { DropdownOptions } from '../types';
+import { shouldDisplaySection } from '../../util/shouldDisplaySection';
 
 import map from '@polkadot/extrinsics';
 
-export default function createOptions (type: 'private' | 'public'): DropdownOptions {
+export default function createOptions (type: string = 'private' || 'public'): DropdownOptions {
```
jacogr: Not sure I understand this one - it is not a general string, rather one of 2 values. It should rather (if an issue) be defined as an opaque type (as it was) and if an issue in use at a single point, casted.

packages/ui-app/src/InputStorage/options/section.ts
```diff
@@ -2,32 +2,21 @@
...
 import map from '@polkadot/storage';
 
-export default function createOptions (): DropdownOptions {
+export default function createOptions (type: string = 'public'): DropdownOptions {
   return Object
     .keys(map)
     .sort()
     .filter((name) => {
-      const section = map[name as Storage$Sections];
+      const section: Section<Storages> = map[name as Storage$Sections];
```
jacogr: Is this type needed here? Type inference will set it up correctly.

packages/ui-app/src/InputExtrinsic/options/section.ts
jacogr: On closer inspection, this is actually completely wrong now. What is defined here now is an optional parameter that (when not supplied) takes the value of ‘public || private’ which evaluated to the former. It is not a type anymore, rather now you can pass in any string value, ie. even ‘foo’ will work and the type checker will be happy.

packages/ui-app/src/InputStorage/options/section.ts
ltfschoen: Agreed, type inference does the trick

packages/ui-app/src/InputStorage/index.tsx
```diff
@@ -44,10 +44,11 @@ class InputStorage extends React.PureComponent<Props, State> {
     super(props);
 
     const { section } = this.props.defaultValue;
+    const type = 'public';
```
jacogr: Is there a reason this is actually passed through? (storage always being public - really don't need to define it explicitly as a variable?)

packages/ui-app/src/InputStorage/options/section.ts
```diff
@@ -4,30 +4,18 @@
 
 import { Storage$Sections } from '@polkadot/storage/types';
 import { DropdownOptions } from '../../InputExtrinsic/types';
+import { shouldDisplaySection } from '../../util/shouldDisplaySection';
 
 import map from '@polkadot/storage';
 
-export default function createOptions (): DropdownOptions {
+export default function createOptions (type: 'public'): DropdownOptions {
```
jacogr: As per previous comment, rather just pass through 'public' on line 18

packages/ui-app/src/InputStorage/index.tsx
ltfschoen: I just noticed it was done this way here https://github.com/polkadot-js/apps/blob/master/packages/ui-app/src/InputExtrinsic/index.tsx#L54 where it's also passed through

packages/ui-app/src/util/shouldDisplaySection.ts
```diff
@@ -0,0 +1,31 @@
+// Copyright 2017-2018 @polkadot/ui-app authors & contributors
+// This software may be modified and distributed under the terms
+// of the ISC license. See the LICENSE file for details.
+
+import isUndefined from '@polkadot/util/is/undefined';
+import isNull from '@polkadot/util/is/null';
+
+function anyMethodToDisplay (methods: any) {
```
jacogr: Almost there. `any` is a bit of a "smell", only used when no options. Here we can actually type it using Section + generics. (The callers all have the same generic-ified Section type)

packages/ui-app/src/util/shouldDisplaySection.ts
```diff
@@ -0,0 +1,31 @@
...
+    .keys(methods)
+    .filter((name) => {
+      const { isDeprecated, isHidden } = methods[name];
+
+      return !isDeprecated && !isHidden;
+    })
+    .length !== 0;
+}
+
+export function shouldDisplaySection (section: any, type: string) {
```
jacogr: As above.

Additionally, type is not a string, it is one of public or private.

ltfschoen:
Creating a new PR as a branch of this repo instead of a branch of my fork

PR #155, opened 2018-07-22
fix: Fixes #94 by adding missing filter for extrinsics timestamp

### LINKED ISSUE #94: Assertion error in the "Extrinsic tab page" when user selects "timestamp" from the selection box labelled "from extrinsic section"

ltfschoen:


jacogr:
Why are there 2 of these?

PR #158, opened 2018-07-24
WIP: Fixes #116 Show Storage query response with AccountId key in table

### LINKED ISSUE #116: Better display of accounts/others in storage
i.e. currently a list of accounts will show the addresses, however probably should be expanded to show the actual identitycon as well.

Where param names as displayed, it can be "_paramName_ = paramValue"

Basically take what we have and turn it from a "text-dump" into something that actually looks and feels usable

ltfschoen:
Screenshot with changes shown:

screen shot 2018-07-24 at 1 14 40 pm

### DIFF
--- a/packages/app-storage/src/Query.tsx
+++ b/packages/app-storage/src/Query.tsx
@@ -14,7 +14,6 @@ import Labelled from '@polkadot/ui-app/Labelled';
 import valueToText from '@polkadot/ui-app/Params/valueToText';
 import classes from '@polkadot/ui-app/util/classes';
 import withStorageDiv from '@polkadot/ui-react-rx/with/storageDiv';
-import IdentityIcon from '@polkadot/ui-react/IdentityIcon';
 
 import translate from './translate';
 
@@ -42,48 +41,14 @@ class Query extends React.PureComponent<Props, State> {
         value as Storage$Key$Value
       );
 
-      const renderAccounts = (accountIds: Array<string>) => {
-        return (
-          <table className='accounts'>
-            <thead>
-              <tr>
-                <th />
-                <th>Account ID</th>
-              </tr>
-            </thead>
-            <tbody>
-              {
-                accountIds && accountIds.length ?
-                  (
-                    accountIds
-                      .map((accountId) => (
-                        <tr key={accountId}>
-                          <td><IdentityIcon size={24} value={accountId} /></td>
-                          <td>{accountId}</td>
-                        </tr>
-                      ))
-                  )
-                : null
-              }
-            </tbody>
-          </table>
-        );
-      };
-
       const lineHeight = key.type[0] === 'AccountId' ? '2em' : '1em';
 
       cache[id] = withStorageDiv(key, { params: values })(
         (value: any) => {
-          const accountIds: Array<string> = valueToText(key.type, value).split(', ');
-          if (key.type[0] === 'AccountId') {
-            return renderAccounts(accountIds);
-          } else {
-            return accountIds;
-          }
+          return valueToText(key.type, value);
         },
         { className: 'ui--output', style: { lineHeight: lineHeight } }
       );
-    //   }
     }
 
     return cache[id];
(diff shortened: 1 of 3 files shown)


jacogr:
Ohhh, pretty :) Will take a loo through the code a bit later, but like what I see UI-wise, it helps a _lot_ :)

packages/app-storage/src/Query.tsx
```diff
@@ -41,11 +43,51 @@ class Query extends React.PureComponent<Props, State> {
         value as Storage$Key$Value
       );
 
-      cache[id] = withStorageDiv(key, { params: values })(
-        (value: any) =>
-          valueToText(key.type, value),
-        { className: 'ui--output' }
-      );
+      // FIXME - Move into ui-app/src/Params/valueToAccountList for reuse
```
jacogr: The idea is right - however, we should actually be extending https://github.com/polkadot-js/apps/blob/master/packages/ui-app/src/Params/valueToText.ts to return React.Node (or text, text just being a subset of Node)

packages/app-storage/src/Query.tsx
```diff
@@ -41,11 +43,51 @@ class Query extends React.PureComponent<Props, State> {
...
+                      ))
+                  )
+                : null
+              }
+            </tbody>
+          </table>
+        );
+      };
+
+      if (key.type[0] === 'AccountId') {
```
jacogr: This is not correct, it only works for the specific instance and not for eg. Referendums where AccountId is not in that position. As per above, it needs to be on a single value (so each account, dfoesn't matter where gets rendered)

In some cases where we have tuples with AccountId first, we will now not display all info.

packages/app-storage/src/Query.tsx
```diff
@@ -41,11 +43,51 @@ class Query extends React.PureComponent<Props, State> {
...
-          valueToText(key.type, value),
-        { className: 'ui--output' }
-      );
+      // FIXME - Move into ui-app/src/Params/valueToAccountList for reuse
+      const renderAccounts = (addresses: Array<Uint8Array | string>) => {
+        return (
+          <table className='accounts'>
+            <thead>
+              <tr>
+                <th>Identitycon</th>
```
jacogr: When displaying (since we should not be using tables, rather render the single values - see other comments), no need to have a header for Icon. (throw-away comment here sin ce the implementation needs some work)

packages/app-storage/src/Query.tsx
jacogr: Correct place is here https://github.com/polkadot-js/apps/blob/master/packages/ui-app/src/Params/valueToText.ts#L64

packages/app-storage/src/Query.tsx
```diff
@@ -41,11 +42,48 @@ class Query extends React.PureComponent<Props, State> {
...
+
+      const lineHeight = key.type[0] === 'AccountId' ? '2em' : '1em';
+
       cache[id] = withStorageDiv(key, { params: values })(
-        (value: any) =>
-          valueToText(key.type, value),
-        { className: 'ui--output' }
+        (value: any) => {
+          const accountIds: Array<string> = valueToText(key.type, value).split(', ');
+          if (key.type[0] === 'AccountId') {
```
jacogr: This should not be in here - the generic valueToText should just render this. There should be no specific rendering in here to do with types at all.

The above check would only work in some specific instances.

jacogr: Basically -

1. valueToText should now return a React.Node
2. with AccountId rendering, pass pack a div/span with the icon & value

This https://github.com/polkadot-js/apps/blob/master/packages/ui-app/src/Params/valueToText.ts#L64 should be the only place where we actually check for the type, nowhere elase.

jacogr:
Could we please move the branch (away from the fork), so it is possible to just switch to it? Just easier without having to clone to take a look and fix the issues.

ltfschoen:
Closing as moving to a PR from a branch of this repo https://github.com/polkadot-js/apps/pull/169, instead of a PR from a branch of my fork of it

PR #161, opened 2018-07-26
Fixes #117 - Telemetry frontend integration

### LINKED ISSUE #117: Make telemetry visible as an app
As the telemetry is available, pull it in as an app - the front-end can be pulled-in, the back-end is hosted, i.e. it should connect to the same one. There are probably some smaller changes required on the telemetry app, i.e. not connecting by default to the host. For the most part, should just be a navigation entry and launching.

Not an immediate, telemetry UI should stabilise first.

@maciejhirsz Any additional comments here? (Don't do anything to change/adapt, but would like to explore the possibility in the future)

ltfschoen:
This PR includes all Telemetry frontend package and Telemetry common package commits up until 5pm on Jul 26, 2018 from the Dotstats repo that I've copy/pasted across https://github.com/polkadot-js/dotstats. Probably not the best solution, as better alternative would be to publish them both to NPM and just import a single Telemetry component.

I've temporarily hid the border around the icons in the Telemetry UI as they were offset for some reason. The map icon is also slightly offset but I've left it alone

I've refactored the code since wouldn't allow use of `export namespace` when building the polkadot-js/apps repo

I haven't figure out how to resolve the following tslinting errors, but they don't prevent building the code and using the app 

* block is empty
* keyCode is deprecated.
* Promises must be handled appropriately

Screenshot below.

screen shot 2018-07-26 at 5 27 18 pm


jacogr:
I'm not convinced we need to pull in the code - surely we can pull from the repo with a `"@dotstats/ui": "polkadot-js/dotstats#<ref>"` (so that is always the source of truth) and just have a thin entry point that launces the app and possibly sets some overrides?

ltfschoen:
Great idea! I'll give it a shot

ltfschoen:
FYI new branch is in process. We'll use to import telemetry frontend luke-117-telemetry-app-github-pull

PR #164, opened 2018-07-30
Add Dockerfile

chevdor:
Add a multistage docker image with the UI.
It builds a docker image based on nginx and weight around 11 MB.
See https://hub.docker.com/r/chevdor/polkadot-ui/

### DIFF
--- a/packages/apps/src/index.tsx
+++ b/packages/apps/src/index.tsx
@@ -36,9 +36,14 @@ function App ({ className, style }: Props) {
   );
 }
 
-createApp(App, {
-  // tslint:disable-next-line
-  url: (process.env.WS_URL === null) || (process.env.WS_URL === '')
+// tslint:disable-next-line
+var url = (process.env.WS_URL === null) || (process.env.WS_URL === '')
     ? undefined
-    : process.env.WS_URL
+    : process.env.WS_URL;
+
+console.log('process.env.WS_URL=', process.env.WS_URL);
+console.log('Web socket url=', url);
+
+createApp(App, {
+  url
 });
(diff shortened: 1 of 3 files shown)


packages/apps/src/index.tsx
```diff
@@ -36,9 +36,12 @@ function App ({ className, style }: Props) {
...
 
-createApp(App, {
-  // tslint:disable-next-line
-  url: (process.env.WS_URL === null) || (process.env.WS_URL === '')
+// tslint:disable-next-line
+var url = (process.env.WS_URL === null) || (process.env.WS_URL === '')
     ? undefined
-    : process.env.WS_URL
+    : process.env.WS_URL;
+console.log('process.env.WS_URL=', process.env.WS_URL);
```
jacogr: Not sure we still want these anymore?

If _really_ needed, spacing before and after would work just to make it readable.

jacogr:
... and I'm assuming 11MB in your comment, not GB :)

chevdor:
Oups, MB indeed!

packages/apps/src/index.tsx
chevdor: The main issue here is that using a build, the WS_URL seems to no longer be read.
I guess this won´t be an issue for full dockerized environments but I did not find a way yet to get the UI within Docker to talk to a random node.

Passing an ENV with docker looks like:
`docker run --rm -d -e WS_URL=http://a.b.c.d:9933 chevdor/polkadot-ui:latest`
but the value does not make it thru.

chevdor: Done

Issue #108, opened 2018-06-28
Convert the hashing to libsodium-wasm for vanity addr


jacogr:
... not a high priority since the functionality is just a "o, this can be cool", however should bring in a nice boost which should make it actually useful. https://github.com/alexcrichton/wasm-sodium

(For the command-line app the generation uses natively compiled sodium (as available), it is faster.)

Probably seeing this bit as a POC, when done properly it is added to the `@polkadot/util-crypto`, however in that case we probably wnat native & wasm support. So for now, trial with just the vanitygen app.

PR #170, opened 2018-08-01
sodium-wasm support

jacogr:
Closes https://github.com/polkadot-js/apps/issues/108

Seems like a speedup in vanity generation of around 10x. (Still slower than command-line version, but a step in the right direction)

### DIFF
--- a/packages/app-vanitygen/src/generator/sodiumKeygen.ts
+++ b/packages/app-vanitygen/src/generator/sodiumKeygen.ts
@@ -1,0 +1,18 @@
+// Copyright 2017-2018 @polkadot/app-vanitygen authors & contributors
+// This software may be modified and distributed under the terms
+// of the ISC license. See the LICENSE file for details.
+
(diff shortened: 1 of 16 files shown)


packages/app-vanitygen/src/generator/sodiumKeygen.ts
```diff
@@ -0,0 +1,17 @@
+// Copyright 2017-2018 @polkadot/app-vanitygen authors & contributors
+// This software may be modified and distributed under the terms
+// of the ISC license. See the LICENSE file for details.
+
+import { Generator$PkFromSeed } from './types';
+
+import bufferToU8a from '@polkadot/util/buffer/toU8a';
+import u8aToBuffer from '@polkadot/util/u8a/toBuffer';
+
+export default function setupFn (sodium: any): Generator$PkFromSeed {
```
jacogr: This is horrible - however, a small nightmare to type both the sodium and libsodium-wrappers completely in the /ts repo.

Issue #175, opened 2018-08-02
Add nominate/unnominate to staking


jacogr:
Didn't make it into https://github.com/polkadot-js/apps/pull/171 since staking getStorage keys are very, very outdated. Needs an update wih all the new keys.

PR #183, opened 2018-08-03
(un)Nominate functionality

jacogr:
Closes https://github.com/polkadot-js/apps/issues/175

polkadot apps portal 2018-08-03 16-44-26

### DIFF
--- a/packages/app-staking/src/Account.tsx
+++ b/packages/app-staking/src/Account.tsx
@@ -76,6 +76,7 @@ class Account extends React.PureComponent<Props, State> {
 
   private renderAccount () {
     const { address, isValidator, name, t } = this.props;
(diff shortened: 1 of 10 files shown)


packages/app-staking/src/Account.tsx
```diff
@@ -114,6 +117,116 @@ class Account extends React.PureComponent<Props, State> {
...
+    if (!isInNominate) {
+      return null;
+    }
+
+    const { t } = this.props;
+
+    return (
+      <div className='staking--Account-nominate'>
+        <div className='ui--row'>
+          <Input
```
jacogr: This really should be a validator dropdown. (However needs to be logged - we currently don't have a dropdown that takes an arbitrary list of accounts)

Issue #190, opened 2018-08-04
Re-lock account after sign/send is completed


jacogr:
User to be promted for password again on subsequent sends.

(Obviously not in the case of §isTesting§ accounts)

Needs - `lockAccount()` in keyring (or sign with password, that does it automatically)

jacogr:
First part hasSecretKey -> isLocked, contained in https://github.com/polkadot-js/apps/pull/200

ltfschoen:
I've attempted to address this issue in branch luke-190-relock-account-after-sign-send-completed https://github.com/polkadot-js/apps/compare/luke-190-relock-account-after-sign-send-completed?expand=1. I've updated create.ts so if create a new account and then open the modal to send a transaction it shows the password prompt (since beforehand it didn't prompt the user for the password straight after creating an account). But strangely if I debug just before calling the lockAccount function with `pair.isLocked()` it says the pair isn't locked... And when I run `pair.lock()` in lock.ts (the lockAccount function) to lock it whether it's already locked or not, it generates error         `Uncaught (in promise) Error: bad secret key size` from nacl-fast.js.

ltfschoen:
Refer to comments here: https://github.com/polkadot-js/apps/pull/314#issuecomment-421962234.
This is being actioned in branch jg-account-lock https://github.com/polkadot-js/apps/compare/jg-account-lock?expand=1, so PR luke-190-relock-account-after-sign-send-completed was closed.
This is dependent on settings app where user can defined unlocking strategy https://github.com/polkadot-js/apps/issues/86

jacogr:
Closing this for now until it pops up.

PR #110, opened 2018-06-28
WIP: Add basic UI for settings

chevdor:
image

PR #198, opened 2018-08-06
Re-lock account after use (as set)

jacogr:
- Allow user to specify unlock strategy, per session or per use
- Lock account per use as required
- Closes https://github.com/polkadot-js/apps/issues/190

Future -

- Once https://github.com/polkadot-js/apps/pull/110 available, allow the user to specify the default


jacogr:
Re-open once https://github.com/polkadot-js/apps/pull/202 is in (and integrated).

Issue #86, opened 2018-06-11
Add settings apps, e.g changing api endpoint


jacogr:
1. Add settings app, `packages/app-settings`
2. Allow configuration of WS endpoint, anything allowed (`ws://|wss://` prefixed)
3. Show 2 pre-configured options, one for poc-1 hosted node (`wss://poc-1.polkadot.io:9944`) and one for locally-running node (`ws://127.0.0.1:9944`)
4. WS URL should be stored in localstorage - by default we probably want to pull in `WS_URL` env variable for the running-instance default.
5. On changing and setting endpoint via `setApiWsUrl` (assuming we wrap with `ui-react-rx/with/api` we probably need to refresh - the API itself won't automatically refresh existing subscriptions (that needs to be fix at the API layer, probably more involved - phase II)

jacogr:
API issue that deals with re-connections - https://github.com/polkadot-js/api/issues/75 (as mentioned above, phase II)

jacogr:
Closed in #359 in "next" branch, due to go live with poc-3.

PR #221, opened 2018-08-14
fix: Fixes Issue #144. Extrinsics with high balance restarts webpage

ltfschoen:
**Screenshots**:

* Default value of 1 testnet DOT
screen shot 2018-08-14 at 5 48 03 pm

* Input validation error shown when no input value
screen shot 2018-08-14 at 5 48 14 pm

* Input validation error shown when user inputs non-numeric values
screen shot 2018-08-14 at 5 48 27 pm

* Throw exception input value into the `isValidBalance` function is in scientific notation
screen shot 2018-08-14 at 5 49 06 pm

* Maximum amount of digits allowed for the balance input value
screen shot 2018-08-14 at 5 51 28 pm

**References:**
* Previous PR that was closed https://github.com/polkadot-js/apps/pull/168
* Original PR from my fork that was closed in favour of a branch from the upstream repo https://github.com/polkadot-js/apps/pull/146


ltfschoen:
@jacogr I've finished incorporating your review comments. If there's an error in the user input we shown them an error message notification. If there's no error and they enter a value in either scientific notation (i.e. 3.4e3) or exponential notation (i.e. 3.4e+3) then I we shown them an information notification below the input field with the converted equivalent integer value (i.e. 3400). Note that I allow decimal places only when they enter a value in scientific notation or exponential notation. If they enter an integer value (i.e. 3400) it shows an information notification with the converted equivalent value in scientific notation (i.e. 3.4e3) below. So the PR supports E-Notation #224.

The changes also require changes to polkadot-js/api, and I have created an PR here [https://github.com/polkadot-js/api/pull/139](https://github.com/polkadot-js/api/pull/139)

ltfschoen:
@jacogr I've tried implementing InputNumber again, hopefully this is what you meant?

packages/ui-app/src/InputNumber.tsx
```diff
@@ -0,0 +1,59 @@
...
+          maxLength={maxLength}
+          onChange={onChange}
+          onKeyDown={onKeyDown}
+          onKeyUp={onKeyUp}
+          placeholder={placeholder}
+          style={style}
+          type='text'
+          withLabel={withLabel}
+        />
+        <Notification
```
jacogr: Notification should be rendered on `Input` itself, not really on the wrapping component. (It is useful for all inputs, e.g. Password, Balance, normal, etc.)

packages/ui-app/src/Notification.tsx
```diff
@@ -0,0 +1,41 @@
...
+export default class Notification extends React.PureComponent<Props> {
+  render () {
+    const { error, info, size = 'tiny' } = this.props;
+
+    if (!error && !info) {
+      return null;
+    }
+
+    return (
+      <div className='ui--Notifications'>
```
jacogr: This one we would want to add the `className` & `style` as passed-in as well. (Sinmce it is a base component and we may want to apply overrides on a per-use basis)

packages/ui-app/src/Notification.tsx
```diff
@@ -0,0 +1,41 @@
...
+        <Message
+          size={size}
+          error={!!error}
+          info={!error ? !!info : undefined}
+        >
+          {error}
+          {!error ? info : null}
```
jacogr: Lots of `!error` here. Rather make a variable `isError` and use that, i.e.

```
const isError = !!error;
const isInfo = !isError && !!info;
```

(The logic here is not easy to follow, some mental gymnastics required.)

packages/ui-app/src/Params/Param/Balance.tsx
```diff
@@ -0,0 +1,192 @@
...
+import { ApiProps } from '@polkadot/ui-react-rx/types';
+import { I18nProps } from '../../types';
+import { Props as BareProps } from '../types';
+
+import BN from 'bn.js';
+import React from 'react';
+import withApi from '@polkadot/ui-react-rx/with/api';
+
+import isValidBalance from '../../util/isValidBalance';
+import Bare from './Bare';
```
jacogr: Move this import to be last, i.e. first imports external to us, then our import.

packages/ui-app/src/Params/Param/Balance.tsx
```diff
@@ -0,0 +1,192 @@
...
+      info: ''
+    };
+  }
+
+  render () {
+    const { className, defaultValue: { value }, isError, label, style, t, withLabel } = this.props;
+    const { error, info } = this.state;
+    const defaultValue = new BN((value as BN).toString(10) || '0').toString(10);
+
+    const maxLengthForLatestChainSpec = 39;
```
jacogr: This is going to create issues. :( For InputNumber, pass through the bitLength and calculate the number there. i.e. 

```
const maxValue = new BN(2).pow(bitLength).subn(1);
const maxLength = maxValue.toString().length;`
```

Additionally, the input value can now actually be checked against this maxValue to see that if it is still indeed in-range.

packages/ui-app/src/Params/Param/Balance.tsx
```diff
@@ -0,0 +1,192 @@
...
+          error={error}
+          info={info}
+          isError={isError}
+          label={label}
+          maxLength={maxLengthForLatestChainSpec}
+          onChange={this.onChange}
+          onKeyDown={this.onKeyDown}
+          onKeyUp={this.onKeyUp}
+          placeholder={t('account.balance.placeholder', {
+            defaultValue: 'Between 1 testnet DOT and the available testnet DOT balance (minus 1) of the account'
```
jacogr: We should no mention testnet, keep it generic otherwise we need to come back to this. Additionally, we can allow 0 as an input. (Just not negative). Why "minus 1"

packages/ui-app/src/Params/Param/Balance.tsx
```diff
@@ -0,0 +1,192 @@
...
+  }
+
+  onChange = (value: string): void => {
+    const { onChange, t } = this.props;
+
+    // prepare value by remove all whitespace
+    value = value.split(' ').join('');
+
+    try {
+      const { isValid, errorMessage, infoMessage, inputConvertedFromScientificNotation } = isValidBalance(value);
```
jacogr: `inputConvertedFromScientificNotation` this is getting very long... something like `isScientific` would do just as well.

packages/ui-app/src/Params/Param/Balance.tsx
```diff
@@ -0,0 +1,192 @@
...
+
+      this.setState({
+        error: !isValid && errorMessage ? t(errorMessage) : '',
+        info: isValid && infoMessage ? t(infoMessage) : ''
+      });
+
+      if (!onChange) return;
```
jacogr: No single-line returns, add `{}`

packages/ui-app/src/Params/Param/Balance.tsx
```diff
@@ -0,0 +1,192 @@
...
+
+      let newValue: BN;
+      if (inputConvertedFromScientificNotation) {
```
jacogr: Blank line before.

packages/ui-app/src/Params/Param/Balance.tsx
```diff
@@ -0,0 +1,192 @@
...
+      } else {
+        newValue = new BN(value || '0');
+      }
+
+      onChange({
+        isValid,
+        value: newValue
+      });
+    } catch (e) {
+      console.error('error: ', e);
```
jacogr: Just console.error(error) would do (Name it `error`, no need to shorten)

packages/ui-app/src/Params/Param/Balance.tsx
```diff
@@ -0,0 +1,192 @@
...
+    const tabKey = 9; // next input field
+    const aKey = 65; // select all from balance input
+    const xKey = 88; // cut balance
+    const cKey = 67; // copy balance
+    const eKey = 69; // scientific or exponential notation
+    const vKey = 86; // paste balance
+    const zeroKey = 48;
+    const zeroNumpadKey = 96;
+
+    const regexE = /[e]/gi;
```
jacogr: Remove extra blank line.

All these can be constants that are defined on top of the file (and named accordingly, i.e. `KEY_E` or `KEY_PLUS`). As it stands each time the method fires the variables are created in-scope.

packages/ui-app/src/Params/Param/Balance.tsx
```diff
@@ -0,0 +1,192 @@
...
+    ) {
+      event.preventDefault();
+      return;
+    }
+
+    // obtain the current cursor index position in the Balance input field
+    const inputCursorIndex = event.target.value.slice(0, event.target.selectionStart).length;
+
+    // allow these keys
+    if ([escapeKey, enterKey, backspaceKey, delKey, tabKey, ctrlKey, altKey, cmdKey, eKey, arrowLeftKey,
```
jacogr: `KEYS_ALLOWED` (Once again, we don't need to re-create the array each time, it is file constant)

packages/ui-app/src/Params/Param/Balance.tsx
```diff
@@ -0,0 +1,192 @@
...
+      return;
+    }
+
+    // allow users to to use cut/copy/paste combinations, but not non-numeric letters individually
+    // allow users to use the + key for exponential notation
+    if (
+      ((event.ctrlKey || cmdKey) && event.which === aKey) || // select all
+      ((event.ctrlKey || cmdKey) && event.which === cKey) || // copy
+      ((event.ctrlKey || cmdKey) && event.which === vKey) || // paste
+      ((event.ctrlKey || cmdKey) && event.which === xKey) || // cut
```
jacogr: `(event.ctrlKey || cmdKey)` is re-used, so something like this would work as well -

``(event.ctrlKey || cmdKey) && [aKey, cKey, vKey, xKey].includes(vKey)`

(Once again in the above example, just make sure we have the proper constant arrays defined.

packages/ui-app/src/Params/Param/Balance.tsx
```diff
@@ -0,0 +1,192 @@
...
+        error: t('Non-numeric values are not permitted')
+      });
+      event.preventDefault();
+      return;
+    }
+  }
+
+  onKeyUp = (event: any): void => {
+    // if user inputs the value of 'E', replace it with lowercase 'e'
+    const regexE = /[E]/gi;
```
jacogr: file-based constant.

packages/ui-app/src/util/isValidBalance.ts
```diff
@@ -0,0 +1,97 @@
...
+
+  const matchE = input.match(/e/gi);
+  const matchPlus = input.match(/\+/gi);
+  const matchEPlus = input.match(/e\+/gi);
+
+  const matchEBeforeNumberOrPlus = input.match(/(e)(?=[0-9+])/ig);
+  const matchEPlusBeforeNumber = input.match(/(e\+)(?=[0-9])/ig);
+
+  if (matchE && matchE.length > 1 || matchPlus && matchPlus.length > 1) {
+    return { isValid: false, errorMessage: 'Scientific notation may only contain one instance of \'e\' for scientific notation or exponential with \'e+\'' };
```
jacogr: - Each prop of the object on a seperate line.
- All messages need to be translatable, these are not

packages/ui-app/src/util/isValidBalance.ts
```diff
@@ -0,0 +1,97 @@
...
+
+    if (matchEPlus && !inputConvertedFromScientificNotationBN.lt(maxBN128Bit)) {
+      return { isValid: false, errorMessage: 'Balance value after converting from exponential notation exceeds maximum for 128 bit' };
+    } else if (!inputConvertedFromScientificNotationBN.lt(maxBN128Bit)) {
+      return { isValid: false, errorMessage: 'Balance value after converting from scientific notation exceeds maximum for 128 bit' };
+    } else {
+      return {
+        isValid: true,
+        infoMessage: inputConvertedFromScientificNotation,
+        inputConvertedFromScientificNotation: inputConvertedFromScientificNotation
```
jacogr: Simpler to do -

```
return {
        isValid: true,
        infoMessage,
        inputConvertedFromScientificNotation
}
```

packages/ui-app/src/util/isValidBalance.ts
```diff
@@ -0,0 +1,97 @@
...
+    if (!inputConvertedFromScientificNotation) {
+      return { isValid: false, errorMessage: errorMessage };
+    }
+
+    const inputConvertedFromScientificNotationBN = new BN(inputConvertedFromScientificNotation);
+
+    if (matchEPlus && !inputConvertedFromScientificNotationBN.lt(maxBN128Bit)) {
+      return { isValid: false, errorMessage: 'Balance value after converting from exponential notation exceeds maximum for 128 bit' };
+    } else if (!inputConvertedFromScientificNotationBN.lt(maxBN128Bit)) {
+      return { isValid: false, errorMessage: 'Balance value after converting from scientific notation exceeds maximum for 128 bit' };
```
jacogr: Same comments are earlier in the file - between this and all the preceding lines the same issue occurs. (Formatting & non-translatable messages)

packages/ui-app/src/util/isValidBalanceExpectedResponses.ts
```diff
@@ -0,0 +1,32 @@
+// Copyright 2017-2018 @polkadot/ui-app authors & contributors
+// This software may be modified and distributed under the terms
+// of the ISC license. See the LICENSE file for details.
+
+import { IsValidWithMessage } from './types';
+
+export enum ErrorMessage {
+  BalanceMustContainOnlySingleEOrPlus = 'Scientific notation may only contain one instance of \'e\' for scientific notation or exponential with \'e+\'',
```
jacogr: No. All messages needs to be translatable when they are used on the UI. As it stand, these are not.

packages/ui-app/src/util/isValidBalanceExpectedResponses.ts
```diff
@@ -0,0 +1,32 @@
...
+  BalanceInExponentialNotationMustHaveExponent = 'Exponential \'e+\' notation when used must be followed by a number',
+  BalanceInScientificNotationMustHaveExponent = 'Scientific notation \'e\' when used must be followed by a number',
+  BalanceMustBeNumber = 'Balance to transfer in DOTs must be a number or expressed in scientific notation (i.e. 3.4e38) or exponential with \'e+\'',
+  BalanceMinimumRequired = 'Balance to transfer in DOTs must be greater than or equal to one',
+  BalanceMustBeFinite = 'Balance to transfer in DOTs must not be infinite',
+  BalanceWithDecimalsOnlyWithE = 'Decimals points are only allowed in scientific notation by using an \'e\' (i.e. 3.4e38) or exponential with \'e+\'',
+  BalanceExceedsMaximum = 'Balance exceeds maximum for 128 bit',
```
jacogr: No. Needs to be specific to chain and specific to the number required. (Translatable param strings cater for this)

packages/ui-app/src/util/isValidBalanceExpectedResponses.ts
```diff
@@ -0,0 +1,32 @@
...
+  BalanceExceedsMaximumScientificNotation = 'Balance value after converting from scientific notation exceeds maximum for 128 bit',
+  BalanceExceedsMaximumExponentialNotation = 'Balance value after converting from exponential notation exceeds maximum for 128 bit'
```
jacogr: There are all overly long and Java-ish: They don't all need the `Balance` prefix, duplication end extra words can be omitted, e.g. `Exceeds` add no real value.

packages/ui-app/src/util/isValidBalanceExpectedResponses.ts
```diff
@@ -0,0 +1,32 @@
...
+export const expectedIsValidResponse = (isValid: boolean = false, errorMessage?: ErrorMessage, infoMessage?: string, inputConvertedFromScientificNotation?: string): IsValidWithMessage => {
+  let response: IsValidWithMessage = { isValid: isValid };
+  if (errorMessage) {
+    response.errorMessage = errorMessage;
+  }
+  if (infoMessage) {
+    response.infoMessage = infoMessage;
+  }
+  if (inputConvertedFromScientificNotation) {
+    response.inputConvertedFromScientificNotation = inputConvertedFromScientificNotation;
```
jacogr: As above, rename this overly long var

packages/ui-app/src/util/scientificNotationToNumber.ts
```diff
@@ -0,0 +1,44 @@
...
+export default function scientificToDecimal (num: string): ScientificNotation {
+  if (/\d+\.?\d*e[\+\-]*\d+/i.test(num)) {
+    const zero = '0';
+    const parts = String(num).toLowerCase().split('e'); // split into coeff and exponent
+    let e = parts.pop(); // store the exponential part
+    if (e) {
+      let l = Math.abs(Number(e)); // get the number of zeros
+      if (isNaN(l)) {
+        return {
+          errorMessage: 'Exponential part of scientific notation is invalid',
```
jacogr: Non-translatable.

packages/ui-app/src/util/scientificNotationToNumber.ts
```diff
@@ -0,0 +1,44 @@
...
+import { ScientificNotation } from './types';
+
+// Convert positive number from scientific notation (i.e. scientificToDecimal('2.594e40'))
+// Reference: https://gist.github.com/jiggzson/b5f489af9ad931e3d186
+export default function scientificToDecimal (num: string): ScientificNotation {
+  if (/\d+\.?\d*e[\+\-]*\d+/i.test(num)) {
+    const zero = '0';
+    const parts = String(num).toLowerCase().split('e'); // split into coeff and exponent
+    let e = parts.pop(); // store the exponential part
+    if (e) {
```
jacogr: Blank line before.

packages/ui-app/src/util/scientificNotationToNumber.ts
```diff
@@ -0,0 +1,44 @@
...
+      num = coeffArray.join('') + new Array(l + 1).join(zero);
+
+      // remove preceding 0's (i.e. 0.1e4 produces 1000 instead of 01000)
+      if (num.substring(0, 1) === '0') {
+        num = num.replace(/^0+/g, '');
+      }
+    } else {
+      return {
+        errorMessage: 'Exponential part of scientific notation is missing',
+        inputConvertedFromScientificNotation: undefined
```
jacogr: Overly long field, non-translatable string.

packages/ui-app/src/util/scientificNotationToNumber.ts
```diff
@@ -0,0 +1,44 @@
...
+      };
+    }
+  }
+  return {
+    errorMessage: undefined,
```
jacogr: No need to add the field here if not used. Just leave out instead of saying `undefined`

packages/ui-app/src/Params/Param/Balance.tsx
jacogr: I'm probably even misunderstanding the use here - so make it clear. If it is a value, call it `value`, if it is a bool, call it `is<Something>`

packages/ui-app/src/Params/Param/Balance.tsx
```diff
@@ -0,0 +1,192 @@
...
+    const { onChange, t } = this.props;
+
+    // prepare value by remove all whitespace
+    value = value.split(' ').join('');
+
+    try {
+      const { isValid, errorMessage, infoMessage, inputConvertedFromScientificNotation } = isValidBalance(value);
+
+      this.setState({
+        error: !isValid && errorMessage ? t(errorMessage) : '',
```
jacogr: ???

The way this is done -

`t(<name-for-the-string>, { defaultValue: <default-english-value> })`

That is the only way it will make sense to translate since `name-for-the-string` gives content, where/what and `default-english-value` shows the translator what the original was.

packages/ui-app/src/util/isValidBalance.ts
jacogr: (The above `t(errorMessage)` approach is not correct, it will just create havoc for a translation pass.)

packages/ui-app/src/Params/Param/Balance.tsx
jacogr: EDIT: Technically the above would work, however where the concern comes in - the mapping now matches against the full text, instead of a static key that should not change. If we now make any changes to that string (which is used in translation matching), the already translated text breaks. (So something as simple as a type breaks anything that has been translated for that text.)

Always stick to key + defaultValue everywhere.

packages/ui-app/src/Params/Param/Balance.tsx
ltfschoen: Ok. I can't recall the reason for the "minus 1". It may have been because of errors in the Polkadot Node logs when I tried sending the full account balance, but I'll have to check again once I revive Alice.
Why would we allow them to enter 0 as an input?

packages/ui-app/src/Params/Param/Balance.tsx
ltfschoen: Sorry, I'm not sure what you mean. Would you prefer I change it to say `event.target.value.indexOf('E')`?

packages/ui-app/src/util/isValidBalance.ts
ltfschoen: I've refactored the code so it returns a translated `errorMessage`, but I'm also now returning with it an untranslated `errorMessageUntranslated` so it doesn't interfere with the unit tests as I had difficulty trying to mock the translate function

packages/ui-app/src/util/isValidBalanceExpectedResponses.ts
ltfschoen: I'm only using `ErrorMessage` as a helper for unit testing purposes in isValidBalance.spec.js,  but not on the UI

ltfschoen:
@jacogr I've addressed the review comments except for the following:

1) We should no mention testnet, keep it generic otherwise we need to come back to this. Additionally, we can allow 0 as an input. (Just not negative). Why "minus 1"

  * **RESPONSE** - Ok. I can't recall the reason for the "minus 1". It may have been because of errors in the Polkadot Node logs when I tried sending the full account balance, but I'll have to check again once I revive Alice.
	Why would we allow them to enter 0 as an input?
	Sorry, rephrase... why would we allow them to enter just a value of 0 as an input (i.e. submit a transaction with a Balance value of 0)?

2) inputConvertedFromScientificNotation this is getting very long... something like isScientific would do just as well.
I'm probably even misunderstanding the use here - so make it clear. If it is a value, call it value, if it is a bool, call it is<Something>

  * **RESPONSE** - Renamed to `num`

3) (event.ctrlKey || cmdKey) is re-used, so something like this would work as well -
``(event.ctrlKey || cmdKey) && [aKey, cKey, vKey, xKey].includes(vKey)`
(Once again in the above example, just make sure we have the proper constant arrays defined.

  * **RESPONSE** - I've encountered a bug where it allows user to enter repeatedly the following non-numeric keys still: a, c, v, x multiple times (since I can't seem to only allow those characters to be presseed when in combination with them pressing CMD or CTRL). Any ideas how I can resolve?

4) file-based constant.

  * **RESPONSE** - Sorry, I'm not sure what you mean. Would you prefer I change it to say `event.target.value.indexOf('E')`?

5) Same comments are earlier in the file - between this and all the preceding lines the same issue occurs. (Formatting & non-translatable messages)

  * **RESPONSE** - I've refactored the code so it returns a translated errorMessage, but I'm also now returning with it an untranslated errorMessageUntranslated along with it so it doesn't interfere with the unit tests. I tried to avoid doing this but had difficulty trying to mock the translate function

6) No. All messages needs to be translatable when they are used on the UI. As it stand, these are not.

  * **RESPONSE** - I'm only using `ErrorMessage` as a helper for unit testing purposes in isValidBalance.spec.js,  but not on the UI

jacogr:
1. You are thinking of Balance as purely used in transfers - it is not. It is used when making proposals, starting referendums and elsewhere as a value that gets added to the `reservedBalance` property of the account. So any value (valid for decoding) is possible.

2. Ok.

3. Debugging it is the best option here. Without doing it myself, cannot provide pointers.

4. file based constants, i.e instead of doing

```
function something () {
  // only used for queries, never modified
  const constant = [...];
  // test on constant
}
```

which now gets created as a new variable (and the garbage collected when out of scope) each time the function is called. You would do -

```
const CONSTANT = [...];

function something () {
  // test on CONSTANT
  ...
}
```

5. Ok, will take a look. (Obviously if used in tests, it would be better to have enum values as numbers instead of doing string checks)

6. As above.

ltfschoen:
@jacogr I've restored functionality allowing input of 0. I fixed a bug I found which occurred when a user enters (for scientific or exponential notation purposes) a decimal value that was too low such that its byteLength was over 128. I've fixed the bug so users cannot enter x,c,v,a characters but can still copy/cut/paste/select all. I've added some enzyme simulation tests.

packages/app-extrinsics/src/Extrinsic.tsx
```diff
@@ -82,7 +84,9 @@ class Extrinsic extends React.PureComponent<Props, State> {
           isValid &&
           !isUndefined(values[index]) &&
           !isUndefined(values[index].value) &&
+          (values[index].type === 'Balance' ? checkValueBitLength(values[index].value as BN) : true) &&
```
jacogr: Nested condition, would make sense to make the code clean and split this condition out into `isValidLength` and then just `&& isValidLength`

packages/ui-app/src/Input.tsx
```diff
@@ -25,18 +28,18 @@ type Props = BareProps & {
   label?: any, // node?
   max?: any,
   min?: any,
+  maxLength?: number,
```
jacogr: Alphabetical ordering.

packages/ui-app/src/Params/Param/Balance.tsx
```diff
@@ -0,0 +1,177 @@
...
+          error={error}
+          info={info}
+          isError={isError}
+          label={label}
+          maxLength={defaultMaxLength}
+          onChange={this.onChange}
+          onKeyDown={this.onKeyDown}
+          onKeyUp={this.onKeyUp}
+          placeholder={t('account.balance.placeholder', {
+            defaultValue: 'Between 1 DOT and the available DOT balance of the account'
```
jacogr: Can be 0. (Additionally this now does not match up with the default shown above)

packages/ui-app/src/util/chainSpec.ts
```diff
@@ -0,0 +1,26 @@
+import BN from 'bn.js';
```
jacogr: License header.

packages/ui-app/src/util/isValidBalance.ts
```diff
@@ -0,0 +1,237 @@
...
+  bitLength = bitLength || BIT_LENGTH_128;
+
+  let errorMessageKey: string = '';
+  let errorMessageUntranslated: string = '';
+
+  let warnMessageKey: string = '';
+  let warnMessageUntranslated: string = '';
+
+  // check value is valid number, whether it be scientific notation, exponential (which allow decimal)
+  if (isNaN(parseFloat(input))) {
```
jacogr: This will fail for very large values >64bits. (JS numbers cannot handle it)

packages/ui-app/src/util/isValidBalance.ts
```diff
@@ -0,0 +1,237 @@
...
+      isValid: false,
+      errorMessage: t(errorMessageKey, {
+        defaultValue: errorMessageUntranslated
+      }),
+      errorMessageUntranslated
+    };
+  }
+
+  // always a string from <input type='number'> but leave as failsafe
+  if (!(typeof input === 'string')) {
```
jacogr: Use `isString`

packages/ui-app/src/util/isValidBalance.ts
```diff
@@ -0,0 +1,237 @@
...
+      isValid: false,
+      errorMessage: t(errorMessageKey, {
+        defaultValue: errorMessageUntranslated
+      }),
+      errorMessageUntranslated
+    };
+  }
+
+  // check value is not decimal and greater than or equal to one, whether it be scientific notation, exponential (which allow decimal)
+  if (Number(parseFloat(input)) < 1 && Number(parseFloat(input)) !== 0) {
```
jacogr: We are using BN for a reason - JS cannot handle these. (Until BigInt is available)

packages/ui-app/src/util/isValidBalanceExpectedResponses.ts
```diff
@@ -0,0 +1,57 @@
+// Copyright 2017-2018 @polkadot/ui-app authors & contributors
+// This software may be modified and distributed under the terms
+// of the ISC license. See the LICENSE file for details.
+
+import { IsValidWithMessage } from './types';
+
+import { TranslationFunction } from 'i18next';
+
+export enum ErrorMessage {
+  MustBeStringType = 'Balance input value must be valid type',
```
jacogr: As soon as a file is available that 

(a) is not a test
(b) has naked strings

We end up with future issues.

packages/ui-app/src/util/isValidBalanceExpectedResponses.ts
```diff
@@ -0,0 +1,57 @@
...
+  if (errorMessageUntranslated) {
+    response.errorMessageUntranslated = errorMessageUntranslated;
+  }
+  if (warningMessage) {
+    response.warnMessage = warningMessage;
+  }
+  if (warningMessageUntranslated) {
+    response.warnMessageUntranslated = warningMessageUntranslated;
+  }
+  if (infoMessage) {
```
jacogr: Blank lines to denote blocks - for the whole function.

packages/ui-app/src/util/scientificNotationToNumber.ts
```diff
@@ -0,0 +1,63 @@
...
+
+      if (isNaN(l)) {
+        errorMessageKey = 'conversion.error.notation.scientific.exponent.invalid';
+        errorMessageUntranslated = 'Exponential part of scientific notation is invalid';
+
+        return {
+          errorMessage: t(errorMessageKey, {
+            defaultValue: errorMessageUntranslated
+          }),
+          errorMessageUntranslated: errorMessageUntranslated
```
jacogr: Why? Only always translated strings.

packages/ui-app/src/util/scientificNotationToNumber.ts
```diff
@@ -0,0 +1,63 @@
...
+      }
+    } else {
+      errorMessageKey = 'conversion.error.notation.scientific.exponent.missing';
+      errorMessageUntranslated = 'Exponential part of scientific notation is missing';
+
+      return {
+        errorMessage: t(errorMessageKey, {
+          defaultValue: errorMessageUntranslated
+        }),
+        errorMessageUntranslated: errorMessageUntranslated
```
jacogr: As above.

packages/ui-app/src/util/types.d.ts
```diff
@@ -12,3 +14,19 @@ export type DropdownOption = {
 export type DropdownOptions = Array<DropdownOption>;
 
 export type SectionVisibilityAll = 'private' | 'public';
+
+export type IsValidWithMessage = {
+  isValid: boolean,
+  errorMessage?: TranslationFunction,
+  errorMessageUntranslated?: string,
```
jacogr: No. Only always translated strings everywhere.

ltfschoen:
Progress update:

**Review comments addressed:**
* Refactor nested condition into `isValidLength` instead of `(values[index].type === 'Balance' ? checkValueBitLength(...`
* Fix alphabetical ordering
* Change to `Between 0 DOT` instead of `Between 1 DOT`
* Add License heaer to chainSpec.ts
* Replace `if (!(typeof input === 'string')) ` with `isString`
* Add blank lines for all blocks in isValidBalanceExpectedResponses.ts

**TODO** - still have to address the following comments mentioned above:
  * This will fail for very large values >64bits. (JS numbers cannot handle it): `if (isNaN(parseFloat(input))) {`
  * We are using BN for a reason - JS cannot handle these. (Until BigInt is available) `if (Number(parseFloat(input)) < 1 && Number(parseFloat(input)) !== 0) {`
  * As soon as a file is available that (a) is not a test (b) has naked strings. We end up with future issues. Regarding:
  `MustBeStringType = 'Balance input value must be valid type',`
  * Why? Only always translated strings in ui-app/src/util/scientificNotationToNumber.ts: `errorMessageUntranslated: errorMessageUntranslated`
  * No. Only always translated strings everywhere. Re ui-app/src/util/types.d.ts: `errorMessageUntranslated?: string,`

ltfschoen:
@jacogr I've made simplifying changes to just focus on the issue at hand. I've removed from this PR any support for input of scientific notation (we can address that in a separate issue and leverage lessons learnt from this task)

packages/app-extrinsics/src/Extrinsic.tsx
```diff
@@ -73,6 +75,10 @@ class Extrinsic extends React.PureComponent<Props, State> {
   }
 
   nextState (newState: State): void {
+    const isValidLength = (values: Array<RawParam>, index: number) => {
+      return (values[index].type === 'Balance' ? checkValueBitLength(values[index].value as BN) : true);
```
jacogr: This would read a bit better split,

```
const isValidLength = (values: Array<RawParam>, index: number) =>
      return values[index].type === 'Balance'
          ? checkValueBitLength(values[index].value as BN)
           : true;
```

packages/app-extrinsics/src/Extrinsic.tsx
```diff
@@ -73,6 +75,10 @@ class Extrinsic extends React.PureComponent<Props, State> {
...
+    };
```
jacogr: With that however, the bitlength check should be elsewhere, not specifically here. If we do it at the base, we don't need to add the same logic to each component that uses it.

this component should not care about lengths and being valid, what is passed to it should already have the required checks applied.

packages/app-extrinsics/src/Extrinsic.tsx
```diff
@@ -82,7 +88,9 @@ class Extrinsic extends React.PureComponent<Props, State> {
           isValid &&
           !isUndefined(values[index]) &&
           !isUndefined(values[index].value) &&
+          isValidLength(values, index) &&
```
jacogr: The component would have already done this, i.e. it cannot be `isValid` and then have an invalid length - `isValid` for the type should indicate that the value is valid and passes all checks, including length (if applicable).

packages/ui-app/src/Input.tsx
```diff
@@ -2,41 +2,44 @@
...
   isEditable?: boolean,
   isError?: boolean,
   isHidden?: boolean,
   label?: any, // node?
   max?: any,
+  maxLength?: number,
   min?: any,
   name?: string,
   onChange: (value: string) => void,
+  onKeyDown?: (event: SUIEvent) => void,
```
jacogr: Don't pass the SUIEvent up to components using this - rather deconstruct the value in here (i.e. this component knows it uses SUI), calling components should not.

packages/ui-app/src/Input.tsx
```diff
@@ -2,41 +2,44 @@
...
+  onKeyUp?: (event: SUIEvent) => void,
```
jacogr: As above per the `on*` definition.

packages/ui-app/src/Input.tsx
```diff
@@ -90,11 +96,32 @@ export default class Input extends React.PureComponent<Props, State> {
...
 
   onChange = (event: React.SyntheticEvent<Element>, { value }: SUIEvent): void => {
     this.props.onChange(value);
   }
+
+  onKeyDown = (event: SUIEvent): void => {
+    const { onKeyDown } = this.props;
+
+    if (onKeyDown) {
+      onKeyDown(event);
```
jacogr: Here grab the values from event (i.e. shift state, key pressed) and pass only those up.

packages/ui-app/src/Input.tsx
```diff
@@ -90,11 +96,32 @@ export default class Input extends React.PureComponent<Props, State> {
...
+    }
+  }
+
+  onKeyUp = (event: SUIEvent): void => {
+    const { onKeyUp } = this.props;
+
+    if (onKeyUp) {
+      onKeyUp(event);
```
jacogr: As per the above comment, make the callers not know about SUI.

packages/ui-app/src/InputNumber.tsx
```diff
@@ -0,0 +1,58 @@
...
+type Props = BareProps & {
+  defaultValue?: string,
+  error?: React.ReactNode,
+  info?: React.ReactNode,
+  isError?: boolean,
+  label?: any,
+  maxLength?: number,
+  onChange: (value: string) => void,
+  onKeyDown?: (event: SUIEvent) => void,
+  onKeyUp?: (event: SUIEvent) => void,
```
jacogr: SUIEvent as per previous comments.

packages/ui-app/src/InputNumber.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+type Props = BareProps & {
+  defaultValue?: string,
+  error?: React.ReactNode,
+  info?: React.ReactNode,
+  isError?: boolean,
+  label?: any,
+  maxLength?: number,
+  onChange: (value: string) => void,
+  onKeyDown?: (event: SUIEvent) => void,
```
jacogr: SUIEvent as per previous comments.

packages/ui-app/src/Notification.tsx
```diff
@@ -0,0 +1,52 @@
+// Copyright 2017-2018 @polkadot/ui-app authors & contributors
+// This software may be modified and distributed under the terms
+// of the ISC license. See the LICENSE file for details.
+
+import { BareProps } from './types';
+
+import React from 'react';
+
+import Message from 'semantic-ui-react/dist/commonjs/collections/Message/index';
+import classes from './util/classes';
```
jacogr: Extra blank line before this line, i.e. split external and per-package imports into seperate sections.

packages/ui-app/src/Params/Param/Balance.spec.js
```diff
@@ -0,0 +1,93 @@
+// Copyright 2017-2018 @polkadot/ui-app authors & contributors
+// This software may be modified and distributed under the terms
+// of the ISC license. See the LICENSE file for details.
+
+import React from 'react';
+import { mount } from '../../../../../test/enzyme';
```
jacogr: Import split, blank line before.

packages/ui-app/src/Params/Param/Balance.tsx
```diff
@@ -0,0 +1,158 @@
...
+          error={error}
+          info={info}
+          isError={isError}
+          label={label}
+          maxLength={defaultMaxLength}
+          onChange={this.onChange}
+          onKeyDown={this.onKeyDown}
+          onKeyUp={this.onKeyUp}
+          placeholder={t('account.balance.placeholder', {
+            defaultValue: 'Between 0 DOT and the available DOT balance of the account'
```
jacogr: Remove the DOT ref. "Positive number" would convey the message just as well.

packages/ui-app/src/Params/Param/Balance.tsx
```diff
@@ -0,0 +1,158 @@
...
+          defaultValue: 'Balance to transfer in DOTs must be a number'
+        })
+      });
+      event.preventDefault();
+      return;
+    }
+  }
+
+  onKeyUp = (event: any): void => {
+    if (KEYS_PRE.includes(event.keyCode)) {
```
jacogr: Just pass the keyCode in from the underlying component. (Additionally, we now have an untyped `: any`, which we really should not have unless it is an exceptional circumstance.

packages/ui-app/src/Params/Param/Balance.tsx
```diff
@@ -0,0 +1,158 @@
...
+      onChange({
+        isValid,
+        value: valueBN
+      });
+    } catch (error) {
+      console.error(error);
+    }
+  }
+
+  onKeyDown = (event: any): void => {
```
jacogr: `any`

packages/ui-app/src/Params/Param/Balance.tsx
```diff
@@ -0,0 +1,158 @@
...
+    const { t } = this.props;
+    const { isPreKeyDown } = this.state;
+
+    // only allow user balance input to contain one instance of '.' for decimals.
+    // prevent use of shift key
+    if (
+      (keydown.isDuplicateDecimalPoint(event)) ||
```
jacogr: Rather not. See the comments about deconstructing the event, passing shift & keyCode. (So only the component ever wrapping SUI knows about SUI.) Standard React events are fine, not SUI-specific events.

packages/ui-app/src/types.d.ts
```diff
@@ -25,3 +25,7 @@ export type BaseContext = {
     }
   }
 };
+
+export type SUIEvent = {
+  value: any
```
jacogr: Don't use `any` unless there is no option. Specify or use standard React events. (They are typed)

packages/ui-app/src/util/chainSpec.ts
```diff
@@ -0,0 +1,30 @@
...
+
+import BN from 'bn.js';
+
+import { BIT_LENGTH_128 } from '../constants';
+
+/* chain specification 'latest' is 128 bit and supports max value of 2^128-1, which is ~3.40×10^38
+ * and may be entered as 3.4e38, 3.4e+38 or 340282366920938463463374607431768211455. we may
+ * show it is equivalent to 340282366920938463463374607431768211455 by calling .toString(10)
+ */
+const maxValue = (bitLength?: number): BN => new BN(2).pow(new BN(bitLength || BIT_LENGTH_128)).subn(1);
```
jacogr: Split over 2 lines for readability, i.e.

```
const maxValue = (bitLength: number = BIT_LENGTH_128): BN =>
  new BN(2).pow(new BN(bitLength)).subn(1);

File: packages/ui-app/src/util/chainSpec.ts [hunk 56]
```diff
@@ -0,0 +1,30 @@
...
+
+const maxLength = (maxValue: BN): number => maxValue.toString().length; // returns 39 for 128 bit arg
```
jacogr: Readability imporvements -

```
const maxLength = (maxValue: BN): number =>
  maxValue.toString().length;
```

Additionally, we want to have one less digit, i.e. let's take 16-bit values, i.e. 2^16 = 65536. Your calculation would return 5 digits, however when I enter `70000` it is an invalid number. Err on the side of conservativeness rather. (Which is fine with 64/128/256 bit numbers)

packages/ui-app/src/util/chainSpec.ts
```diff
@@ -0,0 +1,30 @@
...
+
+const defaultMaxLength = maxLength(maxValue());
+
+// fixes error when enter value like 0.1111111111111111111111111111111, which causes
+// bn.js error in toU8a.js of 'byte array longer than desired length'
+const checkValueBitLength = (value: BN, bitLength?: number): boolean => {
```
jacogr: Would read better as

```
const checkValueBitLength = (value: BN, bitLength?: number): boolean =>
  value.bitLength() <= (bitLength || BIT_LENGTH_128);
```

The `<=` already returns a boolean result, no need to check it.

packages/ui-app/src/util/isValidBalance.ts
```diff
@@ -0,0 +1,73 @@
...
+    return {
+      isValid: false,
+      errorMessage: t('balance.error.format', {
+        defaultValue: 'Balance to transfer in DOTs must be a number'
+      })
+    };
+  }
+
+  const maxBN = maxValue(bitLength);
+
```
jacogr: Extra empty line, remove.

packages/ui-app/src/util/isValidBalance.ts
```diff
@@ -0,0 +1,73 @@
...
+    return {
+      isValid: false,
+      errorMessage: t('balance.error.above.max', {
+        defaultValue: 'Balance above max for {{bitLength}} bit',
+        replace: {
+          bitLength: bitLength
+        }
+      })
+    };
+  } else {
```
jacogr: No need for the nested if, this is simply -

```
} else if (inputBN.isZero()) {
...
}
```

packages/ui-app/src/util/isValidBalance.ts
```diff
@@ -0,0 +1,73 @@
...
+      return {
+        isValid: true,
+        warnMessage: t('balance.warn.zero', {
+          defaultValue: 'Balance to transfer 0 DOTs'
+        })
+      };
+    }
+  }
+
+  const infoMessage = Number(input) <= MAX_SAFE_INTEGER ? Number.parseFloat(input).toExponential(2) : undefined;
```
jacogr: Readability -

```
const infoMessage = Number(input) <= MAX_SAFE_INTEGER
  ? Number.parseFloat(input).toExponential(2)
  : undefined;
```

However, we are still using Number when dealing with BN. The whole idea is that we only use `BN` checks, never anything from `Number`.

packages/ui-app/src/util/keyboard.ts
```diff
@@ -0,0 +1,32 @@
...
+import { KEYS } from '../constants';
+
+const regex = {
+  decimalPoint: /[\.]/gi
+};
+
+// obtain current cursor index position of an input field
+const cursorIndexInputField = (event: any) => event.target.value.slice(0, event.target.selectionStart).length;
+
+const isNotDigit = (event: any): boolean => (event.keyCode < 48 || event.keyCode > 57);
```
jacogr: For the whole file, split the definition of the function from the actual body. (See readability improvement comments earlier)

packages/ui-app/src/util/types.d.ts
```diff
@@ -12,3 +14,10 @@ export type DropdownOption = {
 export type DropdownOptions = Array<DropdownOption>;
 
 export type SectionVisibilityAll = 'private' | 'public';
+
+export type IsValidWithMessage = {
+  isValid: boolean,
+  errorMessage?: TranslationFunction,
```
jacogr: Only strings. Not sure why we are padding the definition of `t` in here.

ltfschoen:
Progress Update:

Remaining review comments:

* [ ] - Don't pass the SUIEvent up to components using this - rather deconstruct the value in here (i.e. this component knows it uses SUI), calling components should not.
* [ ] -  As above per the on* definition.
* [ ] -  Here grab the values from event (i.e. shift state, key pressed) and pass only those up.
* [ ] -  As per the above comment, make the callers not know about SUI.
* [ ] - SUIEvent as per previous comments.
* [ ] - SUIEvent as per previous comments.
* [ ] - Just pass the keyCode in from the underlying component. (Additionally, we now have an untyped : any, which we really should not have unless it is an exceptional circumstance.
* [ ] - any
* [ ] - Rather not. See the comments about deconstructing the event, passing shift & keyCode. (So only the component ever wrapping SUI knows about SUI.) Standard React events are fine, not SUI-specific events.
* [ ] - Don't use any unless there is no option. Specify or use standard React events. (They are typed)
* [ ] - Readability imporvements -
const maxLength = (maxValue: BN): number =>
  maxValue.toString().length;
Additionally, we want to have one less digit, i.e. let's take 16-bit values, i.e. 2^16 = 65536. Your calculation would return 5 digits, however when I enter 70000 it is an invalid number. Err on the side of conservativeness rather. (Which is fine with 64/128/256 bit numbers)

packages/ui-app/src/Input.tsx
ltfschoen: I've had to pass up the `event` itself in Input.tsx:
```
    if (onKeyDown) {
      onKeyDown(event);
...
```

I can't pass up destructured properties of event with the following in Input.tsx, because I'd also have to destructure and pass up the `preventDefault` method of `event`, which I haven't been able to do
```
    if (onKeyDown) {
      onKeyDown(key, preventDefault, shiftKey, (target as HTMLInputElement).value);
...
```

ltfschoen:
@jacogr I've pushed commits that address all your review comments. Note that I've replied directly to your comment "Here grab the values from event (i.e. shift state, key pressed) and pass only those up.".

packages/ui-app/src/InputNumber.tsx
```diff
@@ -0,0 +1,58 @@
...
+  maxLength?: number,
+  onChange: (value: string) => void,
+  onKeyDown?: (event: React.KeyboardEvent<Element>) => void,
+  onKeyUp?: (key: string) => void,
+  placeholder?: string,
+  warn?: React.ReactNode,
+  withLabel?: boolean
+};
+
+class InputNumber extends React.PureComponent<Props> {
```
jacogr: Ok, looking this as an "InputNumber". It is just a text input. As a number input, I would expect it to only take numbers. So the logic we have in Balance that does actual numbers should be in here. The only addition Balance should do is to convert to a Param.

1. This component should do number input checks
2. This component should convert to a number (BN in our case)
3. Users of this can further enhance the values

So it is just a move of logic to the base component, i.e. this one.

packages/ui-app/src/Params/Param/Balance.tsx
```diff
@@ -0,0 +1,158 @@
...
+        info: isValid && infoMessage ? infoMessage : '',
+        warn: isValid && warnMessage ? warnMessage : ''
+      });
+
+      if (!onChange) {
+        return;
+      }
+
+      let valueBN: BN;
+      valueBN = new BN(value || '0');
```
jacogr: This type of thing in `InputNumber` as per comment above.

packages/ui-app/src/Params/Param/Balance.tsx
```diff
@@ -0,0 +1,158 @@
...
+      (keydown.isPaste(event.key, isPreKeyDown))
+    ) {
+      return;
+    }
+
+    // prevent input of non-integer values (allow numeric including from keyboards with numpad)
+    if (keydown.isNonNumeric(event.key)) {
+      this.setState({
+        error: t('balance.error.format', {
+          defaultValue: 'Balance to transfer in DOTs must be a positive number'
```
jacogr: Balance does not have anything to do with "transfer" - transfer is just one extrrinsic, a number of others also take Balances as inputs.

ltfschoen:
@jacogr I'm on a plane that's about to take off, but I've pushed a commit that I think addresses what you mean, could you take a look and let me know if i'm on the right track?

packages/ui-app/src/InputNumber.tsx
```diff
@@ -0,0 +1,159 @@
...
+};
+
+class InputNumber extends React.PureComponent<Props, State> {
+  state: State;
+
+  constructor (props: Props) {
+    super(props);
+
+    this.state = {
+      error: '',
```
jacogr: (nitpick) Since these are marked as optional, we actually don't need to assign values to them here.

packages/ui-app/src/InputNumber.tsx
```diff
@@ -0,0 +1,159 @@
...
+      (keydown.isPaste(event.key, isPreKeyDown))
+    ) {
+      return;
+    }
+
+    // prevent input of non-integer values (allow numeric including from keyboards with numpad)
+    if (keydown.isNonNumeric(event.key)) {
+      this.setState({
+        error: t('balance.error.format', {
+          defaultValue: 'Balance in DOTs must be a positive number'
```
jacogr: Can't we just reject the keys outright? As a normal `type='number'` HTML input behaves. (Not a major issue, but not allowing is always better than allowing and then saying "you are doing it wrong")

packages/ui-app/src/InputNumber.tsx
```diff
@@ -0,0 +1,159 @@
...
+import classes from './util/classes';
+import Input from './Input';
+import { KEYS_ALLOWED, KEYS_PRE } from './constants';
+
+type Props = BareProps & {
+  defaultValue?: string,
+  isError?: boolean,
+  label?: any,
+  maxLength?: number,
+  onChange?: RawParam$OnChange | undefined,
```
jacogr: This should just be a `(value: BN) => void` callback. The Param stuff is only for Params. (e.g. see other Inputs, it only returns the value)

packages/ui-app/src/InputNumber.tsx
```diff
@@ -0,0 +1,159 @@
...
+
+      let valueBN: BN;
+      valueBN = new BN(value || '0');
+
+      if (!onChange) {
+        return;
+      }
+
+      onChange({
+        isValid,
```
jacogr: As per earlier comments on the actual onChange type - we should check the value, if valid, callback with _only_ the value, nothing more required.

packages/ui-app/src/InputNumber.tsx
jacogr: EDIT: This will make the `isValid` pass-up not needed as well since we will always have a valid value here.

packages/ui-app/src/constants.ts
```diff
@@ -0,0 +1,42 @@
...
+  V: 'v',
+  X: 'x',
+  ZERO: '0'
+};
+
+const KEYS_ALLOWED: Array<any> = [KEYS.ARROW_LEFT, KEYS.ARROW_RIGHT, KEYS.BACKSPACE, KEYS.DECIMAL_POINT, KEYS.BACKSPACE, KEYS.ESCAPE, KEYS.ENTER, KEYS.TAB, KEYS.ZERO];
+
+const KEYS_PRE: Array<any> = [KEYS.ALT, KEYS.CMD, KEYS.CTRL];
+
+const MAX_SAFE_INTEGER: number = 9007199254740991; // 2**53-1
```
jacogr: Rathe `const MAX_SAFE_INTEGER: number = 2 ** 53-1;` instead of working it out and commenting on it. Additionally, not convinced at all that we need to re-define what JS already gives us - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER

packages/ui-app/src/util/chainSpec.ts
```diff
@@ -0,0 +1,39 @@
...
+
+/* max length in digits the user may enter in input field is reduced by conservativeness factor.
+ * i.e. 16 bit largest allowable number in decimal is 2**16-1 (65535) having 5 digits.
+ * if user is allowed to enter 5 digits and they enter >65535 then an error notification appears.
+ * but we only allow user to enter 4 digits (i.e. max 9999) when conservativeness factor is set to 1,
+ * so with the conservativeness factor set they are should not encounter the error notification.
+ */
+const maxLength = (maxValue: BN): number => {
+  const conservativenessFactor = 1;
+  return maxValue.toString().length - conservativenessFactor; // returns 38 for 128 bit arg
```
jacogr: (nitpick) blank line before this line.

packages/ui-app/src/util/isValidBalance.ts
```diff
@@ -0,0 +1,90 @@
...
+  } else if (inputBN.isZero()) {
+    return {
+      isValid: true,
+      warnMessage: t('balance.warn.zero', {
+        defaultValue: 'Balance to transfer 0 DOTs'
+      })
+    };
+  }
+
+  const maxSafeIntegerBN = new BN(String(MAX_SAFE_INTEGER));
```
jacogr: Just simply `new BN(Number.MAX_SAFE_INTEGER)` - no coercions, no re-definition of built-in values.

packages/ui-app/src/util/keyboard/isNonNumeric.ts
```diff
@@ -0,0 +1,8 @@
+// Copyright 2017-2018 @polkadot/ui-app authors & contributors
+// This software may be modified and distributed under the terms
+// of the ISC license. See the LICENSE file for details.
+
+export default function isNonNumeric (key: string): boolean {
+  const digits = new Array(9).join().split(',').map((item, index) => String(++index));
+  return !digits.includes(key);
```
jacogr: (nitpick) blank line before

packages/ui-app/src/util/keyboard/isDuplicateDecimalPoint.ts
```diff
@@ -0,0 +1,14 @@
...
+
+import { KEYS } from '../../constants';
+
+const regexDecimalPoint = /[\.]/gi;
+
+export default function isDuplicateDecimalPoint (key: string, value: string): boolean {
+  const inputValue: string = value;
+  const didPressDecimalPoint: boolean = key === KEYS.DECIMAL_POINT;
+  const foundExistingDecimalPoint: boolean = inputValue.match(regexDecimalPoint) ? true : false;
+  return didPressDecimalPoint && foundExistingDecimalPoint;
```
jacogr: (nitpick) blank line before

packages/ui-app/src/Params/Param/Balance.tsx
```diff
@@ -0,0 +1,52 @@
...
+          placeholder={t('account.balance.placeholder', {
+            defaultValue: 'Positive number'
+          })}
+          t={t}
+          withLabel={withLabel}
+        />
+      </Bare>
+    );
+  }
+}
```
jacogr: Great. Now, this component should make it a valid RawParam, i.e. take the BN and add the isValid stuff as required. (Which in most cases should be just always `true`, i.e. what we receive from the InputNumber should be a valid number based on the logic there)

packages/ui-app/src/InputNumber.tsx
jacogr: Additionally, This is a number Input, not a Balance input - so a number can be anything, not just a DOT. (Also remove the DOT references everywhere, for substrate chains they have their own tokens)

packages/ui-app/src/util/isValidBalance.spec.js
```diff
@@ -0,0 +1,23 @@
+// Copyright 2017-2018 @polkadot/ui-app authors & contributors
+// This software may be modified and distributed under the terms
+// of the ISC license. See the LICENSE file for details.
+
+import { BIT_LENGTH_128 } from '../constants';
+import isValidBalance from './isValidBalance';
+import translate from '../translate';
```
jacogr: (nitpick) move translate one up, so external to folder first, then internal to folder.

packages/ui-app/src/InputNumber.tsx
ltfschoen: I've had to change it to `onChange?: (value: { isValid: boolean, value: BN }) => void,` instead of `onChange?: (value: RawParam$OnChange$Value) => void,` (that way we're not using Param types), and because the onChange that is passed as a prop to InputNumber from the Balance component comes from BareProps and is of type `RawParam$OnChange | undefined`, and the `onChange` method that we call on InputNumber returns `{ isValid, value: valueBN }`

ltfschoen:
@jacogr I've tried to address all the review comments. I've left it passing translate down as props from Balance to InputNumber because otherwise I had difficulty getting the unit tests to remain working (both when I wrapped InputNumber with translate, and when I tried creating an InputNumber.test.js and modified version of the same unit tests on that component instead)

packages/ui-app/src/InputNumber.tsx
```diff
@@ -0,0 +1,151 @@
...
+  }
+
+  onKeyUp = (key: string): void => {
+    if (KEYS_PRE.includes(key)) {
+      this.setState({ isPreKeyDown: false });
+    }
+  }
+}
+
+export default InputNumber;
```
jacogr: I see the testing comment, however when we specify `t` the component needs to be wrapped in a translate. We cannot have components using this pass through `t`. So it _needs_ to be wrapped. (The testing setup is incorrect if it is not working, i.e. i18next is not being initialised)

packages/ui-app/src/InputNumber.tsx
jacogr: This is still outstanding. `onChange` should be - `onChange: (value: BN) => void`, not extra `inValid` is should be what it says on the tin, i.e. a number input.

ltfschoen:
@jacogr I've overhauled the code and tried to make it more organised.

I've addressed both your review comments (i.e. only passing the value to `onChange` without isValid by ensuring I make the value valid in the InputNumber component itself, and I'm no longer passing translate down from Balance, I'm just injecting the observable directly into InputNumber and moved the tests into InputNumber and got them working).
I've written a detailed list of other changes in this commit, where most of the work was done https://github.com/polkadot-js/apps/pull/221/commits/ba7619b2a0109dc4c89dab1c0606fb41a7c0cbc6

jacogr:
Ok, As mentioned above. Can be please trim this down and get rid of unneeded stuff that makes it flashy without real benefit -

1. I'm not sure why this is even here. This is the first thing that my non-techy family and friends will ask about. This adds no value at all to the users.

polkadot apps portal 2018-09-19 14-08-45

2. Not sure why this is here - in this case you can still submit, the warning just looks out of place.

polkadot apps portal 2018-09-19 14-09-10

3. Since actual transfers will be done in the transfer app, we need to get it in there -

polkadot apps portal 2018-09-19 14-09-47

4. I'm struggling to see where we are actually now doing errors in the Balance input - it filters everything non-valid.

5. Move all the different files that are only used in the number input into functions inside that input. Then we also get rid of calling external functions (like isNumberValid) with t as a param. Keep it small, keep it together.

ltfschoen:
Progress update:

* ASSUMPTION: InputNumber may be used in future for useful error, warning, and info messages, so leave the boilerplate props along with Notification.

* ASSUMPTION: Displaying an info message in scientific notation i.e. 3e+7
causes more confusion than benefit since most users are non-techy. In future this when we have Issue #86 Settings App, it could be added as an optional extra that could be toggled by the user along with Issue #224 E-Notation
	* [x] - Remove display of scientific notation equivalent
	* [x] - Remove warning 'number.warn.zero' when user enters 0 value
	* [x] - Remove info notification 'number.info.scientific' when user enters valid value
	* [x] - Remove `toScientificNotation` function, which is incorrectly passed the TranslationFunction `t` anyway.
	* [x] - Remove `IsValidWithMessage` type and replace return value with just a boolean of whether its valid or not
	* [x] - Move `isValidNumber` into React componet, and do not pass the TranslationFunctio to it!
	* [x] - Move `isValidKey` into React componet, and do not pass the TranslationFunctio to it!
	* [x] - Move `isNonDecimal` into React componet, and do not pass the TranslationFunctio to it!
	* [X] - Remove `isNonInteger` and isNonInteger.ts and isNonInteger.spec.js since no longer using
	* [X] - Remove `error`, `warn`, `info` from InputNumber state, since we're filtering out everything not valid. Instead add them as optional props (so Balance or another component can display notifications if required)
	* [ ] - Move tests for isValidKey into React component, and remove isValidKey.spec.js
	* [ ] - Move tests for isValidNumber into React component, and remove isValidNumber.spec.js
	* [ ] - Move tests for isNonDecimal into React component, and remove isNonDecimal.spec.js
	* [ ] - Update InputNumber.spec.js to remove useless tests that check that the Notification component isn't shown
	* [ ] - Integrate into the Transfer App `send a value of` where actual transfers will be done (see @jacogr comment and screenshot)
	* [ ] - Move all the different files that are only used in the number input into functions inside that input. Then we also get rid of calling external functions (like isNumberValid) with t as a param. Keep it small, keep it together.
	* [ ] - Make methods `private` as relevat
	* [ ] - Figure out what action to take yet based on @jacogr's comment: "I'm struggling to see where we are actually now doing errors in the Balance input - it filters everything non-valid.". 
	Maybe the Balance isValid needs an isBalanceValid function that checks if the user has sufficient balance and passes error as props to InputNumber?
	* [ ] - Figure out what action to take yet based on @jacogr's comment:  "Seems SUI docs are currently broken. But in the cases where you absolutely need to point to an input with an attachment, their Label does exactly that and allows attachment with a pointer to an Input. So at most Notification should just warp that. And it really should only be used is completely exceptionally circumstances. (And I really couldn't see any such circumstances from playing with the actual inputs. So at this point seems like a good, although unused in practice idea.)"

jacogr:
Second assumption is wrong - it is about allowing the "input of scientific notation number", not the display of them on an otherwise clean input. Basically, the user (in the future), will be able to input either

`1000000` or `1e6` and both would be accepted.

packages/ui-app/src/InputNumber.tsx
```diff
@@ -0,0 +1,188 @@
...
+
+import classes from './util/classes';
+import Input, { KEYS, KEYS_PRE, isCopy, isCut, isPaste, isSelectAll } from './Input';
+import translate from './translate';
+
+type Props = BareProps & I18nProps & {
+  defaultValue?: string,
+  isError?: boolean,
+  label?: any,
+  maxLength?: number,
```
jacogr: I believe we need to pass in bitLength here instead of maxLength. Additionally, put that at 32-bits default (i.e. which is just "normal" numbers, for Balance and friends we will pass through the correct length as required.

packages/ui-app/src/InputNumber.tsx
```diff
@@ -0,0 +1,188 @@
...
+
+    return (
+      <div
+        className={classes('ui--InputNumber', className)}
+        style={style}
+      >
+        <Input
+          {...this.props}
+          defaultValue={defaultValue || '0'}
+          maxLength={maxLength || this.maxLength()}
```
jacogr: Here maxLength is fine, however we use the bitLength as passed in to calculate to pass to Input. So InpoutNumber works with bitLength, the lower-level can work with character lengths.

packages/ui-app/src/InputNumber.tsx
```diff
@@ -0,0 +1,188 @@
...
+        />
+      </div>
+    );
+  }
+
+  private maxValue = (bitLength?: number): BN =>
+    new BN(2).pow(new BN(bitLength || BIT_LENGTH_128)).subn(1)
+
+  private maxLength = (): number => {
+    const conservativenessFactor = 1;
```
jacogr: You can put this one file-level since it never changes. (like you have the default bitlength)

packages/ui-app/src/Params/Param/Balance.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+class Balance extends React.PureComponent<Props> {
+  render () {
+    const { defaultValue: { value } } = this.props;
+    const defaultValue = new BN(value as BN || '0').toString(10);
+
+    return (
+      <Bare {...this.props}>
+        <InputNumber
+          {...this.props}
```
jacogr: Not sure why we are passing the same props to both Bare and InputNumber - rather be explicit here as to what goes where.

packages/ui-app/src/InputNumber.tsx
```diff
@@ -0,0 +1,191 @@
...
+  withLabel?: boolean
+};
+
+type State = {
+  isPreKeyDown: boolean,
+  isValid: boolean,
+  previousValue: string
+};
+
+const BIT_LENGTH_32 = 32; // normal numbers
```
jacogr: Any reason we can't just use 32, 128 where we pass it in?

Elsewhere in the API (encoding/decoding), have actually defined this type of thing in the following manner:

```
type BitLength = 32 | 64 | 128 | 256;

// default here
function foo (bitLength?: BitLength = 128) {
 ...
}

foo(32); // works
foo(64); // works
foo(121); // does not work, TS checks shows an error
```

ltfschoen: @jacogr I've pushed a commit with these changes.

I also created a TS `enum` in enum.ts (since enums don't go in .d.ts files) and used it as shown below, but it wouldn't recognise it in the unit tests

```
// enum.ts
export declare const enum BitLengthOption {
  CHAIN_SPEC = 128,
  NORMAL_NUMBERS = 32
}

// InputNumber.tsx
const DEFAULT_BITLENGTH: BitLength = BitLengthOption.NORMAL_NUMBERS;

// Transfer.tsx / Balance.tsx
const DEFAULT_BITLENGTH: BitLength = BitLengthOption.CHAIN_SPEC;
```

So I reverted to just using key/value pairs, which worked and passes the unit tests. But I wasn't sure where to put it (i.e. if I put it in Balance.tsx, then it would be strange to import it and use BitLengthOption in Transfer.tsx where we are importing InputNumber directly...), so I've just left it in enum.ts for the moment.

```
// enum.ts
export const BitLengthOption = {
  CHAIN_SPEC: 128 as BitLength,
  NORMAL_NUMBERS: 32 as BitLength
};
```

I like doing this because I find it more readable (and avoids me leaving a comments like `// 32 is normal number, 128 is chain spec` above the type definition of BitLength for future reference, but I'm also happy to remove it

ltfschoen:
@jacogr I've pushed a commit with these changes.

I also created a TypeScript `enum` in enum.ts (since enums don't go in .d.ts files, but I wasn't sure where would be an appropriate place for it). I wasn't sure where to put it because if I put it in Balance.tsx, then it would be strange to import it and use BitLengthOption in Transfer.tsx where we are importing InputNumber directly (instead of Balance), so I've just left it in enum.ts for the moment until I get your feedback, and whether to use the enum at all

I like using the `enum` here because I find it more readable (and avoids me leaving a comments like `// 32 is normal number, 128 is chain spec` above the type definition of BitLength for future reference, or writing long variable names, but I'm also happy to remove it

ltfschoen:
@jacogr I've pushed a commit addressing your review comments to use `BitLength` type.

I also created a TypeScript `enum` named `BitLengthOptions` in constants.ts so we can give meaningful names for their usage. I put it in constants.ts because I wasn't sure where to put it because if I put it in Balance.tsx, then it would be strange to import it and use BitLengthOption in Transfer.tsx where we are importing InputNumber directly (instead of Balance), so I've just left it in constants.ts for the moment until I get your feedback, and whether to use the enum at all

I like using the `enum` here because it augments the use of constants to make it more readable (and avoids me leaving a comments like `// 32 is normal number, 128 is chain spec` above the type definition of BitLength for future reference, or writing long variable names, but I'm also happy to remove it

ltfschoen:
@jacogr I've pushed an additional commit that I believe solves the problem of preventing the user from entering a value that's preceded with 0, which is SOLUTION OPTION 3 below (where I've also included other options I considered. 

* RESTRICTED user actions
  * user should not be able to enter a value that displays multiple 0's (i.e. 00), even though a display value of '00' would be submitted as '0', on the basis that allowing it looks unprofessional
    * currently we restrict the user from entering the key 0 when the existing value is 0
  * user should not be allowed to enter or paste a value that is preceeded by 0 (i.e. '01', or '01234'), because this display value would actually be submitted as '1' or '1234' (this could be dangerous because a user may think 01234 would be interpreted as just 0, and actually send/transfer more value than they anticipated!)
    * OPTION 1 - revert to the previous value if the first digit of the new value in the input field is '0' (after they enter a key on the RHS of the previous value of '0', or they paste a value after a previous value of '0', or they enter a key or paste a value that replaces the whole value with a value that is preceded with '0')
      * Cons - when implemented in onKeyDown it produces an ugly UI, it allows the user to enter the wrong values and displays them, and only reverts the changes after they keyup
      * Verdict - IGNORE
    * OPTION 2 - prevent the user from pasting a value if their cursor position is on the RHS of a '0' that is the first character of the input value
      * Cons - confuses the user and they may think they can't paste at all, not realising they can paste on the LHS
      * Verdict - IGNORE
    * **OPTION 3** - allow the user to enter a value or paste, but strip all the preceding '0's from the new value, which can be achieved in keydown. and as the user may have entered or pasted a value of '0' so it became '00' we need to ensure that we're not stripping the value away entirely, so only strip if the new value isn't still a zero value
      * **SOLUTION**:
        * **implemented in onKeyUp**

```
/* if new value equates to '0' in BN when but it's length is >=1 (i.e. '012', '00', etc)
 * then replace the input value with just '0'.
 * otherwise remove the preceding zeros from the new value (i.e. '0123' -> '123')
 * note: edge case glitch occurs if existing value is '0' and you 'hold down' and keep
 * pasting a value of '00' after it, then sometimes when you let go the
 * remaining value shown as '000' or '00000' in the UI, but it's still ok because
 * the actual BN if the user submitted would still be '0', and if they then press any key
 * the UI input value resets to '0'
 */
 (event.target as HTMLInputElement).value = (isNewValueZero && !!newValue.length) ? '0' : newValue.replace(/^0+/, '');
```
        * **implemented in onKeyDown**

This solution is augmented by in onKeyDown which prevents the user from entering the '0' key if the digit in the first index position of the input field is already '0'

```
// prevents entry of zero if initial digit is zero
const isDuplicateZero = previousValue[0] === '0' && event.key === KEYS.ZERO;

if (isDuplicateZero) {
  return false;
}
```


* PERMITTED user actions
  * user should be able to change the value from '0' to '', because even though it displays '', if they press submit then '0' will be submitted (so it's not necesssary to propagate isValid up from InputNumber via onChange to Balance or Transfer. assumes most users know that if a submission button is disabled it means an input is invalid, so to avoid confusion they just need to follow the instruction of entering a 'Positive Number'
    * otherwise we'd have to either:
      * prevent them deleting the '0', which would be horrible as they'd have to move their cursor to index 0 before entering any value
        * unless we automatically removed the '0' value and replaced it with the value they entered (which may confuse some users)

packages/ui-app/src/types.d.ts
```diff
@@ -25,3 +25,5 @@ export type BaseContext = {
     }
   }
 };
+
+export type BitLength = 32 | 64 | 128 | 256;
```
ltfschoen: @jacogr Should this be `8 | 16 | 32 | 64 | 128 | 256` so its the same as https://github.com/polkadot-js/api/commit/60d3d6396a8e48626c74376e5cba0bdd41f17110#diff-0dce4593d48039d48c497cf070a879cfR16

jacogr: I think (and we are going through this in the encoding atm), that all numbers are (currently) 32 (most indexes, i.e. Nonce, PropIndex, ReferendumIndex, ...) or 64 (BlockNumber, ...) or 128 (Balance). We actually don't have any values shuttled around that don't fit here.

For future-proofing, the 8 | 16 | ... is probably better. (Although it can be added as the need arrises.)

jacogr: TL;DR - We can adpat either way, but happy as it is, and won't complain if it is changed either.

Issue #245, opened 2018-08-21
Consistent styling, theme & colours throughout


jacogr:
Basic SideBar styling introduced in https://github.com/polkadot-js/apps/pull/243. Extend this with a common style across the board, at this moment the actual apps feels somewhat misplaced with the default blue + blue styling.

jacogr:
"help wanted" - need somebody with a good eye for detail and styling to help on what works as what doesn't.

Tbaut:
What about something very simple with the same pink and black?
I'd be happy to help.
![image](https://user-images.githubusercontent.com/33178835/44406541-2fa95180-a55c-11e8-8370-88f51f3bc470.png)

jacogr:
PRs are very welcome. I did play a bit with the dark grey buttons and it almost felt a bit too much and almost too dull since the text is exactly the same color. And pink is a highlight color, so should probably not be overused, i.e as links in my view it is too much, it may work as a primary button color, but will need to see what it actually does on-screen.

So what needs to be looked at -

- primary buttons (the current blue ones)
- negative buttons (the current red ones)
- vote (negative & positive) buttons (the current green & red ones)
- links
- not to be forgotten is the switch buttons (i.e. top of accounts, addresses, etc)

Overall it is not too involved, couple of styles to override `.primary`, `.secondary` (if needed), `.negative`, `.positive` & `a`

At this exact point I cannot quite spend time on it myself, the time vs outcome reward is just not there.

Tbaut:
I'll give it a try and see what come out.

PR #259, opened 2018-08-24
Use new colors

Tbaut:
- Fix https://github.com/polkadot-js/apps/issues/245
- Use of the following colors https://coolors.co/4e4e4e-d75ea1-ffffff-96bdc6-98aaae
- For buttons, progress bar, anchors, staking star

image
image

image

image

### DIFF
--- a/packages/app-staking/src/index.css
+++ b/packages/app-staking/src/index.css
@@ -76,7 +76,7 @@
 }
 
 .staking--Account-validating.isValidator {
-  color: red;
+  color: #d75ea1;
   opacity: 1 !important;
 }
 
--- a/packages/ui-app/src/styles/partials/_variables.css
+++ b/packages/ui-app/src/styles/partials/_variables.css
@@ -1,1 +1,16 @@
 $small-corner: .28571429rem;
+
+$green: #77c689;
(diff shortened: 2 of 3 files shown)


Tbaut:
I thought about changing the colors of the lateness progress bar but didn't end up doing it because of the natural human understanding of green, orange, red. The colors should probably be changed for "less loud" ones though.

Tbaut:
- made the anchors a little darker
- toned down the progress bar red, orange, green
- used variables (I used the names from https://coolors.co/ not sure that makes sense..)

selection_161
selection_160selection_159selection_158

jacogr:
Nice!

Tbaut:
Hehe the reason why I chose the name of the colors rather than a descriptive name is because we may want to use the same color for different assets. I'll opt for the go crazy option that solves it then and make the anchor a little darker :)

Tbaut:
BTW I don't have the rights to merge, so feel free to do it :)

jacogr:
Cool. Best is to get it out there and let's see - I think we will obviously need some tweaks, but happy as it stands. So let's see what falls out of the tree when we shake it.

Issue #260, opened 2018-08-24
Cannot use return/enter to submit extrinsics 


phillux:
Upon entering my password, it would be nice to be able to hit return to submit the extrinsic.

It would also improve the UI to make the 'cancel' button outlined and the 'submit' button heavy, so that the user can easily see which button is the default when pressing return.

![plk1](https://user-images.githubusercontent.com/6450379/44576218-47681c00-a78e-11e8-8e47-f08e8aa9153d.png)

jacogr:
Additionally - the password needs to be auto-focussed when the modal opens.

PR #265, opened 2018-08-26
feat: Fixes Issue #260. Allow use Enter to submit extrinsics

ltfschoen:
* Autofocus input field when modal opens
* Tabbing after entering password focusses on Submit where user can press enter to submit
* Pressing Enter whilst password field still focussed clicks the Submit Transaction button


ltfschoen:
Note that I had issues using `setState` in the tests, so I've skipped that test and created an issue here: https://github.com/airbnb/enzyme/issues/1794

packages/ui-app/src/Button/types.d.ts
```diff
@@ -18,6 +18,7 @@ export type ButtonProps = BareProps & {
   isPrimary?: boolean,
   onClick?: () => void | Promise<void>,
   size?: Button$Sizes,
+  tabIndex?: number | string,
```
jacogr: Only `number`?

packages/ui-app/src/Input.tsx
```diff
@@ -27,16 +28,14 @@ type Props = BareProps & {
   min?: any,
   name?: string,
   onChange: (value: string) => void,
+  onKeyDown?: (event: SUIEvent) => void,
   placeholder?: string,
+  tabIndex?: number | string,
```
jacogr: Only `number`?

packages/ui-app/src/Input.tsx
```diff
@@ -97,4 +99,11 @@ export default class Input extends React.PureComponent<Props, State> {
   onChange = (event: React.SyntheticEvent<Element>, { value }: SUIEvent): void => {
     this.props.onChange(value);
   }
+
+  onKeyDown = (event: SUIEvent): void => {
+    const { onKeyDown } = this.props;
+    if (onKeyDown) {
```
jacogr: Extra blank line before.

packages/ui-app/src/Password.tsx
```diff
@@ -2,21 +2,24 @@
...
+  autoFocus?: boolean,
   defaultValue?: any,
   children?: React.ReactNode,
   isDisabled?: boolean,
   isError?: boolean,
   label?: string,
   name?: string,
   onChange: (value: string) => void,
+  onKeyDown?: (event: SUIEvent) => void,
+  tabIndex?: number | string,
```
jacogr: Only `number`?

packages/ui-app/src/Password.tsx
```diff
@@ -61,6 +67,7 @@ export default class Password extends React.PureComponent<Props, State> {
           }
           isPrimary
           onClick={this.onToggleVisible}
+          tabIndex='-1'
```
jacogr: `={-1}` or rather don't specify at all

packages/ui-signer/src/Modal.tsx
```diff
@@ -106,6 +106,7 @@ class Signer extends React.PureComponent<Props, State> {
           <Button
             isNegative
             onClick={this.onCancel}
+            tabIndex='3'
```
jacogr: `={3}`

packages/ui-signer/src/Modal.tsx
```diff
@@ -114,6 +115,7 @@ class Signer extends React.PureComponent<Props, State> {
           <Button
             isPrimary
             onClick={this.onSend}
+            tabIndex='2'
```
jacogr: `={2}`

packages/ui-signer/src/Modal.tsx
```diff
@@ -153,10 +155,13 @@ class Signer extends React.PureComponent<Props, State> {
 
     return (
       <Unlock
+        autoFocus
         error={unlockError && t(unlockError.key, unlockError.value)}
         onChange={this.onChangePassword}
+        onKeyDown={this.onKeyDown}
         password={password}
         value={currentItem.publicKey}
+        tabIndex='1'
```
jacogr: `={1}`

packages/ui-signer/src/Modal.tsx
```diff
@@ -189,6 +194,31 @@ class Signer extends React.PureComponent<Props, State> {
     });
   }
 
+  onKeyDown = (event: any): void => {
+    let buttonEls: any = [];
+    let inputList: any = [];
+    let buttonFilteredEls: any = [];
+
+    if (event.keyCode === 13) {
+      buttonEls = document.getElementsByTagName('button');
```
jacogr: We really, really should not be mixing DOM access with React. The React-way is to use references.

packages/ui-signer/src/Unlock.tsx
```diff
@@ -13,9 +13,12 @@ import keyring from '@polkadot/ui-keyring/index';
 import translate from './translate';
 
 type Props = I18nProps & {
+  autoFocus?: boolean,
   error?: string,
   onChange: (password: string) => void,
+  onKeyDown?: (event: SUIEvent) => void,
   password: string,
+  tabIndex?: number | string,
```
jacogr: `number`

packages/ui-app/src/Button/types.d.ts
ltfschoen: If I change it from `tabIndex?: number | string` to `tabIndex?: number`, it causes linting errors.

```
tabIndex?: number
         ~~~~~~~~
    The expected type comes from property 'tabIndex' which is declared here on type 'IntrinsicAttributes & BareProps & { children?: ReactNode; floated?: "left" | "right" | undefined; icon?: string | undefined; isBasic?: boolean | undefined; isCircular?: boolean | undefined; ... 7 more ...; tabIndex?: number | undefined; } & { ...; }'
```

Because in Input.tsx https://github.com/polkadot-js/apps/blob/master/packages/ui-app/src/Input.tsx#L8 it uses `import SUIInput from 'semantic-ui-react/dist/commonjs/elements/Input/Input';`, where it's defined as 

```
  /** An Input can receive focus. */
  tabIndex?: number | string
```

Here's the line of code in their latest version, which still allows number or string https://github.com/Semantic-Org/Semantic-UI-React/blob/master/src/elements/Input/Input.js#L91

ltfschoen:
@jacogr I've pushed changes as follows:
* Replace use of DOM interaction with use of Refs (i.e. `React.createRef()` as specified here https://reactjs.org/docs/refs-and-the-dom.html instead of `document.getXyz`).
* Replaced tabIndex prop values with number types (i.e. replace `tabIndex='1'` with `={1}`)
* Added missing newline

I've replied to your comment requesting me to change `tabIndex?: number | string,` to `tabIndex?: number`, as doing so causes linting errors.

I've updated the unit tests but I'm skipping the ones that are failing due to not being able to `setState` with Enzyme until this issue raised at the enzyme repo is resolved https://github.com/airbnb/enzyme/issues/1794. Alternative may be to try just using ReactTestUtils instead of Enzyme.

ltfschoen:
@jacogr I think the linting errors were because I hadn't rebuilt with `yarn run build` before running `yarn run check`. I've made the change from `string | number` -> `number` and linting passes

ltfschoen:
@jacogr I've created a branch luke-260-cannot-use-return-enter-to-submit-extrinsics-wip, with two commits https://github.com/polkadot-js/apps/commit/412bfbed811fe4c823ef792f84e234c5f467ae76, and https://github.com/polkadot-js/apps/commit/3e6d02d49b94806f1de634043448f7b55ea08a8a.
So far I've just tried not to expose SUIEvent type in base components (i.e. where it's exposed in https://github.com/polkadot-js/apps/blob/master/packages/ui-app/src/Dropdown.tsx#L28, and https://github.com/polkadot-js/apps/pull/265#discussion-diff-214608444R29).
I've been resolving linting errors for a while now, and now sure how to overcome the latest ones. I also think I must be taking the wrong approach and overcomplicating things, and probably should have asked for help earlier, but I thought my approach made sense at the time. I've been passing down both the event and the data as props even when only the event data is required, but I intended to remove after refactoring if I got it to work. I also wondered whether I should be extending the type `RawParam$OnChange` used for the onChange property of BaseProps in ui-app/src/Params/types.d.ts instead of extending the React event with type InputOnChangeEvent and InputOnChangeEventData in ui-app/src/types.d.ts.

ltfschoen:
TODO - considered these comments https://github.com/polkadot-js/apps/pull/221

ltfschoen:
@jacogr I've pushed commits addressing all the review comments. I've updated the tests but still skipping them as I haven't managed to get enzyme to work (see https://github.com/airbnb/enzyme/issues/1794)

packages/ui-signer/src/Modal.tsx
```diff
@@ -189,6 +201,21 @@ class Signer extends React.PureComponent<Props, State> {
...
 
+  onKeyDown = (event: React.KeyboardEvent<Element>): void => {
+    if (event.keyCode === 13) {
+      const node = this.submitButtonDiv.current;
+
+      if (node instanceof HTMLElement) {
+        const child = node.firstChild;
+
+        if (child instanceof HTMLElement) {
+          child.click();
```
jacogr: Would the simpler solution rather to just do "onSend" here? Instead of 

1. making a reference
2. clicking that?

jacogr: EDIT: I just typed this up elsewhere to explain a bit better.

Instead of -

1. get ref to component and storing it
2. checking for char 13
3. creating an event on that component
4. the event calling the associated handler

We can just do -

1. check for char 13
2. directly call handler

ltfschoen: Absolutely!

ltfschoen:
@jacogr I've addressed your comment.

I also tried to simulate pressing of the Enter key on the password input element, but I only got as far as detecting the Unlock component if I modified the Unlock component so it wasn't wrapped in translate and instead passed `t` down to it via props. But even with the Unlock component I was not able to click the input field. Below is a snippet of the code I tried doing it with

```
...

import Unlock from './Unlock';

...

describe('Signer', () => {
  ...

  it('renders Unlock component with empty password', () => {
    const spy = jest.spyOn(wrapper.instance(), 'onSend');
    wrapper.instance().forceUpdate();
    expect(spy).not.toHaveBeenCalled();

    return new Promise((resolve) => {
      wrapper.setState({
        currentItem: fixtureCurrentItemState,
        password: '123',
        unlockError: null
      }, () => {
        wrapper.setProps({
          queue: fixtureQueueProp
        }, resolve);
      });
    }).then(() => {
      wrapper.update();
      expect(wrapper.state('currentItem')).toEqual(expectedNextCurrentItemState);
      expect(wrapper.find('.ui--signer-Signer')).toHaveLength(1);
      expect(wrapper.find(Unlock).props().password).toEqual('');
      const enterKey = 'Enter';
      wrapper.simulate('keyDown', { key: enterKey });
    }).then(() => {
      wrapper.update();
      expect(spy).toHaveBeenCalled();
      spy.mockRestore();
    });
  });
});
```

chevdor:
Do you plan on supporting ESC as well ?

jacogr:
@ltfschoen Are you happy with the tests as they are (i.e. issues remaining), or need some more time?

ltfschoen:
@jacogr Yes, happy with tests as they are, unfortunately they're not protecting the code as I'd like them to but it's something I'll figure out how to do soon enough

jacogr:
@chevdor Please log that one, I don't believe we have it, neither have I heard that request. (And it is probably wider, i.e. needs to apply to all Modals. However I think this one needs to stay focussed to the task as hand, i.e. the specific feature as requested)

PR #273, opened 2018-09-02
WIP 257 extrinsics no accounts

marnibrewster:
WIP for showing users that they can't execute an extrinsic until there are accounts present. New button links to Accounts page so that users can add an account. 

Not sure if this is the correct use of <Link> for this project. 

Open to all other input - I'm newish to React, blockchain, semanticui, 
and obviously Polkadot! The copy is too wordy, maybe a note explaining that they need accounts would be better. 

screen shot 2018-09-02 at 11 14 56 am

### DIFF
--- a/packages/app-extrinsics/src/index.tsx
+++ b/packages/app-extrinsics/src/index.tsx
@@ -1,8 +1,8 @@
 // Copyright 2017-2018 @polkadot/app-extrinsics authors & contributors
 // This software may be modified and distributed under the terms
 // of the ISC license. See the LICENSE file for details.
-import { BareProps } from '@polkadot/ui-app/types';
 import { QueueProps } from '@polkadot/ui-signer/types';
+import { I18nProps } from '@polkadot/ui-app/types';
 
 import './index.css';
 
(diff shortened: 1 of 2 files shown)


jacogr:
We were all new(ish) to React, Blockchain and Polkadot at some point. :)

Ok, I'll tell you what I would have done (not saying it is correct, but may help getting you a bit familar with the codebase - that is also a monster)

1. First off, need to determine if the user has accounts. In this case we can inject the list of accounts similar to what is done in ui-app/InputAddress. (This part is relatively new, and surely not 100% optimal). So there is an observable that injects the list of accounts -

```
// Observable that updates with the list of active accounts
import accountsSubject from '@polkadot/ui-keyring/observable/accounts';
// generic observable -> props injector
import withObservableBase from '@polkadot/ui-react-rx/with/observableBase';

...
// Component here would be the main component, i.e. where this is actually used
export default withObservableBase(accountsSubject, { propName: 'allAccounts' })(Component);
``` 

2. At this point you should have a list of accounts on the pros object - I would use this in the base component to either display the selection, or if nothing, just display a message (what you have in that regard seems fine) with the link to accounts

So a bit different to ui-app/InputAddress (where the actual already-formatted options are injected for rendering with the Dropdown), but overall the approach is close.

packages/app-extrinsics/src/index.tsx
```diff
@@ -29,3 +50,7 @@ export default class ExtrinsicsApp extends React.PureComponent<Props> {
     );
   }
 }
+
+export { ExtrinsicsApp };
+
+export default withObservableBase(accountsObservable.subject, { propName: 'accountAll' })(ExtrinsicsApp);
```
jacogr: The issue with the `t` return is here - if you wrap this in a `translate()` it will inject the `t` property. As an example, see https://github.com/polkadot-js/apps/blob/master/packages/app-accounts/src/Editor.tsx#L232

packages/app-extrinsics/src/index.tsx
```diff
@@ -1,23 +1,44 @@
...
-export default class ExtrinsicsApp extends React.PureComponent<Props> {
+class ExtrinsicsApp extends React.PureComponent<Props> {
   render () {
+
+    const { accountAll } = this.props;
+
+    if (!accountAll || !Object.keys(accountAll).length) {
+      return (
+        <div>
+          <p>There are no saved accounts. Add some first.</p>
```
jacogr: Make sure that all strings are translatable, i.e. wrap these in `t(...)` (The injection below will solve the issue)

packages/app-extrinsics/src/index.tsx
```diff
@@ -1,23 +1,44 @@
...
+          <Link to='/accounts'>
+            <Button
```
jacogr: Maybe just a link will do just as well? (Have not looked, but it feels more linky-y)

marnibrewster:
Ah, now I understand the t('', ...) ! Thank you! How do these changes look?

packages/app-extrinsics/src/index.tsx
```diff
@@ -1,23 +1,43 @@
...
+import { I18nProps } from '@polkadot/ui-app/types';
 
 import './index.css';
 
 import React from 'react';
+import { Link } from 'react-router-dom';
 import { QueueConsumer } from '@polkadot/ui-signer/Context';
+import accountsObservable from '@polkadot/ui-keyring/observable/accounts';
+import withObservableBase from '@polkadot/ui-react-rx/with/observableBase';
+import translate from './translate';
```
jacogr: (nitpick) Would prefer a split between "stuff external to package" and "stuff inside the package", so a line-break just before would help making the distinction.

packages/app-extrinsics/src/index.tsx
```diff
@@ -1,23 +1,43 @@
...
+type Props = I18nProps & {
+  basePath: string,
+  accountAll?: Array<any>
 };
 
-export default class ExtrinsicsApp extends React.PureComponent<Props> {
+class ExtrinsicsApp extends React.PureComponent<Props> {
   render () {
+
+    const { accountAll, t } = this.props;
```
jacogr: (nitpick) Extra unneeded blank line before this.

PR #287, opened 2018-09-05
marni-WIP-284-added account meta to forget modal

marnibrewster:
Also fixed the translate issue in Forgetting.


packages/app-accounts/src/Forgetting.tsx
```diff
@@ -1,18 +1,22 @@
 // Copyright 2017-2018 @polkadot/app-accounts authors & contributors
 // This software may be modified and distributed under the terms
 // of the ISC license. See the LICENSE file for details.
+import { KeyringPair } from '@polkadot/util-keyring/types';
```
jacogr: (nitpick) blank line between comment and first import (just aids readability)

packages/app-accounts/src/Forgetting.tsx
```diff
@@ -57,20 +61,31 @@ class Forgetting extends React.PureComponent<Props> {
           <Button
             isPrimary
             onClick={doForget}
-            text='Delete'
+            text={t('forget.delete', {
+              defaultValue: 'Delete'
```
jacogr: I should have picked this up before - can we call this "Forget" rather. Basically "Delete" says that is is removed from the chain, rather in this case the account and associated balances will still be there, but this instance just "forgot" about it.

jacogr: Same with the "... you want to delete ..." below, should be "... remove from this browser...", or keeping with "forget".

PR #269, opened 2018-08-31
WIP: Fix copy/paste of addresses

chevdor:
Cheap trick but it works...

PR #291, opened 2018-09-10
fix: Part fix #264. Remove end space. Use consistent monospace

### LINKED ISSUE #264: Addresses are not properly cnp-able across the UI
Version of the UI: polkadot-js-ui v0.19.51

In the past, I could transfer to any account.
This no longer work if the recipient account is not in your address book.

image

ltfschoen:
Addresses _"2. Param rendering needs to remove extra space at end"_ of the to top-of-mind issues mentioned at https://github.com/polkadot-js/apps/issues/264#issuecomment-415955065

* Trims the address of preceding and trailing whitespace
* Shows CopyButton on the right of each address
* Uses Courier/monospace font family so the CopyButton's are vertically aligned when multiple addresses are returned from a query

Screenshot:

screen shot 2018-09-10 at 8 24 03 am

### DIFF
--- a/packages/ui-app/src/Params/Params.css
+++ b/packages/ui-app/src/Params/Params.css
@@ -5,7 +5,7 @@
 @import '../styles/partials/_mixins.css';
 
 .ui--Param-Address {
-  @add-mixin hash-font-family;
+  @add-mixin hash-font-traits;
 }
 
 .ui--Params-Content {
--- a/packages/ui-app/src/Params/Params.css
+++ b/packages/ui-app/src/Params/Params.css
@@ -2,6 +2,11 @@
 /* This software may be modified and distributed under the terms
 /* of the ISC license. See the LICENSE file for details. */
 
+.ui--Param-Address {
+  font-family: 'Courier New', Courier, monospace;
+  margin-right: 0.5rem;
+}
+
 .ui--Params-Content {
   box-sizing: border-box;
   padding-left: 2em;
--- a/packages/ui-app/src/Params/Params.css
+++ b/packages/ui-app/src/Params/Params.css
@@ -4,7 +4,6 @@
 
 .ui--Param-Address {
   font-family: 'Courier New', Courier, monospace;
-  margin-right: 0.5rem;
 }
 
 .ui--Params-Content {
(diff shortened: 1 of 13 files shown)


ltfschoen:
@jacogr I've retained the monospace font, removed the CopyButton, and added an empty div so double clicking the address only selects the address

packages/ui-app/src/Params/Params.css
```diff
@@ -2,6 +2,10 @@
 /* This software may be modified and distributed under the terms
 /* of the ISC license. See the LICENSE file for details. */
 
+.ui--Param-Address {
+  font-family: 'Courier New', Courier, monospace;
```
jacogr: Just make this `monospace` consistent with what we have in the rest of the UI, e.g. https://github.com/polkadot-js/apps/blob/master/packages/ui-app/src/styles/components.css#L54

ltfschoen:
@jacogr I've added a key prop the empty div (that allows the user to double click and select the address) so it doesn't produce missing key error. I've made monospace the preferred font and added Courier fonts as fallback incase the user doesn't support it. I've added the PostCSS Mixins dependency so we can consistently use the same font family or other traits for different hashes that we are displaying across the UI

chevdor:
Feel free to close #269 once this PR is merged.

packages/ui-app/src/styles/partials/_variables.css
```diff
@@ -24,3 +28,6 @@ $button-negative-color-hover: $middle-purple;
 $anchor-color: $light-slate-gray;
 $anchor-color-hover: $cadet-grey;
 $progress-bar-color: $dark-sky-blue;
+
+/* fallback font families included */
+$hash-font-family: monospace, 'Courier New', Courier;
```
jacogr: Actually `monospace` is whatever the browser decides it to be, so actually no need for a fallback, it is the built-in that is always there. https://developer.mozilla.org/en-US/docs/Web/CSS/font-family

So all these mixin changes can be reverted. Just `monospace` is fine and works by default.

ltfschoen:
@jacogr I've reverted the mixins and just used monospace now

PR #293, opened 2018-09-10
Basic transfer app

jacogr:
- allows transfers
- checks fees
- warning when account will cease to exist
- does not allow when no funds available
- when amount < available & reserved, display staking info
- warns when the final recipient amount < existential (i.e. no effect)

polkadot apps portal 2018-09-10 16-46-38


chevdor:
Cool!
It also should not allow or warn if the account is currently staking or nominating.

jacogr:
It only checks the freeBalanceOf for the available amount. But yes, you are correct, it could be 0 (and here it will warn), however a more helpful message may be that the "balance is reserved".

chevdor:
> however a more helpful message may be that the "balance is reserved".

Yes, at least a "check that you are not staking or nominating".

jacogr:
@chevdor I'm sure there are still some edge-cases which I have not thought of or missed. Some playing would be great - as it stands the list of messages and possibilities is longer than my arm. 

(Fee warnings are easy to tweak/extend as it stands. Funnily enough, nominating and the transferring actually does work as expected - checked with an "old nominator" as well as a "new one")

packages/app-transfer/src/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    return (
+      <Transfer />
+    );
+  }
+}
+
+export default withObservableBase(
+  accountObservable.subject, { propName: 'allAccounts' }
```
ltfschoen: We've called this propName `accountAll` elsewhere

PR #296, opened 2018-09-12
Parity Substrate theming

jacogr:
Change build with `UI_THEME=polkadot` (current default) or `UI_THEME=substrate` env variable.

- It swaps in and out the logo
- It changes the page title
- Changes the colors (currently set at defaults)

Future efforts -

- Substrate favicon
- Substrate Identicons (swap to beachball?)
- ...

polkadot apps portal 2018-09-13 10-07-55

Additionally there is the `UI_MODE` env variable. By default it is set to `full`, the intent is that it can be switch off features, i.e. for `UI_MODE=light` it will hide (not compile in) some of the apps, e.g. extrinsics, ... 

Future efforts -

- Build should be smaller, i.e. not include stuff in `UI_MODE=light` (I believe it is only hidden, some jiggling of import/requires required)
- We could (and probably should) re-jig the interface for `light` since there is much less stuff, e.g. do we want the bar on left? etc...
- ...

substrate apps portal 2018-09-13 11-18-58


jacogr:
cc @Tbaut 

- I moved the Polkadot-specific colors (css) into an own partial
- Polkadot-specific overrides are now in a seperate file
- All overrides are within a `.theme--polkadot` block
- There is a empty substrate theme (`.theme--substrate` block - currently they are just the SUI defaults)

Overall, good as a start without breaking anything, but allowing stuff to be swapped in-pout based on ENV config

PR #308, opened 2018-09-13
fix: Fixes #297. Toggles Creator menu tab before initial mount when no addresses exist

### LINKED ISSUE #297: Empty address tab is jumpy on create with no addresses
A picture is worth a thousand words...

https://youtu.be/WqPqzuC6gwo

(Marked this as accounts as well - not tested there, but I'm quite sure the same issue would probably exist?)

ltfschoen:
* Fixes bug introduced by PR #292 that didn't cater for switching to the Creator menu tab on initial mount when no accounts exist.
The user shouldn't need to be notified and have to manually switch menu tabs when there aren't any addresses, the UI should do the work for the user and switch automatically depending on the current state.
It only catered for updates to the available addresses (i.e. forgetting an address when it was the only one remaining would toggle displaying the Creator menu tab)
https://github.com/polkadot-js/apps/pull/292

### DIFF
--- a/packages/app-addresses/src/Editor.tsx
+++ b/packages/app-addresses/src/Editor.tsx
@@ -39,7 +39,8 @@ class Editor extends React.PureComponent<Props, State> {
   }
 
   render () {
-    const { isForgetOpen, current } = this.state;
+    const { current, isForgetOpen } = this.state;
+
     return (
       <div className='addresses--Editor'>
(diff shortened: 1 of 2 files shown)


ltfschoen:
Closing in favour of much better approach https://github.com/polkadot-js/apps/pull/309

PR #309, opened 2018-09-14
Fix Address add flicker

jacogr:
Alternative to closing https://github.com/polkadot-js/apps/issues/297

https://youtu.be/hJAeWznispY

- Allow Tabs component to hide items
- Don't re-create the items on each render
- Change state edit/create based on availablility of allAddresses
- Fix type definition of allAddresses
- autoFocus on address in Create (not having it is quite visible in above video)


packages/app-addresses/src/Editor.tsx
```diff
@@ -12,15 +12,11 @@ import Input from '@polkadot/ui-app/Input';
...
-import addressObservable from '@polkadot/ui-keyring/observable/addresses';
-import withObservableBase from '@polkadot/ui-react-rx/with/observableBase';
 
 import Forgetting from './Forgetting';
 import translate from './translate';
 
-type Props = I18nProps & {
-  addressAll?: Array<any>
-};
+type Props = I18nProps & {};
```
ltfschoen: I think we can change this to just `type Props = I18nProps;`

packages/app-addresses/src/Editor.tsx
```diff
@@ -245,6 +235,4 @@ class Editor extends React.PureComponent<Props, State> {
   }
 }
 
-export default withObservableBase(
-  addressObservable.subject, { propName: 'addressAll' }
-)(translate(Editor));
+export default translate(Editor);
```
ltfschoen: What do you think about passing `t` down as a prop from index.tsx instead of wrapping this component with `translate` too?

jacogr: I really don't want to start passing t around everywhere. So would really want to stay away from `translate(App)` and then passing it along the tree.

So if we have `App -> B(t)` and only `B` actually have translatable text, it already is an issue. (Which is mostly the case everywhere) 

Now if we have `App -> B -> C(t)` and we don't use it in App and don't use is in B, it just becomes this stream of props.

PR #314, opened 2018-09-15
WIP: Issue #190 relock account after sign send completed

ltfschoen:
I've attempted to address this issue in branch luke-190-relock-account-after-sign-send-completed https://github.com/polkadot-js/apps/compare/luke-190-relock-account-after-sign-send-completed?expand=1. I've updated create.ts so if create a new account and then open the modal to send a transaction it shows the password prompt (since beforehand it didn't prompt the user for the password straight after creating an account). But strangely if I debug just before calling the lockAccount function with pair.isLocked() it says the pair isn't locked... And when I run pair.lock() in lock.ts (the lockAccount function) to lock it whether it's already locked or not, it generates error Uncaught (in promise) Error: bad secret key size from nacl-fast.js.


jacogr:
There is already a branch for this that is waiting for the other account changes to go in as to minimize rework - https://github.com/polkadot-js/apps/compare/jg-account-lock?expand=1

1. Adds the ability to save the strategy for each account (i.e. unlock per use or per session)
2. Uses strategy to lock when created (hence waiting, it needs to do it for upload as well)
3. Locks based on strategy in Signer

Going forward, once we have the settings app the user should be able to set the defaults for the unlock strategy

ltfschoen:
Ok cool, I'll keep an eye on your branch and the settings app https://github.com/polkadot-js/apps/issues/86

Issue #93, opened 2018-06-12
Combine "unlock to sign" with packages/ui-signer


jacogr:
Currently we have duplicated logic. It would mean a bit of rework on the signer since it currently has a purpose of unlock, sign & submit, but from an overall perspective we don't want keychain management in multiple places.

jacogr:
Cleaning up the queue. Closing until relevant,.

PR #315, opened 2018-09-16
feat: Fix Issue #184 Allow user to change account password

### LINKED ISSUE #184: Allow user to change account password
1. verify old password
2. encrypt with new password

ltfschoen:
Screenshots of "Change Password" button, associated Modal window, and an example of an error notification:

screen shot 2018-09-16 at 1 12 57 pm

screen shot 2018-09-16 at 1 14 46 pm


packages/app-accounts/src/ChangePasswordButton.tsx
```diff
@@ -0,0 +1,261 @@
...
+import Button from '@polkadot/ui-app/Button';
+import Error from '@polkadot/ui-app/Error';
+import keyring from '@polkadot/ui-keyring/index';
+
+import ChangePasswordModal from './ChangePasswordModal';
+import translate from './translate';
+
+type State = {
+  address: string,
+  error?: FormErrorProps,
```
jacogr: I really don't get this new `FormErrorProps` that is now in here. So if I look at the definition, we have this -

```
type FormErrorProps = React.ReactNode & {
  props: React.ReactNode
};
```

So basically the above State become when expanded -

```
type State = {
  ...
  props: {
   ...
  }
}
```

Could we please re-look at this. It really just seems to add unneeded complexity for no benefit. Let's rather stick to the basics making things less magical and maintainable.

ltfschoen:
@jacogr I've pushed commits that address your review comments

ltfschoen:
@jacogr  I've pushed a commit that tries to refactor the code as much as possible.

* I've reduced the overall new lines of code in the PR from +730 −27 to +663 −27.
* I reduced ChangePasswordButton from 264 -> 178 lines of code, however I also added the helper/changePassword.ts that has 60, so whilst some code was removed, much of it was just rearranged
* I reduced changePassword.ts in ui-keyring from 65 -> 35
* I reduced ChangePasswordButton.spec.js test file from 45 -> 40
* I reduced ChangePasswordModal from 187 -> 183

Below is a count of new lines of code against each file with in latest changes, and a description of changes that I made where it may not be obvious (as some changes weren't just relevant to this task. Note that I've written a detailed commit message in the latest commit 800da2a

* 183  packages/app-accounts/src/ChangePasswordModal.tsx
* 178  packages/app-accounts/src/ChangePasswordButton.tsx
* 60  packages/app-accounts/src/helpers/changePassword.ts
  * helper functions/variables for generating empty state, empty errors, organising all the change password error messages, and for validating the input fields
* 42  packages/ui-app/src/Notification.tsx
  * notification component has also been created in other PRs that haven't yet been approved, but I need it here too
* 40  packages/app-accounts/src/ChangePasswordModal.spec.js
* 35  packages/ui-keyring/src/account/changePassword.ts
  * implement changePassword so new password replaces existing password
* 34  packages/app-accounts/src/index.css
  * styles for the change password modal including responsive media queries
* 22  packages/ui-keyring/src/account/update.ts
  * implement updateAccount method that is called by changePassword
* 17  packages/ui-signer/src/Unlock.tsx
  * pass the error message down as props so it's displayed in the password input field
* 16  packages/app-accounts/src/Editor.tsx
  * add the ChangePasswordButton button. reordering alphabetically
* 16  packages/ui-app/src/types.d.ts
  * declare FormErrors type that is used to show input field validation errors on input fields and overall form validation errors on the Notification component
* 13  packages/ui-app/src/Input.tsx
  *  showing the notification after the input. refactoring and reordering
* 11  packages/ui-keyring/src/index.ts
  * declare changeAccountPassword and updateAccount methods
* 11  packages/ui-keyring/src/types.d.ts
  * declare changeAccountPassword, updateAccount, and AccountResponse types 
  * lock the pair if it's not found to be locked already
* 7  packages/ui-app/src/Password.tsx
  * pass error message down and use classes import for Password component className
* 4  packages/ui-app/src/styles/components.css
  * styles for providing spacing between copy button and AddressMini in the modal 
* 1  packages/ui-keyring/src/account/save.ts
  * fixes bug where account doesn't save correctly after creation

packages/ui-signer/src/Unlock.tsx
```diff
@@ -56,8 +58,9 @@ class Unlock extends React.PureComponent<Props, State> {
           <Password
             autoFocus={autoFocus}
             className='medium'
+            error={error}
             isError={isError}
-            label={t('unlock.password', {
+            label={label || t('unlock.password', {
```
jacogr: Good catch.

packages/ui-signer/src/Unlock.tsx
```diff
@@ -36,18 +37,19 @@ class Unlock extends React.PureComponent<Props, State> {
...
   render () {
-    const { autoFocus, onChange, onKeyDown, password, t, tabIndex } = this.props;
-    const { isError, isLocked } = this.state;
+    const { autoFocus, label, onChange, onKeyDown, password, t, tabIndex } = this.props;
+    const { error, isError, isLocked, pair } = this.state;
 
-    if (!isLocked) {
-      return null;
+    if (pair && !isLocked) {
+      pair.lock();
```
jacogr: Ok, now we are forcing the unlocking each and every time - refrain from doing this until we have https://github.com/polkadot-js/apps/issues/190 in (and then this happens _only_ when the strategy is to unlock per transaction)

packages/ui-keyring/src/types.d.ts
```diff
@@ -35,6 +40,7 @@ export type State = {
 };
 
 export type KeyringInstance = {
+  changeAccountPassword: (t: TranslationFunction, address: string, password: string, newPassword: string) => AccountResponse,
```
jacogr: No. We cannot do this. We should not be passing `t` around between functions, it is specific to a component and internally yo it only, nowhere else.

packages/ui-keyring/src/types.d.ts
```diff
@@ -2,10 +2,15 @@
 // This software may be modified and distributed under the terms
 // of the ISC license. See the LICENSE file for details.
 
-import { KeyringInstance as BaseKeyringInstance, KeyringPair, KeyringPair$Meta } from '@polkadot/util-keyring/types';
+import { TranslationFunction } from 'i18next';
+import { KeyringInstance as BaseKeyringInstance, KeyringPair, KeyringPair$Json, KeyringPair$Meta } from '@polkadot/util-keyring/types';
```
jacogr: (nitpick) extra space before.

packages/ui-keyring/src/index.ts
```diff
@@ -2,23 +2,26 @@
 // This software may be modified and distributed under the terms
 // of the ISC license. See the LICENSE file for details.
 
+import { TranslationFunction } from 'i18next';
```
jacogr: No i18n in here, no passing of `t`

packages/ui-keyring/src/account/update.ts
```diff
@@ -0,0 +1,22 @@
...
+// of the ISC license. See the LICENSE file for details.
+
+import { KeyringPair } from '@polkadot/util-keyring/types';
+import { State } from '../types';
+
+export default function updateAccount (state: State, pair: KeyringPair, password?: string, newPassword?: string): void {
+  const { accounts, keyring } = state;
+  const json = pair.toJson(newPassword);
+
+  if (!json.meta.whenCreated) {
```
jacogr: Update, so it should already be created. Unneeded.

packages/ui-keyring/src/account/update.ts
```diff
@@ -0,0 +1,22 @@
...
+    json.meta.whenCreated = Date.now();
+  }
+
+  if (!json.meta.whenEdited) {
+    json.meta.whenEdited = Date.now();
```
jacogr: It is always edited, so edited should be set.

packages/ui-keyring/src/account/save.ts
```diff
@@ -13,4 +13,5 @@ export default function saveAccount (state: State, pair: KeyringPair, password?:
   }
 
   state.accounts.add(json.address, json);
+  state.keyring.addFromJson(json);
```
jacogr: Reasoning to this change?

packages/ui-keyring/src/account/changePassword.ts
```diff
@@ -0,0 +1,35 @@
...
+import { AccountResponse, State } from '../types';
+
+import updateAccount from './update';
+
+export default function changeAccountPassword (state: State, t: TranslationFunction, address: string, password: string, newPassword: string): AccountResponse {
+  const { keyring } = state;
+  let response = { error: undefined };
+
+  if (!address || !password || !newPassword) {
+    response.error = t('editor.change.password.error.missing.address.password', {
```
jacogr: What does keyring have to do with editor? This is keyring. Additionally, should translation be done, it is in the `ui-keyring` namespace, i.e. namespace is specific to the package. (No cross-sharing)

packages/ui-app/src/types.d.ts
```diff
@@ -25,3 +21,15 @@ export type BaseContext = {
     }
   }
 };
+
+export type FormErrors = {
+  inputError: {
```
jacogr: No nested structures like these in the state please. This just seems, well deep, cumbersome and a maintenance nightmare.

packages/ui-app/src/Password.tsx
```diff
@@ -34,14 +36,15 @@ export default class Password extends React.PureComponent<Props, State> {
...
   render () {
-    const { autoFocus, children, className, defaultValue, isDisabled, isError, label, name, onChange, onKeyDown, style, tabIndex, value, withLabel } = this.props;
+    const { autoFocus, children, className, defaultValue, error, isDisabled, isError, label, name, onChange, onKeyDown, style, tabIndex, value, withLabel } = this.props;
     const { isVisible } = this.state;
 
     return (
       <Input
         autoFocus={autoFocus}
-        className={className}
+        className={classes('ui--Password', className)}
```
jacogr: Good catch.

packages/app-accounts/src/helpers/changePassword.ts
```diff
@@ -0,0 +1,60 @@
...
+    newPassword: '',
+    password: '',
+    showModal,
+    success: undefined
+  });
+};
+
+const errorForKey = (key: string, t: TranslationFunction) => {
+  const errors: any = {
+    'editor.change.password.input.password.differ.error': t('editor.change.password.input.password.differ.error', {
```
jacogr: We fall into this trap with each PR -

1. the keys are _only_ allowed inside the actual React component
2. lets refrain from trying to pass `t` in everywhere, rather keep it in the component and there only

packages/ui-keyring/src/account/save.ts
ltfschoen: Without that change the user will encounter an error `ExtError: No encrypted data available to decode` when they try to change their password immediately after creating an account (unless of course they refresh the page after creating the account).

* Remove all accounts
* Create a new account with password "me"
* Click "Change Password"
* Enter the password "me"
* Enter the new password "me2"
* Click "Submit"
* View the error shown in the modal, which was encountered in the catch block of changePassword function in ui-keyring. It should show in "red" between the input fields and the buttons `ExtError: No encrypted data available to decode`

On a side note, would you like me to show such error notification in the UI as I've done. I thought this would be easier for users to report bugs and less load on tech support in production, as they could just submit a screenshot or copy/paste the error, and we wouldn't have to explain to them how to open Chrome Inspector to access the actual error if I did a console.error or were to throw the error instead.

ltfschoen:
@jacogr I've read through all your review comments and your branch jg-password and I've listed some assumptions and associated actions below. I've also prepared some questions (highlighted in "bold" that I wanted to ask you just to make sure my assumptions are correct before I get stuck into the code.

* Assumptions, Actions, and Questions in response to Usability comments from @jacogr on 20 Sept 2018

  * ASSUMPTION: Assume that users will know it was changed successfully if they click "Change" and the modal disappears
    * ACTION: remove the Success notification entirely
    * ACTION: just close the modal without any feedback if they click "Change" with valid input fields

  * ASSUMPTION: Users do not want to have to retype their password if an error occurs after clicking "Change"
    * ACTION: Retain the password input field values if the user clicks "Change" and a form error occurs so user doesn't have to retype the passwords

  * ASSUMPTION: Modal is too small on-screen, adopt the size of the Forget Modal
    * ACTION: Change modal size to `tiny`
    * **QUESTION**: Should we incorporate basic mobile-first responsive design with media queries in this PR so the modal size is small enough for mobile devices by default, but then increases to `tiny` size if the screen size is >1024px to cater for tablet/mobile?

  * ASSUMPTION: Consistent alignment of form labels across the UI
    * ACTION: Left-align the form labels (instead of centered)

* Assumptions, Actions, and Questions in response to Code comments from @jacogr on 20 Sept 2018

  * ASSUMPTION: Refrain from locking the account in any PR until PR #190 is merged, as this happes only when strategy is to unlock per transaction
    * ACTION: Update Unlock.tsx, removing `pair &&` from Line 51, and removing `pair.lock()` from Line 52

  * ASSUMPTION: TranslationFunction `t` should not be passed between functions as it is specific to a component and internally to it only
    * ACTION: Remove code that passes the TranslationFunction as an argument to ui-keyring functions
    * **QUESTION**:
      * Should we generate translatable error messages in ui-keyring and return them to the calling React component for rendering in the UI?
      If so is it **possible** to load the TranslationFunction directly into ui-keyring (and do you know if this has already been done before elsewhere)? I'm assuming the answer is yes because in relation to the translatable strings I've used in ui-keyring you said "should translation be done, it is in the ui-keyring namespace", but I just want to make sure

  * ASSUMPTION: Spaces required between imported types, imported libraries, and local imports
    * ACTION:
      * Add extra spaces to Line 5/6 of packages/ui-keyring/src/types.d.ts

  * ASSUMPTION: `updateAccount` function in ui-keyring performs updating, not creation
    * ACTION:
      * Remove `whenCreated` from packages/ui-keyring/src/account/update.ts
      * Remove the IF statement check `if (!json.meta.whenEdited) {`, but retain `json.meta.whenEdited = Date.now();`

  * ASSUMPTION: "Change Password" button should be shown between the "Forget" and "Reset" buttons in the UI
    * ACTION: Move it from the LHS to the RHS of the "Forget" button

  * ASSUMPTION: Should be more readable:
    * ACTION:
      * Rename Title from "Change Password" to "Change account password"
      * Rename submission button from "Submit" to "Change"

  * ASSUMPTION: Key namespace of a translatable string should be the package name where it's used. Keys are only allowed inside the actual React component
    * ACTION:
      * Change `t('editor.change.password` to `t('ui-keyring.change.password`, and reflect the change elsewhere where the namespace is incorrect

  * ASSUMPTION: Types that are nested object structures (that contain properties) should not be stored in state regardless of whether we're using React.ReactNode that supports `{}`
    * ACTION: 
      * Remove `export type FormErrors = { inputError: { ...` from packages/ui-app/src/types.d.ts
      * We'll just use the `error` prop to store the error directly as a translated string (whether it be a submission error after clicking the "Change" button, or an invalid input error)

  * ASSUMPTION: Instead of just disabling the "Change" button and illuminating the erroneous input fields in "red" as a result of form validation (if isValid is false due to any of the input fields returning !is<INPUT_FIELD>Valid), we'll assume the user needs basic input field error messages (password cannot be blank, new password should not match existing password, password must not exceed 32 characters) in addition to illuminating the input fields with colour "red" if they're erroneous and disabling the "Change" button, so the user knows what they're doing wrong.
    * Add input field validation check for password not blank, and password must not exceed 32 characters

  * ASSUMPTION: We'll show form submission errors in "red" in the area between the lowest input field and the "Change" button (i.e. incorrect password, ) if they encountered when processing the request in ui-keyring.
  We'll also show input field errors in the same error (before the user clicks the "Change" button), instead of separately below each input field
    * **QUESTION**: Which of the following Actions should I proceed with:
      * Remove showing input field validation errors below each specific input field
      * Show input field validation errors instead in the same place as submission errors, in the area between lowest input field and the "Change" button
      * Do not show input field validation errors (i.e. password should not be blank) at all, it's sufficient for the input field to be illuminated in "red" if they're invalid. Only show the submission errors (i.e. incorrect password)

  * ASSUMPTION: User can submit even if password is the same
    * **QUESTION**: Should we allow the existing password and new password to be the same without showing an input field validation error and disabling the "Change" button?

  * ASSUMPTION: Adopt best practice
    * ACTION: Review branch https://github.com/polkadot-js/apps/compare/jg-password?expand=1, and incorporate relevant techniques
      * Store `isNewValid` and `isOldValid` in state. Update them depending on whether or not the new or old password are incorrect (see https://github.com/polkadot-js/apps/compare/jg-password?expand=1#diff-339e40c834f8f18536f325d36365e2f7R131). `error` prop should just be the translated error message.
      * Remove use of `inputError` since we're not using nested object for the `error` prop anymore
      * Change Modal size from `mini` to `tiny`
      * Update ChangePasswordModal with
        * `{this.renderContentHeader()}`
        * `{this.renderContentAddress()}`
        * `{this.renderContentMessage()}`
        * `{this.renderNotification()}`
      * Do not use Unlock component, just use Password component
      * Add `{this.renderContent()}`, and move Modal.Header and Modal.Content into it and return an array with `key='content'`
      * Remove the `success` prop and do not pass it down to Notification component. Use `error` state from ChangePasswordButton and isValid to determine whether to close the modal upon success
      * Consider how in jg-password the doChange function clocks the account before calling .decodePkcs8. Check if this is required in changePassword.ts of ui-keyring. See https://github.com/polkadot-js/apps/compare/jg-password?expand=1#diff-339e40c834f8f18536f325d36365e2f7R125
      * Adopt validPass function. See https://github.com/polkadot-js/apps/compare/jg-password?expand=1#diff-339e40c834f8f18536f325d36365e2f7R159
      * Update Editor.tsx with state values `isForgetOpen` and `isPasswordOpen`. Instead of passing `<ChangePasswordButton address={address || ''} />` to the Button Group, add a Button for it to the Button Group with a handler `this.togglePass`. See https://github.com/polkadot-js/apps/compare/jg-password?expand=1#diff-98f7f33d50563f190479f76c53484640R71.
      Add `{this.renderModals()}` to only render the modals that depending on value of `isForgetOpen` or `isPasswordOpen` 
      * Rename `updateAccount` to `encryptAccount` in ui-keyring

    * **QUESTION**: Why did you just use Password instead of using Unlock? 

  * ASSUMPTION: User should be able to change their password immediately after creating a new account without encountering error `ExtError: No encrypted data available to decode`
    * ACTION: Retain the addition of `state.keyring.addFromJson(json);` in packages/ui-keyring/src/account/save.ts, as without that change the user will encounter an error `ExtError: No encrypted data available to decode` when they try to change their password immediately after creating an account (unless of course they refresh the page after creating the account).

jacogr:
1. "Should we incorporate basic mobile-first responsive design" - I actually use poc-2.polkadot.io more on my mobile than on my desktop. We are so far away from having anything usable (although with proper navbar placement, `UI_MODE=light` may be closer. Honestly, this is a waste atm - when we design for mobile, modals probably needs to take up the full screen anyway, so whatever tricks we do here is never going to be used.

2. "Should we generate translatable error messages in ui-keyring" - No. Stick to having the messages in the React components only. In all these PRs I've made the comment again and again to not move the i18n messages and not passing `t` around. If there is more granularity needed, ui-keyring can either (a) Throw util/ExtError with a code of (b) return an enum.

3. " Do not show input field validation errors " - Start there, get that in. Then we expand. Iterate on a basic framework and make it better instead of trying to do everything without real-world feedback.

4. "Should we allow the existing password and new password to be the same" - There are actually reasons why (me personally) will re-encrypt with the same password, i.e. when doing so the encryption nonce gets re-generated, even if people were close to cracking to password, at this point they need to start all-over again.

5. "Why did you just use Password instead of using Unlock" - Unlock is for the Signer, it was only ever meant to be used as part of the sending dialog. (Until https://github.com/polkadot-js/apps/issues/93 and there is a proper generic solution in-place, all is needed is to capture the password, nothing more.)

ltfschoen:
Closing in favour of using PR #332 instead. Possible additional changes to PR #332 have been proposed to be merged from PR #335.

Issue #316, opened 2018-09-17
Submit RPC call with default config crashes since accessing property of undefined prop array element 


ltfschoen:
Error is caused because the `values` array property of the `value` prop, which is an empty array, and we are trying to access an element of that array

To reproduce the error:

* Clone latest polkadot-js/apps
* Checkout "master" branch
* Go to http://localhost:3000/#/rpc

Screenshot:
<img width="898" alt="screen shot 2018-09-17 at 1 56 37 pm" src="https://user-images.githubusercontent.com/6226175/45605616-8d5c8b00-ba81-11e8-9c5b-5ee6494e344d.png">

* Click button "Submit RPC call"
* Observe browser crashing
* View error in Chrome Inspector

```
Uncaught TypeError: Cannot read property '0' of undefined
    at Extrinsic.render (Extrinsic.tsx?6628:66)
    at finishClassComponent (react-dom.development.js?d5f4:13393)
    at updateClassComponent (react-dom.development.js?d5f4:13356)
    at beginWork (react-dom.development.js?d5f4:13945)
    at performUnitOfWork (react-dom.development.js?d5f4:16249)
    at workLoop (react-dom.development.js?d5f4:16287)
    at HTMLUnknownElement.callCallback (react-dom.development.js?d5f4:145)
    at Object.invokeGuardedCallbackDev (react-dom.development.js?d5f4:195)
    at invokeGuardedCallback (react-dom.development.js?d5f4:248)
    at replayUnitOfWork (react-dom.development.js?d5f4:15578)
```

Screenshot:
<img width="908" alt="screen shot 2018-09-17 at 1 52 01 pm" src="https://user-images.githubusercontent.com/6226175/45605585-62723700-ba81-11e8-9c6a-d9e3974ed57a.png">

PR #317, opened 2018-09-17
fix: Fixes Issue #316. Assigns default extrinsic when prop value array element undefined

ltfschoen:


ltfschoen:
@jacogr I've pushed commits that update app-rpc instead and reverted the changes previously made to ui-signer. I found that when a user goes to http://localhost:3000/#/rpc, it's configured by default with "author > submitExtrinsic(extrinsic)" (as the rpc area and method name), but with no value for Bytes. So we find that in app-rpc/src/Selection.tsx the state which stores `values: Array<RawParam>` of the encoded message is an empty array that doesn't contain any param values, but it was returning true for isValid. So I've added a variable `hasRawParamValues` to check if the array contains any raw params and set isValid to false if it doesn't so it makes the "Submit RPC Call" button disabled.

ltfschoen:
@jacogr I just noticed that you've created a PR https://github.com/polkadot-js/apps/pull/321 to close issue #316. I tried it and fixes the bug. 
Question: in your PR you check if `rpc.params.length === values.length` (whereas I was just checking if `Object.keys(values).length !== 0`). It occurred to me that if `rpc.params.length` ever equals 0 then we'd encounter the bug again since `values.length` also equals 0 with the default value of Bytes. Is there ever a possibility that `rpc.params.length` equals 0? If so could we change [Line 98](https://github.com/polkadot-js/apps/pull/321/files#diff-b8af1f4d258d2da127437f453fc4914eR98) to `}, values.length && rpc.params.length === values.length && hasNeededKey);`
Apart from that I'm happy to close this PR #317.

ltfschoen:
Closing since other PRs actually resolve the issue

Issue #319, opened 2018-09-17
Voting info cuts with large numbers


jacogr:
A picture is worth a thousand words... (notice the cut on the "3")

<img width="329" alt="polkadot apps portal 2018-09-17 17-24-36" src="https://user-images.githubusercontent.com/1424473/45632850-a1fb4c00-ba9e-11e8-8dfe-e303659e6717.png">

This may help somewhat https://github.com/polkadot-js/apps/issues/270

jacogr:
Closed in #350, due to go live with "next" branch in poc-3.

Issue #270, opened 2018-09-01
Display all balances in µDOT (DOT / 1,000,000)


jacogr:
Adjust all balance displays to show up in µDOT. Just take the number, divided by 1_000_000 and that is the value we display.

1. Extrinsic decoding
2. Balances from storage
3. ui-app/Params/valueToText
4. ...

Next up after this - input into 1.0001 (Not initially, just allow micro inputs)

jacogr:
Closed in #350 due to go live with "next" branch for poc-3

PR #322, opened 2018-09-18
Display balance with μ, k & m

jacogr:
- Changes balance formatting to display with μ, aligning with polkadash.io
- Closes https://github.com/polkadot-js/apps/issues/270
- Closes https://github.com/polkadot-js/apps/issues/319


jacogr:
This works as-is, but I'm holding off to closer to POC-3 with the network reset. 

As it stands, the `μ` display is fine, but unless it goes `kμ` and then `mμ`, I'm a believer that it will just cause frustration for those who are used to having large numbers appear on-screen. (As it stands, for me going from close to a billion to less than 1000, well, it is really very weird weird. `mμ` may work.)

jacogr:
Replaced by #350

PR #326, opened 2018-09-18
FIX#323: remove call to trim on non-string type value

yjkimjunior:


packages/ui-app/src/Params/valueToText.tsx
```diff
@@ -125,7 +125,7 @@ function valueToText (type: Param$Types, value: any, swallowError: boolean = tru
 
     if (type === 'AccountId') {
       return value && value.length
-        ? accountToText(value.trim() as string)
+        ? accountToText(value as string)
```
ltfschoen: Well spotted. But shouldn't we still trim the value later on in `accountToText` on [Line 42](https://github.com/polkadot-js/apps/blob/e73dce5ea5edfc97e400d652ac35a090758d9abf/packages/ui-app/src/Params/valueToText.tsx#L42) once it's assigned to `address` as a string, instead of not trimming the value at all?

jacogr: I'm not convinced on the trim - just based on where the stuff comes from, i.e. these values all come from the API. It should encode without spaces. (If not, we have another bigger problem) They only occur when manually inputted. (So on inputs it is def needed, for values returned via API and then displayed, it is overkill)

PR #330, opened 2018-09-20
Remove public key from address inputs, use ss58 address instead

yjkimjunior:
submitting a WIP commit before making appropriate changes throughout the rest of the InputAddress component users. @jacogr does this look like what you were looking for?


packages/app-accounts/src/Editor.tsx
```diff
@@ -185,10 +185,10 @@ class Editor extends React.PureComponent<Props, State> {
     );
   }
 
-  onChangeAccount = (publicKey: Uint8Array): void => {
-    const current = publicKey && publicKey.length === 32
-      ? keyring.getPair(publicKey)
-      : null;
+  onChangeAccount = (ss58?: string): void => {
```
jacogr: Small nitpick (and yes, should have mentioned it before), it probably would make sense calling this `accountId` (which is basically the terminology used across Substrate/Polkadot)

packages/app-transfer/src/Submit.tsx
```diff
@@ -47,7 +47,7 @@ class Submit extends React.PureComponent<Props> {
     queueExtrinsic({
       extrinsic: extrinsics.staking.public.transfer,
       nonce: accountIndex || new BN(0),
-      publicKey: from,
+      ss58: from,
```
jacogr: Where the rubber meats the road ;)

packages/app-extrinsics/src/Params/Account.tsx
```diff
@@ -28,12 +28,12 @@ export default class Account extends React.PureComponent<Props> {
...
   }
 
-  onChange = (publicKey?: Uint8Array): void => {
+  onChange = (accountId?: string): void => {
     const { onChange } = this.props;
 
     onChange && onChange({
-      isValid: !!publicKey && publicKey.length === 32,
-      value: publicKey
+      isValid: !!accountId && accountId.length === 48,
```
jacogr: Generally ok, however there are some tricks. (I commented on the other PR as well). Polkadot also has the notion of "short addresses", i.e. 1 bytes, 2 byte, 4 byte and 8 byte. Additionally the address contains 2 checksums, so just checking the length misses that.

The best route of attack to check for validity (not the fastest, but safest) is to -

1. Take address, decode it
2. If it throws, not valid
3. Check length (decoded publicKey) for a length of 32 bytes

For now, the decoding does not yet cater for the short addresses, so the first part of the comment is not catered for byt this. However, as soon as the decoder does, we can adapt here. In the meantime, we ensure that it is an actual real address and not just 48 random characters.

packages/ui-identicon/src/index.ts
```diff
@@ -10,7 +10,12 @@ import newShape from './shape/circle';
 import newElement from './svg/element';
 import { SHAPE_COUNT } from './defaults';
 
-export default function identicon (seed: string | Uint8Array, diameter: number = 256, className: string = '', style?: { [index: string]: string }): HTMLElement {
+export default function identicon (seed: string | null, diameter: number = 256, className: string = '', style?: { [index: string]: string }): HTMLElement {
+  if (!seed) {
+    // should return a more descriptive error HTML Element
+    return newContainer(diameter, 'white', className, style);
```
jacogr: +1

packages/ui-react/src/IdentityIcon/Polkadot.tsx
```diff
@@ -9,13 +9,17 @@ import Identicon from 'polkadot-identicon';
 
 type Props = BaseProps & {
   size: number,
-  value: string | Uint8Array
+  value: string | null
```
jacogr: At this level, I would still allow for both Uint8Array and string. (So basically keep at least the definition as-is)

packages/ui-react/src/IdentityIcon/Substrate.tsx
```diff
@@ -12,7 +12,7 @@ import isU8a from '@polkadot/util/is/u8a';
 
 type Props = BaseProps & {
   size: number,
-  value: string | Uint8Array
+  value: string | null
```
jacogr: As above, let's allow at the base-level for both. (Especially around short-addresses, not 100% sure how they would look, plus if we share these libraries, we want to allow some flexibility)

packages/ui-react/src/IdentityIcon/index.tsx
```diff
@@ -13,7 +13,7 @@ import Substrate from './Substrate';
 
 type Props = BaseProps & {
   size?: number,
-  value: string | Uint8Array
+  value: string | null
```
jacogr: As above, allow U8a

packages/ui-identicon/src/index.ts
```diff
@@ -10,7 +10,11 @@ import newShape from './shape/circle';
 import newElement from './svg/element';
 import { SHAPE_COUNT } from './defaults';
 
-export default function identicon (seed: string | Uint8Array, diameter: number = 256, className: string = '', style?: { [index: string]: string }): HTMLElement {
+export default function identicon (seed: string | null, diameter: number = 256, className: string = '', style?: { [index: string]: string }): HTMLElement {
+  if (!seed) {
+    throw new Error('Unable to generate identicon from empty seed');
```
jacogr: Was a bit sceptical, however we _should_ now never get here since we have the traps on the UI components. (And this happens more that we would like, i.e. creating new account, it just calls without regard for validity)

TL;DR Makes sense with the UI traps

packages/app-transfer/src/Submit.tsx
```diff
@@ -18,7 +18,7 @@ type Props = I18nProps & {
   isDisabled: boolean,
   accountIndex?: BN,
   amount: BN,
-  from: Uint8Array,
+  from: string,
```
jacogr: Should `to` not also be accountId type? Not sure where it is converted now - e.g. transfer, the recipient gets selected from the InputAddress, passed through (previously it was U8a, now accountId). Basically my question is - does actual transactions still work? (Either via the extrinsics app or transfer?)

jacogr:
@yjkimjunior Closing this one, since the changes are in the next branch. (Thank you). Just trying to minimise the merge conflicts for when that one needs to merge into master.

PR #332, opened 2018-09-21
Account change password functionality

### LINKED ISSUE #184: Allow user to change account password

jacogr:
- Closes #184
- Replaces and closes #315 
- Needs additional checks & playtime

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,8 +29,8 @@
     "start": "cd packages/apps && webpack-serve --config webpack.config.js --port 3000"
   },
   "devDependencies": {
-    "@polkadot/dev-react": "^0.20.20",
-    "@polkadot/ts": "^0.1.29",
+    "@polkadot/dev-react": "^0.20.21",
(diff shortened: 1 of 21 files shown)


ltfschoen:
Propose to add an InputPassword component (similar to how we've created InputFile and InputNumber in other PRs)?

jacogr:
1. I would need to look again at the nacl libs, and actually just did - it checks the length. The 32 however aligns with create. (It is found here - https://github.com/dchest/tweetnacl-js/blob/master/nacl.js#L926 it is 32 max. However there are obviously ways and means around it, i.e. it can be hashed up-front)

2. There should be no "don't use these character" issues. The nacl u8a (which is what the password gets converted to), can contain anything. So spaces are actually valid. (If it fails, these is an issues somewhere on the UI, not in "this cannot be done")

ltfschoen:
@jacogr I've added a commit that reverts the "don't use spacebar" in passwords restriction

PR #202, opened 2018-08-07
feat: Fixes #74. Allow upload/download JSON/private key from accounts

ltfschoen:
* Upload JSON/private keys. Decrypt in modal by entered password to unlock secret key and store in memory and local storage. Error notifications on password input field when empty or incorrect.
* Download JSON/private keys. Decrypt in modal by entered password to unlock secret key then download encrypted copy. Error notifications on password input field when empty or incorrect.
screen shot 2018-08-07 at 3 33 09 am
* Requires `pair.lock()` https://github.com/polkadot-js/apps/pull/198 functionality to be incorporated at the following FIXME locations in the code so that the secret key is locked again immediately after each individual download or upload, otherwise another user could repeat the process and download or upload with the incorrect password.
  * https://github.com/polkadot-js/apps/compare/luke-74-download-upload-with-password?expand=1#diff-fdef2d47f48cbd87a974f75935042996R37
  * https://github.com/polkadot-js/apps/compare/luke-74-download-upload-with-password?expand=1#diff-fdef2d47f48cbd87a974f75935042996R50
  * https://github.com/polkadot-js/apps/compare/luke-74-download-upload-with-password?expand=1#diff-4460fae0eb0b78fa6efa8b092a9a0991R32
  * https://github.com/polkadot-js/apps/compare/luke-74-download-upload-with-password?expand=1#diff-fcf70b0c98d01219eab7a943d0de67a6R52
* All tests pass and tslint errors resolved

Issue #138, opened 2018-07-12
Class-ify ui-keyring


jacogr:
https://github.com/polkadot-js/apps/blob/master/packages/ui-keyring/src/index.ts#L35

The closure here has gotten a bit out of hand and it not that easy to maintain. (And less familiar as a pattern with passing state through)

jacogr:
Done for util-keyring in https://github.com/polkadot-js/common/pull/215

jacogr:
Should be quick following the above approach once https://github.com/polkadot-js/apps/pull/202 is in (don't want to do it before or during since it adds unneeded rework in there and non-specific items)

jacogr:
Closed in #393

PR #333, opened 2018-09-21
Account backup & restore functionality

### LINKED ISSUE #74: Allow download of JSON/Privatekey from accounts
Keyring supports extraction, just need a way of selecting, unlocking and saving. (Or just move from storage -> file)

jacogr:
- Closes #74
- Replaces and closes #202 
- Steps on the toes of #332 a bit, so would need a merge here or there (depending which is first)
- Needs additional checks & playtime


packages/ui-keyring/src/index.ts
```diff
@@ -32,7 +33,11 @@ const state: State = {
 
 loadAll(state);
 
+// FIXME The quicker we get in https://github.com/polkadot-js/apps/issues/138
```
jacogr: This can actually happen in parallel, quite happy to deal with the merge issues on both this and password to clean up this clutter. And it is not _that_ bad https://github.com/polkadot-js/common/pull/215

packages/app-accounts/src/Backup.tsx
```diff
@@ -0,0 +1,150 @@
...
+  private doBackup = (): void => {
+    const { onClose, pair } = this.props;
+    const { password } = this.state;
+
+    if (!pair) {
+      return;
+    }
+
+    try {
+      if (pair.isLocked()) {
```
ltfschoen: It looks like this should be `if (!pair.isLocked()) {`.

ltfschoen:
Is it necessary to add the Backup icon as well, or is the Backup button sufficient?

packages/ui-keyring/src/index.ts
ltfschoen: I've added a couple of extra commits in the same PR #336 to address the classify ui-keyring Issue #138.

packages/app-accounts/src/Backup.tsx
ltfschoen: I've fixed this in https://github.com/polkadot-js/apps/pull/336

jacogr: O, no, you are right. it is a bug. (Do it late introduce a bug, read it again late, miss the bug.) Bug 2 - Jaco 0

PR #335, opened 2018-09-24
fix: Show errors. Add Password maxLength. Fix classes lib. Fix styles.

ltfschoen:
Relevant extras from PR #315 that will be closed in favour of this PR #332. Each change is split into a separate commit.

* ChangePass component
  * Add className to ChangePass modal component using classes library
  * Reorder props alphabetically
  * Fix by add missing `--ChangePass` component namespace to styles

* Password component
  * Add maxLength property to Password component set to value of 32
  * Fix by use missing classes library in Password component Input
  * **REMOVED** (addresses review comment https://github.com/polkadot-js/apps/pull/332#issuecomment-423878755) initial restriction of keys that was added to onKeyDown event including spacebar

* Unlock component
  * Add missing label prop to allow overriding default label

* ui-keyring/src/account/save.ts
  * Add missing call to keyring.addFromJSON that allows users to change their password immediately after creating an account without encountering error `ExtError: No encrypted data available to decode`. Otherwise user must refresh webpage between creating new account and changing its password.


packages/ui-app/src/Input.tsx
```diff
@@ -72,6 +73,7 @@ export default class Input extends React.PureComponent<Props, State> {
           error={isError}
           hidden={isHidden}
           max={max}
+          maxLength={maxLength}
```
jacogr: +1

packages/ui-app/src/Password.tsx
```diff
@@ -40,12 +41,13 @@ export default class Password extends React.PureComponent<Props, State> {
     return (
       <Input
         autoFocus={autoFocus}
-        className={className}
+        className={classes('ui--Password', className)}
```
jacogr: Nice.

packages/app-accounts/src/index.css
```diff
@@ -2,6 +2,6 @@
 /* This software may be modified and distributed under the terms
 /* of the ISC license. See the LICENSE file for details. */
 
-.accounts--Modal-Address {
+.accounts--ChangePass-Modal-Address {
```
jacogr: I don't think this is correct - it is shared, i.e. 

- ChangePass
- Forgetting
- Backup

packages/app-accounts/src/ChangePass.tsx
```diff
@@ -93,7 +97,7 @@ class ChangePass extends React.PureComponent<Props, State> {
       </Modal.Header>,
       <Modal.Content key='content'>
         <AddressSummary
-          className='accounts--Modal-Address'
+          className='accounts--ChangePass-Modal-Address'
```
jacogr: This class is shared across all modals here (see comment in css)

packages/app-accounts/src/ChangePass.tsx
```diff
@@ -41,11 +42,14 @@ class ChangePass extends React.PureComponent<Props, State> {
   }
 
   render () {
+    const { className } = this.props;
+
     return (
       <Modal
-        size='tiny'
+        className={classes('accounts--ChangePass-Modal', className)}
```
jacogr: (nitpick) Since we never override, we don't need classes (rather let's keep it to ui-* and where exactly needed - we have been adding this all-over in the past and in 99.9% of the cases it was just extra fluff with no benefit. When needed, can add - there are reason we may want to)

packages/ui-signer/src/Unlock.tsx
```diff
@@ -57,7 +58,7 @@ class Unlock extends React.PureComponent<Props, State> {
             autoFocus={autoFocus}
             className='medium'
             isError={isError}
-            label={t('unlock.password', {
+            label={label || t('unlock.password', {
```
jacogr: Great :)

ltfschoen:
@jacogr I've updated the base branch from jg-password  to master since jg-password has now been merged. I've also merged latest master into this branch and addressed your review comments.

PR #338, opened 2018-09-25
fix #266 - helpful error message when input address is not in intentions list

### LINKED ISSUE #266: Cannot nominate any non-intention address
Using polkadot-js-ui v0.19.51

I have an account that wants to nominate :)

I can pretty much nominate any address (I tested a few):
image
If you need to copy/paste , this is the address: `5END82tfD39fvgwxe9qCkZJxtyQCtFJXzeSnBXpAR2D7vhZT `

The next however, does not work!
image
The address is `5HMeowwuJACuZwRLFbNMxgdKKKKGKjiFVUFrPF2tayK5e3rT `

So @Tbaut I am sorry, I cannot nominate you!

The addresses have the same length:
```
OK       5END82tfD39fvgwxe9qCkZJxtyQCtFJXzeSnBXpAR2D7vhZT
NOT OK   5HMeowwuJACuZwRLFbNMxgdKKKKGKjiFVUFrPF2tayK5e3rT
```

yjkimjunior:
#266

### DIFF
--- a/packages/app-staking/src/StakeList/Nominating.tsx
+++ b/packages/app-staking/src/StakeList/Nominating.tsx
@@ -111,19 +111,21 @@ class Nominating extends React.PureComponent<Props> {
         />
         {
           !isNomineeValid
-          ? <InputError label={t('nominator.error', {
-            defaultValue: 'The address you input is not intending to stake, and is therefore invalid. \
-                          Please try again with a different address.'
-          })} />
-          : null
+            ? <InputError
+                label={t('nominator.error', {
+                  defaultValue: 'The address you input is not intending to stake, and is therefore invalid. \
+                                Please try again with a different address.'
+                })} />
+            : null
         }
         {
           !isAddressFormatValid
-          ? <InputError label={t('nominator.error', {
-            defaultValue: 'The address you input does not conform to a recognized address format. \
-                          Please make sure youve entered the address correctly and try again.'
-          })} />
-          : null
+            ? <InputError
+                label={t('nominator.error', {
+                  defaultValue: 'The address you input does not conform to a recognized address format. \
+                            Please make sure youve entered the address correctly and try again.'
+                })} />
+            : null
         }
       </Modal.Content>
     ];
--- a/packages/app-staking/src/StakeList/Nominating.tsx
+++ b/packages/app-staking/src/StakeList/Nominating.tsx
@@ -9,6 +9,7 @@ import React from 'react';
 import Button from '@polkadot/ui-app/Button';
 import Input from '@polkadot/ui-app/Input';
 import Modal from '@polkadot/ui-app/Modal';
+import Labelled from '@polkadot/ui-app/Labelled';
 
 import translate from '../translate';
 
--- a/packages/app-staking/src/StakeList/Nominating.tsx
+++ b/packages/app-staking/src/StakeList/Nominating.tsx
@@ -109,19 +109,29 @@ class Nominating extends React.PureComponent<Props> {
           onChange={this.onChangeNominee}
           value={nominee}
         />
-        { !isNomineeValid ?
-          t('nominator.error', {
-            defaultValue: 'The address you input is not intending to stake, and is therefore invalid. Please try again with a different address.'
-          })
-          :
-          null
+        {
+          !isNomineeValid
+          ? <Labelled>
+              {
+                t('nominator.error', {
+                  defaultValue: '* The address you input is not intending to stake, and is therefore invalid. \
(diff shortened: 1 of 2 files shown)


packages/app-staking/src/StakeList/Nominating.tsx
```diff
@@ -104,6 +106,20 @@ class Nominating extends React.PureComponent<Props> {
...
+          isNomineeValid
+          ?
+          null
+          :
+          <p> The address you input is not intending to stake, and is therefore invalid. Please try again with a different address.</p>
+        }
+        {
+          isAddressFormatValid
+          ?
+          null
```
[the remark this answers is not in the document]
yjkimjunior: not so sure about this one. made some changes here but not so sure how much more readable this is compared to previous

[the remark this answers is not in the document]
jacogr: What we do elsewhere -

```
test // line 1
  ? value_one // new line, 2 spaces before
  : value_two // new line, 2 spaces before
```

packages/app-staking/src/StakeList/Nominating.tsx
```diff
@@ -104,15 +108,32 @@ class Nominating extends React.PureComponent<Props> {
...
+        }
       </Modal.Content>
     ];
   }
 
   private onChangeNominee = (nominee: string) => {
     const { intentions } = this.props;
 
+    const publicKey = addressDecode(nominee);
```
jacogr: Great. Small issue remains, the function throws, so just wrap in try/catch, see decoding at eg. https://github.com/polkadot-js/common/blob/master/packages/util-keyring/src/address/decode.ts#L16

packages/app-staking/src/StakeList/Nominating.tsx
```diff
@@ -104,17 +109,55 @@ class Nominating extends React.PureComponent<Props> {
...
-      isNomineeValid: intentions.includes(nominee),
-      nominee
-    });
+    try {
+      this.setState({
+        isNomineeValid: intentions.includes(nominee),
+        nominee
+      });
+
+      const publicKey = addressDecode(nominee);
```
jacogr: I would just try/catch around this, since the underlying function will throw.

packages/app-staking/src/StakeList/Nominating.tsx
```diff
@@ -104,17 +109,55 @@ class Nominating extends React.PureComponent<Props> {
           onChange={this.onChangeNominee}
           value={nominee}
         />
+        {
+          !isNomineeValid
+          ? <Labelled>
```
jacogr: Not entirely excited about the use of this since we basically ignore what it is for however, we don't have a box-y component, so you made do with what you had. No issues there.

(We probably need something in ui-app that does a "display a box with border", e.g. if you look at Transfer there is something that displays the fees, but completely local to there).

TL;DR - Not excited, but it is ok since there is little alternative atm

yjkimjunior: We could add an optional prop to the Labelled component for the semantic-react-ui's Label props of color and pointing: https://react.semantic-ui.com/elements/label/#types-pointing-colored

jacogr: Ahh, yes, I actually pointed Luke to that a while back. We should obviously not over-use it, but I think it makes perfect sense.

Maybe not use Labelled for that since it is actually different from the SUI "Label", but rather call it something else and just wrap that properly. (Not sure what on the naming). In Labelled we actually fight against the SUI Label, the intent is different.

packages/ui-app/src/InputError.tsx
```diff
@@ -0,0 +1,33 @@
...
+import React from 'react';
+import Label from 'semantic-ui-react/dist/commonjs/elements/Label/Label';
+
+import classes from './util/classes';
+
+type Props = BareProps & {
+  label?: any
+};
+
+const defaultLabel: any = (// node?
```
jacogr: Yes, `React.ReactNode`

packages/ui-app/src/InputError.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+import { BareProps } from './types';
+
+import React from 'react';
+import Label from 'semantic-ui-react/dist/commonjs/elements/Label/Label';
+
+import classes from './util/classes';
+
+type Props = BareProps & {
+  label?: any
```
jacogr: `React.ReactNode` (tyr to steer clear from any as much as possible - yes, it does happen that they are needed)

Issue #276, opened 2018-09-03
Add scrollbar for the nominator list


chevdor:
The more nominators you have the wilder the UI becomes:
![image](https://user-images.githubusercontent.com/738724/44990354-0db6c280-af91-11e8-93a0-c39d4ade6656.png)

I propose the following change to have the UI limited to the following no matter what:
![image](https://user-images.githubusercontent.com/738724/44990392-2921cd80-af91-11e8-8904-356672e918ec.png)

The code shows:
![image](https://user-images.githubusercontent.com/738724/44990413-35a62600-af91-11e8-8915-a03fc2b7d9e2.png)

I´d suggest:
- [ ] to encapsulate all the `div.ui--AddressMini padded` into a new div with class `ui--Nominators` in order to keep the button visible at all time
- [ ] Add the following pros to the new css of `ui--Nominators`
```
overflow-y: overlay;
max-height: 196px;
padding-right: 18px;
```

jacogr:
While in here, think we can also lose some space/padding between items. (From the screenshot it looks like quite a bit, just a 20% trim or so)

chevdor:
I agree, that would not hurt.

jacogr:
Cool. Lemme whip up something - I'm actually in the same position, local node or not. Basically it shows only for validators with a lot of nominators. But will figure something out to reproduce and then update.

jacogr:
Ok, I came up with a hack that at least shows the issue and allows for fixes. (Some dev hacking required). Basically I can make it do this -

<img width="526" alt="polkadot apps portal 2018-09-21 09-45-37" src="https://user-images.githubusercontent.com/1424473/45867223-3c9ba980-bd83-11e8-915b-5907e9cadc2c.png">

So to get this working you need to have nominated somebody on the poc-2 network. (If you need some extra DOTs, give me a shout with your address, either here on on Riot, I'm jacogr:matrix.parity.io)

Anyway, I've mangled the code to make it show up here https://github.com/polkadot-js/apps/blob/master/packages/app-staking/src/StakeList/Account.tsx#L95

Basically did the following in that function, ugly but shows the issue very clearly -

```
private renderNominee () {
    const { stakingNominating } = this.props;

    if (!stakingNominating) {
      return null;
    }

    const n = (
      <AddressMini
        balance={this.balanceArray(stakingNominating)}
        value={stakingNominating}
        withBalance
      />
    );

[...]

PR #346, opened 2018-10-04
trying to get mapping of nominators correct

marnibrewster:
#276 
I'm not sure this is the right solution, I am still confused on how to create practice data to test. Feel free to send a gif of how to populate (while running polkadot node) with practice data!


jacogr:
Merging as-is, tested it _looks_ perfectly fine. @chevdor feel free to comment and/or requests additions when you are not 100% with the solution.

Issue #303, opened 2018-09-13
Display intention/validator name in staking list


jacogr:
Currently we have -

<img width="1061" alt="polkadot apps portal 2018-09-13 17-47-56" src="https://user-images.githubusercontent.com/1424473/45499688-3141f000-b77d-11e8-9aeb-de2e5efb5c9a.png">

So the idea would be to display the name of the address/account when we have that saved locally. (In stead of the generic `intention` or `validator` we have currently). So basically what we would need to do -

1. pull in the ui-keyring package into the staking list
2. check for any address in that and when we found it, extract the name from the meta - so we probably need 2 lookups - 
  - `pair = keyring.getAccount(...)`, if available then `pair.getMeta().name`, alternatively 
  - `pair = keyring.getAddress(...)`, if available get name from meta
  - if nothing else, render the current default strings 

We probably want to (instead of directly using AccountRow), make this whole lookup part of a local component, then

1. we pass in the address (& balance as required for nominators)
2. do the lookup in there as described above
3. do the passing of the name in there (either the value found above, or the fallback/default as passed in as props)

(Just to avoid duplication, alternatively just an `renderAddress` method in the current implementation - probably quicker and not too messy. It is up to the implementor to see what works best here)

jacogr:
cc @chevdor (from TODO in the staking code)

jacogr:
Closed in #448 (next branch)

PR #347, opened 2018-10-04
Marni wip 303 intentionvalidator name

marnibrewster:
#303 

This is a start, I think? 
The address vs. account is still confusing for me - so let me know what needs to be changed or adjusted.

### DIFF
--- a/packages/app-staking/src/Overview/CurrentList.tsx
+++ b/packages/app-staking/src/Overview/CurrentList.tsx
@@ -9,6 +9,7 @@ import BN from 'bn.js';
 import React from 'react';
 import AddressMini from '@polkadot/ui-app/AddressMini';
 import AddressRow from '@polkadot/ui-app/AddressRow';
+import keyring from '@polkadot/ui-keyring/index';
 
 import translate from '../translate';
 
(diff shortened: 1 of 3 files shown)


packages/app-staking/src/Overview/CurrentList.tsx
```diff
@@ -32,6 +33,18 @@ class CurrentList extends React.PureComponent<Props> {
     );
   }
 
+  private renderAddress (address: string) {
+    if (keyring.getAccounts()) {
```
jacogr: We should be able to do` keyring.getAccount(address))`and it is not available, horrible :(

There it will fail though since `getAccounts()` should return something. So the first condition is always true.

I suppose (since we don't have the getAccount, and we really should), we can do -

```
let pair = keyring.getAccounts().filter(account => account.address() === address);

if (!pair) {
  pair = keyring.getAddress(address);
}

return pair
  ? pair.getMeta().name
  : address;
```

jacogr:
- Address - those that you have saved of other people on the system. (Does not belong to you, but you can use it, eg. transfer funds to them). So in my addressbook, I have your address saved - since I sent you some funds

- Account - Here you control the private keys. So you can basically do anything with the, i.e. send funds from them, send funds to them, anything you like.

PR #351, opened 2018-10-07
Fix for disabled chain state button bug

### LINKED ISSUE #324: Same-type param queries yields disabled storage
This is a variant of https://github.com/polkadot-js/apps/issues/286 

1. Go to Chain state
2. Select staking
3. select bondageOf - it should work
4. now swap to currentNominatorsFor

The button is disabled when it should be enabled.

kwingram25:
- Closes #324 

Merges previous and next params when changing key on chain state app, fixes button disabling when switching between keys with same param type.

### DIFF
--- a/packages/app-storage/src/Selection.tsx
+++ b/packages/app-storage/src/Selection.tsx
@@ -73,10 +73,40 @@ class Selection extends React.PureComponent<Props, State> {
   nextState (newState: State): void {
     this.setState(
       (prevState: State) => {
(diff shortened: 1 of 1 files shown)


kwingram25:
Thank you! The merge only affects cases where the two keys have the same type of parameter at the same index - otherwise the parameters are properly reset to the defaults as before. I tried switching between system > accountIndexOf(who) and blockHashAt(blockNumber) and it functions properly

PR #336, opened 2018-09-24
fix: Backup and restore modifications. See detailed commit message

ltfschoen:
Relevant extras from PR #202 that will be closed in favour of this PR #333. I've tried to split each change into a separate commit.

* Backup
  * Reorder alphabetically
  * Add missing className using classes library
  * Remove duplicate try/catch block
  * Move backupAccount logic into ui-keyring backupAccount
  * Fix ui-keyring/src/account/save.ts by adding `state.keyring.addFromJson` so user can download account immediately after creating an account (without having to refresh the webpage) - as was done in PR #332
  * Refactor into `validatePass` class method, as was done in PR #332
    * Future option - move into helper and reuse across ChangePass, Backup, Restore
  * Fixes existing bug `ExtError: No encrypted data available to decode` when try to Backup account **straight after** creating a new account

* Creator
  * Replace onBack with onCreateAccount for readability

* Editor
  * Reorder alphabetically
  * Remove onBack

* Restore
  * Reorder alphabetically
  * Add `private` to class methods
  * Refactor into `validatePass` class method, as was done in PR #332
  * Add className using classes library
  * Restrict user to only selecting .json file formats (non-json files cannot be selected).
  * Add missing `withLabel` prop to InputFile
  * Replace onBack with onRestoreAccount for readability
  * Move restoreAccount logic into ui-keyring. Fix bug `ExtError: Invalid Pkcs8 divider found in body` (see comments of this PR for details). Re-work addAccountPair.

* ui-app InputFile
  * Reorder alphabetically
  * Add `private` to class methods
  * Add `acceptedFormats` prop and missing `withLabel` prop

* ui-app Input
  * Reorder alphabetically
  * Add maxLength of 32

* ui-app Password
  * Reorder alphabetically
  * Add maxLength of 32
  * **REMOVED** (addresses review comment) initial restriction of keys that was added to onKeyDown event including spacebar

* ui-app Params/Param/File.tsx
  * Spread operator to pass down props without default override

* ui-app styles/components.css
  * Reorder alphabetically
  * Fix by adding padding between AddressSummary and CopyButton

* ui-keyring src/index.ts
  * Convert closure to ES6 class to address Issue #138 

A bug was identified but has been resolved, hence the reason for commit #d1737f7 that adds console.error to the catch block in Backup.tsx for debugging purposes.  To reproduce the error:
  1) Forget all accounts
  2) Create an account with password
  3) Backup the account (works the first time you download it)
[...]

PR #386, opened 2018-10-26
[next] refactor: Move backup/restore functionality to Keyring. See details

ltfschoen:
* Refactor moving backup account functionality into Keyring

* Refactor moving restore account functionality into Keyring

* Refactor checking password length. Move into Keyring function `isPassValid`.
Add constant `MAX_PASS_LEN` to Keyring defaults for referring to max password length of 32

* Rename `onBack` to `onCreateAccount` and `onRestoreAccount` as appropriate

* Add explicit type to `allAccounts` of `Array<Object>` instead of `Array<any>`

* Add `private` to class functions as appropriate

* Only allow user to select JSON files, and show label with instructions 
when restoring account by passing props `acceptedFormats='.json'` and `withLabel` 
to InputFile component

* Re-order alphabetically

* Show error for debugging by adding console.error to catch in try/catch blocks

* Note: 
  * This PR is based on work previously done in PR https://github.com/polkadot-js/apps/pull/336, but has been tailored based on previous feedback provided and builds upon "next" branch instead of "master"
  * Excludes resolving Issue #138 by refactoring Keyring into a class. This work will be done in a subsequent commit after this is merged.


packages/app-accounts/src/Editor.tsx
```diff
@@ -15,8 +15,7 @@ import Forgetting from './Forgetting';
 import translate from './translate';
 
 type Props = I18nProps & {
-  allAccounts?: Array<any>,
-  onBack: () => void
+  allAccounts?: Array<Object>
```
jacogr: Sadly, Object is not much better than `any` :( We should rather either just add the `FIXME` for the any, or do it correctly with the correct typed. (Originally I was rushing, hence the `any`, when we fix it up we should not need to come back to fix `Object`)

packages/app-transfer/src/index.tsx
```diff
@@ -15,7 +15,7 @@ import Transfer from './Transfer';
 import translate from './translate';
 
 type Props = I18nProps & {
-  allAccounts?: Array<any>
+  allAccounts?: Array<Object>
```
jacogr: As per the other `any` vs `Object` comment above.

packages/ui-app/src/InputFile.tsx
```diff
@@ -8,11 +8,12 @@ import { BareProps } from '@polkadot/ui-app/types';
 import React from 'react';
 import Dropzone from 'react-dropzone';
 
-import Labelled from './Labelled';
 import classes from './util/classes';
+import Labelled from './Labelled';
 import translate from './translate';
 
 type Props = BareProps & {
+  acceptedFormats?: string,
```
jacogr: I believe it should maybe be `acceptedFormat` since it is a string and singular? (Or rather I'm probably not seeing how to specify multiples then, i.e. would have expected `Array<string>` for multiples)

jacogr: If indeed valid, i.e. can accept `.json;.txt` (all in-one), maybe we should just leave at least a comment above.

ltfschoen:
@jacogr I've pushed a couple of commits that address your review comments

jacogr:
@ltfschoen Do you want this before of after the class refactor? (Since there is a conflict between the 2 - I'm thinking this is probably better as the last PR, i.e. get class in, change this one slightly and then get it in?)

ltfschoen:
@jacogr I'm happy with your recommendation!

packages/app-transfer/src/Fees.tsx
```diff
@@ -80,7 +80,7 @@ class FeeDisplay extends React.PureComponent<Props, State> {
     );
 
     let txfees = fees.baseFee
-        .add(fees.transferFee)
+        .add(fees.transferFee ? fees.transferFee : new Balance(0))
```
jacogr: Not sure why this is in this PR. And not sure what this is supposed to fix since it should be there. However, if this is an issue it means the `= ZERO_FEES` does not do the trick. (And the issue is actually elsewhere in api itself since we get an undefined/null in the structure, which should not be the case at all, we should not work around it)

Have not seen it manifest however.

With all that, still unsure why this is in here...

jacogr: I double-checked api-observable, it has this - https://github.com/polkadot-js/api/blob/master/packages/api-observable/src/Queries.ts#L155

So it should never be null/undefined. If it is, something is going horribly wrong along the way.

ltfschoen: I just pulled and merged in the latest from 'next' branch, removed node_modules, restored the code back to `.add(fees.transferFee)`, and ran `yarn; yarn run build; yarn run check; yarn run start;` and this time no linting errors appeared. Although not sure what I did that triggered the linting error.

PR #387, opened 2018-10-28
[next] refactor: Issue #138. Convert closure into ES6 class object. Move methods into class

ltfschoen:


packages/ui-app/src/InputAddress/index.tsx
```diff
@@ -189,7 +187,7 @@ class InputAddress extends React.PureComponent<Props, State> {
       if (accountId) {
         if (!matches.find((item) => item.key === RECENT_KEY)) {
           matches.push(
-            createOptionHeader('Recent')
+            keyring.createOptionHeader('Recent')
```
jacogr: These I'm a bit sceptical about - well, not just here, overall :) It never felt like a good 100% fit, but was rather put there since it is the closest to a fit. So not convinced the option stuff (which is really UI-related, the rest of keyring is not). So maybe would rather keep this in it's own class? Any other suggestions?

packages/ui-keyring/src/index.ts
```diff
@@ -4,83 +4,429 @@
...
+    if (!singletonInstance) {
+      singletonInstance = this;
+    } else {
+      return;
+    }
+
+    // NOTE Everything is loaded in API after chain is received
+    // this.loadAll();
+
+    this.id = ++Keyring.counter;
```
jacogr: All this should not been needed. Since we don't export `class Keyring` and rather export the only instance we create, there will only be one in existence. So basically,

```
// not exported
class Keyring {
...
}

// since instance
const keyring = new Keyring();

// export the only instance ever
export default keyring;
```

would do the trick

packages/ui-keyring/src/index.ts
```diff
@@ -4,83 +4,429 @@
...
+      .forEach(({ json: { meta: { isRecent = false } }, option }: SingleAddress) => {
+        if (isRecent) {
+          options.recent.push(option);
+        } else {
+          options.address.push(option);
+        }
+      });
+  }
+
+  private emptyOptions (): KeyringOptions {
```
jacogr: I would suggest (as per the comments above), to just keep this to the add, remove, etc. logic - rather keep the observable stuff seperate?

packages/ui-keyring/src/index.ts
```diff
@@ -4,83 +4,429 @@
...
+          options.testing.push(option);
+        }
+      });
+  }
+}
+
+Keyring.counter = 0;
+
+export {
+  Keyring
```
jacogr: Rather don't export at all. (Only the instance, not the class - no hoops needed above then)

ltfschoen: I needed to export the class for the unit tests https://github.com/polkadot-js/apps/pull/387/files/f45d32fb7cb33d5bf14676cbb57072d22edabdee#diff-ef3c29ea5508f636947d9a7544f8bf7dR5

ltfschoen: But exporting the class for the unit tests exposes it to someone creating multiple instances of it

ltfschoen: Actually, since `initOptions` is moving elsewhere I won't need to test that in Keyring class anymore

packages/ui-app/src/InputAddress/index.tsx
ltfschoen: Yes, I've created a `KeyringOption` class for the options/ folder in ui-keyring

ltfschoen:
@jacogr I've pushed commits that address all your review comments.

packages/ui-keyring/src/index.ts
```diff
@@ -4,83 +4,310 @@
...
-} as KeyringInstance);
+class Keyring implements KeyringInstance {
+  private state: State;
+
+  constructor () {
+    // state must be always defined in the constructor even if trying to create second instance to overcome TSLint error
+    this.state = {
+      accounts: observableAccounts,
+      addresses: observableAddresses,
+      keyring: testKeyring()
```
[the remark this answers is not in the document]
jacogr: Although not 100% correct, I would just `this.state = {... } as State` here and let things crash and burn if loadAll was not called (e.g. keyring === undefined`. It is the first things that _should_ be done to setup, I'm guessing loadAll is actually now (at this moment) actually `init`. And unless called at the correct point after the chain is loaded and the prefixes set in the API, the class is not usable.

ltfschoen:
@jacogr I've pushed a commit that moves the `state` back out of the Keyring instance. It loads the test accounts again.

packages/ui-keyring/src/index.ts
```diff
@@ -7,36 +7,28 @@ import { SingleAddress } from './observable/types';
...
-    this.state = {
-      accounts: observableAccounts,
-      addresses: observableAddresses,
-      keyring: testKeyring()
-    };
-
-    // NOTE Everything is loaded in API after chain is received
-    // this.loadAll();
-  }
+const state: State = {
```
jacogr: We could just have these on `this`. (Which is closer to OOP). Then we just go `this.keyring` instead of `const { keyring } = state;`

So we just need some small hoops for `keyring` itself since it is actually not initialised in the constructor and we will make the linter unhappy. I suggest the following -

```
class Keyring ... {
  private _keyring: KeyRingInstance: // not 100% sure of the type

  get keyring (): KeyringInstance {
    if (this._keyring) {
      return this._keyring;
    }

    throw new Error('Keyring should be initialised via 'loadAll' before use');
  } 

  ...
  someFunction () {
    this.keyring.doSomething();
  }
}
```

ltfschoen:
@jacogr I've pushed a commit that addresses your review comment

Issue #294, opened 2018-09-11
Signer dialog extended to be more informative


jacogr:
In short, currently the information displayed is very technical and not much use to an actual user. Make it informative. 

A random bunch of thoughts -

- [x] Display the actual account we are sending with with and `AddressSummary` component
- [x] Decode the actual extrinsic to be sent - we currently do something similar in the block explorer, instead of a bytes stream (somewhat useful to maybe devs), display what we are actually doing
- ...

Some bonus points -

1. Function to allow calculate the length of an extrinsic (before submission - e.g. in the Transfer app this is hardcoded)

jacogr:
Extrinsic decoding done as part of the next branch (not a PR as of yet)

jacogr:
Still needs fee value display in the signer.

jacogr:
Closing. Only fees outstanding on the signer, these are tracked in https://github.com/polkadot-js/apps/issues/71

PR #389, opened 2018-10-28
[next] feat: Relates to #294. Show fees in Extrinsics Signer dialog

ltfschoen:
* Obtain fees from observable in ui-signer Extrinsic component

* Use `FeeDisplay` component in ui-signer Extrinsic component to render fees

* Pass `hasAvailable` value obtained from `FeeDisplay`'s onChangeFee up to Modal component and store in state there to disable the "Sign and Submit" button if the sender has insufficient funds

* Fix `accountNonce` since was crashing the browser when tried to click "Submit Transaction" in the Extrinsics tab

* Add `ui--signer-Signer-FeeDisplay` styles to indent the fees that are shown

* TODO
	* Reuse from common location (but not sure where is appropriate) instead of obtaining from app-transfer the FeesDisplay
		```
		import { Fees } from '@polkadot/app-transfer/types';

		import FeeDisplay from '@polkadot/app-transfer/Fees';
		```

Screenshot:
screen shot 2018-10-28 at 17 09 00


packages/ui-signer/src/Extrinsic.tsx
```diff
@@ -3,30 +3,47 @@
 // of the ISC license. See the LICENSE file for details.
 
 import { I18nProps } from '@polkadot/ui-app/types';
+import { RxFees } from '@polkadot/api-observable/types';
 import { QueueTx } from './types';
+// TODO - reuse instead of obtaining from app-transfer
```
jacogr: 100% on this TOD. Should rather have it in signer (and let Transfer pull that from that package), instead of having the dependency from ui into app.

packages/ui-signer/src/Extrinsic.tsx
```diff
@@ -40,10 +57,23 @@ class Transaction extends React.PureComponent<Props> {
...
+                You are about to sign a message from <span className='code'>{accountId}</span> calling <span className='code'>{section}.{method}</span> with an index of <span className='code'>{nonce.toString()}</span>
               </Trans>
             </p>
-            <Call value={extrinsic} />
+            <Call value={extrinsic}/>
+            <div
+              className={'ui--signer-Signer-FeeDisplay'}
+              style={style}
+            >
+              <FeeDisplay
```
jacogr: This is not quite correct - but it does work... So the issue is that this component takes the `transferFee` into account as well, but it is only applicable for, well, balances.transfer. It does work at this point since it is set to `0` on-chain, so the match work. However as soon as that value changes, the values for non-transfers will be incorrect.

packages/ui-signer/src/Modal.tsx
```diff
@@ -5,7 +5,10 @@
 import { ApiProps } from '@polkadot/ui-react-rx/types';
 import { I18nProps, BareProps } from '@polkadot/ui-app/types';
 import { QueueTx, QueueTx$Id, QueueTx$MessageSetStatus, QueueTx$Result, QueueTx$Status } from './types';
+// TODO - reuse instead of obtaining from app-transfer
```
jacogr: As per other comments, the dep needs to be flipped - app-* can depend on ui-*, however ui-* should never depend on app-*.

ltfschoen:
@jacogr I've pushed a couple of commits that I believe address your review comments

packages/ui-signer/src/Extrinsic.tsx
```diff
@@ -3,32 +3,52 @@
...
+      ? accountNonce
+      : new BN(0);
 
     if (!extrinsic) {
       return null;
     }
 
+    const methodInstance = new Method(extrinsic, extrinsic.meta);
+    const amount = methodInstance.args[1].raw;
+    const recipientId = methodInstance.args[0].raw.toString();
```
jacogr: We should never access `.raw` outside of types. As it stands Amuary is working on removing all `.raw`

packages/ui-signer/src/Extrinsic.tsx
```diff
@@ -3,32 +3,52 @@
...
+    const nonce = accountNonce
+      ? accountNonce
+      : new BN(0);
 
     if (!extrinsic) {
       return null;
     }
 
+    const methodInstance = new Method(extrinsic, extrinsic.meta);
+    const amount = methodInstance.args[1].raw;
```
jacogr: This is _only_ the case for transfer, not guaranteed for any other extrinsics.

packages/ui-signer/src/Extrinsic.tsx
```diff
@@ -3,32 +3,52 @@
...
+    const recipientId = methodInstance.args[0].raw.toString();
+
     const { method, section } = Method.findFunction(extrinsic.callIndex);
 
+    const isTransfer = fees && fees.transferFee && section === 'balances' && method === 'transfer';
+
+    if (isTransfer) {
+      fees.transferFee = new Balance(0);
```
jacogr: I'm not sure I get this - `transferFee` is from the state, we cannot assign it? Especially not for transfer where it needs to use the on-chain value?

packages/ui-signer/src/Extrinsic.tsx
ltfschoen: I discussed this with Amaury, which resulted in the following PR to update user-facing docs https://github.com/polkadot-js/api/pull/375, and I'll use that technique instead of using `.raw`

packages/ui-signer/src/Extrinsic.tsx
ltfschoen: Agreed, as some extrinsics don't have an amount (i.e. `setCode`)

packages/ui-signer/src/Extrinsic.tsx
ltfschoen: Agreed, I'm surprised it didn't result in a linting error.

jacogr: Do you actually get null/undefined/something-invalid transferFee through? If so, there is an issue in api-obvervable.

ltfschoen: Yes, `transferFee` is `null`, so it's not being set to the default value of `new Balance(0)` further up when destructuring from props since that'd only work if `transferFee` was `undefined`, hence the reason why I'm trying to set it to `new Balance(0)` here instead. I'll have a look into it further.

jacogr:

PR #390, opened 2018-10-29
[next] Support latest substrate master

jacogr:
- Compact field encoding support
- Add Compact rendering support
- Aura support


jacogr:
Tough one, adding the "So Not Merge" label for now. The issue is that the latest Substrate master has `Header` with `Compact<BlockNumber>`, whereas older version such as BBQ does not. For the most part, not an issue - however bringing this in means that the `explorer` won't work on BBQ - the hashes are calculated wrong. (This is the only area where we have issue with support)

jacogr:
Ok, this is not a perfect solution, but to unblock the dam, just display an error message when unable to find the block

PR #392, opened 2018-10-31
[next] Block explorer event display

jacogr:
polkadot apps portal 2018-11-01 12-23-37


packages/app-explorer/src/Events.tsx
```diff
@@ -0,0 +1,75 @@
...
+    if (recentEvents.length === 0) {
+      return (
+        <div>{t('events.none', {
+          defaultValue: 'no non-internal events available'
+        })}</div>
+      );
+    }
+
+    return (
+      <pre>{JSON.stringify(recentEvents, null, 2)}</pre>
```
yjkimjunior: had a question about system events in general as I'm thinking about how this event data can be displayed in a more useable fashion. The data is just a number or hex so I'm not sure what I'm looking at when I read `{
    "index": "0x0200",
    "data": [
      7401
    ]
  },` for example.

jacogr: Yeap. This is still WIP - basically the Events has a section + method + arguments (each with a type). Next couple of pushes will start making sense of the data - for now (so I can actually see something), was just dumping the stringified JSON to make sure it does come in.

PR #404, opened 2018-11-05
moved shared function to parent

yjkimjunior:
reusing duplicated code

### DIFF
--- a/packages/app-staking/src/index.tsx
+++ b/packages/app-staking/src/index.tsx
@@ -86,6 +86,7 @@ class App extends React.PureComponent<Props, State> {
         </header>
         <Component
           balances={validatingBalances}
+          balanceArray={this.balanceArray}
           intentions={intentions}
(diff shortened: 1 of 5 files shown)


jacogr:
Nice, was not quite sure what is the right approach, this one works for me.

packages/app-staking/src/index.tsx
```diff
@@ -86,16 +86,29 @@ class App extends React.PureComponent<Props, State> {
...
       </main>
     );
   }
 
-  onMenuChange = (action: Actions) => {
+  private onMenuChange = (action: Actions) => {
     this.setState({ action });
   }
+
+  private balanceArray (_address: AccountId | string, _balances: RxBalanceMap): Array<Balance> | undefined {
```
jacogr: Not 100% sure we need to pass the balances through as well - we have the balances here as part of `validatingBalances`

Issue #304, opened 2018-09-13
Track extrinsics, don't allow until accountIndex is updated


jacogr:
Currently you can submit and then before the tx is included, submit again - in-between these the accountIndex is not updated since it is auto-pulled from the state.

Rather, keep track of the tx, don't allow sending while still in progress, or display a warning message around it, i.e. the second will replace the first. (Or whichever gets found and included first will go through)

We need https://github.com/polkadot-js/apps/issues/218

jacogr:
cc @chevdor (as discussed)

jacogr:
Closed in #406

PR #406, opened 2018-11-06
[WIP] Track extrinsics, don't allow until accountIndex is updated

yjkimjunior:
#304

Currently debugging but approaching this by:
1. onQueueExtrinsic -> check whether there is already a txn in queue with the same accountNonce
2. If yes -> then take current txn and set status to `blocked`
3. When result comes back from `submitAndWatchExtrinsic` set status to `queued`, with the nonce incremented and then re-sendExtrinsic

Another (easier) way [ the current WIP ] would be just to not queue txns before accountNonce is incremented, tell the user to wait a bit, and then tell them when they can try again.

### DIFF
--- a/packages/ui-signer/src/Extrinsic.tsx
+++ b/packages/ui-signer/src/Extrinsic.tsx
@@ -40,7 +40,7 @@ class Transaction extends React.PureComponent<Props> {
           <div className='expanded'>
             <p>
               <Trans i18nKey='decoded.short'>
-                You are about to sign a message from <span className='code'>{accountId}</span> calling <span className='code'>{section}.{method}</span> with an index of <span className='code'>{accountNonce.toString()}</span>
+                You are about to sign a message from <span className='code'>{accountId}</span> calling <span className='code'>{section}.{method}</span> with an index of <span className='code'>{accountNonce ? accountNonce.toString() : 0}</span>
               </Trans>
             </p>
             <Call value={extrinsic} />
--- a/packages/ui-signer/src/Queue.tsx
+++ b/packages/ui-signer/src/Queue.tsx
@@ -80,16 +80,9 @@ export default class Queue extends React.Component<Props, State> {
   }
 
   queueAdd = (value: QueueTx$Extrinsic | QueueTx$Rpc): QueueTx$Id => {
-    if (this.isDuplicateNonce(value)) {
-      this.queueSetStatus(nextId, 'error');
-      // TODO: Notify user to wait for prev txn to be finalized
-      console.error('Give it a second will ya');
-      return nextId; // unincremented id from the previous since we short circuit
-    }
-
     const id: QueueTx$Id = ++nextId;
 
-    console.log(`txn ${JSON.stringify(value)} with nonce -> `, value.accountNonce);
+    const _this = this;
 
     this.setState(
       (prevState: State): State => ({
@@ -99,7 +92,7 @@ export default class Queue extends React.Component<Props, State> {
             ? (value as QueueTx$Rpc).rpc
             : jsonrpc.author.methods.submitAndWatchExtrinsic,
           id,
-          status: 'queued'
+          status: _this.isDuplicateNonce(value) ? 'error' : 'queued'
         }])
       } as State)
     );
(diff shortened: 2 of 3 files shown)


yjkimjunior:
Just marking duplicate nonce txns as blocked, then calling queueUnclog on accountNonce update.

packages/ui-signer/src/Extrinsic.tsx
```diff
@@ -40,7 +40,7 @@ class Transaction extends React.PureComponent<Props> {
           <div className='expanded'>
             <p>
               <Trans i18nKey='decoded.short'>
-                You are about to sign a message from <span className='code'>{accountId}</span> calling <span className='code'>{section}.{method}</span> with an index of <span className='code'>{accountNonce.toString()}</span>
+                You are about to sign a message from <span className='code'>{accountId}</span> calling <span className='code'>{section}.{method}</span> with an index of <span className='code'>{accountNonce ? accountNonce.toString() : 0}</span>
```
jacogr: Still worried about this `accountNonce ? ` - it should be set here. If not, we have an underlying bug.

packages/ui-signer/src/Queue.tsx
```diff
@@ -81,17 +88,19 @@ export default class Queue extends React.Component<Props, State> {
     }
   }
 
-  private queueAdd = (value: QueueTx$Extrinsic | QueueTx$Rpc): QueueTx$Id => {
+  private queueAdd = (value: QueueTx$Extrinsic | QueueTx$Rpc | QueueTx): QueueTx$Id => {
     const id: QueueTx$Id = ++nextId;
     const rpc: RpcMethod = (value as QueueTx$Rpc).rpc || SUBMIT_RPC;
 
+    const _this = this;
```
jacogr: Do we need this? With the `=>` functions, it is assigned properly.

packages/ui-signer/src/Queue.tsx
```diff
@@ -115,4 +124,21 @@ export default class Queue extends React.Component<Props, State> {
       values
     });
   }
+
+  queueUnclog = (accountNonce: BN): void => {
+    const { queue } = this.state;
+    let _this = this;
```
jacogr: `this`?

packages/ui-signer/src/Extrinsic.tsx
yjkimjunior: Moved this to a separate issue so it can be tracked. #450

packages/ui-signer/src/types.d.ts
```diff
@@ -9,52 +9,52 @@ import { Base, UInt } from '@polkadot/types/codec';
 import { Hash, Extrinsic } from '@polkadot/types';
 import { RawParam$Value } from '@polkadot/ui-app/Params/types';
 
-export type QueueTx$Status = 'finalised' | 'usurped' | 'dropped' | 'broadcast' | 'cancelled' | 'completed' | 'error' | 'incomplete' | 'queued' | 'sending' | 'sent';
+export type AccountInfo = {
+  accountId?: string | null,
+  accountNonce: UInt | BN
```
jacogr: Not a major thing, but since we are fiddling here, it can now just be `accountNonce: BN` (Uint now extends BN, so no need to import that anymore)

packages/ui-signer/src/types.d.ts
```diff
@@ -9,52 +9,52 @@ import { Base, UInt } from '@polkadot/types/codec';
...
   rpc: RpcMethod,
   values: Array<any>
 };
 
-export type PartialQueueTx$Rpc = ParialAccountInfo & {
-  rpc: RpcMethod,
-  values: Array<any>
+export type PartialAccountInfo = {
+  accountId?: string | null,
+  accountNonce?: UInt | BN | null
```
jacogr: As above, just `BN | null` (since we are making changes, not that anybody would have guessed this)

packages/ui-signer/src/Queue.tsx
```diff
@@ -55,6 +56,12 @@ export default class Queue extends React.Component<Props, State> {
     );
   }
 
+  private isDuplicateNonce = (value: QueueTx$Extrinsic | QueueTx$Rpc | QueueTx): boolean => {
+    const { queue } = this.state;
+
+    return queue.filter(item => item.accountNonce === value.accountNonce).length > 0;
```
jacogr: Since it is a BN, you would need to do `item.accountNonce.eq(value.accountNonce)`, `===` would just check for exact instances

jacogr: Additionally, it is unique per account, so this will do a global check, i.e. when you use 2 different accounts and have the same nonce, this would be an issue.

PR #407, opened 2018-11-06
[next] Handle null nonce in extrinsics (in addition to undefined)

jacogr:


packages/ui-signer/src/Queue.tsx
```diff
@@ -98,18 +99,18 @@ export default class Queue extends React.Component<Props, State> {
...
+      extrinsic
     });
   }
 
-  queueRpc = ({ accountNonce = new BN(0), accountId, rpc, values }: QueueTx$Rpc): QueueTx$Id => {
+  queueRpc = ({ accountId, accountNonce, rpc, values }: PartialQueueTx$Rpc): QueueTx$Id => {
     return this.queueAdd({
-      accountNonce,
       accountId,
+      accountNonce: accountNonce || new BN(0),
```
jacogr: The actual fix (same thing that is done on line 105) - the rest are just plumbing to get the types correct

Issue #72, opened 2018-06-08
Allow for Array<*> types in Param inputs


jacogr:
Currently we are not allowing params to be inputted

jacogr:
Expanding... 

This one is a bit involved, but do-able - requiring both logic & UI adaptations. Basically each Param gets rendered via https://github.com/polkadot-js/apps/blob/master/packages/ui-app/src/Params/index.tsx and each Param then takes care of the encoding of the values. So there needs to be a component that handles `Array<*>` and does the right thing, managing the sub-component indexes, their values and passing the encoded values up the tree.

Which sounds complicated, however it is the same thing that actually happens with the Params component. It manages a bunch of sub-params, each an entry in an array. Where the Arrays do become difficult (in this example) is (apart from the fact that Arrays can contain others or Tuples) that it needs to be expandable, i.e. each array can have more or less entries.

I would suggest - as an initial stab, remove some of the UI complexity, i.e. when an Array value is required -

1. Initialise the array (initValues) with a single entry of the correct type
2. Initially only cater Arrays where it only has a fixed length of one entry
3. Once the single entry is working, making it dynamic

As a bonus, with 1 & 2 in place, it is enough to cater for 99% of the use-case currently available.

jacogr:
Closed in #441 (next branch)

PR #421, opened 2018-11-08
[next] WIP: Fixes #417 but does not cater for #72.

### LINKED ISSUE #417: Display parameters for storage queries
Bring it back, currently missing after API upgrades for PoC-3 - https://github.com/polkadot-js/apps/pull/416/files#diff-b548c114245b6470a658f3c2840c5a26R58

ltfschoen:
* Fixes #417 (well it appears to fix it as shown in screenshot below), but only caters for Chain State queries with 0 or 1 parameter,
for >=2 parameters to be catered for we need to address #72

screen shot 2018-11-08 at 11 37 47

* Incomplete is resolving TSLint errors
(i.e. `Argument of type 'RawParam & TypeDef' is not assignable to parameter of type 'MethodParams'.
Property 'params' is missing in type 'RawParam & TypeDef'.`) when trying to combine the method
parameter's value and type into a single object (since the method parameter's value and type are
passed into the `onAdd` separately contained in the `values` and `params` of the `this.state`)

* Incomplete is knowing whether to apply the `isValid` property further.

* Incomplete is checking the functionality is correct in the "Raw key" tab of http://localhost:3000/#/chainstate
(where a hex-encoded storage key is provided) since functionality has only been tested in the "Module" tab.


jacogr:
Atm all the params are either only 0 or 1 (in this case of params for a map), so we will be safe for the time being

ltfschoen:
there's actually one that takes two parameters `councilVoting > councilVoteOf((Hash, AccountId)): bool`

jacogr:
Ahhh, well, that is actually still only 1 param, it is a Tuple. (Long-standing issue logged around those for support as well, same as Arrays, Params don't allow input yet)

packages/app-storage/src/Query.tsx
```diff
@@ -53,16 +53,16 @@ class Query extends React.PureComponent<Props, State> {
...
-      //   </span>
-      // ));
-      : [];
+    const { params } = value;
+    // FIXME - we need to cater for multiple params. see code prior to this commit
+    // FIXME - we haven't used params.isValid, should we?
+    // i.e. councilVoting > councilVoteOf(Hash, AccountId)
+    const inputs = params.map(({ isValid, type, value }) => (
+        <span key={`param_${type}`}>
+          {type}={valueToText(type, params[0].value)}
```
jacogr: So not sure here why we refer to `params[0]` when mapping through `params` - value is exploded

packages/ui-app/src/Params/types.d.ts
```diff
@@ -16,6 +17,7 @@ export type RawParam$Values = RawParam$Value | RawParam$ValueArray;
 
 export type RawParam = {
   isValid: boolean,
+  key: StorageFunction,
```
jacogr: This is not always the case, e.g. for Extrinsics, RawParam is used and it is not storage at all.

packages/app-storage/src/Selection/Modules.tsx
```diff
@@ -101,11 +110,18 @@ class Modules extends React.PureComponent<Props, State> {
...
+    // if the method has a parameter, store the value and
+    //  type of the method's parameter in the same object
+    if (values.length && params.length) {
+      newParams.push(Object.assign({}, values[0], params[0]['type']));
+    }
 
     onAdd({
       key,
-      params: values
+      params: newParams
```
jacogr: It would seem like all this bubbles up now, as opposed to just the `isValid` & `value`?

jacogr:

ltfschoen:
replaced by #496

PR #425, opened 2018-11-08
[next] Bare-bones QR display & reading

jacogr:
Does not do much yet... first bits would be to add external account, then interfacing with the signer app via the signer

Currently all the hooks are in place to -

- create accounts from QR
- provide extrinsics via QR to the signer

These are all untested (some issues will be in there), testing to start as soon as the infrastructure is ready. (Good progress is being made on the Signer itself in this regard)


jacogr:
Closing to be rebased on master

Issue #348, opened 2018-10-05
Browser crashes when run Raw RPC > state > getStorage or getStorageHashAt


ltfschoen:
When I navigate to http://poc-2.polkadot.io/#/rpc, and choose "state" > "getStorage", the browser crashes with the following error shown in the browser console:
<img width="494" alt="screen shot 2018-10-05 at 16 16 07" src="https://user-images.githubusercontent.com/6226175/46543951-2a108c80-c8ba-11e8-8e72-620befe9e3b8.png">

When I navigate to http://poc-2.polkadot.io/#/rpc, and choose "state" > "getStorageHashAt", the browser crashes with the following error shown in the browser console:
<img width="490" alt="screen shot 2018-10-05 at 16 18 08" src="https://user-images.githubusercontent.com/6226175/46544017-575d3a80-c8ba-11e8-95aa-9ff6304d2595.png">

jacogr:
This will be fixed for poc-2 - the API as it stands is completely backwards incompatible. (As you linked the API issue, that is the actual cause - once updated the next branch will pull it in.)

jacogr:
Should be done in next branch now since it pulls in the latest API

PR #430, opened 2018-11-09
refactor(ui-apps)

behzad888:
In semantic-UI dropdown we don't need to `componentDidUpdate` to change local item anymore.

### DIFF
--- a/packages/ui-app/src/Dropdown.tsx
+++ b/packages/ui-app/src/Dropdown.tsx
@@ -30,25 +30,6 @@ type SUIEvent = {
 };
 
 export default class Dropdown<Option> extends React.PureComponent<Props<Option>> {
-  componentDidMount () {
-    this.componentDidUpdate({} as Props<Option>);
-  }
-
-  componentDidUpdate (prevProps: Props<Option>) {
-    const { defaultValue, value } = this.props;
-    const startValue = isUndefined(value)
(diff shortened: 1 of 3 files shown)


behzad888:
@jacogr what about other changes?
I would be appreciated if you could review them.

behzad888:
Whoops!

You have used PureComponent such as `DropDown` in `InputAddress` and used it in `Editor`. If you have a mistake in the kid most Component lifecycle all of the render methods will be called double time. 
You are right, we can use triggers and I think we can use `OnChange` method of the `InputAddress` in its `componentDidMount`. If you agree with me we can keep up this PR open and refactor all components that used `DropDown` and then we should refactor all parent components.

@jacogr what do you think?

jacogr:
First point - quite happy to have things done correctly.

On the second point - look at the "next"branch and the next deployed version https://polkadot.js.org/apps/next/ - it is one one that has all the current changes - basically with the upcoming PoC-3, the API has been completely reworked. (That crash is issue https://github.com/polkadot-js/apps/issues/348). It is a bit messy since PoC-2 (current) and PC-3 (next, around 2 weeks from now) are completely incompatible.

For major work, I would base it on the next branch, until that gets merged into master. (Master does not get much love atm)

jacogr:

PR #391, opened 2018-10-31
[next] add user notifications in App-Accounts

yjkimjunior:
Addresses #199 

still needs a couple small fixes:
- ~~notification should disappear~~
- ~~generalize Status component and move to `ui-app`~~ https://github.com/polkadot-js/apps/pull/413
- integrate with other the apps where appropriate, as per jaco's comments

PR #447, opened 2018-11-13
[next] Backup warning modal on account creation #394

### LINKED ISSUE #394: Warning Modal to Backup Seed Phrase Should Have The Seed Phrase in it
screen shot 2018-11-02 at 1 30 31 pm

Just need to add the seed phrase component from create screen and hover to copy.

Stefie:
- Added the `<AddressSummary />` component (as suggested by @yjkimjunior)
- Added forced backup file download on save (instead of showing additional download button)
- Changed text from 'Warning' to 'Important notice'
- Fixes #394

TODO: Implement user feedback as notification pop-up in the style of https://github.com/polkadot-js/apps/pull/391 once #391 is merged

### DIFF
--- a/packages/app-accounts/src/Creator.tsx
+++ b/packages/app-accounts/src/Creator.tsx
@@ -182,17 +182,6 @@ class Creator extends React.PureComponent<Props, State> {
           </Input>
         </div>
         <div className='ui--row'>
-          <Input
-            className='full'
-            isError={!isNameValid}
-            label={t('creator.name', {
-              defaultValue: 'name the account'
-            })}
-            onChange={this.onChangeName}
-            value={name}
-          />
-        </div>
-        <div className='ui--row'>
           <Password
             className='full'
             isError={!isPassValid}
--- a/packages/app-accounts/src/Creator.tsx
+++ b/packages/app-accounts/src/Creator.tsx
@@ -13,6 +13,7 @@ import { encodeAddress } from '@polkadot/keyring';
 import keyring from '@polkadot/ui-keyring/index';
 
 import translate from './translate';
+import FileSaver from 'file-saver';
 
 const BipWorker = require('worker-loader?name=[name].[hash:8].js!./bipWorker');
 
@@ -181,6 +182,17 @@ class Creator extends React.PureComponent<Props, State> {
           </Input>
         </div>
         <div className='ui--row'>
+          <Input
+            className='full'
+            isError={!isNameValid}
+            label={t('creator.name', {
+              defaultValue: 'name the account'
+            })}
+            onChange={this.onChangeName}
+            value={name}
+          />
+        </div>
+        <div className='ui--row'>
           <Password
             className='full'
             isError={!isPassValid}
@@ -194,6 +206,7 @@ class Creator extends React.PureComponent<Props, State> {
         <Modal
           dimmer='inverted'
           open={showWarning}
+          size='small'
         >
           {this.renderModalContent()}
           {this.renderModalButtons()}
@@ -220,7 +233,7 @@ class Creator extends React.PureComponent<Props, State> {
             isPrimary
             onClick={this.onCommit}
             text={t('seedWarning.continue', {
-              defaultValue: 'Continue'
+              defaultValue: 'Create and backup account'
             })}
           />
         </Button.Group>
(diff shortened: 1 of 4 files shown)


packages/app-accounts/src/Creator.tsx
```diff
@@ -345,6 +354,16 @@ class Creator extends React.PureComponent<Props, State> {
       ? keyring.createAccountMnemonic(seed, password, { name })
       : keyring.createAccount(formatSeed(seed), password, { name });
 
+    try {
+      const json = pair.toJson(password);
```
Stefie: Is the `meta.whenCreated` timestamp property in the backup file needed somewhere? 
It's only being added `onCreateAccount()` / `saveAccount()` so it's not part of the `pair` object here. If it has to be included, I'd have to wait until the account has been created and then re-request it with keyring.getPair()

jacogr: It is mostly app-specified. At this point it really does not get used anywhere, although it could be useful at some point, i.e. we could quite feasibly show on the Account overview when the account was created and/or updated.

packages/app-accounts/src/Creator.tsx
```diff
@@ -230,17 +232,24 @@ class Creator extends React.PureComponent<Props, State> {
...
         {t('seedWarning.header', {
-          defaultValue: 'Warning'
+          defaultValue: 'Important notice!'
         })}
       </Modal.Header>,
       <Modal.Content key='content'>
+        <p>
         {t('seedWarning.content', {
-          defaultValue: 'Before you continue, make sure you have properly backed up your seed in a safe place as it is needed to restore your account.'
+          defaultValue: 'We will provide you with a generated backup file after your account was created. Plesae make sure to save this file in a secure location as it is the only way to restore your account. As long as you have access to your account you can always redownload this file later.'
```
jacogr: Small typo `Plesae`

packages/app-accounts/src/Creator.tsx
jacogr: Put another way - it is part of the metadata (i.e. `pair.meta`), so it can really be anything that we deem useful. (Or even the user does, i.e. at some point it would bvre great if they can add their own fields)

The only real part of meta that is currently visible to the user is the `name` property.

packages/app-accounts/src/Creator.tsx
Stefie: Oh, that's fixed now

packages/app-accounts/src/Creator.tsx
Stefie: Hmm... it's also the only property that is not generated on keyring.createAccountMnemonic/ keyring.createAccount but on keyring.saveAccount, which is called in both create functions.

Would it be ok to move the check from `saveAccount` to a separate method which then can be called from within the create functions?  I just pushed a suggestion. Like this, the `whenCreated` timestamp would be part of the initial `pair` object that is being returned.

packages/app-accounts/src/Creator.tsx
```diff
@@ -230,17 +232,24 @@ class Creator extends React.PureComponent<Props, State> {
...
         {t('seedWarning.header', {
-          defaultValue: 'Warning'
+          defaultValue: 'Important notice!'
         })}
       </Modal.Header>,
       <Modal.Content key='content'>
+        <p>
         {t('seedWarning.content', {
-          defaultValue: 'Before you continue, make sure you have properly backed up your seed in a safe place as it is needed to restore your account.'
+          defaultValue: 'We will provide you with a generated backup file after your account was created. Please make sure to save this file in a secure location as it is the only way to restore your account. As long as you have access to your account you can always redownload this file later.'
```
yjkimjunior: grammar nitpick - ...after your account ~~was~~ _is_ created...

Stefie: That's totally fine! As a non-native speaker I'm happy get corrected on things like that

packages/app-accounts/src/Creator.tsx
```diff
@@ -230,17 +232,24 @@ class Creator extends React.PureComponent<Props, State> {
...
         {t('seedWarning.header', {
-          defaultValue: 'Warning'
+          defaultValue: 'Important notice!'
         })}
       </Modal.Header>,
       <Modal.Content key='content'>
+        <p>
         {t('seedWarning.content', {
-          defaultValue: 'Before you continue, make sure you have properly backed up your seed in a safe place as it is needed to restore your account.'
+          defaultValue: 'We will provide you with a generated backup file after your account is created. Please make sure to save this file in a secure location as it is the only way to restore your account. As long as you have access to your account you can always redownload this file later.'
```
yjkimjunior: Style nitpick - how about we separate informational text from the call to action text? Just for readability. And some padding between the message and the identicon. 
screen shot 2018-11-13 at 9 41 05 am

packages/app-accounts/src/Creator.tsx
jacogr: 100%, makes sense. (Good to have eyes on that)

packages/app-accounts/src/Creator.tsx
Stefie: I added the option to use  <Modal.Highlight />, it has the same properties as https://react.semantic-ui.com/elements/header/.
screen shot 2018-11-13 at 19 35 48

Stefie: Hmm, that didn't work out as expected. I can't rename the semantic-ui-react modules, the build is failing because of that. I'm on it...

PR #448, opened 2018-11-13
[next] Display intention/validator name in staking list #303

### LINKED ISSUE #303: Display intention/validator name in staking list

Stefie:
- Added missing getAccount() method to ui-keyring package (from address and public key)
- Added getDisplayName() as method to `CurrentList`
- Fixes #303

### DIFF
--- a/packages/app-staking/src/Overview/CurrentList.tsx
+++ b/packages/app-staking/src/Overview/CurrentList.tsx
@@ -8,6 +8,7 @@ import { RxBalanceMap } from '@polkadot/api-observable/types';
 import React from 'react';
 import { AccountId, Balance } from '@polkadot/types';
 import { AddressMini, AddressRow } from '@polkadot/ui-app/index';
+import keyring from '@polkadot/ui-keyring/index';
 
 import translate from '../translate';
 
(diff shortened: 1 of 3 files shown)


Stefie:
Nice, thanks! 
But I think I don't have permissions to merge my pull requests. So you probably need to do that...

Issue #459, opened 2018-11-23
Suggest: bytes state need have the button of spread and copy


jiangfuyao:
<https://polkadot.js.org/apps/next/#/chainstate>

![_ _20181123153334](https://user-images.githubusercontent.com/34789555/48932246-ca3b7700-ef35-11e8-8006-21d89dbdb5b9.png)
The “bytes” state show the abbreviated content, need have the button of spread and copy data.
likes:
![_ _20181123153613](https://user-images.githubusercontent.com/34789555/48932315-2900f080-ef36-11e8-8fc5-eec6978d925a.png)

jiangfuyao:
Or, have the copy runtime code and dowlond file.

jiangfuyao:
suggest: converting  to hex in the UI with wasm. local speed.

eimfach:
Hello,
I would like to implement this, I am missing the 'substrate' query state entry in my local application though.
I am running it like this ```WS_URL=wss://poc-2.polkadot.io:9944 yarn run start```

jiangfuyao:
And the "address" need the copy button. If not , very inconvenient.  

![_ _20181204152911](https://user-images.githubusercontent.com/34789555/49425953-e9f65900-f7d9-11e8-8a67-152ccb47990f.png)

![_ _20181204152931](https://user-images.githubusercontent.com/34789555/49425959-ee227680-f7d9-11e8-963c-7f31546a6192.png)

jacogr:
All address icons are clickable and copies the address - just over over the icon and click. (The issue here is that is is not 100% obvious and discoverable)

jacogr:
Spread introduced in #465

jacogr:
With spread in-place, we have a way of making available the full data and manually copying. Closing although technically not 100% completed.

Issue #460, opened 2018-11-23
Chain state echo: get bytes need remove the prefix


jiangfuyao:
<https://polkadot.js.org/apps/next/#/chainstate>
When to get the bytes data of chain state, the result will have prefix, not equal to the put bytes data.

**put** :

jacogr:
The display should match up here - the first is without length prefix, the second with. It is confusing, it should be handled consistently across the UI.

jiangfuyao:
And, I find: when the bytes (about) > 70k, the prefix is 3 bytes.  <70k is 2 bytes perfix.

jiangfuyao:
### Some time have 4 bytes prefix: 
When I put this png Img for **test**:
![example](https://user-images.githubusercontent.com/34789555/48981063-e55fdf80-f10b-11e8-88ff-a11a2d43796e.png)

**echo for put_data Uint8Array**:
![_ _20181125234210](https://user-images.githubusercontent.com/34789555/48981095-2f48c580-f10c-11e8-8a60-505881d70b9e.png)
**echo for get_data Uint8Array**:
![_ _20181125234000](https://user-images.githubusercontent.com/34789555/48981104-47b8e000-f10c-11e8-9c2e-c0c7186006df.png)

have **4** bytes prefix, is not 2.

#### UI for Uint8Array to hex:
-------------**put_data hex is:** 0x89504e470d0a1a0a0000000d494844520000059b00000325.......
**get_data hex is:** 0xaa09090089504e470d0a1a0a0000000d494844520000059b00000325.......

jacogr:
100%. It is a display issue.

On the put side, it displays the value without the prefix. However on the get side, it displays it with. (Now just need to make it consistent)

The number of length bytes will vary on the actual size. For small Vec<u8>, it will have a single byte, this can be up to 5 characters for very large values (it is always either 1, 2, 4 or 5 depending on the size being encoded - it is Compact<u32> encoding)

PR #465, opened 2018-11-26
#459 Add spread and copy buttons for substrate.code query

eimfach:
Hey there,

I implemented the spread and copy buttons for the substrate.code query..
The buttons have yet no functionality because I don't understand how the queries functionality is wired. I saw some code and modules which handle this functionality and ended there: https://github.com/polkadot-js/api/blob/master/packages/type-storage/src/substrate.ts

So yes, this would be my 'first revision' approach to this issue.

Greetings,

Robin

@tomusdrw

### DIFF
--- a/packages/app-storage/src/Query.tsx
+++ b/packages/app-storage/src/Query.tsx
@@ -26,8 +26,7 @@ type ComponentProps = {};
 
 type State = {
   inputs: Array<any>, // node?
-  Component: React.ComponentType<ComponentProps>,
-  showFullResultContent?: boolean
+  Component: React.ComponentType<ComponentProps>
 };
 
 type CacheInstance = {
--- a/packages/app-storage/src/Query.tsx
+++ b/packages/app-storage/src/Query.tsx
@@ -14,7 +14,7 @@ import { withObservableDiv } from '@polkadot/ui-react-rx/with/index';
 import { isU8a, u8aToHex, u8aToString } from '@polkadot/util';
 
 import translate from './translate';
-import { Bytes } from '@polkadot/types';
+import { ReactNodeArray } from 'prop-types';
 
 type Props = I18nProps & {
   onRemove: (id: number) => void,
@@ -102,9 +102,9 @@ class Query extends React.PureComponent<Props, State> {
   }
 
   private renderButtons () {
-
-    const specificFunctionalButtons = [];
-    const defaultButton =
+    const { key } = this.props.value as StorageModuleQuery;
+    const buttons = [] as ReactNodeArray;
+    const closeButton =
       <Button
         icon='close'
         isNegative
@@ -122,11 +122,11 @@ class Query extends React.PureComponent<Props, State> {
         text='copy'
       />;
 
-    if (this.props.value.key.meta.type.raw == 'Bytes') {
-      specificFunctionalButtons.push(spreadButton, copyButton);
+    if (key.meta.type.toString() === 'Bytes') {
+      buttons.push(spreadButton, copyButton);
     }
-    // array spread not usable ?
-    return [].concat(specificFunctionalButtons, defaultButton);
+
+    return buttons.concat([ closeButton ]);
   }
 
   private renderInputs () {
@@ -137,7 +137,7 @@ class Query extends React.PureComponent<Props, State> {
         <div className='ui--Param-text name'>:</div>
       );
     }
-
+    console.log(inputs);
     return [
       <div key='open' className='ui--Param-text name'>(</div>,
       inputs,
--- a/packages/app-storage/src/Query.tsx
+++ b/packages/app-storage/src/Query.tsx
@@ -100,8 +100,8 @@ class Query extends React.PureComponent<Props, State> {
     );
   }
 
-  private renderButtons() {
-    const { value: { key: {method, section} } } = this.props;
+  private renderButtons () {
+    const { value: { key: { method, section } } } = this.props;
 
     const specificFunctionalButtons = [];
     const defaultButton =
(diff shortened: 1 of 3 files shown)


packages/app-storage/src/Query.tsx
```diff
@@ -93,17 +93,42 @@ class Query extends React.PureComponent<Props, State> {
...
+        text='spread'
+      />;
+
+    const copyButton =
+      /* needs an copy content action (wasm byte code) */
+      <Button
+        text='copy'
+      />;
+
+    if (section === 'substrate' && method === 'code') {
```
jacogr: Due to the completely dynamic nature of Substrate, we would only check for specific sections in exceptional circumstances. Basically what we would rather do is when we render something of type `Bytes`, provide the copy functionality.

This dynamic nature is quite important, if we take a look at https://github.com/polkadot-js/apps/issues/460, we see that a developer is using the base libraries to build something and has a need in this case for copy as well. (In his case, it is also WASM and also quite large)

eimfach:
@jacogr Can you explain to me why it is necessary to create a component cache for the query result data in the actual state of a Query component (Why not cache the raw data) ? This looks a little overdone to me. Anyways I have a solution in mind and prepared to create the spread action, the caching through ```getDerivedStateFromProps``` is quite an obstacle though ...

eimfach:
@jacogr Could you have a look ?

packages/app-storage/src/Query.tsx
```diff
@@ -24,35 +26,67 @@ type ComponentProps = {};
...
         ? key.meta.type.toString()
         : 'Data';
-
-      cache[query.id] = withObservableDiv('rawStorage', { params: [key, ...values] })(
-        (value: any) =>
-          valueToText(type, value),
-        { className: 'ui--output' }
+      const defaultProps = { className: 'ui--output' };
+
+      /* render function to create an element for the query results which is plugged to the api */
```
jacogr: (insubstantial nitpick) Can we change this to be `//` comments - it just used everywhere else. (There are a number of these)

packages/app-storage/src/Query.tsx
```diff
@@ -93,17 +127,43 @@ class Query extends React.PureComponent<Props, State> {
         >
           <Component />
         </Labelled>
-        <Labelled className='storage--actionrow-button'>
-          <Button
-            icon='close'
-            isNegative
-            onClick={this.onRemove}
-          />
+        <Labelled className='.storage--queryrow-button'>
```
jacogr: Extra leading `.` to be removed

packages/ui-app/src/Params/valueToText.tsx
```diff
@@ -154,7 +154,7 @@ function valueToText (type: string, value: any, swallowError: boolean = true): R
     : div(
       {},
       ['Bytes', 'Data'].includes(type)
-        ? u8aToHex(value.toU8a(true), 512)
+        ? u8aToHex(value.toU8a(true), contentShorten ? 512 : value.length)
```
jacogr: The second part we can keep at `-1` instead of `value.length`

jacogr: That is an interesting one, good to know. With the size of the buffers, the conversion does take some time - with `-1` it should be auto-detected. (here https://github.com/polkadot-js/common/blob/master/packages/util/src/u8a/toHex.ts#L20)

So there _may_ be an underlying bug if they are not acting equivalently or causes lock-ups.

packages/app-storage/src/Query.tsx
```diff
@@ -93,17 +127,43 @@ class Query extends React.PureComponent<Props, State> {
...
+          {this.renderButtons()}
         </Labelled>
       </div>
     );
   }
 
+  private renderButtons () {
+    const { id, key } = this.props.value as StorageModuleQuery;
+
+    const buttons = [] as ReactNodeArray;
```
jacogr: Can we just use something like `Array<React.ReactNode>>` instead of pulling in `prop-types` (Thus far have not been using it at all since we rely on the TS definitions)

packages/app-storage/src/Query.tsx
```diff
@@ -24,35 +26,67 @@ type ComponentProps = {};
 
 type State = {
   inputs: Array<any>, // node?
-  Component: React.ComponentType<ComponentProps>;
+  Component: React.ComponentType<ComponentProps>
 };
 
-const cache: Array<React.ComponentType<ComponentProps>> = [];
+type CacheInstance = {
+  component: React.ComponentType<any>,
```
jacogr: Prefer `Component` naming to indicate it is a class.

packages/ui-app/src/Params/valueToText.tsx
eimfach: Hm it looks more like that if basically bitLength > -1 is given, the Uint8Array is simply 'stringified' by a reducer function rather that anything auto-detected ... Point being that there is a whole lot of integers converted and joined to hex strings which does block the main thread... Looks very costly.. Also I don't really know why Chrome does behave like this.. it feels like that the whole main thread keeps blocked even after 10 minutes and longer and even the DevTools Console is not accessible ..

I tried it with firefox and all went well, also I excluded host implementation problems with ```Uint8Array.prototype.reduce``` or ```Uint8Array.prototype.toString``` on my chrome (version 70.0.3538.110) in case they would not be catched by the transpiler beforehand ..

Also, ```@polkadot/util/u8a/toHex.js u8aToHex()``` has no solution to delegate the workload, it simply doesn't care how large the incoming data actually is. This could be an async operation with a promise returned for example...

Maybe a Service Worker ? :P

eimfach: / Or use wasm .. ?

jacogr: I'm sure the implementation there can be optimised. Service workers are a bit of an issue, since the utils are environment-agnostic. (Unless we do a specific impl. for the UI itself - which defeats the purpose of having utils.)

eimfach: I thought of an another basic js approach which is to delegate workloads to the event loop.
What we could do is to have a helper which creates a task queue which does create chunks of workloads and delegates them to the event loop (via setTimeout for example). Everything would be handled by promises step by step and the UI would not be blocked when a heavier workload is introduced.

Also there are different approaches, for example this one:
https://github.com/mozilla/task.js

PR #492, opened 2018-12-05
update the remote server location, closes #487

### LINKED ISSUE #487: You are not connected to a node. Ensure that your node is running and that the Websocket endpoint is reachable.
hi, i am a getting acquainted to this technology. i have been running the frontend for a while and it always connected to polkadot successfully. today it seems not to work, it might be due to me using a different network, although `ping` works reliably. did anything change? i am running the last `master`, f9afbeea9873d93cb166fc3df2cc17ee508e7bd5

danse:
also changed a path and fixed a typo

### DIFF
--- a/examples/tut-001.md
+++ b/examples/tut-001.md
@@ -37,14 +37,14 @@ Once the app is created, it needs to be hooked up to the rest, allowing TypeScri
 At this point the application will be available and included, even though it does not have a [navigation entry](../packages/apps/routing) yet. Adding the navigation -
 
 - Add a reference to the application to the [apps package.json](../packages/apps/package.json)
-- Add a route file the defines the [app, name & icon](../packages/apps/routing/example.ts)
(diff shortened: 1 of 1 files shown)


danse:
you are right, `poc-2` works. i collected some changes as a side effect of starting the tutorial :)

jacogr:
Thank you! 

This is a tough one though, the examples are _really_ outdated :( Actually so outdated that they are being dropped, i.e. we don't have them in the "next" branch anymore. (Which will become the next master branch as soon as Po-C-3 is launched.)

I'm still a bit up in the air as to what to do with the specific examples and how to replace them. For the API specifically, we now have examples on use, e.g. https://polkadot.js.org/api/examples/promise/

The UI ones are much tougher, there are 2 things at play -

1. People want to contribute to this repo - here the touch-this-touch that README will be more appropriate

2. People wanting to build their own UIs - we have been mulling around something like a react-create-app (or similar) around this.

PR #495, opened 2018-12-06
Extra information on address decode errors

jacogr:
Trying to get a handle and issue from @rphmeier (which I cannot quite reproduce atm - at the very least this will give more info)


ltfschoen:
Where was the issue recorded?

PR #496, opened 2018-12-06
[next] fix: Relates to #417. Display single parameters and tuples in storage query results

ltfschoen:
Extension of the closed PR https://github.com/polkadot-js/apps/pull/421.

Try in http://localhost:3000/#/chainstate using councilVoting's councilVoteOf which takes a tuple as a parameter


packages/app-storage/src/Query.tsx
```diff
@@ -85,16 +86,81 @@ class Query extends React.PureComponent<Props, State> {
...
+        inputs.push(<span key={`param_unknown`}>unknown</span>);
+        return;
+      }
+
+      // Case 1: single parameter, i.e.
+      //   isValid: true,
+      //   info: 1, // i.e. not a tuple
+      //   type: "AccountId",
+      //   value: "C123"
+      if (param.info === 1) {
```
jacogr: Use the type enum for these.

packages/app-storage/src/Query.tsx
```diff
@@ -85,16 +91,69 @@ class Query extends React.PureComponent<Props, State> {
...
+      //
+      // param = {
+      //   isValid: true,
+      //   value: ['0xABC', 'C123', '3'],
+      //   info: 3, // tuple with three elements
+      //   type: '(Hash, AccountId, BlockNumber)',
+      //   sub: [ { info: 1, type: 'Hash' }, { info: 1, type: 'AccountId' }, { info: 1, type: 'BlockNumber' }]
+      // };
+      if (param.info && param.info >= StorageQueryParameter.Tuple && param.sub && param.sub.length === param.info) {
+        console.log('mult');
```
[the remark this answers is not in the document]
ltfschoen: Sorry, fixed it

packages/ui-app/src/Params/types.ts
```diff
@@ -15,6 +15,9 @@ export type RawParam$Values = RawParam$Value | RawParam$ValueArray;
 
 export type RawParam = {
   isValid: boolean,
+  info?: number,
```
jacogr: Ok, if we have to do this, I would expect it to actually have something like a `typeDef` on this object, or even `type` containing the actual type information.

packages/app-storage/src/Query.tsx
```diff
@@ -37,6 +38,62 @@ type CacheInstance = {
 
 const cache: Array<CacheInstance> = [];
 
+enum StorageQueryParameter {
```
jacogr: Why not the actual types enum - here we now, when changing it in types, remember to change things here. Fragile :(

packages/ui-app/src/Params/types.ts
ltfschoen: oh i see what you mean now, i found a similar example here polkadot-js/api/types/codec/createType.d.ts

ltfschoen:
@jacogr i've pushed a couple of commits that i believe address your review comments. i've excluded implementations and associated tests that cater for `Compact` and `Vec` as I haven't got them working yet

packages/ui-app/src/Params/types.ts
```diff
@@ -13,7 +13,7 @@ export type RawParam$ValueArray = Array<RawParam$Value | Array<RawParam$Value>>;
 
 export type RawParam$Values = RawParam$Value | RawParam$ValueArray;
 
-export type RawParam = {
+export type RawParam = TypeDef & {
```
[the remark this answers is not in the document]
ltfschoen: i tried creating a commit https://github.com/polkadot-js/apps/commit/ab948b660dcc1545545f6ca1902dcf8e5a86c6ed in separate branch 'luke-417-display-params-for-storage-queries-3-typedef' to try using using a `typeDef` property internally, but I get errors when i run it because `param` doesn't have a `typeDef` property.

when i check the changes with `yarn run check` it doesn't show any errors so i thought the changes would also resolve the CI errors, but when i tried querying storage in the UI at http://localhost:3000/#/chainstate it gives error `Uncaught TypeError: Cannot read property 'info' of undefined` since `param` doesn't have a `typeDef` property, and it compiles from TypeScript to JS as `const paramInfo = param.typeDef.info;`

jacogr:
Closing. Instead of taking this approach, rather combine it all into https://github.com/polkadot-js/apps/issues/648 and just use `valueToText` to display. (No duplicated logic)

PR #497, opened 2018-12-06
[next] fix: Fixes #493. Contain logs digest item within div

### LINKED ISSUE #493: Logs (Seal) spills over on next (Substrate master)
Additionally, the log should actually be decoded into the correct params here

polkadot apps portal 2018-12-06 13-11-28

ltfschoen:
screenshot 2018-12-07 at 00 35 29

### DIFF
--- a/packages/app-explorer/src/index.css
+++ b/packages/app-explorer/src/index.css
@@ -57,7 +57,7 @@
   }
 
   .value {
-    font-family: monospace;
+    word-break: break-all;
   }
 }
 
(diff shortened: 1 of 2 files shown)


packages/app-explorer/src/index.css
```diff
@@ -55,6 +55,10 @@
       white-space: nowrap;
     }
   }
+
+  .value {
+    font-family: monospace;
```
jacogr: The logs aren't decoded properly here, it should have the actual values, i.e. the same as transactions/events. So monospace is not correct, neither is the actual dropdown above. (The wrapping however is)

PR #498, opened 2018-12-07
[next] Cleanup React & API errors

jacogr:


packages/ui-react-rx/src/with/observableBase.tsx
```diff
@@ -40,7 +40,12 @@ export default function withObservableBase<T, P> (observable: Observable<P>, { r
         this.setState({
           subscriptions: [
             observable
-              .pipe(map(transform))
+              .pipe(
+                map(transform),
+                catchError(() =>
+                  of(undefined)
```
jacogr: Handle errors, don't break component

PR #538, opened 2018-12-27
Tone down colors (secondary + primary)

jacogr:


jacogr:
Still not perfect, but less distracting.

Issue #500, opened 2018-12-07
Warn/block on exceptionally large payloads


jacogr:
... instead of sending to the Node, having it fail and then ending up with a spinner on-screen. The maximum payload size is now set in Substrate, which will drop connection if exceeded. (The message not even parsed, so no response at all - just a dropped connection)

https://github.com/paritytech/substrate/pull/1231/files#diff-435a815e40d1829b9fd17fb03813d2edR36

Rather be a bit more conservative here with a max 10MB payload size (`const MAXIMUM_PAYLOAD = 10 * 1024 * 1024`), we can increase if users validly bump against the limit. (It is mostly WASM contracts as well as runtimes that could get here - as of now the runtime code is an order of a magnitude smaller)

Related to https://github.com/paritytech/substrate/issues/1160

jacogr:
- Closed in #621 
- Follow-up v2 in #649

PR #621, opened 2019-01-19
Warn user when RPC request size is >10MB

maxwellfoley:
Responding to issue #500


maxwellfoley:
I ended up moving the new code to a different component entirely because I couldn't figure out how to get the function "section" name from the extrinsic object in order to properly print the warning.

maxwellfoley:
👍

Issue #644, opened 2019-01-25
Allow user to manually enter node url/ip in settings


xlc:
It will be good if user can enter url/ip of their node in the settings. Maybe popup some warning if trying to access ws endpoint from https UI.

jacogr:
So would suggest -

- selection for pre-defined vs custom
- custom allows entry into input 
- validation checks for ws:// (loclahost/127.0.0.1 only) or wss:// prefixes
- save this to localstorage along with setting (pre/custom)
- inject URL upon startup (apps)

maxwellfoley:
submitted a PR here: https://github.com/polkadot-js/apps/pull/657

jacogr:
Closed in #657 - thanks a million!

PR #657, opened 2019-01-27
Allow custom node URLs in settings (issue #644)

maxwellfoley:


jacogr:
Brilliant, thank you.

packages/app-settings/src/index.tsx
```diff
@@ -178,9 +210,40 @@ class App extends React.PureComponent<Props, State> {
...
+  }
+
   private save = (): void => {
+    const { onStatusChange } = this.props;
     const { settings: { types, typesError } } = this.state;
 
+    // validate custom node url
+    const apiUrl = this.state.settings.apiUrl;
+
+    if (this.state.settings.customNode) {
```
ltfschoen: Are we satisfying all the of RFC 6455 "4.1 Client Requirements" https://tools.ietf.org/html/rfc6455#section-4.1, such as validating the WebSocket URIs by parsing it before connecting to it?

Are we always validating server data by safely parsing messages received on the client-side?

ltfschoen: * Are we transporting any MSRP traffic over WebSockets? If so, should we only allow WSS and have the client authenticate the server by validating the server's certificate? Even if the user is using WSS should we not indicate end-to-end security to the user as specified by RFC 7977 "9. Security Considerations" https://tools.ietf.org/html/rfc7977#page-24?
* Are there any risks associated with using a custom WSS or WS connection that we should communicate to the user?
* If they connect using WS then what are they exposing to a potential MITM attack?
* Are we exposing the user to [Cross-Site WebSocket Hijacking (CSWSH)](https://content-security-policy.com/)?
* Do we need to add a CSP like `<meta http-equiv="Content-Security-Policy" content="connect-src 'self'">` to prevent WS requests, CORS, EventSource requests from anywhere other than the current server?

jacogr:
And also deployed to https://poc-3.polkadot.io/#/settings

PR #661, opened 2019-01-30
display params when querying chain storage

maxwellfoley:
issue #417

Right now I only have tested this with the types that are currently used as parameters for the various functions available, I can't guarantee that it will work for any conceivable type that may one day be used as a parameter to a function - should ensuring that be within the scope of this issue? If so I can make this more robust. 

I also made some calls on graphic design that included re-arranging some of the existing output so that the placement of the input parameters would make visual sense. Obviously this is up to taste so let me know if this is to your liking or not.

### DIFF
--- a/packages/app-storage/src/Query.tsx
+++ b/packages/app-storage/src/Query.tsx
@@ -16,6 +16,8 @@ import { isU8a, u8aToHex, u8aToString } from '@polkadot/util';
 import translate from './translate';
 import { RenderFn, DefaultProps, ComponentRenderer } from '@polkadot/ui-api/with/types';
 
+import './Query.css';
+
 type Props = I18nProps & {
   onRemove: (id: number) => void,
   value: QueryTypes
@@ -107,17 +109,20 @@ class Query extends React.PureComponent<Props, State> {
     return (
       <div className='storage--Query storage--actionrow'>
         <div className='storage--actionrow-value'>
+          {this.keyToName(key)}
+          <hr/>
           <Labelled
             label={
-              <div className='ui--Param-text'>
-                {this.keyToName(key)}: {
-                  isU8a(key)
-                    ? 'Data'
-                    : key.meta.type.toString()
-                }
+              <div>
+                {this.renderInputs()}
               </div>
             }
           >
+          <label>
+          Output: ({
+            isU8a(key) ? 'Data' : key.meta.type.toString()
+          })
+          </label>
             <Component />
           </Labelled>
         </div>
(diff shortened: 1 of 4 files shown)


jacogr:
Sorry, I am _very_ behind in reviewing here, as well as the previous PR - my humble apologies.

packages/app-storage/src/Query.tsx
```diff
@@ -130,6 +135,58 @@ class Query extends React.PureComponent<Props, State> {
...
+
+    // if function has inputs, render them
+    if ((value as StorageModuleQuery).params.length > 0) {
+      const type = isU8a(key) ? 'Data' : key.meta.type.asMap.key.toString();
+      const params = (value as StorageModuleQuery).params;
+      return (
+        <div>
+          Inputs: ({type})
+          {
+            params.map((obj) => {
```
jacogr: This would actually be better in valueToText itself so it is re-usable. Basically here we should be really light on logic and just pass the stuff we get in as-is for formatting. (In this app we are, sadly, already quite logic heavy I believe - blame me.)

packages/ui-params/src/valueToText.tsx
```diff
@@ -150,6 +151,12 @@ function valueToText (type: string, value: any, swallowError: boolean = true, co
   //   }
   // }
 
+  // When displaying parameters of type Hash, these will sometimes come in the
```
jacogr: Ahh, ok. So basically what happens - when it comes from external sources (i.e. via the API), it would be a Codec type. (So Hash). However the UI actually captures these as Uint8Array, hence the discrepancy.

maxwellfoley:
Ok. The reason I put the logic of breaking up the tuples in Query.tsx and not valueToText.tsx is for reasons related to visual presentation but looking at it now I think I can move the logic into valueToText and still have the same effect, will do.

jacogr:
Just traveling atm. Will get to a proper look ASAP. Took a brief look and it looks sane. (Arrays can be handled the same way, so it won’t be an nightmare to add)

packages/app-storage/src/Query.tsx
```diff
@@ -130,6 +135,36 @@ class Query extends React.PureComponent<Props, State> {
...
+              // check if the type name references a tuple, e.g. "(Hash, AccountId)"
+              return (
+                <div className='query-input'>
```
jacogr: We would need a `key` attribute here to keep React happy and now spam to console, small enough.

packages/app-storage/src/Query.tsx
```diff
@@ -107,17 +109,20 @@ class Query extends React.PureComponent<Props, State> {
     return (
       <div className='storage--Query storage--actionrow'>
         <div className='storage--actionrow-value'>
+          {this.keyToName(key)}
+          <hr/>
```
jacogr: Not 100% on-board with this layout. And it is probably 100% to-do with my updated label placement which used to be on-top, but is now on the side. I believe it would be more useful having it on a single line with something like -

`system.accountNonce(5HYmsxGRAmZMjyZYmf7uGPL2YDQGHEt6NjGrfUuxNEgeGBRN): Index`

or even 

`system.accountNonce(AccountId: 5HYmsxGRAmZMjyZYmf7uGPL2YDQGHEt6NjGrfUuxNEgeGBRN): Index`

In the above we will probably start running into some issues anyway, so we will probably need to forgo the label for the output (which is now empty anyway) allow the box to stretch the whole length and then keep the placement of the above where you currently have `Output`

If it becomes messy with the full display (e.g. as per the full address above), give that part a 8rem width and set overflow so it renders something like -

`system.accountNonce(AccountId: 5HYmsxG...): Index`

(And feel free to swap the type or value above)

jacogr:
Closing as stale. Feel free to re-open at any point, this is a great start.

PR #666, opened 2019-01-31
Javascript & API console app

jacogr:
POC, so not much, but it runs, here subscribing to new blocks -

image

- v0.1, so only access is directly via `/#/js`
- Probably will only have available when connected to local dev chains (???)
- very basic editor (top and bottom lines are not editable, it is what is "injected", i.e. you can use `await`, have api, keyring, util and hashing injected)
- `console.{log,error}` works as expected (need to trap more of those)
- need a way to stop functions executing, no idea how as of yet
- possibly hook into the signer for transactions (? - v2, initially assume dev mode)
- save current & resume on re-load, save snippets (maybe, v2)
- nice would be an actual interactive tutorial here...


xlc:
Another way is just expose all the relevant objects to global and just Chrome debug console, which have good snippet support and good js editor

jacogr:
That needs to be done regardless. Thanks for the reminder, it has been in the back of my mind, but I have not snuck it in somewhere yet.

PR #673, opened 2019-02-04
Forward-port low-hanging breaks

### LINKED ISSUE #671: Account page redesign feedback
Relates to user feedback here https://matrix.to/#/!HzySYSaIhtyWrwiwEV:matrix.org/$1549062202142806JJnlI:matrix.org?via=matrix.org&via=matrix.parity.io&via=cryptonomic.tech&via=t2bot.io

Currently when the browser window width  is less than < 1280px, it shows across two rows as shown below, with opportunities to improve the second row (i.e. reduce the left padding, align it to the left, reduce the width so the 'using my account' selection box doesn't have so much whitespace between the identicon and the address.

screen shot 2019-02-02 at 10 44 54 am
 
And when the browser width becomes more than > 1280px, there are opportunities to also reduce the width so the 'using my account' selection box doesn't have so much whitespace between the identicon and the address, and aligning the labels to the left (whilst having both selection boxes being vertically aligned.

screen shot 2019-02-02 at 10 44 44 am

**Proposed wireframes**

not-split

split

jacogr:
(From the still WIP `jg-ui-break` branch)

- `ui--grid` and `ui--row` now defaults to `nowrap` (this solves breaks in eg. Accounts, Closes #671)
- Hide some stuff in apps based on media-query size (ui-app/styles/media.css)
  - explorer - summary items + author for a block (this solves my specific phone-view use-case)
  - transfer - hide recipient on medium, sender on small - the inputs are ok-ish
  - 2-column apps - on small screens lose 2-column layout (media flex)
- Half-ports, i.e. not close to 100%, but better than current master -
  - linked dropdowns (e.g. chain state section  + method), has smaller sizing allowance for first item, allowing for better with overlaps on small screens, but nowhere close to perfect

Nowhere near comprehensive, these are just the forward-ported items (really quick basics) that more-or-less aligns with the "show when available" approach taken with apps. Going through app by-app atm as a first step.

What is not covered, but hopefully will be when the rest is available (after a short break) -

- linked dropdown proper solution (probably want 33/67% here in smaller cases)
- overlaps in medium + medium items, i.e. something like the account info on extrinsics page
- active proposals and referendums are still on default layout only
- modals not adjusted

### DIFF
--- a/packages/ui-app/src/styles/media.css
+++ b/packages/ui-app/src/styles/media.css
@@ -2,41 +2,53 @@
 /* This software may be modified and distributed under the terms
 /* of the Apache-2.0 license. See the LICENSE file for details. */
 
+/* block by default, flex as per media queries below */
+.ui--flex-large,
+.ui--flex-medium,
+.ui--flex-small {
+  display: block;
+}
+
+/* hide all by default, add as per media queries below */
 .ui--media-large,
 .ui--media-medium,
 .ui--media-small {
+  height: 0;
   visibility: hidden;
   width: 0;
 }
 
-/*
-  ##Device = Desktops
-  ##Screen = 1281px to higher resolution desktops
-*/
 @media (min-width: 1281px) {
+  .ui--flex-large {
+    display: flex;
+  }
+
   .ui--media-large {
+    height: auto;
     visibility: visible;
     width: auto;
   }
 }
 
-/*
-  ##Device = Laptops, Desktops
-  ##Screen = B/w 1025px to 1280px
-*/
 @media (min-width: 1025px) {
+  .ui--flex-medium {
+    display: flex;
+  }
+
   .ui--media-medium {
+    height: auto;
     visibility: visible;
     width: auto;
   }
 }
 
-/*
-  ##Device = Tablets, Ipads (portrait)
-  ##Screen = B/w 768px to 1024px
-*/
 @media (min-width: 768px) {
+  .ui--flex-small {
+    display: flex;
+  }
+
   .ui--media-small {
+    height: auto;
     visibility: visible;
     width: auto;
   }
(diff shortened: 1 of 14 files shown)


packages/ui-app/src/styles/media.css
```diff
@@ -0,0 +1,43 @@
+/* Copyright 2017-2019 @polkadot/ui-app authors & contributors
+/* This software may be modified and distributed under the terms
+/* of the Apache-2.0 license. See the LICENSE file for details. */
+
+.ui--media-large,
+.ui--media-medium,
+.ui--media-small {
```
Stefie: Your naming convention for class names is different in most other files:
The first letter after the `--` should be uppercase: `.{appname}--{Specific-class}`

jacogr: Not for ui— (eg. ui—row,grid,...)

Those are “globals” that apply anywhere.

jacogr:
Needs some playing here before merging - just to make sure the low hanging doesn’t break anything. Eg. Staking may not display actions properly with the half-port. So first a play through is required, may need to bring in more than anticipated.

Stefie:
I played through with the latest master and everything works fine!

One side-note: I generally liked it better, when the info in boxes in the top row stayed visible on smaller screens (on explorer, transfer, democracy) because I'd always favour having the same information on different screen sizes. Maybe making the elements more dense / the font smaller could be an option instead of hiding them.

jacogr:
100% at this point it is too rough, some info is very much extra, but it would be preferable if it can stay and be displayed smaller or differently. 

Opted for the rough-cut version initially (def. don’t see tho as the end result) for ease of implementation and getting the ugly reflows gone. 

For summaries we need to follow a different approach I think, they take up a lot of space but don’t really need to - so no reason why it can’t fit with a cleverer way of displaying.

Issue #676, opened 2019-02-06
Display all numbers in the same unit for consistency


Stefie:
From the [Polkado Watercooler Riot chat](https://matrix.to/#/!fOOzymDEHiIIUtmlBE:matrix.org/$1549377929898489zhgwz:matrix.org?via=matrix.org&via=matrix.parity.io&via=status.im):

> i tried to send 10DOTs but it says i dont have the necessary balance
<img width="1059" alt="image" src="https://user-images.githubusercontent.com/125398/52338343-70156280-2a0a-11e9-86da-64dc38ed604b.png">

We're currently using different units to display the account balance, the fees and transfer value, which can be confusing.
We should think of a way to display all values in the same unit to make the ui more intuitive.

https://github.com/polkadot-js/apps/issues/270 might be related?

jacogr:
270 is not quite related. That was in poc-2 where we didn’t have using yet. So the suggestion was to add micro instead of the configurable (basically set per chain in the spec file) balances. 

It would probably help if the text display does “mDOT” in the above screens. Or even “milli DOT”

PR #677, opened 2019-02-06
Show balance units with denomination (inline-text)

jacogr:
Closes https://github.com/polkadot-js/apps/issues/676

polkadot apps portal 2019-02-06 20-32-43

### DIFF
--- a/packages/ui-app/src/util/formatBalance.ts
+++ b/packages/ui-app/src/util/formatBalance.ts
@@ -18,6 +18,7 @@ interface BalanceFormatter {
   getDefaultDecimals (): number;
   getOptions (decimals?: number): Array<SiDef>;
   setDefaultDecimals (decimals: number): void;
+  setDefaultUnits (unit: string): void;
 }
 
 const SI: Array<SiDef> = [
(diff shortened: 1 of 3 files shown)


packages/ui-app/src/util/formatBalance.ts
```diff
@@ -29,7 +30,7 @@ const SI: Array<SiDef> = [
   { power: -9, value: 'n', text: 'nano' },
   { power: -6, value: 'µ', text: 'micro' },
   { power: -3, value: 'm', text: 'milli' },
-  { power: 0, value: '-', text: '----' }, // position 8
+  { power: 0, value: '-', text: 'Unit' }, // position 8
   { power: 3, value: 'k', text: 'Kilo' },
```
Stefie: This should get the 'more than unsubstantial' label, but is there a reason why Kilo is the only unit larger than X / DOT where the token symbol is lowercase (k), all the others are uppercase (M, G, etc). I was asking myself the same question this morning, when I was looking into this issue...

jacogr: https://en.wikipedia.org/wiki/International_System_of_Units#Prefixes

Stefie: Thanks!

PR #678, opened 2019-02-07
better handling of large balance values

xlc:


packages/ui-app/src/util/formatBalance.ts
```diff
@@ -45,7 +45,7 @@ const SI_MID = 8;
 let defaultDecimals = 0;
 
 export function calcSi (text: string, decimals: number = defaultDecimals): SiDef {
-  return SI[(SI_MID - 1) + Math.ceil((text.length - decimals) / 3)];
+  return SI[(SI_MID - 1) + Math.ceil((text.length - decimals) / 3)] || SI[SI.length - 1];
```
jacogr: First off, sorry about the delay. (As you are aware I'm normally a bit faster...) 

I'm not 100% on this one. It basically means we don't take the chain decimals into account for the fallback. A fallback should be provided (no doubt), however it will always go to largest here and not make sense with the decimals used? 

(Have really not seen this issue represent itself, so cannot quite play to see what is the effect and if my fears are unfounded. So just unsure here)

jacogr: Ok, thinking through it again - it basically will (when overflowing), choose the last value available. So my fears there are unfounded.

So should be fine, will just display >3 decimals on the top-end.

Issue #675, opened 2019-02-05
Create API for app-js that can be cleaned up


jacogr:
Basically we need some way of terminating all the subscriptions started in the app-js when the user stops or runs a new one. Currently since the subs are created inside the function, they will never be cleaned. (Good enough as a first step, but will need cleanups at some point)

Basically, we would need to track and have some way to destroy, off the top of my head what can work -

- Create our own instance of an API-alike object we pass in, basically just a new object with all the api object properties copied
- for rpc, tx, derive & query, we do some magic - create a proxy that we expose here, each actually storing references to the results of the calls (i.e. `Promise<unsubfn>` internally
- when we call a `destroy` on this object, it loops through that list, cleanup up all subs

(Probably worth some expanded description above, will do so when I have a gap, for now just jotting down to not get lost - maybe the info is enough, probably not)

Not immediately critical, but we would need to start cleanups at some point for proper use.

jacogr:
On second thought, the idea of having a destoy-able API is spot-on, approach above is probably not the best.

- make API clone-able (provider, instance)
- disconnect (requires https://github.com/polkadot-js/api/pull/664)

This way we use something relatively standard, instead of cherry-picking.

jacogr:
Integrated.

PR #681, opened 2019-02-08
Pull in API with clone & disconnect, use in #/js

jacogr:
- Creates a specific API instance for the JS console (via clone), allowing it to be disposed of (via disconnect), cleaning all subscriptions and state associated with it
- Add stop next to run command
- Closes https://github.com/polkadot-js/apps/issues/675


jacogr:
Closed in favour of #686 (merging into @Stefie 's branch directly)

Issue #670, opened 2019-02-01
Split js Editor/Output areas into tabs


jacogr:
This makes https://github.com/polkadot-js/apps/issues/668 more usable as well

Create 2 tabs, one for the editor (taking up the full-screen real-estate available) and one for the Output component, once again taking up all space. (IT would be cool if this could auto-scroll to the bottom as well as output is added)

Additionally, when we execute, clear the console so each execution starts fresh. (Auto-swap to "Output" when execute, would work well here)

Stefie:
Closing in favour of: https://github.com/polkadot-js/apps/issues/692

Issue #668, opened 2019-02-01
Add selectable examples to js console


jacogr:
Port the examples (promise), from https://polkadot.js.org/api/ to the js console accessible via http://localhost:3000/#/js 

(The current default code included is basically a port of https://polkadot.js.org/api/examples/promise/02_listen_to_blocks/ - init skipped since we already have an instance of the api configured and passed-in)

To make this usable (off the top of my head) -

- Add top-level selector to swap between the samples, dropdown is fine
- Add examples in app-js/snippets
- ... make sure they all work :)

That is the first step to pre-select. ~~When anything edited, add it to a "Recent" (or something properly named) entry and allow selection thereof. Save these into localstorage. (If available and user comes back, load that one as the default and select it - so we always come back to the user code)~~
(EDIT SD: Moved to https://github.com/polkadot-js/apps/issues/685)

After this we should have -

- a nice toolset of available examples (we can edit/clarify the above quite a bit, even add some more - there are gaps, i.e. no derive atm, which should also go into the api examples)
- ~~have a way for the user to access his last-edited code~~

It moves us closer to having the interface sort-of usable and adding value to actual "I just want-to-play" coders. It is not a replacement for anything resembling building real-apps, but nice as a testing and checking env.

Stefie:
Closing. Remaining todo's moved to:
https://github.com/polkadot-js/apps/issues/685 and
https://github.com/polkadot-js/apps/issues/692

PR #684, opened 2019-02-09
Sd 668 app console

Stefie:
Adresses https://github.com/polkadot-js/apps/issues/668

- [x] Expose @polkadot/api, @polkadot/util-crypto, @polkadot/util, @polkadot/keyring to console
- [x] Resizable columns for editor  and output in pure CSS (t.b. discussed, alternative to https://github.com/polkadot-js/apps/issues/670)
- [x] Working examples for api calls
- [x] Save last selection in local storage

I didn't implement to save the users edited code since I'm not convinced that that's a behaviour a user would expect. I was playing around with it but it felt weird to be shown half my finished code experiments when returning to the page.

I'd move it to a different ticket either way since this PR is already getting quite big.


packages/app-js/src/Intro.tsx
```diff
@@ -0,0 +1,30 @@
...
+import React, { Fragment } from 'react';
+
+import translate from './translate';
+
+type Props = I18nProps;
+
+const Intro = (props: Props) => {
+  const { t } = props;
+  return (
+    <Fragment>
```
jacogr: Prefer a simple `<>` (with terminating `</>`) - no need to import `Fragment`

packages/app-js/src/index.tsx
```diff
@@ -3,57 +3,83 @@
...
-    code: '',
-    logs: []
+    code: local ? local.code : snippets[0].code,
+    logs: [],
+    snippet: local ? local.value : snippets[0].value
   };
 
   render () {
-    const { logs } = this.state;
+    customWindow.api = this.props.api;
```
jacogr: I'm still wondering about the exposure the API and keyring on window since it means that anything has access to it, including extensions and scripts. I'm still not convinced on it, it opens up potential avenues of attack/inspection without real additional benefit. 

There are benefits in fooling about, but doing this by default when we don't have to, I'm not sure about. (Been giving it some thought after it come up)

packages/app-js/src/index.tsx
```diff
@@ -3,57 +3,83 @@
...
-            className='action-button'
-            isCircular
-            isPrimary
-            icon='play'
-            onClick={this.runJs}
+        <header className='container'>
+          <Intro />
+          <Dropdown
+            className='js--Dropdown'
+            onChange={(value) => this.selectExample(value)}
```
jacogr: Rather use `this.selectExample` instead of creating the closure each render

packages/app-js/src/index.tsx
```diff
@@ -3,57 +3,83 @@
...
+            options={snippets.map(({ code, ...options }) => ({ ...options }))}
+            label={'Select example'}
```
jacogr: (nitpick) wrap in `t`

packages/app-js/src/index.tsx
```diff
@@ -3,57 +3,83 @@
...
-            isCircular
-            isPrimary
-            icon='play'
-            onClick={this.runJs}
+        <header className='container'>
+          <Intro />
+          <Dropdown
+            className='js--Dropdown'
+            onChange={(value) => this.selectExample(value)}
+            options={snippets.map(({ code, ...options }) => ({ ...options }))}
```
jacogr: As per above, probably better calculating these once and keeping in the state.

packages/app-js/src/index.tsx
```diff
@@ -75,11 +102,24 @@ class App extends React.PureComponent<Props, State> {
       window: null
     };
 
-    const exec = `(async ({${Object.keys(injected).join(',')}}) => { try { ${code} } catch (error) { console.error(error); } })(injected);`;
+    const exec = `(async ({${Object.keys(injected).join(',')}}) => {
+      try {
```
jacogr: Ahhh, actually made a comment on this here https://github.com/polkadot-js/apps/pull/681/files#diff-7c043ef600c68ef2f18082701888270eR109 (when I saw that one-liner again)

The issue is that the `catch (error) { console.log(error) }` currently suck - we actually want the full trace with line-number and the ability to map-back.

packages/app-js/src/snippets/extrinsics-examples.ts
```diff
@@ -0,0 +1,37 @@
+// Copyright 2017-2019 @polkadot/app-js authors & contributors
+// This software may be modified and distributed under the terms
+// of the Apache-2.0 license. See the LICENSE file for details.
+import { Snippet } from '@polkadot/app-js/types';
+
+export const extrinsicMakeTransfer: Snippet = {
```
jacogr: Nice! Love the extended info.

packages/app-js/src/snippets/index.ts
```diff
@@ -0,0 +1,31 @@
+import { Snippet } from './../types';
```
jacogr: Header missing.

packages/app-js/src/index.tsx
Stefie: Yeah, that's one thing I wanted to ask. It's also the reason why I only implemented it in here. 
It's really convenient to explore the api without having to look at the docs, though. I was thinking that we could maybe just expose it when connected to a dev chain?

jacogr: Happy with dev-only exposure. However we should expose it via the API itself then instead of the js part, probably around here - https://github.com/polkadot-js/apps/blob/master/packages/ui-api/src/Api.tsx#L92

So expose once initialised and we are on dev chain (we should be able to use this https://github.com/polkadot-js/apps/blob/master/packages/ui-api/src/util/isTestChain.ts - looking at it now, now 100% on the `dev|loc` match, explicit via `local` and `development` is probably better. I know why it is short, dev chains originally retuned `dev` instead of `development`)

jacogr: I also think the API can be a bit more explorable by default and without tools, i.e. if we add something like

```js
api.tx.listAll() // shows all section.methods with description
```

and apply it consistently on `tx`, `query`, `derive` (this one is probably tougher) and `rpc`, we also have a way of getting info for non-nice-editor enabled people. (Without actually adding bloat)

Outside the scope of this, but may be worth a thought... Another one which apps and Edgeware both seems to use is a `api.tx[method][section]` pattern where section + method is selectable by the user, here something like `api.tx.call(section, method)` can actually be cleaner. Also out of scope here... but since I started mucking in `api.tx`, thought I may as well share what I wondered about...

jacogr:
Ok, played a bit (and made a PR onto this branch to pull in the clone/disconnect changes) -

- Not a fan of the split-screen view. It looks nicer but both are way too small - all examples scroll out of the screen, so the first thing anybody needs to do is resize and then resize again when running
- The resize is quite weird - it feels like it goes in the wrong direction and would have been expected on the left on the edge
- I am not sure the Intro adds much since we have partial duplication with the code comment - links are nice, but I think the orginal with everything in the comment feels a bit better
- It works well with selection, and is really nice to see

packages/app-js/src/Editor.tsx
```diff
@@ -7,74 +7,63 @@ import { BareProps } from '@polkadot/ui-app/types';
...
-      });
+      this.editor.onUpdate(this.props.onEdit);
     });
 
     onEdit(code);
   }
 
-  render () {
-    const { children } = this.props;
+  componentWillReceiveProps (nextProps: any) {
```
jacogr: Legacy, avoid. https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops

`getDerivedStateFromProps` is the current recommended version

packages/app-js/src/types.ts
```diff
@@ -8,3 +8,17 @@ export type Log = {
...
+
+export type Snippet = {
+  text: string;
+  value: string;
+  code: string;
+  label?: object;
+};
+
+export interface CustomWindow extends Window {
+  api?: any;
```
jacogr: If we do this, no `any` (I've done some for Snipper where object is used in the PR into this)

packages/app-js/src/snippets/wrapping.ts
```diff
@@ -1,12 +1,9 @@
-// Copyright 2017-2019 @polkadot/app-123code authors & contributors
+// Copyright 2017-2019 @polkadot/app-js authors & contributors
 // This software may be modified and distributed under the terms
 // of the Apache-2.0 license. See the LICENSE file for details.
 
-export default `// all code is wrapped within an async closure, allowing access to:
-//     api (@polkadot/api), hashing (@polkadot/util-crypto),
-//     keyring (@polkadot/keyring) and util (@polkadot/util)
-// (async ({ api, hashing, keyring, util }) => {
+export default `// (async ({ api, hashing, keyring, util }) => {
```
jacogr: As per comment, would suggest reverting.

packages/app-js/src/Intro.tsx
Stefie: According to the React docs `<></>` is just syntactic sugar for `<React.Fragment />`, not the other way round.
https://reactjs.org/blog/2017/11/28/react-v16.2.0-fragment-support.html

> n React, this desugars to a <React.Fragment/> element, as in the example from the previous section.

packages/app-js/src/index.tsx
Stefie: > probably around here - https://github.com/polkadot-js/apps/blob/master/packages/ui-api/src/Api.tsx#L92

Good idea, I can have a look at it once this PR is merged.

Stefie: > ```js
> api.tx.listAll() // shows all section.methods with description
> ```

I like that! But that would also go directly in the api and not in here, right?

packages/app-js/src/index.tsx
Stefie: You did that in your PR.

packages/app-js/src/index.tsx
Stefie: You also took care of that in your PR. (At least the 1-liner part)

packages/app-js/src/snippets/index.ts
Stefie: You also took care of that in the PR, so I'm closing this.

packages/app-js/src/Intro.tsx
jacogr: 100%, is is syntactic sugar only and equivalent - which also means we don't need to import `Fragment`, we can just use it.

packages/app-js/src/index.tsx
jacogr: 100%, directly on the API.

packages/app-js/src/Intro.tsx
Stefie: I prefer using `<Fragment>` directly because I don't really like the empty html tag syntax. But since this is purely down to personal preferences, I'm happy to change it to `<></>`  – your project, your preferences.

Stefie: ...and when I'm removing the `<p></p>` because we're not exposing the api here in apps but in the api itself, a wrapper is not needed anyways.

packages/app-js/src/index.tsx
Stefie: I put it down on my list to create tickets in /api for that.

packages/app-js/src/Intro.tsx
jacogr: Yes, this one it is a bit of bike-shedding, no doubt. I'm seemingly alone in my love for the "ugly" version :)

jacogr:
Coming back to the drag being weird. Maybe if the resize is on the editor part it will behave as expected. (dragging is just the wrong way around on the icon)

Not sure what the best is here, all options seems to suck.

1. vertical split, not quite right - started of with columns, but too small
2. horizontal split, mean that people will have the resize the whole time to see stuff

jacogr:
Sorry about the close, GH acting weird or me going blind. Probably the latter.

Stefie:
> Coming back to the drag being weird. Maybe if the resize is on the editor part it will behave as expected. (dragging is just the wrong way around on the icon)
> 
> Not sure what the best is here, all options seems to suck.

The drag is just an easy to implement/ pure html/css solution using the `resizable` attribute. I'm totally with you that it does feel a bit unintentional and rough. we could always go for a JS solution, I just wanted to give this a try.

> 1. vertical split, not quite right - started of with columns, but too small
Yes...

> 2. horizontal split, mean that people will have the resize the whole time to see stuff
That's definitely a screen size/ resolution problem. I worked fine on my machine :) Until I've been playing around with my screen resolution and then it didn't look so good anymore....

> 3. tabs, not everything is on the same screen - it helps when it auto-switches, but still not optimal (but you can see a lot at least)
I don't like the tabs approach. It would be really nice to see the code and the results on one screen without needing to switch tabs. It's also not common practice. I had these examples in mind when I implemented the resizable columns

https://jsbin.com/?html,js,output
https://jsfiddle.net/
https://codesandbox.io/s/8y05z5p4q8?from-embed
https://www.w3schools.com/tryit/

packages/app-js/src/index.tsx
Stefie: Since they're static and not getting updated (as of now) i just moved the declaration from the options in  return() to the render function. 
If we decide to save  the edited code as a dropdown option I'd move it to the state.

packages/app-js/src/index.tsx
Stefie: Note: I removed it

packages/app-js/src/types.ts
Stefie: Removed from this branch in favour of implementing it in the api itself

jacogr:
Love the CSS-only solution. (Would hate do do drag via JS, just feel bulky)

Maybe dropping the font-size then - I do agree that the split-screen is more intuitive, just it is horribly when you get in and have to resize as a first step since the example (the API makes more long lines) is out of the screen.

So maybe font-size + 60/40?

packages/app-js/src/Editor.tsx
Stefie: I've rewritten it to getDerivedStateFromProps

packages/app-js/src/snippets/wrapping.ts
Stefie: Moved to https://github.com/polkadot-js/apps/issues/692

PR #686, opened 2019-02-10
Pull api clone/disconnect into sd-668

jacogr:
Pull changes from #681 into #684


packages/app-js/src/index.tsx
```diff
@@ -55,22 +73,34 @@ class App extends React.PureComponent<Props, State> {
...
+                isCircular
+                isDisabled={!isRunning}
+                isNegative
+                icon='close'
+                onClick={this.stopJs}
+              />
+            </div>
           </Editor>
-          <Output className='js--Output' logs={logs}>
+          <Output logs={logs}>
```
Stefie: `className='js--Output'` got lost here. Some of the styling is relying on this class.

packages/app-js/src/index.tsx
```diff
@@ -102,15 +135,24 @@ class App extends React.PureComponent<Props, State> {
...
 
-    const exec = `(async ({${Object.keys(injected).join(',')}}) => {
-      try {
-        ${code}
-      } catch (error) {
-        console.error(error);
-      }
-    })(injected);`;
+    // squash into a single line so exceptions (with linenumbers) maps to the same line/origin
+    // as we have in the editor view (TODO: Make the console.error here actually return the full stack)
```
Stefie: Nice, haven't thought of that!

packages/app-js/src/index.tsx
jacogr: Ahh, it is just removing duplication - it is applied in the `Output` (and `Editor`) component already, so the above was actually a duplication.

packages/app-js/src/index.tsx
jacogr: This happens when you think of it, squash it into a single line and never comment on it. My bad.

packages/app-js/src/index.tsx
Stefie: Ah, ok, closing!

Issue #692, opened 2019-02-11
App-JS Follow up tasks


Stefie:
Follow-Up tasks from https://github.com/polkadot-js/apps/pull/684

- [x]  Adjust font-size on `Output` & `Editor` https://github.com/polkadot-js/apps/pull/697
-  [x] change default width of columns from 50|50 to 60|40 https://github.com/polkadot-js/apps/pull/697
- [x] Put `resizable` attribute on `Editor` instead of `Output`
- [x]  Potentially move npm module links from `Intro` back to code comments

jacogr:
I think 12px and 60% works (well, at least on my 13")

(May want to adjust the line height to 18px as well, both for the gutter and editor portion)

Stefie:
Fixed by https://github.com/polkadot-js/apps/pull/703

PR #697, opened 2019-02-11
Adjust js Editor/Output sizing

jacogr:
- Part of #692
- 12px font, 18px line-height
- adjust editor/output areas to fit screen bounds
- 60/40 initial split for editor/output

### DIFF
--- a/packages/app-js/src/index.css
+++ b/packages/app-js/src/index.css
@@ -6,21 +6,23 @@
   display: flex;
   flex-direction: column;
   height: 100vh;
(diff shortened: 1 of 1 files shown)


jacogr:
I know, however... also don't want you to waste time when you can be doing more fun stuff instead of just mucking about with UI.

jacogr:
... and still no freaking idea how to make the resize work properly in a simple manner. My small kingdom for a draggable bar between the panes as is standard. Way above my CSS-foo though.

PR #703, opened 2019-02-13
App-js follow up tasks

Stefie:
Adresses https://github.com/polkadot-js/apps/issues/692


packages/app-js/src/Editor.tsx
```diff
@@ -64,10 +55,10 @@ export default class Editor extends React.PureComponent<Props> {
 
   componentDidUpdate () {
     const { code, onEdit, snippet } = this.props;
-
```
jacogr: The extra break between the var defs and the code block here, doesn't cost anything :)

Stefie: added.

Issue #685, opened 2019-02-09
Let user save code examples to localStorage


Stefie:
~~When anything edited, add it to a "Recent" (or something properly named) entry and allow selection thereof. Save these into localstorage. (If available and user comes back, load that one as the default and select it - so we always come back to the user code)~~

Copied over from https://github.com/polkadot-js/apps/issues/668

**Updated description:**
- [x] Let user actively save code snippets to local storage
- [x] Show saved snippets in dropdown with `custom` label
- [x] Let users delete snippets from inside dropdown

PR #718, opened 2019-02-14
Save code snippets to local storage

Stefie:
Addresses https://github.com/polkadot-js/apps/issues/685

~~First proposal without the delete functionality.~~


packages/app-js/src/index.tsx
```diff
@@ -158,12 +217,40 @@ class App extends React.PureComponent<Props, State> {
...
+    const { code, customExamples, snippetName } = this.state;
+
+    const snapshot: Snippet = {
+      ...customExample,
+      code,
+      text: snippetName,
+      value: `custom-${Date.now()}`
+    };
+
+    localStorage.setItem('polkadot-app-js-examples', JSON.stringify([snapshot, ...customExamples]));
```
jacogr: May make sense to have these strings (key names) as constants, in the file, i.e.

```
const STORE_EXAMPLES = 'polkadot-app-js-examples';
```

(just protects agains possible typos)

packages/app-js/src/index.tsx
```diff
@@ -37,32 +39,53 @@ type Injected = {
...
-    snippet: ''
+    options: [],
+    snippet: '',
+    snippetName: ''
   };
 
   componentDidMount () {
-    const local = snippets.find(obj => obj.value === localStorage.getItem('app-js-snippet'));
+    const localData = {
+      examples: localStorage.getItem('polkadot-app-js-examples'),
```
jacogr: Rather use the `store` module. Interface is the same, but generally it is easier to use since it serializes/deserializes objects. (Which also means you can actually combine the different keys into a single object, as it makes sense)

packages/app-js/src/snippets/storage-examples.ts
```diff
@@ -68,7 +68,7 @@ const ALICE = '5GoKvZWG5ZPYL1WUovuHW3zJBWBP5eT8CbqjdRY4Q6iMaDtZ';
 
 // Retrieve the initial balance.
 let previous = await api.query.balances.freeBalance(ALICE);
-console.log('ALICE has a balance of' + previous);
+console.log('ALICE has a balance of ' + previous);
```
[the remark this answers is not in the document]
Stefie: Added empty line before log

packages/app-js/src/index.tsx
Stefie: Good idea, I moved it.

packages/app-js/src/LocalStorage.tsx
```diff
@@ -0,0 +1,106 @@
+// Copyright 2017-2019 @polkadot/app-js authors & contributors
+// This software may be modified and distributed under the terms
+// of the Apache-2.0 license. See the LICENSE file for details.
+
+import { BareProps, I18nProps } from '@polkadot/ui-app/types';
+
+import React from 'react';
+import { Button as SUIB, Popup } from 'semantic-ui-react';
+import { Button, Input } from '@polkadot/ui-app/index';
```
Stefie: The `ui-app` `<Button />` doesn't work well together with the `<Popup />` Component, the event triggers are not working properly.  I was trying to debug it, but then went for the easy fix of just using the original Semantic UI components...

packages/app-js/src/index.tsx
Stefie: Is there enough benefit of using the store module over the way it's implemented?
I'm reading from the store once onMount and I'm updating localStorage only in three places, which works perfectly fine with getItem, setItem and the JSON methods.
Adding a module for that seems a bit of an overload

packages/app-js/src/LocalStorage.tsx
jacogr: That is weird, since it is basically just a wrapper. So no ideas. 

Just put a FIXME and let somebody else take a look. No need to continue bumping heads on it, the workaround works.

packages/app-js/src/Editor.tsx
```diff
@@ -7,73 +7,45 @@ import { BareProps } from '@polkadot/ui-app/types';
...
     return (
-      <article className='container js--Editor'>
-        <div
-          className=''
-          id={this.id}
-        />
-        {this.props.children}
-      </article>
+      <div
+        className=''
```
jacogr: No sure we need to specify className here at all?

packages/app-js/src/LocalStorage.tsx
```diff
@@ -0,0 +1,108 @@
...
+      props: { isCustomExample, removeSnippet, t },
+      state: { isOpen, snippetName }
+     } = this;
+
+    if (isCustomExample) {
+      return (
+        // FIXME: The <Popup /> event trigger on='hover' does not work together with the ui-app
+        // <Button /> component. That's why the original Semantic UI component is being used here,
+        <Popup
+          content='Delete this custom example'
```
jacogr: Wrap in `t`

Stefie: done.

packages/app-js/src/Editor.tsx
Stefie: No, not at all. Removed.

Issue #546, opened 2018-12-30
Display Params Structs broken into their parts (like Tuple)


jacogr:
![image](https://user-images.githubusercontent.com/1424473/50547965-d907fe80-0c44-11e9-9fda-d671f244f743.png)

jacogr:
Basically to fix this, we would need to ui-app/Params to display (disabled) structs as split into the parts. (For the time being, we can still allow hex input for values)

jacogr:
This specifically goes into "Unknown", where a rendered does not already exist. (Static render only... at this point, it is most probably a specific type that is a Struct)

PR #728, opened 2019-02-18
Break struct/classes down into components

jacogr:
Closes https://github.com/polkadot-js/apps/issues/546


jacogr:
I think this is a dead-end as it stands. Basically need to do inspection of the objects to get this working and it certainly is not an easy and foolproof feat.

The approach here works, but the sane way is most probably waiting a bit for V2 metadata.

Issue #719, opened 2019-02-15
Allow sharing of snippets


jacogr:
Basically like the Typescript playground -

- add base-64 encoded hash to URL
- read this on load
- copy to clipboard

Stefie:
A quicker first-step implementation would be to just share the `value` property of the pre-baked examples in the url and pass that as the `value` to the dropdown on `componentDidMount`

jacogr:
Yes, agreed. The sharing for me it more for custom examples, i.e. write something, share with others.

Stefie:
Ah ok, got it.

PR #740, opened 2019-02-20
Share functionality for app-js api playground

Stefie:
Implements a sharing functionality for code examples.

Resolves https://github.com/polkadot-js/apps/issues/719

[Test example 1 on http://localhost:3000](http://localhost:3000/#/js/share/Ly8gRXhhbXBsZSBjaGFuZ2Ugc3Rha2luZyBwcmVmZXJlbmNlcwoKY29uc3QgQUxJQ0UgPSAnNUdvS3ZaV0c1WlBZTDFXVW92dUhXM3pKQldCUDVlVDhDYnFqZFJZNFE2aU1hRHRaJzsKCmNvbnN0IGludGVudGlvbnMgPSBhd2FpdCBhcGkucXVlcnkuc3Rha2luZy5pbnRlbnRpb25zKCk7CmNvbnN0IHN0YWtlSW5kZXggPSBpbnRlbnRpb25zLmluZGV4T2YoQUxJQ0UpOwpjb25zdCByZWdpc3RlclByZWZlcmVuY2VzID0geyJ1bnN0YWtlVGhyZXNob2xkIjo1LCJ2YWxpZGF0b3JQYXltZW50IjoxMDAwMDAwMDAwMDAwfTsKY29uc29sZS5sb2coJ2ludGVudGlvbnM6JywgaW50ZW50aW9ucyk7CmNvbnNvbGUubG9nKCdpbmRleDonLCBpbnRlbnRpb25zLmluZGV4T2YoQUxJQ0UpKTsKCmNvbnN0IHRyYW5zZmVyID0gYXBpLnR4LnN0YWtpbmcucmVnaXN0ZXJQcmVmZXJlbmNlcyhzdGFrZUluZGV4LCByZWdpc3RlclByZWZlcmVuY2VzKQoKLy8gU2lnbiBhbmQgU2VuZCB0aGUgdHJhbnNhY3Rpb24KdHJhbnNmZXIuc2lnbkFuZFNlbmQoQUxJQ0UsICh7IGV2ZW50cyA9IFtdLCBzdGF0dXMsIHR5cGUgfSkgPT4gewogIGlmICh0eXBlID09PSAnRmluYWxpc2VkJykgewogICAgY29uc29sZS5sb2coJ2NoYW5nZSByZWdpc3RlclByZWZlcmVuY2VzIHdpdGggaGFzaCAnICsgc3RhdHVzLmFzRmluYWxpc2VkLnRvSGV4KCkpOwogIH0gZWxzZSB7CiAgICBjb25zb2xlLmxvZygnU3RhdHVzIG9mIHRyYW5zZmVyOiAnICsgdHlwZSk7CiAgfQoKICBldmVudHMuZm9yRWFjaCgoeyBwaGFzZSwgZXZlbnQ6IHsgZGF0YSwgbWV0aG9kLCBzZWN0aW9uIH0gfSkgPT4gewogICAgY29uc29sZS5sb2cocGhhc2UudG9TdHJpbmcoKSArICcgOiAnICsgc2VjdGlvbiArICcuJyArIG1ldGhvZCArICcgJyArIGRhdGEudG9TdHJpbmcoKSk7CiAgfSk7Cn0pOw==)

[Test example 2 on http://localhost:3000](http://localhost:3000/#/js/share/Ly8gRXhhbXBsZSB2b3RpbmcgY291bnQKCmNvbnN0IHZvdGluZ1BlcmlvZCA9IGF3YWl0IGFwaS5xdWVyeS5kZW1vY3JhY3kudm90aW5nUGVyaW9kKCkKY29uc3QgbGF1bmNoUGVyaW9kID0gYXdhaXQgYXBpLnF1ZXJ5LmRlbW9jcmFjeS5sYXVuY2hQZXJpb2QoKQoKY29uc29sZS5sb2coIkB2b3RpbmdQZXJpb2Q6Iix2b3RpbmdQZXJpb2QpCmNvbnNvbGUubG9nKCIqbGF1bmNoUGVyaW9kOiIsbGF1bmNoUGVyaW9kKQoKYXdhaXQgYXBpLmRlcml2ZS5jaGFpbi5iZXN0TnVtYmVyKChiZXN0TnVtYmVyKT0+ewogICAgY29uc3Qgdm90aW5nQ291bnRkb3duID0gdm90aW5nUGVyaW9kIC0gYmVzdE51bWJlci5tb2Qodm90aW5nUGVyaW9kKS5hZGRuKDEpCiAgICBjb25zdCBsYXVuY2hDb3VudGRvd24gPSBsYXVuY2hQZXJpb2QgLSBiZXN0TnVtYmVyLm1vZChsYXVuY2hQZXJpb2QpLmFkZG4oMSkKICAgIGNvbnNvbGUubG9nKCJAdm90aW5nQ291bnRkb3duOiIsdm90aW5nQ291bnRkb3duKQogICAgY29uc29sZS5sb2coIipsYXVuY2hDb3VudGRvd246IixsYXVuY2hDb3VudGRvd24pCn0pOw==)


packages/app-js/src/index.tsx
```diff
@@ -61,31 +64,56 @@ class App extends React.PureComponent<Props, State> {
...
+      shareLink: ''
     };
   }
 
   componentDidMount () {
+    const { location } = this.props;
+
+    const sharedExample: Snippet = location.hash.length ?
+    ({
+      code: atob(location.hash.substr(1)),
```
jacogr: We can use the router matching, i.e. https://github.com/polkadot-js/apps/blob/master/packages/app-explorer/src/BlockByHash/index.tsx#L22

Stefie: Only when using `<Switch>` and `<Route>` as far as I know https://github.com/polkadot-js/apps/blob/master/packages/app-explorer/src/index.tsx#L60

It does return an empty object in my case because I'm not rendering different Routes (as of now)...

jacogr: We can add the one with a route, most apps have that anyway. (May not have lots of different routes yet, but it has the benefit of aligning across the codebase instead of doing something special here.) No harm either way, but we can remove some basic heavy-lifting.

packages/app-js/src/index.tsx
```diff
@@ -264,10 +287,49 @@ class App extends React.PureComponent<Props, State> {
...
+    } = this;
+    const base64code = btoa(code);
+
+    if (base64code !== location.hash.substr(1)) {
+      history.push(`${basePath}#${base64code}`);
+    }
+    this.copyToClipboard(`${window.location.origin}/#${location.pathname}#${base64code}`);
+  }
+
+  private copyToClipboard = (link: string): void => {
```
jacogr: Can't we use the CopyToClipboard component here rather?

packages/app-js/src/index.tsx
Stefie: true, I added it.

packages/app-js/src/index.tsx
Stefie: I tried that... But I don't want to calculate the base64 value on every edit so that I can pass it as `value` for the `text` prop to the component.
That would be necessary because the CopyToClipboard component doesn't have an eventHandler so I can't calculate the base64 value of the code on click

jacogr: Grrr... yes, just looked it would have been great if the `text` param could have been string or function-returning-string

Stefie: Yup, it would ;). But that's how I ended up with the solution from Hackernoon. There really should be an easier native solution :|

packages/app-js/src/Playground.tsx
```diff
@@ -0,0 +1,341 @@
...
+
+        return { logs: logs.slice(0) };
+      });
+    };
+  }
+
+  private decodeBase64 = (base64: string): string => {
+    try {
+      const decoded = atob(base64);
+      return ({
```
jacogr: (nitpick) Probably could use an extra blank line before

packages/app-js/src/Playground.tsx
```diff
@@ -0,0 +1,341 @@
...
+      props: { history, match: { params: { base64 } } },
+      state: { selected: { code } }
+    } = this;
+    const base64code = btoa(code);
+    const path = `/js/share/${base64code}`;
+
+    if (base64code !== base64) {
+      history.push(path);
+    }
+    this.copyToClipboard(`${window.location.origin}/#${path}`);
```
jacogr: (nitpick) just a small line break after the if block

packages/app-js/src/Playground.tsx
```diff
@@ -0,0 +1,341 @@
...
+    el.setAttribute('readonly', '');
+    el.style.position = 'absolute';
+    el.style.left = '-9999px';
+    document.body.appendChild(el);
+
+    const existingSelection = document.getSelection()!;
+    const selected = existingSelection && existingSelection.rangeCount > 0
+        ? existingSelection.getRangeAt(0)
+        : undefined;
+    el.select();
```
jacogr: (nitpick) blank line before here (and on 327)

packages/app-js/src/index.tsx
```diff
@@ -1,273 +1,30 @@
-// Copyright 2017-2019 @polkadot/app-js authors & contributors
+// Copyright 2017-2019 @polkadot/app-explorer authors & contributors
```
jacogr: (small) app-js

Issue #760, opened 2019-02-26
Restore account from backup file does not work.


mnaamani:
app-accounts is not correctly validating an account backup file. Unable to restore account from backup.

PR #761, opened 2019-02-26
Fix app-accounts Restore feature

mnaamani:
Fixes file validity check, solving #760

The encoding.content value is not just the string 'pkcs8' but an array `['pkcs8', 'ed25519']`

https://github.com/polkadot-js/common/blob/503dc99815700c79d01c09dc337c5bc8bebd38cc/packages/keyring/src/pair/toJson.ts#L20


jacogr:
Thank you very much! Will get it in ASAP.

Issue #674, opened 2019-02-04
Start using styled-components 


jacogr:
The ui repo uses these exclusively. Start slowly converting (as it makes sense) across all apps as well as components such as those in ui-app. 

No need for a Big Bang, rather touch and adapt as it makes sense. (Some styles, eg. Semantic overrides, UI grids and rows, UI media and utility makes sense to keep, component-specifics can be adapted)

jacogr:
Closing this one, since it is not a once-off all-or-nothing.

PR #777, opened 2019-03-03
added simple styled-components for divs in app-settings (#674)

mohsen-norouzi:


packages/app-settings/src/index.tsx
```diff
@@ -34,8 +35,8 @@ class App extends React.PureComponent<Props, State> {
 
     return (
       <main className='settings--App'>
-        <div className='ui--row'>
-          <div className='full'>
+        <RowDiv>
+          <FullDiv padding='0 0.25rem'>
```
jacogr: I would not pass the padding in here at all, since we row + full/medium these components everywhere, it will just create an absolute maintenance burden trying to keep it updated.

mohsen-norouzi: But our purpose seemed to be removing all custom styles and using styled-components instead. right? Also having many custom styles and overrides isn't cool.

I added the padding and other parameters, just because I had to, so to make it look as before. I wanted to remove them.

Using a styled-component for all the elements would be nice and clean. 
Something like what these guys have made:
https://github.com/aichbauer/styled-bootstrap-components

jacogr: Yes, it would be - however we also need to be really careful -

So the first stage would be to make sure we can get rid of all the styles in `app-*`, the only component doing some of that is `app-explorer` - but even there are is still a style mess left.

`ui-app` is more tricky - when we drop something, we need to really drop something, so we don't need to worry about "what is the way to do this", there should only be one way. So doing `row` part-way, just makes it more difficult to maintain. 

(And I'm still up in the air here, there is a fine line between ` div + class` and adding additional imports. We will most probably end up with small separate components for these anyway, but the road we take need to be clear and without potholes along the way.)

mohsen-norouzi: Yes you're right. 
We need to discuss about the steps to make the changes.
Let me know whenever you're ready to discuss.

jacogr: The place to start with `styled-component` conversions would be in all the `app-*` areas, but until those are clean `ui-app` can be left as-is.

So for instance, just picking one, where it will make an impact (and clean up things immensely - maybe even some component splits), https://github.com/polkadot-js/apps/blob/master/packages/app-accounts/src/Vanity/index.css

After `app-*` (on even interleaving it), areas like `ui-signer` (bundle of stuff there) and `ui-params`.

`ui-app` really needs to be last since it has impact across everything and some parts will no doubt be tough, i.e. there are components, overrides for semantic-ui, etc. (The component itself, not touching global styles, are the first step there - basically anything that makes an appearance in `ui-app/styles/components.css`)

Issue #776, opened 2019-03-03
Settings App - Custom type definition JSON upload Feedback 


yjkimjunior:
Under `https://polkadot.js.org/apps/#/settings/developer`, upon uploading a custom type definition json, there is no Status feedback to let you know whether it was successful or not. Although it does log it in the console.

jacogr:
It does show the types in the upload box (as determined), but yes, more feedback would be good. (At least a status message that pops up and says "ok" or "it failed")

Issue #758, opened 2019-02-26
set_storage interface requires JSON key-value file


rphmeier:
Ideally I would be able to edit it as a list in-UI if I wanted to.

jacogr:
Related to the app-extrinsics, specifically `consensus.setStorage` (only used by admins to change state entries). We added the JSON upload of k/v pairs to aid in making it simpler (and before the days of being able to handle arrays and Tuples...)

Stefie:
The solution I'm implementing in #794 could still work. I could even try to turn the <Editor /> into a UI component

jacogr:
These should not be edited by hand. It is hex storage entries. What should happen is that the KeyValue component render a selectable storage key with a (based on type) value. Effectively we need an Array that displays the same type of entry components we have in the chain state queries.

allowing JSON editing here is asking for trouble. The JSON upload already assumes you know what you are doing - getting it wrong can brick the chain.

jacogr:
Probably, after typing out the above... not quite "easy-as-pie" :) The UI here worries me a bit as wel since we have 2 parts -

- entering the key - which is the same work that happens in `app-storage`
- entering the value (based on key type)

And then over an above this we have the Array wrapping.

Stefie:
Do you happen to have an example JSON file at hand? :

jacogr:
The input that go there, would look something like this -

```json
{
      "0x2ec6e5652282d579398fb8fdfa531ef6": "0x0000000000000000",
      "0x4664fb5d4e16f894df23cadb3faaa9a6": "0x04000000",
      "0xfc2dc4b8bb0b9ca8f01a73a726f7c7f5": "0x00e1000000000000",
      "0xd437fe93b0bd0a5d67d30d85d010edc2": "0x40420f00",
      "0x3a617574683a00000000": "0x4bd3620064cda1f4cf405bf9ab565c9bad69446034c48884ffc5363a5286b145",
      "0x99e2aba8a2b7c8ccba2d740fb86adb0c": "0x00",
      "0x52b963fbdb3d6e1b03808fc20071f07f": '0x0027060000000000",
[...]

PR #794, opened 2019-03-07
Editor for custom types and ActionStatus

Stefie:
- [x] Added a code editor to add and edit JSON manually
- [x] Added global ActionStatus onSave (Fixes https://github.com/polkadot-js/apps/issues/776)
- [x] Validate JSON onChange inEditor
- [x] Split app-settings into separate components


jacogr:
Doesn't address #758 - that is related to the app-extrinsics, specifically `consensus.setStorage` (only used by admins to change state entries)

Stefie:
> Doesn't address #758 - that is related to the app-extrinsics, specifically consensus.setStorage
Ah, ok, I didn't get that from the description. I removed the reference.

jacogr:
Will update that with the above info - too much JSON everywhere :)

packages/ui-app/package.json
```diff
@@ -18,6 +18,7 @@
     "@polkadot/ui-settings": "^0.28.8",
     "@polkadot/ui-util": "^0.28.8",
     "@types/chart.js": "^2.7.45",
+    "@types/classnames": "^2.2.7",
```
jacogr: Could just use https://github.com/polkadot-js/apps/blob/master/packages/ui-app/src/util/classes.ts as is done everywhere.

Stefie: Totally! I didn't see that. I got confused because I saw the classnames module in the node_modules folder so I just assumed that it's used elsewhere.  Probably just as a dependency by another module

Stefie:
This is ready to go from my side...

jacogr:
The "cleanup settings" is especially nice - finally it is easy to find stuff without trudging through thousands of lines of code

Issue #806, opened 2019-03-12
Unable to enter an amount in InputNumber


adetante:
Version : `master`
Browser : `Chrome 72.0.3626.121 / Mac OS`

I can not input a number into InputNumber. The key is just rejected.
It seems to be due to this test : https://github.com/polkadot-js/apps/blob/master/packages/ui-app/src/InputNumber.tsx#L158

With Chrome on Mac, when I enter the value `1`, the KeyboardEvent has the following properties: `shiftKey = true` and `key = "1"`.
So `isValidKey` returns `false` and `onKeyDown` prevents the input.

I do not undestand this test about `shiftKey` in the `isValidKey` function.

PR #807, opened 2019-03-12
Remove InputNumber shift check

jacogr:
Closes https://github.com/polkadot-js/apps/issues/806 (tested on Chrome, things like Shift-2 (`@`) is still not entered.

### DIFF
--- a/packages/ui-app/src/InputNumber.tsx
+++ b/packages/ui-app/src/InputNumber.tsx
@@ -154,11 +154,6 @@ class InputNumber extends React.PureComponent<Props, State> {
       return false;
     }
 
-    // prevent use of shift key
-    if (event.shiftKey) {
-      return false;
-    }
-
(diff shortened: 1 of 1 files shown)


Stefie:
`type='number'` on the `<Input />` was not an option?

jacogr:
It used to be an input number - however with actual length checks, allowing anything > 2^53 - 1 (balance is a u64, other numbers are currenntly up to u128) and allowing to specify the units (e.g. balance), it ended up as a frustrating dead-end. (And text with patterns also doesn't quite do the trick)

Stefie:
Makes sense... thanks for the info!

jacogr:
I absolutely HATE this component and the complexity thereof - the sane approach would have been `text` with a pattern, however the browser doesn't actually disallow inputs based on the pattern, only validates against it. 

I am hoping that at some point, now that `BigInt` is starting to make it into browsers that maybe, just maybe at some there will be a new input type for `BigInt` and we can just drop all this craziness.

PR #813, opened 2019-03-13
Bump UI, HDKD keyring (new substrate crypto)

jacogr:
- Pulls in sr25519 HD derivation, https://github.com/polkadot-js/common/pull/361 & https://github.com/polkadot-js/schnorrkel-js/pull/8
- Needs latest substrate master (after this commit https://github.com/paritytech/substrate/pull/2015)

Current status -

- ~~InvalidTransaction from all ed25519 transfers and on follow-up sr25519 transfers, needs https://github.com/paritytech/substrate/issues/2022 fixed (both ed25519 signature verification & balances being under-reported)~~
- ~~this pops up very intermittently - https://github.com/polkadot-js/api/issues/782 (well, 50% shot of hitting it, so not _that_ intermittent), needs https://github.com/paritytech/substrate/issues/2021 fixed~~

To be merged when the following is available -

- ~~https://github.com/paritytech/substrate/pull/2046~~ (`CantPay` fix)
- ~~https://github.com/paritytech/substrate/pull/2028~~ (allow ed25519 in addition to sr25519)


burdges:
I never put hard derivations for Ed25519 into https://github.com/w3f/hd-ed25519 but someone could do so.  It's way uglier code than Sr25519 but exactly the same issues apply:  You must hash only the curve point from the `ExpandedSecretKey` because if you hash the `SecretKey` then you cannot obtain it from a soft derived key and if you include the nonce seed form the `ExpandedSecretKey` then you'll break whenever someone changes it for some reasons.  I still have no idea why anyone ever wants a hard derivation after a soft derivation, but hey whatever.  :)

jacogr:
The JS side implements only hard for ed25519, same way as here https://github.com/paritytech/substrate/blob/master/core/primitives/src/ed25519.rs#L366

(Soft/Hard for sr25519 is now done via w3f/schnorrkel#d3289df76b8ae6dfb68e733204c5c009df5343a9, both JS & substrate, test vectors here - https://github.com/paritytech/substrate/wiki/Secret-URI-Test-Vectors)

Issue #828, opened 2019-03-22
Update Alice/Bob addresses for sr25519 (js transfer example)


jacogr:
They have been updated for latest substrate master -

- JS app
- transfer example
- copy addresses as generated by the UI

jacogr:
Closed in #839

PR #839, opened 2019-03-24
Updated addresses for sr25519

JoshOrndorff:
This is my attempt to solve https://github.com/polkadot-js/apps/issues/828 .

### DIFF
--- a/packages/app-js/src/snippets/transfer.ts
+++ b/packages/app-js/src/snippets/transfer.ts
@@ -3,7 +3,7 @@
 // of the Apache-2.0 license. See the LICENSE file for details.
 
 export default `// transfer
-const sender = '5GoKvZWG5ZPYL1WUovuHW3zJBWBP5eT8CbqjdRY4Q6iMaDtZ';
+const sender = '5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKv3gB';
 const recipient = '5GTZEKgRKX2a5hepNjwTBzR5nR6UqtEsMZ4TnL1Yuah1N2Tu';
 
 const nonce = await api.query.system.accountNonce(ALICE),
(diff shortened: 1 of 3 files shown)


packages/app-js/src/snippets/transfer.ts
```diff
@@ -3,7 +3,7 @@
 // of the Apache-2.0 license. See the LICENSE file for details.
 
 export default `// transfer
-const sender = '5GoKvZWG5ZPYL1WUovuHW3zJBWBP5eT8CbqjdRY4Q6iMaDtZ';
+const sender = '5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKv3gB';
 const recipient = '5GTZEKgRKX2a5hepNjwTBzR5nR6UqtEsMZ4TnL1Yuah1N2Tu';
```
JoshOrndorff: I wasn't sure there this recipient address came from.

JoshOrndorff: I wasn't sure there this recipient address came from.

jacogr: Think it was just a random sample address, @Stefie ? I suppose we could have just created a random address ourselves with something like, but probably overkill as a sample -

```js
const { randomAsU8a } = hashing;

...
// a random recipient address, from 32-byte public key
const recipient = keyring.encodeAddress(randomAsU8a());
...
```

JoshOrndorff:
Looks like I can't merge it myself.

packages/app-js/src/snippets/transfer.ts
Stefie: @jacogr You added this example when you added the Signer interface, so if you don't know it I think we can change it to sth. random or use BOB

Issue #834, opened 2019-03-23
Node selection is wrongly highlighted


jacogr:
It should have the same color as the rest of the page.

<img width="1428" alt="Polkadot:Substrate Portal 2019-03-23 18-13-59" src="https://user-images.githubusercontent.com/1424473/54869357-86e3fe80-4d97-11e9-9152-42f2d80ab1eb.png">

Stefie:
Which color / part of the node selection field do you mean? I don't really see anything off here...

jacogr:
Look at the background surrounding the field - it is white. Transparent would solve it, i.e. so it fits in with the rest of the fields where it does not sit in a white box.

PR #841, opened 2019-03-25
Adjust/remove settings endpoint background

jacogr:
Fixes https://github.com/polkadot-js/apps/issues/834

Previously -

Polkadot:Substrate Portal 2019-03-25 16-08-56

With this PR -

Polkadot:Substrate Portal 2019-03-25 16-07-59

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,11 +10,11 @@
     "packages/*"
   ],
   "resolutions": {
-    "@polkadot/api": "^0.51.0-beta.1",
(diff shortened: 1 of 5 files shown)


Stefie:
Your screen must have a really sharp contrast! I see what you changed here but the background of the tab didn't look different on my screen. I could only see the difference after switching to a non-default color profile

jacogr:
I look at home with my 2014 MacBook Air with non-retina, chances are I won't spot the difference either.

PR #883, opened 2019-04-03
Add recently offline badge to validators

### LINKED ISSUE #875: Show Offline warnings on validators
Steps - use https://polkadot.js.org/api/METHODS_STORAGE.html#staking - `recentlyOffline` and display indicator on the validators

kwingram25:
Closes #875 

Adds a lower-right badge for staking validators showing number of offline instances, which expands to message with instances and first block number. It was originally to open on hover but this seems more mobile friendly. I have combined the nominators summary and badge into a new combined child node.

Please double check that I have the mapping from recentlyOffline() through bonded() correct! At the moment there are two validators with the badge visible. Also any corrections for css/js project standards I'm not aware of :)

### DIFF
--- a/packages/app-staking/src/Overview/Address.tsx
+++ b/packages/app-staking/src/Overview/Address.tsx
@@ -49,6 +49,7 @@ class Address extends React.PureComponent<Props, State> {
 
   private onClickBadge = (): void => {
     const { badgeExpanded } = this.state;
+
     this.setState({ badgeExpanded: !badgeExpanded });
   }
 
--- a/packages/app-staking/src/Overview/Address.tsx
+++ b/packages/app-staking/src/Overview/Address.tsx
@@ -4,11 +4,11 @@
 
 import { DerivedBalancesMap } from '@polkadot/api-derive/types';
 import { I18nProps } from '@polkadot/ui-app/types';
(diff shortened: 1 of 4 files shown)


packages/app-staking/src/Overview/Address.tsx
```diff
@@ -37,29 +47,70 @@ class Address extends React.PureComponent<Props> {
       : defaultName;
   }
 
+  private onClickBadge = (): void => {
+    const { badgeExpanded } = this.state;
+    this.setState({ badgeExpanded: !badgeExpanded });
```
jacogr: (nitpick) Prefer a blank line separating blocks, i.e. just a single space after the consts. (And just a nitpick, it doesn't aid functionality, however just makes it _slightly_ easier to read and find patterns)

packages/app-staking/src/types.ts
```diff
@@ -1,9 +1,9 @@
 // Copyright 2017-2019 @polkadot/app-staking authors & contributors
 // This software may be modified and distributed under the terms
 // of the Apache-2.0 license. See the LICENSE file for details.
-
+import BN from 'bn.js';
```
jacogr: (nitpick) Blank line before.

PR #888, opened 2019-04-04
Add value field to destination dropdown for bonding settings

### LINKED ISSUE #887: Payment destination dorp down menu always shows: Stash account (increase..)
Running on a node on tag v4.1.

I can't select `Controler`or `Stash account (does not increase..`:
image

it seems the `aria-checked` isn't updated  and is responsible for what is displayed.

Tbaut:
The title says it all, I guess it was simply forgotten. 
- closes #887 

question: should I change my linter settings somehow to make sure my PR reflect what I really changed?


jacogr:
Non-tested is more like it. Thank you.

Tbaut:
Should be better now :)

PR #894, opened 2019-04-05
Refined InputNumber

### LINKED ISSUE #886: Improve balance input field - allow decimal, change w/dropdown
Some improvements I'm going to work on for the balance input field used in most places, open to feedback:

- Default to DOT instead of confusing femto, milli etc.
- Change to incrementable number input (defaulting to increment of 1?)
- When changing units with a number already inputted, calculate the number in the new unit and automatically change it
- Allow decimal use (possibly only on DOT or above to avoid confusion? potential issues over diff uses of , and .)

kwingram25:
Closes #886 

- Change InputNumber to default to whole unit (DOT)
- Update input value to reflect new unit when selected from dropdown
- Regex-based key/paste interception and sanitizing
- Prevent edge cases (too small or too big for max bit length)
- Distinguishes between , and . for decimal character (needs EU tester)

Originally was going to use a number input but there was some trouble with allowing region-specific decimal separators

### DIFF
--- a/packages/ui-app/src/Input.tsx
+++ b/packages/ui-app/src/Input.tsx
@@ -31,6 +31,8 @@ type Props = BareProps & {
   onBlur?: (event: React.KeyboardEvent<Element>) => void,
   onKeyDown?: (event: React.KeyboardEvent<Element>) => void,
   onKeyUp?: (event: React.KeyboardEvent<Element>) => void,
+  onKeyPress?: (event: React.KeyboardEvent<Element>) => void,
+  onPaste?: (event: React.ClipboardEvent<Element>) => void,
   placeholder?: string,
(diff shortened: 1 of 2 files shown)


jacogr:
The fact that I can send `1.234` by default, instead of swapping to milli and entering `1234` has my attention :) Nice!

jacogr:
Something seems to go wrong on InputNumber - just grabbed one area where we use InputNumber (in this case staking), and get this funny: https://www.youtube.com/watch?v=6YAuL82c_C4&feature=youtu.be

Does not happen on polkadot.js.org/apps so something went wonky

kwingram25:
Thanks for catching that! - changed so inputs without SI default to integer-only

Issue #896, opened 2019-04-05
Staking overview should display bonded value


jacogr:
Currently it displays the account balance, but what it should be doing is display the actual bonded amount. (same with the nominators for a specific account)

![image](https://user-images.githubusercontent.com/1424473/55626693-7aef3800-57ac-11e9-8b95-77648e51dad3.png)

jacogr:
We can just retrieve the balance from the staking_ledger and pass the value through to the balance prop. (Currently we pass balanceArray there, which is actually useless and not populated)

jacogr:
Would almost suggest, looking at it and giving at-the-fingertips info to extends this -

![image](https://user-images.githubusercontent.com/1424473/55690009-433ee680-598c-11e9-85f3-dee832f437ed.png)

Instead of just showing balance (which is from the stash), add a row underneath that shows "bonded xxxxx" . As suggested, the bonded value can be retrieved from staking.ledger with the controller, like we retrieve the stash already -

![image](https://user-images.githubusercontent.com/1424473/55690033-813c0a80-598c-11e9-8f7f-fb6f472cf964.png)

(Active amount)

jacogr:
And as we go down the rabbit hole, it gets worse - so, in poc-2/poc-3 we basically had as a display, something like: "my balance (+ nominators balance)"

"my balance === actual balance, bonded in this case"
"(+ nominators balance === cimalive total for all nominators)"

So more rabbit hole stuff -

- needs to be displayed on the account actions for the stash
- in accounts, when a stash, display the bonded value as well

(And these should also be displayed when the account is viewed in the addressbook or my accounts)

PR #915, opened 2019-04-07
Display bonded amount on overview

jacogr:
Sorry @kwingram25 - I KNOW this is on your list, but needed a view while trying to help. However, this is only one part, only on overview. (We still need it on accounts, which is probably more problematic since we need to display it on the stash - and query from the controller.)

- Added extraInfo to AddressSummary & AddressRow
- query bonded value from staking.ledgerr
- pass this amount through
- Part of https://github.com/polkadot-js/apps/issues/896

image


kwingram25:
I was just about to do a pull - actually had the wrong value though, I was using "total"!

jacogr:
Cool. Viewing this as a top-gap until we have @kwingram25's  PR (which does it right with a component that is available everywhere)

PR #922, opened 2019-04-08
Add help overlay for staking (basic intro)

jacogr:
- Part of https://github.com/polkadot-js/apps/issues/900
- Simple markdown content
- HelpOverlay component that can be placed on any page

TODO -

- The content is very basic, needs to be expanded
- Probably want to have seperate info for the overview & account actions
- It is not pretty, it is functional :)

Polkadot:Substrate Portal 2019-04-08 11-30-34
Polkadot:Substrate Portal 2019-04-08 11-30-17


packages/app-staking/src/md/basic.md
```diff
@@ -0,0 +1,32 @@
+# Staking
+
+Welcome to the staking module. Here you can see an overview of all the validators on the network and participate in the network as either as validtor (running a node that helps secure the network), or a nominator (adding funds to help secure the network).
```
Tbaut:
```suggestion
Welcome to the staking module. Here you can see an overview of all the validators on the network and participate in the network as either as validator (running a node that helps secure the network), or a nominator (adding funds to help secure the network).
```

packages/app-staking/src/md/basic.md
```diff
@@ -0,0 +1,32 @@
...
+
+Stakers are nominators earn rewards: at the end of an era the rewards are split between the validators and nominators for that validator, in proportion to the amount bonded by each individual. Simmilarly, at the end of a session set of new validators are selected.
```
Tbaut: Did you mean: "Stakers **and** nominators"? I'd keep the notion of "validator" instead of "staker" in that case.
```suggestion
Validators and nominators earn rewards at the end of an era: the rewards are split between the validators and nominators for that validator, in proportion to the amount bonded by each individual.
```
I think " Similarly, at the end of a session set of new validators are selected." doesn't belong to this section dedicated to rewards.

packages/app-staking/src/md/basic.md
```diff
@@ -0,0 +1,32 @@
...
+
+# Bonding
+
+Bonding funds is the first step that either a validator or nominator performs. It instructs the network to lock up a portion of funds that is used for security. As a validator or nominator, these funds are placed at risk, i.e. you can be slashed and lose a portion if the validator node misbehaves. Similarly, as a nominator, choosing a misbehaving validator will also have a slashing effect. At the same time, should the validators perform well, rewards are payed to the validators and nominators for authoring.
```
Tbaut: I think we don't need to have the notion of "instructing the network" to abstract the fact that those are transactions.

```suggestion
Bonding funds is the first step that either a validator or nominator performs. It locks up a portion of funds that is used to secure the network. These funds are placed at risk, i.e. you can be slashed and lose a portion if the validator node misbehaves. Validators and nominators share both rewards and the slashing effects. Choosing a well behaving validator to nominate is crucial.
```

packages/app-staking/src/md/basic.md
```diff
@@ -0,0 +1,32 @@
...
+
+Nomination (using the controller account as set as part of the bonding) is a process of selecting a number of validators (or potential validators) that you deem trustworthy. Once the validators are selected, the bonded value is assigned to participate in the network security.
+
+At any point you could stop nominating (using the controller) or top up the funds bonded (from the stash).
+
+# Validating
+
+Validators run nodes that author blocks. The primary requirement here is the ability to run a node that is available 24/7 and is well-connected to the network.
+
+In addition to the stash and contrroller acount described above, a validator has to inicate an additional key -
```
Tbaut:
```suggestion
In addition to the stash and controller account described above, a validator has to indicate an additional account called `Session key` -
```

jacogr: Blame Tim Cook - my "r" is either nor "r-ing" and when it finally does, decides to make up for previous missteps by _helpfully_ adding more than one character at a time.

packages/app-staking/src/md/basic.md
```diff
@@ -0,0 +1,32 @@
...
+
+- **Session** The seed for this key is passed to the node. It can be the same as the controller, but it is never recommended that it matches the stash. Since this account does not need to send transactions, it does not need to have funds available. (If the mode is ever compromised, the controller can set an new session key)
+
+As with the nomination operations, you can stop validation at any time using the controller. (Be if for maintence, upgrades, or any other reason)
```
Tbaut: typos
```suggestion
As with the nomination operations, you can stop validation at any time using the controller. (Be it for maintenance, upgrades, or any other reason)
```

jacogr: Sorry. Too many in here.

jacogr:
@Tbaut Thanks for taking the time with the staking "ramblings".

packages/app-staking/src/md/basic.md
```diff
@@ -0,0 +1,32 @@
...
+
+At any point you could stop nominating (using the controller) or top up the funds bonded (from the stash).
+
+# Validating
+
+Validators run nodes that author blocks. The primary requirement here is the ability to run a node that is available 24/7 and is well-connected to the network.
+
+In addition to the stash and controller account described above, a validator has to indicate an additional account called `Session key` -
+
+- **Session** The seed for this key is passed to the node. It can be the same as the controller, but it is never recommended that it matches the stash. Since this account does not need to send transactions, it does not need to have funds available. (If the mode is ever compromised, the controller can set an new session key)
```
Tbaut: What do you think about talking about `account` instead of key for session? The `session key` being simply the seed of that account.
```suggestion
- **Session** The seed of this account should be passed to the node using the `--key` parameter. The session account does not need to have funds as it does not need to send any transaction. The best practice is to create a dedicated account to be used as session account. Although a single account can theoretically be used as both session and controller, it is not recommended to do so. Having a dedicated session account would prevent the theft of funds should the validator node be compromised and the `--key` leaked.
```

jacogr: Happy with that, I was to-and-fro-ing a bit on that one. But re-reading above, since the "seed is passed", account fits in better with the terminology used all-over.

jacogr:
@rossbulat If you are ever looking for something to do between 00:01 and 02:01  - this component would work really well as a slide in-out like with the sidebar and would fit in with those actions added. (Just have not deconstructed that myself to add here. Will get there, but if you don't feel that bored at some point...)

PR #923, opened 2019-04-08
Add BondedRx component & display bonded value, refactor RecentlyOffline

### LINKED ISSUE #896: Staking overview should display bonded value

kwingram25:
Closes #896 

Adds new reactive component to fetch address' controller and then its bonded value

(Accidentally included some work towards showing recentlyOffline in more places as well)

### DIFF
--- a/packages/app-staking/src/Account/index.tsx
+++ b/packages/app-staking/src/Account/index.tsx
@@ -87,6 +87,7 @@ class Account extends React.PureComponent<Props, State> {
           name={name}
           value={accountId}
           identIconSize={96}
+          withBonded
         >
           <div className='staking--Account-expand'>
             {this.renderButtons()}
@@ -188,7 +189,8 @@ class Account extends React.PureComponent<Props, State> {
             <AddressMini
               key={index}
               value={nomineeId}
-              withBalance
+              withBalance={false}
+              withBonded
             />
           ))
         }
@@ -263,7 +265,7 @@ class Account extends React.PureComponent<Props, State> {
     return (
       <div className='staking--Account-detail'>
         <label className='staking--label'>{t('stash')}</label>
-        <AddressMini value={stashId} />
+        <AddressMini withBalance={false} withBonded value={stashId} />
       </div>
     );
   }
--- a/packages/app-staking/src/Account/index.tsx
+++ b/packages/app-staking/src/Account/index.tsx
@@ -92,6 +92,8 @@ class Account extends React.PureComponent<Props, State> {
           value={accountId}
           identIconSize={96}
           withBonded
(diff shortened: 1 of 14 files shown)


packages/app-staking/src/Account/index.tsx
```diff
@@ -188,7 +189,8 @@ class Account extends React.PureComponent<Props, State> {
             <AddressMini
               key={index}
               value={nomineeId}
-              withBalance
+              withBalance={false}
```
jacogr: Happy to try, however think the balance is quite useful - 

- you can see (on the contrroller) if you have enough funds to send transactions
- you can see (on the stash) if there is extra funds that you may want to bond

packages/app-staking/src/Account/index.tsx
```diff
@@ -263,7 +265,7 @@ class Account extends React.PureComponent<Props, State> {
     return (
       <div className='staking--Account-detail'>
         <label className='staking--label'>{t('stash')}</label>
-        <AddressMini value={stashId} />
+        <AddressMini withBalance={false} withBonded value={stashId} />
```
jacogr: Here I believe just the bonded value is all-ok. (Since we are really just interested in that)

packages/app-staking/src/RecentlyOffline.tsx
```diff
@@ -0,0 +1,59 @@
...
+            }
+          })}
+        </div>
+      </div>
+    );
+  }
+
+  private toggleOpen = (): void => {
+    const { open } = this.state;
+    this.setState({ open: !open });
```
jacogr: Probably prefer the form -

```
setState(({ open }) => ({ open: !open }))
```

(Not that there are any other state changes going on, but probably more along the lines of best practice?

(Not a difference here, but just in case, maybe...)

packages/ui-app/src/AddressSummary.tsx
```diff
@@ -224,6 +231,7 @@ export {
 
 export default translate(
   withCalls<Props>(
-    ['derive.accounts.idAndIndex', { paramName: 'value' }]
+    ['derive.accounts.idAndIndex', { paramName: 'value' }],
+    'query.session.validators'
```
jacogr: Do we need the validators here? Don't see where we are using it.

packages/app-staking/src/Account/index.tsx
jacogr: Ummm... scratch that, need to look carefully next time - if does show for the accounts actions.

Issue #381, opened 2018-10-25
Contract management application (substrate-only)


jacogr:
- Allow deployment of contracts
- Allow view of deployed contracts (including addresses)
- Allow interaction with deployed contracts

jacogr:
Need #482

jacogr:
Requires API https://github.com/polkadot-js/api/issues/429

jacogr:
Needs https://github.com/polkadot-js/api/issues/807

PR #928, opened 2019-04-08
Substrate contracts, basics

jacogr:
- Closes https://github.com/polkadot-js/apps/issues/381
- As it stands https://youtu.be/mLabqSG74J8

TODO -

- putCode and create tested and working (erc20 failure in the video is a known issue - the output had some memory issues... things are still adapting on the platform)
- have not been able to test call e2e, there is an issue with the substrate dispatch (being worked on)


jacogr:
Removed it from the sidebar so it is not visible, but still reachable via `/#/contracts` - the reason being that it can be merged sooner rather than later (even when not perfect), so it can also be passed to the contract team to play with and provide actual feedback on operation.

PR #929, opened 2019-04-09
RecentlyOffline added to Account Actions

kwingram25:
Closes #881 

Relocates RecentlyOffline and adds to mini addresses on app-staking Account Actions page for controller, stash, etc.

It uses a ReactTooltip instead of expander due to space limitations; not sure if we want to use just that version everywhere?

Screen Shot 2019-04-08 at 9 32 24 PM


packages/app-staking/src/Account/index.tsx
```diff
@@ -58,12 +60,28 @@ class Account extends React.PureComponent<Props, State> {
...
     bondedId: null,
+    recentlyOffline: {},
     sessionId: null,
     stashId: null
   };
 
-  static getDerivedStateFromProps ({ session_nextKeyFor, staking_bonded, staking_ledger }: Props, state: State): Partial<State> {
+  static getDerivedStateFromProps ({ session_nextKeyFor, staking_bonded, staking_ledger, staking_recentlyOffline = [] }: Props, state: State): Partial<State> {
     return {
+      recentlyOffline: staking_recentlyOffline.reduce(
```
jacogr: Can we not rather calculate this on the main index and pass it in? (So instead of each account having a sub, which is cached, and doing the same thing)

jacogr:
So apart from it not working for me on the overview, the tooltip would actually work quite well (I believe) with the `ui-app/Labelled` component where we display the help. (Instead of the custom we have in there now)

kwingram25:
Ah, I broke the original one while fixing the travis errors! Moved recentlyOffline state to index.tsx

PR #953, opened 2019-04-10
Allow exports of dumb components

amaurymartiny:
Other projects might (hopefully) want to re-use components from apps. Right now, it's kind of hard because they need to provide a React.Context with an api object, which is not what every project wants to do.

This PR allows exporting the dumb components. Dumb components are easy to share/test/re-use, pass in props and they render correctly.

```javascript
import Chain from 'ui-reactive/Chain'; // Smart HOC'd component 
import { Chain } from 'ui-reactive/Chain'; // Dumb component
```

### DIFF
--- a/packages/ui-reactive/src/BestNumber.tsx
+++ b/packages/ui-reactive/src/BestNumber.tsx
@@ -15,7 +15,7 @@ type Props = BareProps & CallProps & {
   chain_bestNumber?: BlockNumber
 };
 
-class BestNumber extends React.PureComponent<Props> {
+export class BestNumber extends React.PureComponent<Props> {
   render () {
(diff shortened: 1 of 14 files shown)


packages/ui-reactive/src/BestNumber.tsx
```diff
@@ -28,7 +28,7 @@ class BestNumber extends React.PureComponent<Props> {
           chain_bestNumber
             ? formatNumber(chain_bestNumber)
             : '-'
-          }{children}
+        }{children}
```
amaurymartiny: not sure why my cmd+s did this... though I would argue this one looks better

jacogr: And more correct. Been seeing some funky spacing not being complained about lately :(

PR #1021, opened 2019-04-15
Replace withCall -> withCalls

### LINKED ISSUE #1009: Convert all withCall to withCalls<Props>
Goal:

Make everything consistent, so there is basically one way of doing things.

Background:

`withCalls` is a relative new addition (replacing `withMulti(withCall(abc), withCall(def), ...)`and actually have some additional properties for types and is slightly more future-proof when changing components. 

Generally the `Props` gets lost with `withCall` and is handled (slightly) better in the case of `withCalls<Props>` - nowhere near perfect, we really would like to have decorators at some point. (But more or less given up on that for now, the TS support is just lacking and types become an absolute nightmare)

TL;DR `withCalls` has _some_ benefits, at the very least when you add more stuff, you can just add it

Actions:

- For all `withCall`, replace it with `withCalls<Props>(...)`
- Adjust props as needed, I've noticed there are places it doesn't even get specified, e.g. https://github.com/polkadot-js/apps/blob/master/packages/ui-reactive/src/Balance.tsx#L12 - no `params` in props, but actually used

kwingram25:
Closes #1009 

Replace all instances of withCall with withCalls<Props>

### DIFF
--- a/packages/app-explorer/src/BlockQuery/index.tsx
+++ b/packages/app-explorer/src/BlockQuery/index.tsx
@@ -5,7 +5,7 @@
 import { BareProps } from '@polkadot/ui-app/types';
 
 import React from 'react';
-import { withCall, withMulti } from '@polkadot/ui-api';
+import { withCalls, withMulti } from '@polkadot/ui-api';
 import { isHex } from '@polkadot/util';
 import { BlockNumber } from '@polkadot/types';
 
@@ -83,5 +83,5 @@ class Entry extends React.Component<Props, State> {
 
 export default withMulti(
   Entry,
-  withCall('derive.chain.bestNumber')
+  withCalls<Props>('derive.chain.bestNumber')
 );
--- a/packages/app-extrinsics/src/Balance.tsx
+++ b/packages/app-extrinsics/src/Balance.tsx
@@ -6,7 +6,7 @@ import { BareProps, CallProps } from '@polkadot/ui-api/types';
 
 import React from 'react';
 import { Balance } from '@polkadot/types';
-import { withCall } from '@polkadot/ui-api';
+import { withCalls } from '@polkadot/ui-api';
 import { InputBalance } from '@polkadot/ui-app';
 
 type Props = BareProps & CallProps & {
@@ -30,4 +30,6 @@ class BalanceDisplay extends React.PureComponent<Props> {
   }
 }
 
-export default withCall('query.balances.freeBalance', { paramName: 'params' })(BalanceDisplay);
+export default withCalls<Props>(
+  ['query.balances.freeBalance', { paramName: 'params' }]
+)(BalanceDisplay);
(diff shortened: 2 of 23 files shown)


packages/apps/src/SideBar/index.tsx
```diff
@@ -190,19 +191,17 @@ class SideBar extends React.PureComponent<Props> {
...
-    //   <Menu.Item className='apps--SideBar-Item'>
-    //     <a
-    //       className='apps--SideBar-Item-NavLink'
-    //       href='https://github.com/w3f/Web3-wiki/wiki/Polkadot'
-    //     >
-    //       <Icon name='book' /> Wiki
-    //     </a>
-    //   </Menu.Item>
-    // );
+    return (
```
Stefie: Could you revert this change? I think it‘s part of your PR #1020

kwingram25: oops! Thank you

jacogr:
Cool. Any major issue along the way?

packages/app-extrinsics/src/Balance.tsx
```diff
@@ -30,4 +30,9 @@ class BalanceDisplay extends React.PureComponent<Props> {
   }
 }
 
-export default withCall('query.balances.freeBalance', { paramName: 'params' })(BalanceDisplay);
+export default withMulti(
```
jacogr: Not sure we need a withMulti here? I am guessing some types started complaining.

packages/app-explorer/src/BlockQuery/index.tsx
```diff
@@ -83,5 +83,5 @@ class Entry extends React.Component<Props, State> {
 
 export default withMulti(
   Entry,
-  withCall('derive.chain.bestNumber')
+  withCalls<Props>('derive.chain.bestNumber')
```
jacogr: Probably can get rid of withMulti here as well?

packages/app-extrinsics/src/Balance.tsx
kwingram25: Yes, some of the reactive components threw issues when exported as withCalls(...) rather than withCall() or withMulti() - 'params' isn't supported?

Type '{ params: string | undefined; }' is not assignable to type 'IntrinsicAttributes & Pick<Props, "className" | "style" | "children" | "label" | "callResult" | "callUpdated" | "callUpdatedAt" | "accounts_idAndIndex"> & { children?: ReactNode; }'. Property 'params' does not exist on type 'IntrinsicAttributes & Pick<Props, "className" | "style" | "children" | "label" | "callResult" | "callUpdated" | "callUpdatedAt" | "accounts_idAndIndex"> & { children?: ReactNode; }'.

jacogr:
Yes, assumed as much. The previous iteration (just due to TS being quite horrible in HOC support) actually mostly ende up with any-ing of the props. withCalls is slightly more correct (but not perfect).

jacogr:
So where params pop up, it actually means that is probably doesn’t appear in Props (And should really)

kwingram25:
Resolved params problems, but I had to add callOnResult type to Nonce for one instance - I can't tell if theres a prop type somewhere that it needs specifically

PR #1022, opened 2019-04-15
Mark raw seed as invalid if smaller than 32 char

Tbaut:
- closes 1019

The bug reported in the issue happened when `seed===''`.

### DIFF
--- a/packages/app-accounts/src/Creator.tsx
+++ b/packages/app-accounts/src/Creator.tsx
@@ -71,7 +71,7 @@ function isHexSeed (seed: string): boolean {
 }
 
 function rawValidate (seed: string): boolean {
-  return ((seed.length > 0) && (seed.length <= 32)) || isHexSeed(seed));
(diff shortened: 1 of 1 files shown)


jacogr:
Ummm. Ok, so for text seeds, which is handled in the same box we do allow eg. “Alice” (the fromUri does all padding as required). 

So there def needs to be sanity as introduced here, not quite sure we want to go back to manual padding by the user. 

Will pull down this and take a peek anyway later today.

packages/app-accounts/src/Creator.tsx
```diff
@@ -71,7 +71,7 @@ function isHexSeed (seed: string): boolean {
 }
 
 function rawValidate (seed: string): boolean {
-  return seed.length <= 32 || isHexSeed(seed);
+  return seed.length >= 32 || isHexSeed(seed);
```
jacogr: Finally played around a bit, think this should rather be `return seed.length > 0 && (seed.length <= 32 || isHexSeed(seed));`

Tbaut:
Thanks

PR #1023, opened 2019-04-16
Refactor TxButton and expand use

### LINKED ISSUE #1010: Convert all tx sending to TxButton
Goal:

Single, simple way of submitting transactions

Background:

Initially all txs were created manually, e.g. https://github.com/polkadot-js/apps/blob/master/packages/app-transfer/src/Transfer.tsx#L73-L82 & https://github.com/polkadot-js/apps/blob/master/packages/app-transfer/src/Submit.tsx#L29-L36

There is a better way (and linking here to a complex) example https://github.com/polkadot-js/apps/blob/master/packages/app-contracts/src/Code.tsx#L110-L120 or just simply, https://github.com/polkadot-js/apps/blob/master/packages/app-staking/src/Account/Bond.tsx#L74-L82

Apart from being simpler, it has less imports, less moving parts, less stuff to trip over and also keeps things DRY

Actions:

- Find all places where we generate the tx and place it on the queue, just convert to tx button (contracts & staking are exceptions, all other apps rely on the queue mechanism)
- (Optionally) Adjust TxButton to also have some pizzazz - when busy, mark as `isBusy`, show busy indicator on it with an icon.
- (I believe I marked all affected apps)

kwingram25:
Closes #1010 

- Replace instances of QueueConsumer/queueExtrinsic with TxButton
- Refactor TxButton to accept either method string and params or prepared IExtrinsic / SubmittableExtrinsic and accountNonce

I believe the types and execution check out, not sure yet how to test for democracy app however. Also not 100% sure how SubmittableExtrinsic differs from IExtrinsic - any reason to use one over the other?


jacogr:
Summitable extends extrinsic - extrinsic is the base type, summit table adds helpers in the API so you can easily sign and submit. 

Generally it can be Extrinsic, but the API adds the sign and signAndSend, which is used in the signer. 

So where we contrarily these (everywhere), they are normally Submittable - when retrieved as part of an API call (eg block details) they are the lower form

kwingram25:
Added disabled & loading indicator to TxButton while tx is broadcasting

packages/ui-app/src/TxButton.tsx
```diff
@@ -29,16 +34,59 @@ type Props = ApiProps & {
...
+
+      extrinsic = api.tx[section][method](
+        ...(isFunction(params) ? params() : params)
+      );
+    } else {
+      extrinsic = tx as SubmittableExtrinsic;
+    }
+
+    let isSending = false;
+    let queuedTx = txqueue.find(({ extrinsic: ex }) => ex === extrinsic);
```
jacogr: const?

packages/ui-app/src/TxButton.tsx
```diff
@@ -48,24 +96,20 @@ class TxButtonInner extends React.PureComponent<Props & InjectedProps> {
...
-    );
-
     queueExtrinsic({
       accountId,
       extrinsic,
       txFailedCb: onFailed,
       txSuccessCb: onSuccess,
-      txUpdateCb: onUpdate
+      txUpdateCb: onUpdate,
+      ...(accountNonce ?
```
jacogr: We could have left this our, only used in transfer. I'm guessing it is actually a hold-over from before the API automagically loaded the nonce.

packages/ui-app/src/TxButton.tsx
```diff
@@ -29,16 +34,59 @@ type Props = ApiProps & {
...
+    if (typeof tx === 'string') {
+      const [section, method] = tx.split('.');
+
+      assert(api.tx[section] && api.tx[section][method], `Unable to find api.tx.${section}.${method}`);
+
+      extrinsic = api.tx[section][method](
+        ...(isFunction(params) ? params() : params)
+      );
+    } else {
+      extrinsic = tx as SubmittableExtrinsic;
```
jacogr: Not 100% sure about this - this is used in extrinsic app and that one is a bit special, and probably the only one with any reason to pull in the queue itself. Maybe rather make it explicit and pass in an (optional) `extrinsic` prop?

kwingram25: Done - extrinsic either passed or generated from tx string

kwingram25:
Noticed that "sending" instead of "broadcast" is used now - is that everywhere or is broadcast sometimes still applied?

jacogr:
So "sending" is an internal status, "broadcast" comes from the node, i.e. https://github.com/polkadot-js/api/blob/master/packages/types/src/rpc/ExtrinsicStatus.ts#L79 & https://github.com/paritytech/substrate/blob/0e6a407a13fbc5a5ad200645aed72bb8a8e528d7/core/transaction-pool/graph/src/watcher.rs#L38

jacogr:
... sending always happens, broadcast may not happen. (i.e. if connected to a dev chain with no network or if connected to a validator that manages to grab the next block)

PR #1026, opened 2019-04-16
Hide Sidebar scrollbar on Firefox (like chrome)

### LINKED ISSUE #1025: Visible sidebar scrollbar on Firefox even when not needed
The image says it all. On chrome the sidebar is hidden (explicitly with a `::-webkit-scrollbar`)

image

Tbaut:
- closes #1025 

Couldn't find any better hack than what's said in [this thread](https://stackoverflow.com/questions/20997183/how-to-hide-scrollbar-in-firefox) to make the behaviour on Firefox the same as chrome.

### DIFF
--- a/packages/apps/src/SideBar/SideBar.css
+++ b/packages/apps/src/SideBar/SideBar.css
@@ -69,8 +69,9 @@
       display: flex;
       flex-direction: column;
       height: 100vh;
-      overflow-y: scroll;
+      overflow-y: auto;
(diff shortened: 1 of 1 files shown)


Stefie:
See my comment in the original issue, I can't reproduce the bug in FF.

jacogr:
Depends on OS as well, ie. Mac is notorious (;)) for hiding any and all scroll bars when it can.

PR #1027, opened 2019-04-16
Make sure labels aren't above help-hover

### LINKED ISSUE #1024: Help ovelay is under the field's next label
This is visible as soon as there is a help overlay that is large enough to reach the following label. Example when bounding:
image

For some reason, the following label (`z-index:1`) is above the .help-hover (`z-index:10`)

Tbaut:
- closes #1024

It turned out to be more complex than I expected. TL;DR It was due to the fact that a parent element had an `opacitiy`<1 in our case the `label` and also had a z-index. (a [good article](https://philipwalton.com/articles/what-no-one-told-you-about-z-index/) about it).

I modified the color of `label` to make it match what it looked like before (with the opacity). Here is how things look like now (basically the same, the background of the help-hover is a little darker, which makes the help text easier to read). 
image

### DIFF
--- a/packages/ui-app/src/Labelled.tsx
+++ b/packages/ui-app/src/Labelled.tsx
@@ -31,7 +31,6 @@ const Wrapper = styled.div`
     margin: 0.25rem 0 0 0;
     padding-right: 0.5rem;
     position: relative;
-    z-index: 1;
 
     .help-hover {
       background: #222;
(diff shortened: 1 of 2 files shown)


jacogr:
Yes. The opacity, in hindsight, was an easy way in, but problematic down the line. Think this is the right approach, or rather I was thinking about the same route - so if wrong, we both are ;)

PR #1030, opened 2019-04-17
Add help on critical apps

### LINKED ISSUE #946: Add extra info and help to critical apps
Goal:

We want to make the app itself as self-explanatory as possible, which means aligning all different sections to provide at-the-fingertip information to all users.

Background:

The staking app has been leading the way is making it simpler to understand what is going on and by providing help for all of the input fields via "(?)".

Actions:

(For all of: democracy, transfer, accounts & addresses)

- Add "(?)" `help` property to all input fields, with relevant information that explains the use
- Add a help overlay (as the `/#/staking` and `/#/contracts` apps has)
- The help overlay should have a layman description of the app and what you can aim to achieve - additionally it should describe any terms that are used and not part of common-day vocabulary

Tbaut:
- closes #946 
- Shorten labels as more info is now available

Todo
add (?) for inputs in
- [x] accounts
- [x] address book
- [x] transfer
- [x] democracy

add overlay in
- [x] accounts
- [x] address book
- [x] transfer
- [x] democracy


jacogr:
Brilliant! This really is a lot of cleanup (already at this stage), but critical for ease of use.

Tbaut:
I need some more info on the key derivation, I basically copied the comments from the code.. not sure it's enough. If it is, I'd simple mention it in the (?), without needing to talk about it in the overlay.

Tbaut:
If I understood correctly, there will be only 1 referendum to vote on at a time. I've fixed a couple typos as we use referenda as the plural of referendum.

The whole governance is quite complex and I kept things simple for a start, we should make it evolve as we add features to the UI.

Tbaut:
Requesting a review from @lsaether

packages/app-accounts/src/md/basic.md
```diff
@@ -0,0 +1,40 @@
...
+You can recover an account from its:
+- seed or mnemonic:  
+  From the "Create account" tab, type your seed or mnemonic in the associated field.
+
+- backup file (also called JSON keystore file) and the account's password:
+  From the "Restore account" tab. Upload your backup file and type in the password associated.
+
+# Minimum allowed balance
+
+Accounts with a balance lower than the minimal amount, 100 mili Unit as of writing are considered as inexistent for the network. If an account's balance ever drops below this amount, it is removed from the network. In this application, it will still be visible, but with a balance of 0.
```
lsaether:
```suggestion
Accounts with a balance lower than the minimal amount, 100 milliDOTs as of writing are considered as nonexistent for the network. If an account's balance ever drops below this amount, it is removed from the network. In this application, it will still be visible, but with a balance of 0.
```

Units would be better for Substrate chains but maybe confusing for people needing help with Polkadot?

packages/app-transfer/src/md/basic.md
```diff
@@ -0,0 +1,22 @@
+# Transfer
+
+This tab allows you to send funds from one account you own to another account.
+
+# Minimum allowed balance
+
+Accounts with a balance lower than the minimal amount, 100 mili Unit as of writing, are considered as inexistent for the network. If an account's balance ever drops below this amount, it is removed from the network. In this application, it will still be visible, but with a balance of 0.
```
lsaether: same as above

packages/app-democracy/src/md/basic.md
```diff
@@ -0,0 +1,13 @@
+# Democracy
+
+This application lets you view proposals and vote for or against a referendum.
+Anyone can create a proposal by depositing minimum value in a certain period of time (No. of Blocks). If someone likes the proposal, they could deposit the same amount of tokens to support it.
```
lsaether:
```suggestion
Anyone can create a proposal by bonding the minimum deposit for a certain period of time (No. of Blocks). If someone likes the proposal, they could deposit the same amount of tokens to support it. The minimum deposit is currently 5 DOTs.
```

packages/app-transfer/src/md/basic.md
Tbaut:
```suggestion
Accounts with a balance lower than the minimal amount, 100 miliDOTs as of writing, are considered as inexistent for the network. If an account's balance ever drops below this amount, it is removed from the network. In this application, it will still be visible, but with a balance of 0.
```

packages/app-accounts/src/md/basic.md
Stefie: Hmm, but the help, like the whole interface,  is not only for Polkadot, but also for Substrate chains. So keeping it more generic with 'Unit' makes more sense. Having the help mentioning DOTs on Embric Elm and other Substrate chains is equally confusing (and wrong).

Tbaut: Right, since Polkadot users might be the ones to be the most confused (Substrate users know about Polkadot, but Polkadot users might not know about Substrate), I've catered for both using " miliUnits (miliDOTs for Polkadot)"

PR #1036, opened 2019-04-20
Sudo app added

### LINKED ISSUE #974: Add a basic sudo app
Goal: 

To help people managing chains (typically dev) to easily make changes to the chain.

Background:

The sudo module allows you to submit root transactions, provided you control the sudo key. this is very useful for performing upgrades or setting specific values - and typically is only used in dev environments, although some test chains do enable them.

Actions:

- Add a sudo app
- It is only visible when: the chain has the `tx.sudo.setKey` function _and_ the user has the key for `query.sudo.key` in their keyring. (The latter we don't allow filtering on the sidebar yet)
- First step: extend the sidebar filtering to perform a query and based on the value return true/false
- The sudo module should have, as a first step, the ability to set the sudo key `tx.sudo.setKey`
- Additionally it should allow for making sudo calls, submitted via the sudo key

For the latter, we probably want to curate some frequently used (_if_ available, i.e. `consensus.setCode`, `consensus.setStorage` - this we want to piggy-back on https://github.com/polkadot-js/apps/issues/973)

kwingram25:
Closes #974

Sudo app allowing use of setKey() and sudo(proposal)
- Checks for ownership of existing key and changes view/permissions accordingly


jacogr:
Cool. Excited to play with this one.

packages/app-sudo/src/Propose.tsx
```diff
@@ -0,0 +1,105 @@
+// Copyright 2017-2019 @polkadot/app-123code authors & contributors
+// This software may be modified and distributed under the terms
+// of the Apache-2.0 license. See the LICENSE file for details.
+import { I18nProps } from '@polkadot/ui-app/types';
```
jacogr: (nit) empty line before

packages/app-sudo/src/index.css
```diff
@@ -0,0 +1,25 @@
+/* Copyright 2017-2019 @polkadot/app-123code authors & contributors
+/* This software may be modified and distributed under the terms
+/* of the Apache-2.0 license. See the LICENSE file for details. */
+
+.sudo--summary {
```
jacogr: Since it is new, probably would have gone for styled-components here rather instead of an additional CSS file.

packages/app-sudo/src/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+      </main>
+    );
+  }
+}
+
+export default withMulti(
+  App,
+  translate,
+  withApi,
+  withCall(
```
jacogr: `withCalls`?

jacogr:
Referring to this above -

image

jacogr:
One more tweak - I re-assigned the key to an external, and then refreshed - in this case, the app shouldn't show anymore.

image

(Just after re-assignment is ok, but when refreshing, the sidebar should actually have a check for this app - which means extending the logic there for checking somewhat. Possibly have a check function that passes through and api and returns async true/false .... lemme know if I'm way too cryptic here)

Issue #971, opened 2019-04-11
Allow filtering in staking overview (by type)


jacogr:
Goal:

Allow the user to have a better view of the information presented.

Background:

In https://github.com/polkadot-js/apps/pull/965 some really basic filtering was introduced for the accounts section of staking. We need to expand this to filter the ever growing list of validators. (This is related to https://github.com/polkadot-js/apps/issues/951, but an basic no-frills addition to use already-available information for a up-front filter)

Actions:

- Add a filter
- Pass through the filter to the actual validators rendered
- Suggestions for filter categories:
  - All validators/intentions (default)
  - Stashes with no nominators
  - Stashes with nominators
  - Stashes with no warnings
  - Stashed with warnings
  - Stashes that my accounts nominate

Stefie:
Solved with @Lowhearth 's PR that got merged this morning.

PR #1039, opened 2019-04-21
Added a simple filter to the staking overview (#971) .

Lowhearth:
Hello!
I'm in love with this project and I'd like to help as much as I can, this is my first attempt resolving the issue  #971.

Implementation of a basic "non-tag" filtering in the staking overview page inspired in #965:

Screenshot 2019-04-22 at 00 20 56

Screenshot 2019-04-22 at 00 21 15

### DIFF
--- a/packages/app-staking/src/Overview/CurrentList.tsx
+++ b/packages/app-staking/src/Overview/CurrentList.tsx
@@ -13,7 +13,6 @@ import { Dropdown } from '@polkadot/ui-app';
 import translate from '../translate';
 import Address from './Address';
 
-
 type Props = I18nProps & {
   balances: DerivedBalancesMap,
   current: Array<string>,
@@ -67,12 +66,12 @@ class CurrentList extends React.PureComponent<Props, State> {
       <Wrapper>
         <div className='filter'>
           <Dropdown
-                help={t('Select which validators/intentions you want to display, all, only the ones you nominated, only the ones that has/no nominators, only the ones that has/no warnings.')}
-                label={t('filter')}
-                onChange={this.onChangeFilter}
-                options={filterOptions}
-                value={filter}
-              />
+            help={t('Select which validators/intentions you want to display, all, only the ones you nominated, only the ones that has/no nominators, only the ones that has/no warnings.')}
+            label={t('filter')}
+            onChange={this.onChangeFilter}
+            options={filterOptions}
+            value={filter}
+          />
         </div>
         <div className='validator--ValidatorsList ui--flex-medium'>
           <div className='validator--current'>
--- a/packages/app-staking/src/Overview/CurrentList.tsx
+++ b/packages/app-staking/src/Overview/CurrentList.tsx
@@ -26,6 +26,7 @@ type State = {
   filter: ValidatorFilter,
   filterOptions: Array<{ text: React.ReactNode, value: ValidatorFilter }>
 };
+
 const Wrapper = styled.div`
   .filter {
     display: flex;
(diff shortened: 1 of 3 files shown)


jacogr:
Nice. Thank you. Any help is really appreciated.

jacogr:
Sorry for the radio silence from me - traveling. Hopefully somebody else will take a look before I get there.

packages/app-staking/src/Overview/CurrentList.tsx
```diff
@@ -20,17 +22,66 @@ type Props = I18nProps & {
...
+      filter: 'all',
+      filterOptions: [
+        { text: t('Show all validators and intentions'), value: 'all' },
+        { text: t('Show only my nominations'), value: 'iNominated' },
+        { text: t('Show only with nominators'), value: 'hasNominators' },
+        { text: t('Show only without nominators'), value: 'noNominators' },
+        { text: t('Show only with warnings'), value: 'hasWarnings' },
+        { text: t('Show only without warnings'), value: 'noWarnings' }
+      ]
+    };
```
Stefie: The `filterOptions` don't change so you could declare them as a const instead of part of the component state.

packages/app-staking/src/Overview/CurrentList.tsx
```diff
@@ -20,17 +22,66 @@ type Props = I18nProps & {
...
+    const { t } = this.props;
+    const { filter, filterOptions } = this.state;
     return (
-      <div className='validator--ValidatorsList ui--flex-medium'>
-        <div className='validator--current'>
-          {this.renderCurrent()}
+      <Wrapper>
+        <div className='filter'>
+          <Dropdown
+            help={t('Select which validators/intentions you want to display, all, only the ones you nominated, only the ones that has/no nominators, only the ones that has/no warnings.')}
```
Stefie: The `help` text has some grammatical errors. Once you click on the dropdown the options are pretty straight forward so you don't really need to list them all in the help text. I'd keep it short:
`Select which validators/intentions you want to display.`

packages/app-staking/src/Overview/CurrentList.tsx
Lowhearth: I'm not really sure how to do it since I need  `t()` from props.
The only possible alternative  I can come up with is something like `this` 
```
class CurrentList extends React.PureComponent<Props, State> {
  state: State;
  filterOptions: FilterOptions;

  constructor (props: Props) {
    super(props);

    const { t } = props;

    this.state = {
      filter: 'all'
    };

    this.filterOptions = [
      { text: t('Show all validators and intentions'), value: 'all' },
      { text: t('Show only my nominations'), value: 'iNominated' },
      { text: t('Show only with nominators'), value: 'hasNominators' },
      { text: t('Show only without nominators'), value: 'noNominators' },
      { text: t('Show only with warnings'), value: 'hasWarnings' },
      { text: t('Show only without warnings'), value: 'noWarnings' }
    ]
  }
```
but I'm not declaring them as const this way.
¿How would you do it?

packages/app-staking/src/Overview/CurrentList.tsx
Lowhearth: Totally agree, thanks!

packages/app-staking/src/Overview/CurrentList.tsx
Stefie: Good point! And we're actually putting translated strings into the state in other components, so just leave it as it is.

Stefie:
Nice, thanks for your contribution!
Is it good to go from your side? If yes I'll merge it later.

Lowhearth:
That's OK.
Thanks your for reviewing my code!

PR #1045, opened 2019-04-25
Upgrade dependencies to match Metadata v4 

Stefie:
For 1.0 release


yarn.lock
```diff
@@ -1762,7 +1762,7 @@
     "@types/memoizee" "^0.4.2"
     memoizee "^0.4.14"
 
-"@polkadot/api@^0.77.0-beta.4":
+"@polkadot/api@^0.77.0-beta.4", "@polkadot/api@^0.77.1":
```
amaurymartiny: There shouldn't be any mention of `0.77.0-beta.4` inside yarn.lock anymore. Maybe update the resolutions field in package.json too?

Stefie: Ah, dammit....

Stefie: As in: Yes, totally. Should be fixed now

package.json
```diff
@@ -10,9 +10,9 @@
     "packages/*"
   ],
   "resolutions": {
-    "@polkadot/api": "^0.77.0-beta.4",
+    "@polkadot/api": "^0.77.1",
     "@polkadot/keyring": "^0.76.1",
-    "@polkadot/types": "^0.77.0-beta.4",
+    "@polkadot/types": "^0.77.1",
```
Stefie: @amaurymartiny  It already is on 0.77.1 ?

package.json
```diff
@@ -10,9 +10,9 @@
     "packages/*"
   ],
   "resolutions": {
-    "@polkadot/api": "^0.77.0-beta.4",
+    "@polkadot/api": "^0.77.1",
```
Stefie: api too

amaurymartiny:
Ah sorry I wasn't clear, I was talking about https://github.com/polkadot-js/apps/blob/master/package.json#L2

Issue #859, opened 2019-03-27
User should be able to unbond & withdraw funds


jacogr:
(Split from #715 with the outstanding items)

Goal:

Allow the user to properly manage their funds in the staking module and allow them to move them between accounts. 

Background:

Currently the user can bond, add additional funds, but cannot withdraw. 

Pre-requisites:

A good understanding of the functions available inside the staking module https://polkadot.js.org/api/METHODS_EXTRINSICS.html#staking

Actions:

(All these apply to the controller)

- Allow the user to unbond, via `unbond` - clicking creates a pop-up that allows the suer to specify the amount he/she wants to unbond. (This should be checked against the available)
- Allow the user to withdraw via `withdrawUnbonded` - this option should _only_ be available when there are funds that have been unbonded already
- Additionally, when the user has funds that have been unbonded, but are not available yet - show the information for these on the overview for the account - fund total and when (block) these will be available. (Information available from the staking ledger)

jacogr:
The tricky bits here are really UX - we don't want to end up with a deluge of buttons. Probably want to re-think things here a bit and have 2 sub-sections, one for the actual actions such as "setting keys" (also see <to be logged>) and one for fund management.

Tbaut:
I've been thinking about it in the past days, studied the apis and want to take a stab at it.

>- Allow the user to unbond, via unbond - clicking creates a pop-up that allows the suer to specify the amount he/she wants to unbond. (This should be checked against the available)

Yes and remove the "Bound additional" when there's nothing more you can add from the stash.


>- Allow the user to withdraw via withdrawUnbonded - this option should only be available when there are funds that have been unbonded already

I was talking with @kianenigma to see if this can be automated on the backend side as this is a pretty bad UX that's not understandable by users. 

> -Additionally, when the user has funds that have been unbonded, but are not available yet - show the information for these on the overview for the account - fund total and when (block) these will be available. (Information available from the staking ledger)

I couldnt'd come with something nicer than:
![image](https://user-images.githubusercontent.com/33178835/56821767-87b4f800-684f-11e9-990f-f52f922018aa.png)

Tbaut:
Circling back on the discussion I had with @Kianenigma. The reason to have a `withdrawUnbonded` is to release the workload on the engine, and move this hurdle to the UI (the hurdle being: check whenever some funds are free to be released and call the extrinsic `withdrawUnbonded`). We will therefore not be able to push back on that.

[...]

PR #1057, opened 2019-04-30
Add unbond button

Tbaut:
Addresses the `unbond` part in https://github.com/polkadot-js/apps/issues/859

This PR also addresses 
- a little quirk that the "Bond Additional" button was still visible although there wasn't anything possible to bond.
- removes an unused `validators: Array<string>` type.
- ordering imports alphabetically 

I'll prepare a separate PR to address the `withdrawUnbonded` part of https://github.com/polkadot-js/apps/issues/859 to make things easier to review.

### DIFF
--- a/packages/app-staking/src/Account/Unbond.tsx
+++ b/packages/app-staking/src/Account/Unbond.tsx
@@ -12,6 +12,7 @@ import translate from '../translate';
 
 type Props = I18nProps & {
   controllerId?: AccountId | null,
+  isOpen: boolean,
   onClose: () => void
 };
 
@@ -23,9 +24,13 @@ class Unbond extends React.PureComponent<Props, State> {
   state: State = {};
 
   render () {
-    const { controllerId, onClose, t } = this.props;
+    const { controllerId, isOpen, onClose, t } = this.props;
     const { maxUnbond } = this.state;
     const canSubmit = !!maxUnbond && maxUnbond.gtn(0);
+
+    if (!isOpen) {
+      return null;
+    }
 
     return (
       <Modal
--- a/packages/app-staking/src/Account/Unbond.tsx
+++ b/packages/app-staking/src/Account/Unbond.tsx
@@ -6,12 +6,13 @@ import { I18nProps } from '@polkadot/ui-app/types';
 
 import BN from 'bn.js';
 import React from 'react';
+import { AccountId } from '@polkadot/types';
 import { Button, InputAddress, InputBalance, Modal, TxButton } from '@polkadot/ui-app';
 
 import translate from '../translate';
 
 type Props = I18nProps & {
-  accountId: string,
+  controllerId?: AccountId | null,
   isOpen: boolean,
   onClose: () => void
 };
@@ -24,7 +25,7 @@ class Unbond extends React.PureComponent<Props, State> {
   state: State = {};
 
   render () {
-    const { accountId, isOpen, onClose, t } = this.props;
+    const { controllerId, isOpen, onClose, t } = this.props;
     const { maxUnbond } = this.state;
     const canSubmit = !!maxUnbond && maxUnbond.gtn(0);
 
@@ -49,7 +50,7 @@ class Unbond extends React.PureComponent<Props, State> {
             />
             <Button.Or />
             <TxButton
-              accountId={accountId}
+              accountId={controllerId && controllerId.toString()}
               isDisabled={!canSubmit}
               isPrimary
               label={t('Unbond')}
@@ -64,7 +65,7 @@ class Unbond extends React.PureComponent<Props, State> {
   }
 
   private renderContent () {
-    const { accountId, t } = this.props;
+    const { controllerId, t } = this.props;
 
     return (
       <>
@@ -74,7 +75,7 @@ class Unbond extends React.PureComponent<Props, State> {
         <Modal.Content className='ui--signer-Signer-Content'>
           <InputAddress
             className='medium'
-            defaultValue={accountId}
+            defaultValue={controllerId && controllerId.toString()}
             isDisabled
             label={t('controler account')}
           />
--- a/packages/app-staking/src/Account/Unbond.tsx
+++ b/packages/app-staking/src/Account/Unbond.tsx
@@ -2,12 +2,11 @@
 // This software may be modified and distributed under the terms
 // of the Apache-2.0 license. See the LICENSE file for details.
 
-import { I18nProps } from '@polkadot/ui-app/types';
-
-import BN from 'bn.js';
-import React from 'react';
 import { AccountId } from '@polkadot/types';
+import BN from 'bn.js';
 import { Button, InputAddress, InputBalance, Modal, TxButton } from '@polkadot/ui-app';
+import { I18nProps } from '@polkadot/ui-app/types';
+import React from 'react';
 
 import translate from '../translate';
 
--- a/packages/app-staking/src/Account/Unbond.tsx
+++ b/packages/app-staking/src/Account/Unbond.tsx
@@ -12,7 +12,6 @@ import translate from '../translate';
 
 type Props = I18nProps & {
   controllerId?: AccountId | null,
-  isOpen: boolean,
   onClose: () => void
 };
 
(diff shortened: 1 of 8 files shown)


Tbaut:
Thanks, looking forward for the review, I wonder if there is a nice was to group the `staking.ledger` calls added here: https://github.com/polkadot-js/apps/pull/1057/files#diff-11cd827cd9b9d5e0d8503e090149dcacR464 I couldn't figure this out.

jacogr:
Not really, unless you return both in a single call. 

```js
['query.staking.ledger', {
      paramName: 'controllerId',
      propName: 'stashBonded',
      transform: (ledger: Option<StakingLedger>) =>
        ledger.unwrapOr({ active: null, total: null })
    }],
```

It is generally not too bad since the calls are cached on the API layer, so at the very least it doesn't do multiples.

packages/app-staking/src/Account/index.tsx
```diff
@@ -118,6 +121,19 @@ class Account extends React.PureComponent<Props, State> {
...
   }
 
+  private renderUnbond () {
+    const { controllerId } = this.props;
+    const { isUnbondOpen } = this.state;
+
+    return (
+        <Unbond
+          controllerId={controllerId}
+          isOpen={isUnbondOpen}
```
Stefie: There's  no need to pass`isOpen` down to the component. You can just conditionally render  `<Unbond>` it here in the index.tsx (see comment above)

packages/app-staking/src/Account/index.tsx
```diff
@@ -69,6 +71,7 @@ class Account extends React.PureComponent<Props, State> {
         {this.renderBondExtra()}
         {this.renderNominating()}
         {this.renderSessionKey()}
+        {this.renderUnbond()}
```
Stefie: Rather do `{this.state.isUnbondOpen  && this.renderUnbond()}` (or deconstruct this.state.isUnbondOpen before calling it).
 
There's no need to mount the `Unbond ` component, pass down the `isUnbondOpen` prop and then let the component return null if `isUnbondOpen` is false. It's better to just not call `this.renderUnbond()`  which mounts `<Unbond>`  at all if `isUnbondOpen` is false.

Same is true for all the other calls, so If you're up for refactoring tha in a second PR, feel free :).

packages/app-staking/src/Account/Unbond.tsx
```diff
@@ -0,0 +1,99 @@
...
+  state: State = {};
+
+  render () {
+    const { controllerId, isOpen, onClose, t } = this.props;
+    const { maxUnbond } = this.state;
+    const canSubmit = !!maxUnbond && maxUnbond.gtn(0);
+
+    if (!isOpen) {
+      return null;
+    }
```
Stefie: See my comments regarding `isOpen` on the index.js. It's cleaner to conditionally render this component in the index.js instead of retuning `null` here.

packages/app-staking/src/Account/Unbond.tsx
```diff
@@ -0,0 +1,99 @@
...
+import BN from 'bn.js';
+import React from 'react';
+import { AccountId } from '@polkadot/types';
+import { Button, InputAddress, InputBalance, Modal, TxButton } from '@polkadot/ui-app';
+
+import translate from '../translate';
+
+type Props = I18nProps & {
+  controllerId?: AccountId | null,
+  isOpen: boolean,
```
Stefie: Remove prop (see comment on index.js)

Tbaut:
Thanks for the reviews, although for this PR I didn't originally want to refactor the isOpen things, it's small enough that I'm confident not to break everything :)

packages/app-staking/src/Account/index.tsx
```diff
@@ -294,13 +299,14 @@ class Account extends React.PureComponent<Props, State> {
     }
 
     return (
+      isNominateOpen ?
```
jacogr: There _are_ actually very good reasons why some of these are done this way. 

In the cases where we do queries and display the info to the user, we want to fire them in the background, so there is no actual delay for the display of the info. It has the additional benefit that the subscriptions are not re-created each time as well.

(The latter probably only until we get multi support inside the API)

packages/app-staking/src/Account/index.tsx
```diff
@@ -58,18 +59,20 @@ class Account extends React.PureComponent<Props, State> {
...
     }
 
     return (
       <article className='staking--Account'>
-        {this.renderBond()}
-        {this.renderBondExtra()}
-        {this.renderNominating()}
-        {this.renderSessionKey()}
-        {this.renderValidating()}
+        {isBondOpen && this.renderBond()}
```
jacogr: Re-copying the comment here.

There are actually very good reasons why some of these are done this way with the passing through of `isOpen`.

In the cases where we do queries and display the info to the user, we want to fire them in the background, so there is no actual delay for the display of the info. It has the additional benefit that the subscriptions are not re-created each time as well.

(The latter probably only until we get multi support inside the API)

... and all obviously not obvious as to why

Tbaut: Damn I answered on the original comment but it didn't get through.

Thanks for the info, very interesting. Before I revert it all you say "why **some** of these are done this way". Should I keep this "new" behaviour (without passing the isOpen) for some components?

jacogr:
I would honestly just pass through isOpen on all

packages/app-staking/src/Account/index.tsx
Stefie: Hmm, but none of them are actually doing any queries (as in API queries) in the background, or am I missing something?
```
{this.renderBond()} -> src/Account/Bond.tsx
{this.renderBondExtra()} -> src/Account/BondExtra.tsx
{this.renderNominating()} -> src/Account/Nominating.tsx
{this.renderSessionKey()} -> src/Account/SessionKey.tsx
{this.renderValidating()} -> src/Account/Validating.tsx
```
The only component doing an API call in ./Accounts is the `ValidateController.tsx` which is used in `Account/Bond.tsx` and called from within `renderBond()` .

The way it is implemented we're doing an an api all in `ValidateController` for every account listed under the  'Account Actions' tab, even if the user never opens the Bond overlay. That would make one unneeded call for every account as opposed to maybe re-creating a subsription for the same account in case the user is opening the same overlay twice (or more)...

Tbaut:
We're back with the original logic and a comment that is hopefully clear enough.

Stefie:
@jacogr  Did you read my comment above? We'Re not using `withCalls()` in any of the components where i suggested the change.

```
{this.renderBond()} -> src/Account/Bond.tsx
{this.renderBondExtra()} -> src/Account/BondExtra.tsx
{this.renderNominating()} -> src/Account/Nominating.tsx
{this.renderSessionKey()} -> src/Account/SessionKey.tsx
{this.renderValidating()} -> src/Account/Validating.tsx
```

jacogr:
I always prefer consistency over guessing here. Additionally, we are only about 50% through as it stands - we had to get staking updated fast. So for the modals which I can think of offhand - 

Bond - should have the free and bonded balances as well, and only allow access to what is available. Here we also need to retrieve any previously set preferences (not sure but I believe I dropped that at the last moment)

Bond additional - Same in regards to funds. (Although, thankfully for the time being the max is also adjusted on the node)

Session - should additionally check keys for all accounts, so that you don’t assign. 

The 2 bigger ones for validation and nomination are a bit more problematic. Here we need more info on a per account basis, and atm the current approach is not quite optimal for that. 

(Not sure if I missed some atm)

Stefie:
Hmm... still not really convinced by that approach. 
If we'll need the infos you listed for all accounts in the future, we should do the api calls in the index.tsx and just pass the results down as props to the components instead of mounting all possible modal components without displaying them just to do the api calls. That still feels like a unnecessary detour to me...

But that discussion hasn't much to do with @Tbaut's PR anymore (sorry for delaying your merge :))

jacogr:
In all honesty, trying to do too much atm (like changing the modals) is wasted work. All the modals probably need to go the way of the dodo and we need to have a unified experience for all fund management (for instance). 

Part of the reason the above “this is extra stuff” is not listed - as soon as the basics are in place (which is frustrating since even the code needs a LOT of cleanup), we need to see how to migrate to a proper, simple, UX on a per account basis. 

The reality is that as we go through and get the basics in place, we are actually seeing just how massive the complexity problem here is. (And that is even without corner cases). So it is exceptionally valuable from that POV since there it is easy talking without using, and when you actually use it (and other does), you spot where things are just not simple and clear. 

TL;DR staking UX, o, staking UX - I wish I could love you dearly.

Issue #1002, opened 2019-04-12
Create landing page, home, dashboard


jacogr:
Goal:

To allow the user easy access to functionality, an overview of crritical information and an alternative navigation mechanism.

Background:

This steams from https://github.com/polkadot-js/apps/issues/900 - basically (with some limited input), there seems to be appetite for a usable dashboard. Willing to see where it takes us, but really very speculative.

Actions:

- Create dashboard
- Show an icon for each of the available apps, with name
- In addition for each, show to-the-point info, these may include:
  - Explorer: Just the current blocknumber
  - Accounts: Number accounts or, rather, last selected account (we have this as part of InputAddress)
  - Democracy: Number of active referendums
  - Staking: Total number of validators
  - Contracts (substrate-only) Number of active (our) contracts
  - ...

Will need to see how it looks and how to fit it is. Don't just want something for the sake of having something, but really willing to explore the concept

PR #1059, opened 2019-04-30
Add app dashboard (skeleton)

jacogr:
Part of https://github.com/polkadot-js/apps/issues/1002

- Move routing info to `@polkadot/apps-routing` (so the next item doesn't depend on `@polkadot/apps`)
- Add basic `@polkadot/app-dashboard` (hidden for now - once in would need playing and figuring out placement, overlay, etc)

Cutting it right here, although it is not complete and only the skeleton. Prefer to circle back to it (or anybody can pick it up) since rather focus on the higher priority functional items atm.

Immediate items after this:

- Some kind of usable ordering
- Info on a per-app basis (this will sprawl a bit, basically need a dashboard-summary per app that cane be displayed)
- Make it look less blue and actually pleasing to the eye


jacogr:
Yes, the easist way to access the "hidden" apps is via URL as well - so `/#/123code` or `/#/dashboard` are the current non-visible entries. (Same actually goes for other apps as well, i.e. you _can_ access contracts on Alex... won't work, but it will show up)

Tbaut:
oh yeah good to know, didn't realize that.

Issue #1061, opened 2019-04-30
Faulty example


chevdor:
The `Get chain state Info` example throws:
```
TypeError: blockPeriod.toNumber is not a function
    at eval (eval at Playground.runJs (webpack-internal:///../app-js/src/Playground.tsx), <anonymous>:17:54)
```
![image](https://user-images.githubusercontent.com/738724/56985406-8e61a900-6b88-11e9-92f1-0e7e8b338426.png)

chevdor:
This one as well:
![image](https://user-images.githubusercontent.com/738724/56985528-d97bbc00-6b88-11e9-90ea-e4e8c6925384.png)

jacogr:
Dangit. Those were changed with the timePeriod changes. (And is an option now - however it is deprecated. So should be swapped)

PR #1063, opened 2019-05-01
Fix examples

Stefie:
Fixes https://github.com/polkadot-js/apps/issues/1061

### DIFF
--- a/packages/app-js/src/snippets/storage-examples.ts
+++ b/packages/app-js/src/snippets/storage-examples.ts
@@ -17,10 +17,8 @@ const [minimumPeriod, validators, transferFee] = await Promise.all([
   api.query.balances.transferFee()
 ]);
 
-const period = util.formatNumber(minimumPeriod.toNumber() * 2);
-
-console.log('minimumPeriod: ' + period);
-console.log('transferFee: ', transferFee);
+console.log('minimum period between blocks: ' + minimumPeriod);
+console.log('transfer fee: ', transferFee);
 
 if (validators && validators.length > 0) {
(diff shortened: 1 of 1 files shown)


packages/app-js/src/snippets/storage-examples.ts
```diff
@@ -10,13 +10,16 @@ export const storageGetInfo: Snippet = {
...
-  api.query.timestamp.blockPeriod(),
+
+const [minimumPeriod, validators, transferFee] = await Promise.all([
+  api.query.timestamp.minimumPeriod(),
   api.query.session.validators(),
   api.query.balances.transferFee()
 ]);
 
-console.log('blockPeriod in seconds: ' + blockPeriod.toNumber());
+const period = util.formatNumber(minimumPeriod.toNumber() * 2);
```
Tbaut: (nit) Could we add a comment to explain the `* 2`, e.g it's related to aura implementation. This confused me at first :)

jacogr:
The Aura stuff is problematic - we only get that info on a per block basis (where we have access to the engine type, or rather 4 byte magic) at some point the *2 is going to bite. (Not here, elsewhere)

packages/app-js/src/snippets/storage-examples.ts
Stefie: Good point. My example is wrong either way... It should either be minimumPeriod in the console.log or block period ===  minimumPeriod*2 (with aura)

Stefie:
Taking Jacos comment into account I changed the example to just getting the minimum period

Issue #1017, opened 2019-04-13
Settings / Developer page should have link to the documentation of the type definition syntax 


xlc:
Assuming there is a type definition syntax page, it should be linked in the developer page so people can read it there.

If there isn't one, we need one.

jacogr:
100%, good point - I actually missed this case in my “male help available everywhere” issue.

jacogr:
Extend, properly link, make into HelpOverlay (?) - https://docs.substrate.dev/docs/substrate-runtime-recipes#section-polkadot-ui

jacogr:
Closed in https://github.com/polkadot-js/apps/pull/1068

PR #1068, opened 2019-05-02
Markdown for the settings app

jacogr:
Should round out https://github.com/polkadot-js/apps/issues/1017


jacogr:
cc @xlc: anything specifically that I missed in the actual types that jumps to mind?

PR #1069, opened 2019-05-02
feat: Learning app

ltfschoen:
This opportunity was inspired by a suggestion from @gautamdhameja.
The idea was to create a learning app, that would focus on allowing anyone to undertake learning and interact using just their mobile device.
There'd be different Tabs including 'tracks', 'courses', and 'units'.
The 'tracks' page would list learning pathways available (i.e. validating, staking, etc), which may comprise of entire 'courses' and/or individual 'units'.
The 'courses' page would list courses
 The 'units' page would list units (i.e. with different learning approaches such as multi-choice, multiple-answers, app-js integration, multiple terminal integration, chat integration, etc).

Currently it just a basic 'tracks' page and hard-coded data, but it could be something interesting for someone else to pick up.

It's on-ice, as it's not a priority, unless otherwise advised


jacogr:
This is really cool. 

However, it is not something I'm interested into bringing into apps - keeping this current (as seen in the past with the short-lived examples), is just way too much of a drain on resources. It is the first thing to drop when the going gets tough.

TL;DR - Really cool, needed, but not in apps.

PR #1070, opened 2019-05-02
Allow to withdrawUnbonded + additional info

### LINKED ISSUE #859: User should be able to unbond & withdraw funds

Tbaut:
- closes #859

image

- [x] display the amount available to transfer (e.g what is not locked)
- [x] display the amount that is being unlocked (1 per line)
- [x] display the total amount "unlockable" with a `withdrawUnbonded` button when possible


jacogr:
@xlc's API is pulled into apps master (I'm guessing you have a dep here)

Tbaut:
Thanks, no more console.error :hear_no_evil:

Tbaut:
What definitely still needs tweaking:
- the design and colors for the `balance`,`available`, `bonded`, `unlockable`, `locked`
- the names (`unlockable` is probably not the best)

packages/ui-reactive/src/Unlocking.tsx
```diff
@@ -0,0 +1,121 @@
...
+    paramName: 'params',
+    propName: 'controllerId',
+    transform: (value) =>
+      value.unwrapOr(null)
+  }],
+  ['query.staking.ledger', {
+    paramName: 'controllerId',
+    propName: 'unlockings',
+    transform: (ledger: Option<StakingLedger>) =>
+    ledger.unwrapOr({ unlocking: null }).unlocking
```
jacogr:
```js
['query.staking.ledger', {
    paramName: 'controllerId',
    propName: 'unlockings',
    transform: (ledger: Option<StakingLedger>) =>
      ledger.unwrapOr({ unlocking: null }).unlocking
```

packages/ui-reactive/src/Available.tsx
```diff
@@ -0,0 +1,57 @@
...
+export class AvailableDisplay extends React.PureComponent<Props> {
+
+  render () {
+    const { balances_freeBalance, balances_locks, chain_bestNumber, children, className, label = '', style } = this.props;
+    let maxLock = new BN(0);
+
+    if (Array.isArray(balances_locks)) {
+      // only get the locks that are valid until passed the current block
+      const totals = balances_locks.filter((value, index) => chain_bestNumber && value.until.gt(chain_bestNumber));
+      // get the maximum of the locks according to https://github.com/paritytech/substrate/blob/master/srml/balances/src/lib.rs#L699
```
Tbaut: A probably better link to explain the need to use `max` here: https://crates.parity.io/srml_balances/index.html#terminology
> Multiple locks always operate over the same funds, so they "overlay" rather than "stack".

jacogr:
It works exactly like it says on the tin. We certainly have an overdose of balances now, so that needs a re-visit. (But I believe out of the scope here - we knew it would be an issue, get it up and then see how to best display everything into a smaller space)

packages/ui-reactive/src/Available.tsx
```diff
@@ -0,0 +1,57 @@
...
+      maxLock = max(totals.map(function (o) { return o.amount; }));
+    }
+
+    const available = maxLock.eq(new BlockNumber('0xffffffffffffffffffffffffffffffff')) ? new Balance(0) : balances_freeBalance && balances_freeBalance.sub(maxLock) ;
```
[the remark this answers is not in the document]
Tbaut: From my tests `reservedBalance` is included in the `freeBalance` so that we don't need to read it explicitly. What needs to be taken into account however is `vesting`, the problem is that it can't be tested.

packages/ui-app/src/util/max.ts
```diff
@@ -0,0 +1,11 @@
+// Copyright 2017-2019 @polkadot/ui-app authors & contributors
```
jacogr: @kwingram25 Added these to `@polkadot/util` here https://github.com/polkadot-js/common/pull/401 - however it is not pulled into apps master anyway, just something to keep in mind when/if we start using his derive functions as they are available

PR #1076, opened 2019-05-03
Add maximum balance button to app-transfer and app-staking

### LINKED ISSUE #1043: Be able to send the max amount
I'd like to be able to send the maximum amount of Units possible considering the fees.
Not only useful for transfer but also for bonding / unbounding
This could take the form of a max button that could look like the following.

image

kwingram25:
Closes #1043 

- Adds MAX button to auto-fill maximum available balance when using Transfer and Staking app
- Adds checks for too-high values to InputNumber if maximum is specified

### DIFF
--- a/packages/app-staking/src/Account/Bond.tsx
+++ b/packages/app-staking/src/Account/Bond.tsx
@@ -76,7 +76,7 @@ class Bond extends React.PureComponent<Props, State> {
 
   render () {
     const { accountId, isOpen, onClose, t } = this.props;
-    const { bondValue, controllerError, controllerId, destination, extrinsic, maxBalance } = this.state;
+    const { bondValue, controllerError, controllerId, extrinsic, maxBalance } = this.state;
     const hasValue = !!bondValue && bondValue.gtn(0) && (!maxBalance || bondValue.lt(maxBalance));
     const canSubmit = hasValue && !controllerError && !!controllerId;
 
@@ -171,13 +171,14 @@ class Bond extends React.PureComponent<Props, State> {
   private nextState (newState: Partial<State>): void {
     this.setState((prevState: State): State => {
       const { api } = this.props;
-      const { bondValue = prevState.bondValue, controllerId = prevState.controllerId, destination = prevState.destination, maxBalance = prevState.maxBalance } = newState;
+      const { bondValue = prevState.bondValue, controllerError = prevState.controllerError, controllerId = prevState.controllerId, destination = prevState.destination, maxBalance = prevState.maxBalance } = newState;
       const extrinsic = (bondValue && controllerId)
         ? api.tx.staking.bond(controllerId, bondValue, destination)
         : null;
 
       return {
         bondValue,
+        controllerError,
         controllerId,
         destination,
         extrinsic,
--- a/packages/app-staking/src/Account/Bond.tsx
+++ b/packages/app-staking/src/Account/Bond.tsx
@@ -187,11 +187,11 @@ class Bond extends React.PureComponent<Props, State> {
   }
 
   private setMaxBalance = () => {
-    const { api, system_accountNonce = ZERO, balances_fees = ZERO_FEES, balances_votingBalance = ZERO_BALANCE } = this.props;
+    const { api, system_accountNonce = ZERO, balances_fees = ZERO_FEES, balances_all = ZERO_BALANCE } = this.props;
     const { controllerId, destination } = this.state;
 
     const { transactionBaseFee, transactionByteFee } = balances_fees;
-    const { freeBalance } = balances_votingBalance;
+    const { freeBalance } = balances_all;
 
     let prevMax = new BN(0);
     let maxBalance = new BN(1);
@@ -245,7 +245,7 @@ export default withMulti(
   withApi,
   withCalls<Props>(
     'derive.balances.fees',
-    ['derive.balances.votingBalance', { paramName: 'accountId' }],
+    ['derive.balances.all', { paramName: 'accountId' }],
     ['query.system.accountNonce', { paramName: 'accountId' }]
   )
 );
--- a/packages/app-staking/src/Account/Bond.tsx
+++ b/packages/app-staking/src/Account/Bond.tsx
@@ -192,8 +192,6 @@ class Bond extends React.PureComponent<Props, State> {
 
     const { transactionBaseFee, transactionByteFee } = balances_fees;
     const { freeBalance } = balances_votingBalance;
-    // api.derive.balances.votingBalance(accountId!, ({ freeBalance: senderBalance }) => {
-      // api.derive.balances.votingBalance(recipientId!, ({ freeBalance: recipientBalance }) => {
 
     let prevMax = new BN(0);
     let maxBalance = new BN(1);
@@ -222,8 +220,6 @@ class Bond extends React.PureComponent<Props, State> {
       extrinsic,
       maxBalance
     });
-      // });
-    // });
   }
 
   private onChangeController = (controllerId: string) => {
--- a/packages/app-staking/src/Account/Bond.tsx
+++ b/packages/app-staking/src/Account/Bond.tsx
@@ -8,7 +8,6 @@ import { CalculateBalanceProps } from '../types';
 
 import BN from 'bn.js';
 import React from 'react';
-import { AccountId } from '@polkadot/types';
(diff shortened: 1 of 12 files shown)


jacogr:
Playing now :)

packages/app-staking/src/Account/Bond.tsx
```diff
@@ -136,17 +164,84 @@ class Bond extends React.PureComponent<Props, State> {
...
+    });
+  }
+
+  private setMaxBalance = () => {
+    const { api, system_accountNonce = ZERO, balances_fees = ZERO_FEES, balances_votingBalance = ZERO_BALANCE } = this.props;
+    const { controllerId, destination } = this.state;
+
+    const { transactionBaseFee, transactionByteFee } = balances_fees;
+    const { freeBalance } = balances_votingBalance;
+    // api.derive.balances.votingBalance(accountId!, ({ freeBalance: senderBalance }) => {
```
jacogr: Stray comments

packages/app-staking/src/Account/Bond.tsx
```diff
@@ -136,17 +164,84 @@ class Bond extends React.PureComponent<Props, State> {
...
+        .add(transactionByteFee.muln(txLength));
+
+      maxBalance = new BN(freeBalance).sub(fees);
+    }
+
+    this.nextState({
+      extrinsic,
+      maxBalance
+    });
+      // });
```
jacogr: As above (remove)

packages/app-staking/src/Account/BondExtra.tsx
```diff
@@ -83,16 +109,79 @@ class BondExtra extends React.PureComponent<Props, State> {
...
+      return {
+        maxAdditional,
+        extrinsic,
+        maxBalance
+      };
+    });
+  }
+
+  private setMaxBalance = () => {
+    const { api, system_accountNonce = ZERO, balances_fees = ZERO_FEES, balances_votingBalance = ZERO_BALANCE } = this.props;
```
jacogr: This one is problematic - votingBalance includes the already stashed balance. so basically is is votingBalance - bondedBalance that can be used.

jacogr: ... while problematic, the node will adjust for the max funds, however basically what we display won't be that value.

packages/app-staking/src/Account/Unbond.tsx
```diff
@@ -76,23 +90,58 @@ class Unbond extends React.PureComponent<Props, State> {
...
+
+      return {
+        maxUnbond,
+        maxBalance
+      };
+    });
+  }
+
+  private setMaxBalance = () => {
+    console.log('setMaxBalance2');
```
jacogr: Stray log.

packages/app-transfer/src/Transfer.tsx
```diff
@@ -117,11 +142,62 @@ class Transfer extends React.PureComponent<Props, State> {
...
+            const txLength = SIGNATURE_SIZE + accountNonce.length + (
+              extrinsic
+                ? extrinsic.encodedLength
+                : 0
+            );
+
+            const fees = transactionBaseFee
+              .add(transactionByteFee.muln(txLength))
+              .add(transferFee)
+              .add(senderBalance.isZero() ? creationFee : ZERO)
```
Tbaut: If the `senderBalance.isZero()` then there's nothing that can be sent I guess or did I miss something?

packages/app-transfer/src/Transfer.tsx
```diff
@@ -117,11 +142,62 @@ class Transfer extends React.PureComponent<Props, State> {
...
+              .add(recipientBalance.isZero() ? creationFee : ZERO);
+
```
Tbaut: We need to take into account the locked balance (available from balances.locks). In my tests so far, the max of a stash account fully bonded was wrong because the account is actually unable to transfer any fund. I've done something similar (need review) here: https://github.com/polkadot-js/apps/blob/da179c7550f7b210fa3610ff37a8375f72896900/packages/ui-reactive/src/Available.tsx

Tbaut:
After looking deeper into it, the other forms of "locked funds" (other than what is in `balances.locks`) could be:
- `reserved`: some funds get reserved when proposing/seconding in democracy or submitting candidacy in council module. These funds are directly deduced from `free_balance` (but can additionally be viewed in `balances.reservedBalance`).
- `vested`: we need to cater for it and check `balances.vesting` to calculate how much has been vested and is therefore available for transfer.

packages/app-staking/src/Account/BondExtra.tsx
```diff
@@ -3,30 +3,59 @@
...
 
 import { I18nProps } from '@polkadot/ui-app/types';
+import { ApiProps } from '@polkadot/ui-api/types';
+import { CalculateBalanceProps } from '../types';
 
 import BN from 'bn.js';
 import React from 'react';
 import { Button, InputAddress, InputBalance, Modal, TxButton } from '@polkadot/ui-app';
+import { Option, StakingLedger } from '@polkadot/types';
+import { IExtrinsic } from '@polkadot/types/types';
```
jacogr: (nit) As per previous type import comment.

packages/app-transfer/src/Transfer.tsx
```diff
@@ -138,11 +163,62 @@ class Transfer extends React.PureComponent<Props, State> {
...
+    if (!accountId || !recipientId) {
+      return;
+    }
+
+    void api.query.system.accountNonce(accountId, (accountNonce) => {
+
+      const { transferFee, transactionBaseFee, transactionByteFee, creationFee } = balances_fees;
+
+      void api.derive.balances.all(accountId, ({ availableBalance: senderBalance }) => {
```
jacogr: Not sure about the `void api.derive.balances`? Why the `void`?

packages/app-staking/src/Account/Bond.tsx
```diff
@@ -3,16 +3,23 @@
...
+import { CalculateBalanceProps } from '../types';
 
 import BN from 'bn.js';
 import React from 'react';
 import { AccountId } from '@polkadot/types';
+import { IExtrinsic } from '@polkadot/types/types';
 import { Button, InputAddress, InputBalance, Modal, TxButton, Dropdown } from '@polkadot/ui-app';
+import { withCalls, withApi, withMulti } from '@polkadot/ui-api';
+import { compactToU8a } from '@polkadot/util';
+import { SIGNATURE_SIZE } from '@polkadot/ui-signer/Checks';
```
jacogr: This is dangerous since it assumes that we are always signing with the publicKey - rather expose a `calcSignatureLength` that returns the value. (That we we don't need to find it all-over when we extend and cater for smaller transactions)

packages/app-transfer/src/Transfer.tsx
kwingram25: Linting error without it :(

kwingram25:
@Tbaut not sure what I'm missing from Bond.tsx but it's consistently too high by 1 Unit:

```
      const fees = transactionBaseFee
        .add(transactionByteFee.muln(txLength));

      maxBalance = new BN(freeBalance).sub(fees);
```

packages/app-staking/src/Account/Bond.tsx
```diff
@@ -139,21 +166,91 @@ class Bond extends React.PureComponent<Props, State> {
...
+      extrinsic = controllerId && destination
+        ? api.tx.staking.bond(controllerId, prevMax, destination)
+        : null;
+
+      const txLength = calcSignatureLength(extrinsic, system_accountNonce);
+
+      const fees = transactionBaseFee
+        .add(transactionByteFee.muln(txLength));
+
+      console.log(fees.toString());
```
Tbaut: Do we really want to keep those logs?

packages/app-staking/src/Account/Bond.tsx
```diff
@@ -139,21 +166,91 @@ class Bond extends React.PureComponent<Props, State> {
...
+    const { controllerId, destination } = this.state;
+
+    const { transactionBaseFee, transactionByteFee } = balances_fees;
+    const { freeBalance } = balances_all;
+
+    let prevMax = new BN(0);
+    let maxBalance = new BN(1);
+    let extrinsic;
+
+    console.log(Object.entries(balances_all)
```
Tbaut: Talking about this one as well.

Tbaut:
@kwingram25 the bonding problem I had is gone :+1:

Issue #1060, opened 2019-04-30
Extra padding for non-header apps


jacogr:
Goal:

Consistency.

Background:

The spacing looks off where there are apps without a header. The content is too close to the actual tabs. Compare an app with header to one without:

<img width="651" alt="Polkadot:Substrate Portal 2019-04-30 18-30-09" src="https://user-images.githubusercontent.com/1424473/56977725-273af900-6b76-11e9-8215-1c571840ecaa.png">
<img width="648" alt="Polkadot:Substrate Portal 2019-04-30 18-29-28" src="https://user-images.githubusercontent.com/1424473/56977726-27d38f80-6b76-11e9-93dd-a17bcbe20f1f.png">

Actions:

- Adjust for consistent spacing, with or without

ltfschoen:
@jacogr ok to close this?

PR #1077, opened 2019-05-03
refactor: Relates to #1060. Consistent top padding across apps

ltfschoen:
Removes min-height and padding-top from CardSummary for consistent vertical spacing between the tab menu items and the contents below across different apps, and some responsive tweaks so not too much padding on mobile devices.

### DIFF
--- a/packages/apps/src/Apps.tsx
+++ b/packages/apps/src/Apps.tsx
@@ -38,7 +38,7 @@ const Wrapper = styled.div`
     text-align: center;
 
     ${media.TABLET`
-      margin-bottom: 2.8rem;
+      margin-bottom: 2rem;
    `}
   }
 `;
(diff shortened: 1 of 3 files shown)


packages/apps/src/Apps.tsx
```diff
@@ -34,11 +34,11 @@ const Wrapper = styled.div`
   min-height: 100vh;
 
   header {
-    margin-bottom: 0.8rem;
+    margin-bottom: 1.4rem;
     text-align: center;
 
     ${media.TABLET`
-      margin-bottom: 1.1rem;
+      margin-bottom: 2.8rem;
```
jacogr: This "feels" a bit too much, `2.25rem` or even, `2rem` "feels" much bit better. (Specifically in the case where we have no summary on top, the spacing is too much - 2 seems to give a happy medium)

Apart from that, lgtm

ltfschoen: i've pushed a commit changing it to 2rem

Issue #1083, opened 2019-05-03
Clicking on "yay" actually makes you vote "nay"


Tbaut:
on Substrate, didn't try on Alexander you can see in the modal and on the graph:
![](https://i.imgur.com/LSFrmkS.gif)

jacogr:
Sheiße, introduced here and I must have been drunk when I checked -  https://github.com/polkadot-js/apps/pull/1023/files#diff-9ea1956c2d649bb9a8d2369b24f39afaR31

The good news is that it is a simple fix, swap the 0 and -1

PR #1084, opened 2019-05-03
Swap yay/nay to be correct

jacogr:
Closes https://github.com/polkadot-js/apps/issues/1083


jacogr:
It in unintuitive to say the least - however, my fault entirely since I believe there are 3 people who know about the high-bit thing (since it looks/feels/sounds weird) and one of them checked this, knew about it and somehow missed it.

PR #1086, opened 2019-05-05
Add total issuance summary card

### LINKED ISSUE #1058: Show total insurance
Goal:

Make the explorer more informative.

Actions:

As a summary card, display the `query.balances.totalIssuance` on the explorer overview

kwingram25:
Closes #1058 

Open to feedback on placement & visibility, it is somewhat ugly at the moment


jacogr:
Alternatively, can calc SI ourselves, set withSi false and just concat the unit? https://github.com/polkadot-js/common/blob/master/packages/util/src/format/formatBalance.ts#L46 (More-or-less what is done in the number inputs)

PR #1098, opened 2019-05-07
Move Chart to styled-components

jacogr:
Some movement in https://github.com/polkadot-js/apps/issues/674 (inspired by https://github.com/polkadot-js/apps/pull/1097)

### DIFF
--- a/packages/ui-app/src/Chart/Base.tsx
+++ b/packages/ui-app/src/Chart/Base.tsx
@@ -1,0 +1,37 @@
+// Copyright 2017-2019 @polkadot/ui-app authors & contributors
+// This software may be modified and distributed under the terms
+// of the Apache-2.0 license. See the LICENSE file for details.
+
+import { BareProps } from '../types';
+
+import React from 'react';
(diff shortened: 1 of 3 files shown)


packages/ui-app/src/Chart/Base.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+const Wrapper = styled.div`
+  position: relative;
+  display: inline-block;
+  padding: 1em 1em 0;
+  height: 15vw;
+  width: 15vw;
+`;
+
+export default class BaseChart extends React.PureComponent<Props> {
```
jacogr: Keeping the base wrapper separate (if we add multiple chart types, as was the intent)

jacogr:
Grrr... it should be `Wrapper` since it is a component.

packages/ui-app/src/Chart/Base.tsx
jacogr: `Wrapper` :) capitals...

PR #1099, opened 2019-05-07
Convert some apps to styled components

jacogr:
Just a very quick non-comprehensive run-through as part of https://github.com/polkadot-js/apps/issues/674

- 123-code
- Transfer
- Settings
- Extrinsics (removed unused)
- Toolbox (one item left)


jacogr:
Yes, also did the same - thanks for that, looked at each one, but couldn't spot anything out of the ordinary, but was still worried.

Issue #1004, opened 2019-04-12
Toolbox sign layout is slightly broken


jacogr:
Goal:

Apply the same level of care to all screens.

Background:

Not much to be said, apart from

![image](https://user-images.githubusercontent.com/1424473/56058272-3aba2780-5d61-11e9-9fdf-4ff12de5bd09.png)

Let your eye be drawn to the right of the screen and the boxes flowing outside. The same applies to

![image](https://user-images.githubusercontent.com/1424473/56058336-676e3f00-5d61-11e9-8627-2ee0e83e2e48.png)

Actions:

- Fix it :)

PR #1106, opened 2019-05-08
Adjust toolbox layout (.small is problematic)

jacogr:
Closes https://github.com/polkadot-js/apps/issues/1004

- Remove .small, replace by .medium & split into rows
- Initially tried to adjust the min-width of the content holder, but then it breaks the text on resize - the label just takes up too much space

Polkadot:Substrate Portal 2019-05-08 12-26-27


jacogr:
Please log - honestly the toolbox is one of the original pieces of code and it could use a helluva lot of updates :( (I'm not even sure it does sr25519)

PR #1111, opened 2019-05-08
[#943] Make hash/number query available globally to explorer

### LINKED ISSUE #943: Make hash/number query available globally to explorer
Goal:

We want to move to consistent queries/filtering of information across all apps. So when a user wants to query, we want the ability globally available on the app.

Background:

Currently the block query is fully split into a seperate tab. So when the user wants to query, it takes 2 clicks - first to the query tab, the typing the value and submitting for refresh.

Pre-requisites:

Requires https://github.com/polkadot-js/apps/issues/942 to be done first (so the block query is not empty by default - it only contains the query bar atm when entering)

Actions:

- Move the query from the block query to the nav bar - ensure it is visible on all tabs. (And no overlaps with the tabs)
- Rename block query to block info
- Make sure the queries still work, for both hashes and numbers

When done, the layout should look something along the lines of -

https://github.com/polkadot-js/apps/issues/942

e18r:
Fixes #943 

* Relocated the `Query` component from `BlockQuery` in the `app-explorer` package to `Tabs` in the `@ui-app` package.
* Renamed the `BlockQuery` component to `BlockInfo`.
* Redesigned the `Tabs` component so that it incorporates this new element without losing generality for use in other contexts. Added some styling elements.
* No need to lift the state up to the partent component as originally suggested, because the content of the query is not shared until it is submitted.


jacogr:
Thanks a million. I will look first thing in the morning (EU time zone). The description sounds spot-on, will review and play through.

packages/ui-app/src/styles/semantic.css
```diff
@@ -190,6 +190,10 @@
   }
 }
 
+.ui--query {
```
jacogr: Please move this specifically to the explorer (This area is reserved for specific semantic-ui overrides). Additionally, would prefer to stick to `rem` where possible.

packages/ui-app/src/Tabs.tsx
```diff
@@ -35,6 +37,9 @@ export default class Tabs extends React.PureComponent<Props> {
             .filter(({ name }) => !hidden.includes(name))
             .map(this.renderItem)
         }
+        { query && (
```
jacogr: (nit) Extra unneeded space before `query`

packages/ui-app/src/Tabs.tsx
```diff
@@ -18,12 +19,13 @@ export type TabItem = {
...
   hidden?: Array<string>,
-  items: Array<TabItem>
+  items: Array<TabItem>,
+  query?: boolean
 };
 
 export default class Tabs extends React.PureComponent<Props> {
   render () {
-    const { className, hidden = [], items, style } = this.props;
+    const { className, hidden = [], items, style, query } = this.props;
```
jacogr: (nit) Alphabetical ordering

packages/app-explorer/src/index.tsx
```diff
@@ -56,11 +56,12 @@ class ExplorerApp extends React.Component<Props, State> {
           <Tabs
             basePath={basePath}
             items={items}
+            query={true}
```
jacogr: If this is always true, do we want/need to pass it in? (With this actually we can forgo showing it on the NodeInfo page, which is probably the better experience)

e18r:
@jacogr I'll address your comments now.

packages/app-explorer/src/index.tsx
e18r: We need it because this component is also used by the other apps to show their tabs. I can try to hide the Query component when the user clicks on NodeInfo, but, precisely given its generality, the result might not be super clean.

jacogr: That is fine, you are right, probably more trouble than it is worth.

packages/ui-app/src/styles/semantic.css
e18r: OK. I don't have so much experience with CSS namespacing and overriding, so please don't hesitate to comment on the result if it's not exactly what you expect. That way I get to learn.

jacogr: No issues, just move it to `app-explorer/index.css` (And adjust the naming as well to just align with what is in there)

e18r:
@jacogr I did all the changes you requested. As said in the comment, hiding the Query component in the NodeInfo tab ended up being a little dirty. Please check it out, as well as the styling, and let me know if anything isn't quite right.

jacogr:
Now for the first time it is really irritating now having query on enter on the field. It disturbs workflow. (Small enhancement after this)

e18r:
@jacogr I'm not sure I understood your last comment. Please let me know if there's anything left to do. Otherwise please don't forget to mark the issue as done in Gitcoin.

jacogr:
So the last comment - if you actually is the functionality now, you realize that the fact that there is no submission on enter is an serious (pre-existing) usability issue, which now actually creeps in. 

So would need to be addressed after this. (As a small enhancement)

Gitcoin workflow will fall in place as soon as merged, which should be tomorrow if nobody else has comments on the PR. 

(Always prefer more eyes on all PRs - so generally don’t merge immediately to give people time to look)

jacogr:
@Tbaut Will include it in #1119

PR #1123, opened 2019-05-10
Rework inherent submission (allow inherent submission via extrinsics app)

jacogr:
cc @thiolliere

Enabled again.

image

image


jacogr:
Yes, basically... but not really. Inherents are unsigned, but typically added by the block producers, i.e. like the timestamp extrinsic in all blocks. (Or parachains setHeads for polkadot) 

I think the naming is off, but "Submit unsigned" and "submit signed transaction" gets unwieldly :( So not sure wtf.

Issue #1119, opened 2019-05-09
Add explorer query on enter


jacogr:
In the block hash/number query input box, allow enter to start the navigation. (As some only by the button atm). Keep the button in-place as is, just enhance the input with Enter. 

As per comments in #1111 - it aids the actual usability and allows quick navigation without reaching for the mouse again.

PR #1127, opened 2019-05-10
Explorer query updates

jacogr:
Closes https://github.com/polkadot-js/apps/issues/1119

- Allow onEnter to submit
- Incorporate @Tbaut disabled suggestion
- Move Query back to explorer (it is only valid there)
- Adjust layout (internal button, let SUI do the heavy lifting)


jacogr:
So the rounded thing (I believe) is a bug in the CSS somewhere. When you look at SUI + button attached, they look correct normally - https://react.semantic-ui.com/elements/input/

For the spacing, the FilterOverlay takes the global (?) placement into account. Which is not here yet - was meant to add the slide-in help for explorer today, but ran out of time.

Issue #1120, opened 2019-05-10
JS app needs some layout improvement


xlc:
![Screen Shot 2019-05-10 at 1 54 01 PM](https://user-images.githubusercontent.com/1201310/57497345-66620a80-732b-11e9-8d6d-fcec37e69ac8.png)

The buttons are blocking me

jacogr:
Suggest moving the buttons out of the content areas. (We lose some vertical space, but in most cases the scripts are on the smaller side.)

Stefie:
@xlc I can have a quick look

xlc:
Fixed by #1128

PR #1128, opened 2019-05-10
Move buttons above the editor

Stefie:
Addresses https://github.com/polkadot-js/apps/issues/1120

Screenshot 2019-05-10 at 13 46 08

### DIFF
--- a/packages/app-js/src/index.css
+++ b/packages/app-js/src/index.css
@@ -21,6 +21,12 @@
   }
 }
 
+.js--Buttons {
+  display: flex;
+  justify-content: space-between;
+  margin-bottom: 0;
+}
+
 .js--Content {
   align-content: stretch;
   align-items: stretch;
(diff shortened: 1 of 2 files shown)


packages/app-js/src/index.css
```diff
@@ -21,6 +21,12 @@
   }
 }
 
+.js--Buttons {
+  display: flex;
+  justify-content: space-between;
+  margin-bottom: 0;
```
jacogr: Suggest we align there right, it looks sightly better. (Not hung-up either way about it)

Stefie: I agree, but that doesn't work with the CSS resize solution.
If the buttons are placed inside the resizable container, they can't be positioned absolute outside the container because of the overflow setting. They just disappear.

Stefie: What I could do is just add a padding-top inside the editor...

Stefie: Screenshot 2019-05-10 at 17 53 22

jacogr: Like the second one, as weird as it sounds...

Issue #1074, opened 2019-05-02
Users should not be allowed to select the same account for stash and controller


Tbaut:
As discussed with some of you, having `controller` === `stash` when bonding funds is not only dangerous but also annoying when testing and not paying attention to the warning.

We should prevent users from selecting the same account for `stash` and `controller` in the bonding modal.

jacogr:
The whole idea behind only warning is that we just want to provide the info, not play dad. However, since staking is nowhere close to simple, not against the idea of playing cop and disallowing these completely - the 99.9% won't be affected and the 0.1% can still do it elsewhere.

Related to https://github.com/polkadot-js/apps/issues/926 - these are some other really problematic cases. (Mostly as found by @kwingram25 as he was going through)

jacogr:
The warnings were added here - https://github.com/polkadot-js/apps/pull/963

chevdor:
The 'bad' choice is the dangerous one unfortunately. 

By default, bonding on Account X will propose X as controller.

If you make it quick and confirm, your account will be locked: once your stash is bonded, your controller (the same account) will no longer be able to 'control' anything any more since no more tx will be possible.

Stefie:
We have the same `<InputAddress>` component behaviour when setting a session key.

Maybe we should add a warning? error there as well?

<img width="710" alt="Screenshot 2019-05-10 at 15 07 42" src="https://user-images.githubusercontent.com/125398/57529536-74b52400-7335-11e9-82ae-25aae9b7a767.png">

PR #1130, opened 2019-05-10
Turn bond controller selection into non-submit errors

jacogr:
Closes https://github.com/polkadot-js/apps/issues/1074

### DIFF
--- a/packages/app-staking/src/Account/Bond.tsx
+++ b/packages/app-staking/src/Account/Bond.tsx
@@ -51,7 +51,7 @@ class Bond extends React.PureComponent<Props, State> {
     const { accountId, isOpen, onClose, t } = this.props;
     const { bondValue, controllerError, controllerId, destination } = this.state;
     const hasValue = !!bondValue && bondValue.gtn(0);
-    const canSubmit = hasValue && !!controllerId && !controllerError;
+    const canSubmit = hasValue && !controllerError;
 
     if (!isOpen) {
       return null;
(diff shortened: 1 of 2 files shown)


packages/app-staking/src/Account/ValidateController.tsx
```diff
@@ -5,52 +5,55 @@
...
+      } else if (bondedId) {
+        return t('A controller account should not map to another stash. This selected controller is a stash, controlled by {{bondedId}}', { replace: { bondedId } });
+      } else if (stashId) {
+        return t('A controller account should not be set to manages multiple stashes. The selected controller is already controlling {{stashId}}', { replace: { stashId } });
+      }
+
+      return null;
+    })();
+
+    if (prevState.error !== error) {
```
jacogr: I am no sure this is not an anti-pattern - i.e. determining the stats value and passing it through to the parent (and not the first time doing this in apps either... Other ways?)

packages/app-staking/src/Account/Bond.tsx
```diff
@@ -40,16 +41,17 @@ class Bond extends React.PureComponent<Props, State> {
...
     };
   }
 
   render () {
     const { accountId, isOpen, onClose, t } = this.props;
-    const { bondValue, controllerId, destination } = this.state;
+    const { bondValue, controllerError, controllerId, destination } = this.state;
     const hasValue = !!bondValue && bondValue.gtn(0);
-    const canSubmit = hasValue && !!controllerId;
+    const canSubmit = hasValue && !!controllerId && !controllerError;
```
Stefie: `controllerId` is actually always true once all components are mounted, so there's no real need for the check.

Issue #1126, opened 2019-05-10
False negative to verify signatures for sr25519 accounts


Tbaut:
Either I haven't understood the concept, or something's not right here:

in `/#/toolbox/sign`
- using `Alice`
- data: `bla`
- signature: `0x789751a608ad71...`

Now delete and type `bla` again --> the signature is now totally different.
I would expect to have the same signature for the same message.

Verifying the signature tells it's not signed from `Alice` either.

edit:
the different signature for the same message is expected but the verify signature should work.

jacogr:
Expected. sr25519 signatures are non-deterministic.

jacogr:
Yeap, bug - it always used ed25519 to verify

https://github.com/polkadot-js/apps/blob/master/packages/app-toolbox/src/Verify.tsx#L140

For those, it should just use the keypair verify function (like it does on sign).

Tbaut:
Thanks modifying the title and overall message as the signing isn't the problem.

jacogr:
Ahhh, dangit. So, this is the issue -

- for built-in accounts, it can use the keypair verify 
- for addresses, it cannot do that.

So what I'm suggesting here -

- verify via sr25519 (which we don't do), if fine, show ok and do nothing more
- verify via ed25519, if fine, show ok
- show error if none of the above 2 passes

PR #1131, opened 2019-05-10
Support sr25519 signatures (toolbox verify)

jacogr:
Closes https://github.com/polkadot-js/apps/issues/1126

### DIFF
--- a/packages/app-toolbox/src/Verify.tsx
+++ b/packages/app-toolbox/src/Verify.tsx
@@ -62,11 +62,42 @@ class Verify extends React.PureComponent<Props, State> {
   }
 
   render () {
+    const { t } = this.props;
+    const { cryptoOptions, cryptoType, data, isHexData } = this.state;
+
     return (
       <div className='toolbox--Verify'>
-        {this.renderInput()}
         {this.renderAddress()}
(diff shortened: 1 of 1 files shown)


packages/app-toolbox/src/Verify.tsx
```diff
@@ -48,51 +62,34 @@ class Verify extends React.PureComponent<Props, State> {
...
     return (
-      <>
+      <div className='toolbox--Verify'>
+        {this.renderAddress()}
         <div className='ui--row'>
           <Input
             autoFocus
             className='full'
-            label={t('using the following data (hex or string)')}
+            help={t('The data that was signed. This is used in combination with the signature for the verification. It can either be hex or s string.')}
```
Tbaut: typo 
```suggestion
            help={t('The data that was signed. This is used in combination with the signature for the verification. It can either be hex or a string.')}
```

jacogr:
Since we do cater for the need-to-know-it-all users, I'd keep it. (Not against the idea to hid the info in basic operation mode though - which needs to be become the default as some point relatively soon.) As for the input, tweaks to the CSS is welcome, but basically the read-only is like this everywhere. (light gray-look-ish maybe?)

PR #1132, opened 2019-05-10
Only allow ed25519 session keys in modal

jacogr:
Check sessionkey type before allowing it to be set

### DIFF
--- a/packages/app-staking/src/Account/ValidateSession.tsx
+++ b/packages/app-staking/src/Account/ValidateSession.tsx
@@ -13,7 +13,8 @@ import translate from '../translate';
 type Props = I18nProps & {
   controllerId: string,
   onError: (error: string | null) => void,
-  sessionId: string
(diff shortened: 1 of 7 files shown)


packages/app-staking/src/Account/ValidateSession.tsx
```diff
@@ -0,0 +1,66 @@
...
+    try {
+      const pair = keyring.getPair(sessionId);
+
+      error = pair.type !== 'ed25519'
+        ? t('The selected account is not an ed25519 (Edwards) account as required for validation')
+        : null;
+    } catch (e) {
+      // this _should_ never happen...
+      error = t('The account {{sessionId}} is not a valid account', { replace: { sessionId } });
+    }
```
Stefie: Maybe add another warning here if a stash account/ the stash account of this triage has been selected?

Issue #1134, opened 2019-05-10
Error when opening `nominate` modal


Stefie:
How to reproduce:
- Create a stash account and a controller
- Click the `Nominate` button

<img width="1662" alt="Screenshot 2019-05-10 at 15 18 58" src="https://user-images.githubusercontent.com/125398/57530300-3caee080-7337-11e9-8583-0106759ef636.png">

jacogr:
Cannot replicate, also on dev - https://youtu.be/WCpMfixrskM

Stefie:
Could you reload the page and try again?

jacogr:
Remember that when you just start the local dev, the fist load of `yarn start` is actually and old cached version for some reason.... so could very well be a local issue. (Not saying there is no issue, but have been refreshing like crazy...)

Stefie:
I have this issue since yesterday and did quite a few fresh installs since then (everything @beta). But it doesn't happen on.... https://polkadot.js.org/apps/#/staking/actions

Stefie:
And i just looked at your screencast. 
You don't see any errors in the UI, it's just the console that is getting spammed with error messages on load.
Everything is working fine once the dropdown component received the `isMultiple` prop. But that doesn't happen on load which throws the console error.

jacogr:
Ahhh... ok, now I see it.

PR #1136, opened 2019-05-10
InputAddress should handle defaultValue of U8a

jacogr:
Closes https://github.com/polkadot-js/apps/issues/1134

It handles value properly, but not quite defaultValue when Uint8Array is supplied. (Which is problematic, since AccountId is very valid as an input)


jacogr:
Should have been fixed a long time ago - actually very weird to find it since originally we only worked with u8a all the way through.

PR #1137, opened 2019-05-10
Additional help messages

jacogr:
- Settings > General
- Toolbox (excluding verify, which was done in https://github.com/polkadot-js/apps/pull/1131)

### DIFF
--- a/packages/app-settings/src/General.tsx
+++ b/packages/app-settings/src/General.tsx
@@ -67,6 +67,7 @@ class General extends React.PureComponent<Props, State> {
           <div className='medium'>
             <Dropdown
               defaultValue={uiTheme}
+              help={t('The logo and colors for the app along with the identity icon theme.')}
               label={t('default interface theme')}
               onChange={this.onChangeUiTheme}
               options={uiSettings.availableUIThemes}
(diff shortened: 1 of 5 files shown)


packages/app-settings/src/General.tsx
```diff
@@ -75,6 +76,7 @@ class General extends React.PureComponent<Props, State> {
           <div className='medium'>
             <Dropdown
               defaultValue={uiMode}
+              help={t('Adjust the mode from basic (with a limited number of end-user-only apps) to full (with all basic & advanced apps available)')}
```
Tbaut: I'd say that both beginners and advances users are end-users. Should we go as far as calling them user friendly? I'd say yes :)
```suggestion
              help={t('Adjust the mode from basic (with a limited number of beginner-user-friendly apps) to full (with all basic & advanced apps available)')}
```

packages/app-toolbox/src/Unlock.tsx
```diff
@@ -129,6 +129,7 @@ class Unlock extends React.PureComponent<Props, State> {
             <Password
               className='medium'
               isError={!!unlockError}
+              help={t('The account password as was specified in the creation of this account.')}
```
Tbaut: I'm not native but that sounded a bit weird. Gave my best shot.
```suggestion
              help={t('The account\'s password specified at the creation of this account.')}
```

Issue #1135, opened 2019-05-10
Group unlockings together when the era is the same


Tbaut:
The current implementation is a little dump and won't group the unlocking lines together.
![image](https://user-images.githubusercontent.com/33178835/57538804-712b9800-7349-11e9-8fe7-85b5cc5e64aa.png)

PR #1138, opened 2019-05-10
Group unlocking by era

Tbaut:
closes https://github.com/polkadot-js/apps/issues/1135

I'd take any good advice on how to do this in a more elegant manner, I couldn't figure out how to do this nicely with a `reduce` instead of a `map`+ `for..of`

before / after
image

image

### DIFF
--- a/packages/ui-reactive/src/Unlocking.tsx
+++ b/packages/ui-reactive/src/Unlocking.tsx
@@ -2,7 +2,7 @@
 // This software may be modified and distributed under the terms
 // of the Apache-2.0 license. See the LICENSE file for details.
 
-import { AccountId, AccountIndex, Address, Option, StakingLedger, UnlockChunk } from '@polkadot/types';
+import { AccountId, AccountIndex, Address, Balance, BlockNumber, Option, StakingLedger, UnlockChunk } from '@polkadot/types';
 import { BareProps, CallProps } from '@polkadot/ui-api/types';
 import BN from 'bn.js';
 import { formatBalance } from '@polkadot/util';
@@ -36,6 +36,29 @@ export class UnlockingDisplay extends React.PureComponent<Props> {
     : null ;
   }
 
+  private groupByEra (list: UnlockChunk[]) {
+    let eraMap: Map<string, Balance> = new Map<string, Balance>();
+
+    list.map((el) => {
+      // create a Map with unique eras as keys (Maps need a string|number as key)
+      if (!eraMap.has(el.era.toString())) {
+        eraMap.set(el.era.toString(),el.value);
+      } else {
+        // if the Map already has an entry for the era, sum the value
+        const curr = eraMap.get(el.era.toString()) || new Balance(0);
+        eraMap.set(el.era.toString(), curr.add(el.value));
+      }
+    });
+
+    const groupedList: UnlockChunk[] = [];
+    // reconstruct the UnlockChunk array
+    for (let [eraString, sum] of eraMap) {
+      groupedList.push(new UnlockChunk({ era: new BlockNumber(eraString), value: sum }));
+    }
+
+    return groupedList;
+  }
+
   private remainingBlocks (era: BN) {
     const { chain_bestNumber, session_eraLength } = this.props;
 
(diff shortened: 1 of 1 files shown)


packages/ui-reactive/src/Unlocking.tsx
```diff
@@ -36,6 +36,29 @@ export class UnlockingDisplay extends React.PureComponent<Props> {
     : null ;
   }
 
+  private groupByEra (list: UnlockChunk[]) {
+    let eraMap: Map<string, Balance> = new Map<string, Balance>();
```
jacogr: Difficult since you return a map. A reduce on a object would have worked.

```js
return list.reduce((map, { era, value }) => {
  const key = era.toString();

  if (!map[key)) {
    map[key] = value;
  } else {
    map[key] = map[key].add(value);
  }

  return map;
}, {} as { [index: string]: BN });
```

Tbaut: thanks for the guidance I'll give it another shot

packages/ui-reactive/src/Unlocking.tsx
```diff
@@ -52,24 +70,28 @@ export class UnlockingDisplay extends React.PureComponent<Props> {
...
-        {locked.map((unlocking,index) => (
+      <>
+        { groupedUnlockings && [...groupedUnlockings].map(([eraString, value],index) => (
           <div
             className={className}
             style={style}
             key={index}
           >
-            {labels.locked}{formatBalance(unlocking.value)} ({this.remainingBlocks(unlocking.era).toString()}{labels.remaining})
+            {labels.locked}{formatBalance(value)} ({this.remainingBlocks(new BlockNumber(eraString)).toString()}{labels.remaining})
```
jacogr: This is probably better suited via a,

```js
t('locked {{balance}} ({{remaining}} blocks left), {
  replace: {
    balance: formatBalance(value),
    remaining: this.remainingBlocks(new BlockNumber(eraString))
  }
});
```

i.e. use interpolation on the i18n string

Tbaut: thanks it's much nicer that way indeed.

Tbaut:
I just realized now that the example you gave on `reduce` can actually directly be copy pasted, playing with it now. Not familiar with those maps and Maps.. Is there any benefit to use one or the other? I found it not super easy to get a mental model and to have to deconstruct to be able to `.map()` it..

jacogr:
So when you do `const something = {}` it is actually an `Object`. `Map` is relatively new addition with `get`/`set`. For the most part they can be used interchangeably, old-skool used to be all `{}` over the place.

Here is a decent read - https://medium.com/front-end-weekly/es6-map-vs-object-what-and-when-b80621932373

Fot this specific case since you used it as a temp structure, I believe the Object variant is probably a slightly better choice since you actually used the Map as a scratch-pad. When things are longer-lived and you can live with iterators to get all or know the fields, `Map` could be better.

packages/ui-reactive/src/Unlocking.tsx
```diff
@@ -39,18 +39,17 @@ export class UnlockingDisplay extends React.PureComponent<Props> {
...
-        if (!map.has(key)) {
-          map.set(key, value);
-        } else {
-          const curr = map.get(key) || new Balance(0);
-          map.set(key, curr.add(value) as Balance);
-        }
-
-        return map;
-      }, new Map<string, Balance>())
+       list.reduce((map, { era, value }) => {
```
jacogr: This could simply be (preferred) `return list.reduce` (without the `return ()` wrapper)

packages/ui-reactive/src/Unlocking.tsx
```diff
@@ -74,18 +73,17 @@ export class UnlockingDisplay extends React.PureComponent<Props> {
     const filteredUnlockings = unlockings.filter((chunk) => this.remainingBlocks(chunk.era).gtn(0));
     // group the Unlockchunks that have the same era and sum their values
     const groupedUnlockings = filteredUnlockings.length ? this.groupByEra(filteredUnlockings) : undefined;
-
     return (
       <>
-        { groupedUnlockings && [...groupedUnlockings].map(([eraString, value],index) => (
+        { groupedUnlockings && Object.keys(groupedUnlockings).map(eraString => (
```
jacogr: (nit) Errant space, i.e. rather `{grouperUnlockings && `

Tbaut:
Thanks for the guidance and patience :)

Issue #76, opened 2018-06-08
Add additional metadata on accounts/addresses


jacogr:
Goal:

We want to be able to manage the accounts and addresses better, so the user actually have a better way of categorizing their accounts and, at first glance, be able to see what the account/address relates to.

Background:

With the number of icons, accounts and addresses in the system, it is quite difficult to, at first glance, be able to see what this relates to. On Alexander, my setup has 12+ accounts (for various purposes), and it can be difficult to spot the "right one".

The suggestion here is to to add additional metadata to the saved keyring infomation, initially just focussed on labels that the user can add.

Actions:

- Add the ability for the user to define custom labels (used either for accounts of addresses) - this can simply be an input with "pick from existing" or "add new"
- Defined labels are saved in storage (basically anything defined in the accounts/addresses) 
- Re-use this component on accounts and addresses - as mentioned above, just another add/edit input
- Save this information along with the account. The Keyring supports the meta fields so a "labels" array can just be added with no other overhead.
- For the AddressSummary and AddressRow components, show the actual attached labels (as available) - this way it is useful wherever it is used in the UI

Optionally:

- With the AddressRow, AddressSummary components, allow in-place adding of labels. So without adding an account to my addressbook, I can label it with a "+" (This can actually be a "quick add to addresses" functionality - so specify labels & name, save.

Follow-ups:

Once the above is in-place, we can start doing filtering through the UI based on labels. i.e. on staking, I can filter to see only the validators I am interested in, based on their labelling.

jacogr:
Closing, we have tags.

PR #1142, opened 2019-05-11
Allow tags in Account/Address edit

jacogr:
- Part of https://github.com/polkadot-js/apps/issues/76
- Adds tags to accounts & addresses
- Does not (yet) display them in AddressSummary or AddressRow (probably needs some tweaks here, it is becoming, erm, large...)

Polkadot:Substrate Portal 2019-05-11 10-11-06

- Add InputTags component
- Tags show and are editable for all normal accounts (tags can be added by the user)
- As for now, no tags on built-in testing accounts (name editing, which was not available across refreshes, is also disabled as of now - this could be re-visited)


packages/ui-app/src/InputTags.tsx
```diff
@@ -0,0 +1,81 @@
...
+  placeholder?: string,
+  value?: Array<string>
+};
+
+type State = {
+  options: Array<Option>
+};
+
+function loadTags (): Array<string> {
+  return store.get('tags') || ['Default'];
```
Tbaut: Any reason to have a `Default` tag?

jacogr: I just wanted "something" so you don't start off with everything completely blank. (The way the SUI add-your-won dropdown works it is def. not intuitive)

jacogr:
So yes, there are 3 parts outstanding -

- contracts - that needs a serious re-look, so I didn't initially wanted to make things "worse"
- accounts/address add - you are 100% spot-on about having the tags there. As it stands however, accounts feels like way too may stuff there already. (Addresses however happy to add here immediately)
- Show the tags as part of the Row/Summaries (Which in itself is really simple, however since AddressSummary is really overflowing, also balked at that for the time being)

Issue #1114, opened 2019-05-09
Event data is not updating


shawntabrizi:
I created a simple "adder" module to show the error here.

This module simply adds one to a number and emits an event with the new number.

```rust
use support::{decl_module, decl_storage, decl_event, StorageValue, dispatch::Result};
use system::ensure_signed;

pub trait Trait: system::Trait {
	type Event: From<Event<Self>> + Into<<Self as system::Trait>::Event>;
}

decl_storage! {
	trait Store for Module<T: Trait> as TemplateModule {
		MyNumber get(my_number): u32;
	}
}

decl_module! {
	pub struct Module<T: Trait> for enum Call where origin: T::Origin {

		fn deposit_event<T>() = default;

		pub fn add_one(origin) -> Result {
			let who = ensure_signed(origin)?;

			let my_number = Self::my_number();
			let my_new_number = my_number.checked_add(1).ok_or("Overflow")?;

			<MyNumber<T>>::put(my_new_number);
			Self::deposit_event(RawEvent::NewNumber(my_new_number, who));
			Ok(())
		}
	}
}

decl_event!(
	pub enum Event<T> where AccountId = <T as system::Trait>::AccountId {
		NewNumber(u32, AccountId),
	}
);
```

When I call this function multiple times, I get multiple events, but the data in the events are not updated:

<img width="1421" alt="image" src="https://user-images.githubusercontent.com/1860335/57457399-c1cec280-726f-11e9-8cc8-fb2f845088ab.png">

If I refresh the page, and then call the event again, I get the right number... but then that value repeats...

<img width="1421" alt="image" src="https://user-images.githubusercontent.com/1860335/57457556-065a5e00-7270-11e9-94ce-396e0923ccdf.png">

Let me know if you need help reproducing.

jacogr:
Also happens with something simple such as sessions on a dev chain.

<img width="425" alt="Polkadot:Substrate Portal 2019-05-12 08-43-18" src="https://user-images.githubusercontent.com/1424473/57578722-0d60b680-7492-11e9-8d1c-c40e7d771075.png">

jacogr:
Bumped to P1, since same-type events were _always_ displayed with the wrong info.

PR #1144, opened 2019-05-12
Fix event display indexing & availability

jacogr:
- Closes https://github.com/polkadot-js/apps/issues/1114
- Fix event key for React (_not_ via index)
- Move events state to root (this way it is not reset to [] when swapping between block & overview)
- Below screenshot contrasts with one in the issues (for sessions), where it always duplicated

Polkadot:Substrate Portal 2019-05-12 09-20-14


packages/app-explorer/src/Events.tsx
```diff
@@ -54,35 +44,38 @@ class Events extends React.PureComponent<Props> {
...
 
+    const hash = xxhashAsHex(stringToU8a(JSON.stringify(record)));
+
     return (
-      <article
-        className='explorer--Container'
-        key={index}
+      <div
+        className={eventClassName}
+        key={hash}
```
jacogr: Instead of attaching the hash here, rather create a unique id as it gets added to the array - that way when we do have overlapping events, the rendering will still point to the right items in the queue - and we won't get duplicate key errors. (Have not seen them, but it certainly can happen now since events, system with no nonce, can repeat)

Issue #1139, opened 2019-05-10
Can't sign a message with an ed25519


Tbaut:
As I was testing https://github.com/polkadot-js/apps/pull/1131#pullrequestreview-236200522 I encountered the mentioned bug.

When selecting an account created with `ed25519` the signature field disappears, the dropdown for accounts switches back to `Alice`

Tbaut:
I'm happy to take a look at that as it's not super urgent and probably not super hard. Will have questions though :)

jacogr:
Are we sure it is not absolutely terrible UX here? When selecting a locked account, the "sign" button disappears, and is replaced with an unlock account button. So you first need to unlock before you can sign

Tbaut:
>Are we sure it is not absolutely terrible UX here? When selecting a locked account, the "sign" button disappears, and is replaced with an unlock account button. So you first need to unlock before you can sign

Oh I didn't know this flow (and still havn't because of the bug). Good to know so that I'll know right away what we're talking about we digging into it :)

jacogr:
That explains the disappearance of the signature (can only be done in real-time when unlocked) -

![image](https://user-images.githubusercontent.com/1424473/57546564-49dec600-735d-11e9-92a1-61706050aac7.png)

However, cannot select accounts here properly - all all. So there is some dropdown issue.

So for the above, maybe the signature itself can show, be disabled and have a placeholder that says "account needs to be unlocked". Dunno, whatever works and look much better, would be happy with that. (This stuff has not been properly looked at in ages, it is always on the list and never gotten to - but needs to work 100%)

PR #1157, opened 2019-05-13
Fix signing tool

Tbaut:
fix https://github.com/polkadot-js/apps/issues/1139
added a help label for account.

Question before I implement it:
Rather than letting users click on the "Unlock" button, I'd fire up directly the unlock modal if a locked account is selected. If it's cancelled, then the unlock button will still be there. I would add a message also to say that the data can't be signed unless the account is unlocked (again it would only be visible if the modal is cancelled). What do you think?

### DIFF
--- a/packages/app-toolbox/src/Sign.tsx
+++ b/packages/app-toolbox/src/Sign.tsx
@@ -36,7 +36,7 @@ const Wrapper = styled.div`
     background-color: #0f0e0e7a;
   }
 
-  .unlock-overlay-warning {
+ .unlock-overlay-warning {
     display: flex;
     align-items: center;
     justify-content: center;
--- a/packages/app-toolbox/src/Sign.tsx
+++ b/packages/app-toolbox/src/Sign.tsx
@@ -185,8 +185,6 @@ class Sign extends React.PureComponent<Props, State> {
         };
       }
     );
-
-    console.log('state',this.state);
   }
 
   toggleUnlock = (): void => {
--- a/packages/app-toolbox/src/Sign.tsx
+++ b/packages/app-toolbox/src/Sign.tsx
@@ -35,12 +35,14 @@ const Wrapper = styled.div`
     left:0;
     background-color: #0f0e0e7a;
   }
+
   .unlock-overlay-warning {
     display: flex;
     align-items: center;
     justify-content: center;
     height:100%;
   }
+
   .unlock-overlay-content {
     color:#fff;
     text-align:center;
@@ -79,8 +81,8 @@ class Sign extends React.PureComponent<Props, State> {
       <div className='toolbox--Sign'>
         {this.renderAccount()}
         <Wrapper>
-        {this.renderInput()}
-        {this.renderSignature()}
+          {this.renderInput()}
+          {this.renderSignature()}
           <div className='unlock-overlay' hidden={!isLocked}>
             {this.renderUnlockWarning()}
           </div>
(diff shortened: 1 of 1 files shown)


jacogr:
Sounds fine. I think it is a bit weird to select and have a modal pop-up without your explicit consent, but let's play and see.

packages/app-toolbox/src/Sign.tsx
```diff
@@ -61,15 +57,16 @@ class Sign extends React.PureComponent<Props, State> {
...
     const { t } = this.props;
-    const { defaultValue } = this.state;
+    const { currentPair } = this.state;
 
     return (
       <div className='ui--row'>
         <InputAddress
           className='full'
-          defaultValue={defaultValue}
+          defaultValue={currentPair}
```
jacogr: Not sure how this works - InputAddress takes a string or UInt8Array.

Tbaut:
Agreed although if users select an account it's most probably because they want to sign something with it, by poping up the modal it saves a click and some reading that then need to unlock the account. We'll see how unexpected this modal ends up being.

Tbaut:
I am surprised that [`onChangeAddress`](https://github.com/polkadot-js/apps/blob/tbaut-fix-siging/packages/app-toolbox/src/Sign.tsx#L204) is called when rendering the page, without any user interaction.

From my understanding It is supposed to get called when the input field value changes, so when `currentPair` changes. The later doesn't change according to my logs... So why does `onChangeAddress` get called?

jacogr:
It actually gets called consistently with all inputs - for address it means that it will initialize with the last selected value. Since the components are typically used uncontrolled (ie. No value prop) this makes sense, otherwise the user will either start with an empty selection or (just as bad) start from the top again.

jacogr:
I think the problem, not selection is because we are passing thought he default. It will be on the last account, so actually no need to pass it at all.

Tbaut:
Thanks, the input is actually behaving as expected now (as weird as it can seem it also did with my 1st commit).

I want to toggle the unlock modal when a locked account is selected but only if the users manually select one, not if the default account is locked (when the page is loaded). In that case they would see a warning and could click on the `Unlock` button if needed. To implement this I wanted to set a state variable at `onChangeAddress` to toggle the unlock modal only if users do change the address, but that's simply not possible this way.

Tbaut:
>so actually no need to pass it at all.

Indeed, it also works well this way.

Tbaut:
What do you think of that? It's in my last commit, if you want to play with. Probably needs some color tweaks:

image

jacogr:
That actually works. Nicely descriptive, still allows selection.

packages/app-toolbox/src/Sign.tsx
```diff
@@ -48,36 +73,41 @@ class Sign extends React.PureComponent<Props, State> {
   }
 
   render () {
+    const { isLocked } = this.state;
+
     return (
       <div className='toolbox--Sign'>
         {this.renderAccount()}
+        <Wrapper>
         {this.renderInput()}
```
jacogr: Indent this and the next line.

packages/app-toolbox/src/Sign.tsx
```diff
@@ -2,27 +2,55 @@
...
+
+  .unlock-overlay {
+    position: absolute;
+    width: 100%;
+    height: 100%;
+    top:0;
+    left:0;
+    background-color: #0f0e0e7a;
+  }
+  .unlock-overlay-warning {
```
jacogr: (nit) Extra blank line before (and before the next class following)

packages/app-toolbox/src/Sign.tsx
```diff
@@ -86,13 +116,18 @@ class Sign extends React.PureComponent<Props, State> {
...
-      <Button.Group>
-        <Button
-          isPrimary
-          onClick={this.toggleUnlock}
-          label={t('Unlock account')}
-        />
-      </Button.Group>
+      <div className='unlock-overlay-warning'>
+        <div className='unlock-overlay-content'>
+          You need to unlock this account to be able to sign data.<br/>
```
jacogr: i18n text via `t`

Issue #1044, opened 2019-04-25
Overview of your accounts


Tbaut:
I've needed to have an overview of all my accounts and see their balance at glance. Right now `Accounts` doesn't allow this, `Staking` does ironically.

Here is an idea. ~~We could show an "Edit" button on mouse over to keep the screen lighter.~~ edit: add the ability to edit the name inline.
Additionnaly, we could add filters similar to what we have in the `Staking` app. (#971)

![image](https://user-images.githubusercontent.com/33178835/56729015-789d4f80-6754-11e9-9c23-9c5ea08e522b.png)

jacogr:
Same with addresses. Additionally, currently we have different types of displays in different areas, i.e. staking vs accounts - should be aligned.

jacogr:
Edit is (currently) only for name - that can actually be done in-line, i.e. click on it, it becomes an edit box, unclick, it saves.

Forget/Backup/Password we can probably do with appear buttons - maybe even gray buttons that show their colors when the card is active, that way there is not much jumping going on

nczhu:
Caught up with @Tbaut on this last week. Put some of what we brainstormed down to Sketch. Keen to hear your feedback on this, in light of providing a more beginner-friendly UI for upcoming testnet(s).

1st draft includes:
- Breakdown of types of balances: free (avail, lock, vesting) vs reserved
- Inline edit icon
- Tags
- An empty state related to #1006 

![1  Accounts: empty state](https://user-images.githubusercontent.com/23445933/57706471-1a360380-7666-11e9-844b-864c841a4afc.png)

![2  Accounts: overview](https://user-images.githubusercontent.com/23445933/57706459-1609e600-7666-11e9-9bb1-51bcee2b1286.png)

![3  Accounts: overview 2](https://user-images.githubusercontent.com/23445933/57706484-1e622100-7666-11e9-964c-35c8dde82e4c.png)

Tbaut:
Lots of great ideas. I think it's worth putting this on a meta issue to discuss further. (this one is specific for a first step implementation that I'm actually working on).

PR #1082, opened 2019-05-03
Grayscale article buttons

jacogr:
A trial/test/play spawned from this comment - https://github.com/polkadot-js/apps/issues/1044#issuecomment-489099134

Currently only on staking -

Polkadot:Substrate Portal 2019-05-03 15-49-58

PR #1163, opened 2019-05-14
Account overview

Tbaut:
- closes https://github.com/polkadot-js/apps/issues/1044

- [x] All accounts can be seen on the overview tab.
- [x] The name is editable and forced in upper case in the input field (otherwise it looked weird to have it lower case in the input and upper in the view)
- [x] New Name is saved on press `Enter or click away (Blur)
- [x] slightly changed the look&feel of the unlock button as I was at it.
- [x] view and edit tags
- [x] overview of account's balance on the right
- [x] Forget & backup buttons
- [x] remove the edit tab


Tbaut:
Should be good for a first review.
As marked inline as a comment I've seen that `renderAccountIndex` isn't used anywhere on master. That's maybe a mistake. I've seem the index on Alexander though (polkadot-js.org), which confuses me as I thought it was running master?

Tbaut:
I'd be glad to have an explanation or a link regarding our class name convention. I couldn't really figure out whether we should simple name classes `ui--tagName--function`, or if things are actually more complex than that.

I'll work on the design of the input, lower height, removing the border probably.

jacogr:
What hapenned to the AccountIndex?

Tbaut:
This is actually a question I ask 

>As marked inline as a comment I've seen that renderAccountIndex isn't used anywhere on master. That's maybe a mistake. I've seem the index on Alexander though (polkadot-js.org), which confuses me as I thought it was running master?

What did I miss, why isn't it in master?

jacogr:
Alex is on the 0.4 Polkadot branch. Only FF runs master (EE runs 1.0 substrate)

But basically, may be that you have not sent anything yet - so under edit for me -

image

jacogr:
So played around a tiny bit - I think we would be better off with a layout like we have in staking.

On the left:

- icon
- name
- accountId
- accountIndex
- tags
- available balance (i.e. can be spent)

On the right:

- all balances broken down
- ability to unlock balances
- crypto type (this is useful, since we will probably continue to have a different type for sessions)
- buttons such as forget (trash icon) & backup

Tbaut:
Something like this?
image
image

I really think crypto and number of tx is interesting rarely and could be hidden a little further. but as a start.. 

I'd do this in another PR then only for look&feel

jacogr:
Yeap, that looks ugly, but sane ;)

Tbaut:
If you're ok with this PR I can iterate on this and get in the direction of the Mockup

packages/app-accounts/src/Account.tsx
```diff
@@ -0,0 +1,46 @@
...
+  max-width: 19%;
+
+  .ui--AddressSummary{
+    justify-content: center;
+  }
+`;
+
+class Account extends React.PureComponent<Props> {
+
+  render () {
```
jacogr: (nit) Extra blank line before

packages/app-accounts/src/Account.tsx
```diff
@@ -0,0 +1,46 @@
...
+  flex: 1 1;
+  min-width: 19%;
+  max-width: 19%;
+
+  .ui--AddressSummary{
+    justify-content: center;
+  }
+`;
+
+class Account extends React.PureComponent<Props> {
```
jacogr: just export `default class Account` (no HOCs)

packages/app-accounts/src/Overview.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+    .spacer {
+      flex: 1 1;
+      margin: .25rem;
+      padding: 1rem 1.5rem;
+    }
+  }
+`;
+
+class Overview extends React.PureComponent<Props> {
```
jacogr: Same default export as previously (not going to comment when I see more, or maybe will just mark with something - instead of the same comments)

packages/app-accounts/src/Overview.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+  constructor (props: Props) {
```
jacogr: (nit) Remove extra blank line before

packages/app-accounts/src/Overview.tsx
```diff
@@ -0,0 +1,56 @@
...
+    super(props);
+  }
+
+  render () {
+    const accounts = keyring.getAccounts();
```
jacogr: This should be injected by the observable, so when things are added it is immediately available (which is important once the add is there) Just look for `allAccounts`, like it typically gets injected.

packages/app-accounts/src/index.tsx
```diff
@@ -45,6 +46,10 @@ class AccountsApp extends React.PureComponent<Props, State> {
     this.state = {
       ...baseState,
       tabs: [
+        {
```
jacogr: In the final version should remove edit.

packages/ui-app/src/AddressSummary.tsx
```diff
@@ -2,20 +2,21 @@
 // This software may be modified and distributed under the terms
 // of the Apache-2.0 license. See the LICENSE file for details.
 
-import { I18nProps } from './types';
-
-import BN from 'bn.js';
-import React from 'react';
 import { AccountId, AccountIndex, Address } from '@polkadot/types';
+import BaseIdentityIcon from '@polkadot/ui-identicon';
```
jacogr: Move alphabetical -

- types go first (anything `@polkadot/<package/types` or `../types`, etc.
- blank line
- external non-polkadot imports
- polkadot imports
- blank line
- LOCAL IMPORTS

packages/ui-app/src/AddressSummary.tsx
```diff
@@ -2,20 +2,21 @@
...
 import { withCalls } from '@polkadot/ui-api';
-import BaseIdentityIcon from '@polkadot/ui-identicon';
 
 import AvailableDisplay from './Available';
-import { classes, getAddrName, toShortAddress } from './util';
 import BalanceDisplay from './Balance';
 import BondedDisplay from './Bonded';
+import { classes, getAddrName, toShortAddress } from './util';
 import IdentityIcon from './IdentityIcon';
+import { I18nProps } from './types';
```
jacogr: Move to types section.

packages/ui-app/src/AddressSummary.tsx
```diff
@@ -70,58 +84,92 @@ class AddressSummary extends React.PureComponent<Props> {
...
       ? accountId.toString()
       : DEFAULT_ADDR;
-    const name = getAddrName(address, false, defaultName);
 
-    return (
-      <div className='ui--AddressSummary-data'>
-        <div className='ui--AddressSummary-name'>
+    const name = getAddrName(address, false, defaultName) || '';
+
+    let className = 'ui--AddressSummary-name';
```
jacogr: can just use `ui-app/util/classes` - it removes blanks and concats

packages/ui-app/src/AddressSummary.tsx
```diff
@@ -70,58 +84,92 @@ class AddressSummary extends React.PureComponent<Props> {
...
+          onChange={this.onChangeName}
+          onKeyDown={this.handleKeyDown}
+          withLabel={false}
+        />
+        {this.renderSaveIcon()}
+      </>
+       :
+        <div
+          className={className}
+          onClick={ isEditable ? this.toggleEditor : undefined }
```
jacogr: Extra ` ` before and after `{}`'s

packages/ui-app/src/AddressSummary.tsx
```diff
@@ -70,58 +84,92 @@ class AddressSummary extends React.PureComponent<Props> {
...
+
+    return resultingDom;
+  }
+
+  protected handleKeyDown = (e: React.KeyboardEvent<Element>) => {
+    const { key } = e;
+    switch (key) {
+      case 'Enter':
+      case 'Escape':
+        this.saveName();
```
jacogr: Would not save on `Esc` - just use the `onEnter` callback in `Input` rather

packages/ui-app/src/AddressSummary.tsx
```diff
@@ -183,39 +226,29 @@ class AddressSummary extends React.PureComponent<Props> {
...
-
-    return (
-      <UnlockingDisplay
-        className='ui--AddressSummary-available'
-        params={accountId}
-      />
-    );
+  protected renderEditIcon () {
+    return <Button
+            className='editButton'
```
jacogr: (nit) Funky indentation here.

packages/ui-app/src/AddressSummary.tsx
```diff
@@ -261,20 +294,52 @@ class AddressSummary extends React.PureComponent<Props> {
       </Nonce>
     );
   }
+  protected renderSaveIcon () {
+    return <Button
```
jacogr: (nit) Indentation.

packages/ui-app/src/AddressSummary.tsx
```diff
@@ -261,20 +294,52 @@ class AddressSummary extends React.PureComponent<Props> {
...
   }
+
+  protected saveName = () => {
+    const { value } = this.props;
+    const { newName } = this.state;
+
+    const trimmedName = newName.trim();
+
+    // Save only if the name was changed or if it's no empty.
+    if (trimmedName !== '') {
```
jacogr: `if (value && trimmedName) {` (empty string is false-y)

packages/ui-app/src/AddressSummary.tsx
```diff
@@ -261,20 +294,52 @@ class AddressSummary extends React.PureComponent<Props> {
...
+      const currentKeyring = value && keyring.getPair(value.toString());
```
jacogr: Would rather not enable the editor if not in the keyring. Here we are enabling, but then may not save it.

packages/ui-app/src/Input.tsx
```diff
@@ -172,6 +173,14 @@ export default class Input extends React.PureComponent<Props, State> {
     }
   }
 
+  private onBlur = (event: React.KeyboardEvent<Element>): void => {
+    const { onBlur } = this.props;
+
+    if (onBlur) {
+      onBlur(event);
```
jacogr: Not sure we need to pass the `event` back up - it makes sense for key-presses, but for the rest probably not.

packages/ui-app/src/AddressSummary.tsx
Tbaut: Agreed, I guess having the keyring in the state wouldn't make sense as it doesn't change. I couldn't think of a better way than checking for `keyring.getPair(value.toString()` when first toggling the editor. lmk if there's a better ways.

packages/ui-app/src/AddressSummary.tsx
Tbaut: oh that's nice, hadn't noticed it before.

packages/ui-app/src/AddressSummary.tsx
Tbaut: I hope I got that right, so _external non-polkadot imports_ and _polkadot imports_ are not mixed, but they are **not** separated by a blank like eithr (e.g how it basically way before I messed things up). I'll do that and you can still tell me if that's wrong.

Tbaut:
Thanks. I didn't want to let this PR grow too much but with this review in between I'm happy to add the tags with edition, backup & forget.

jacogr:
tags, you can forgo, add, send, forget as well if that helps - but do think that we need to at least get the layout right as a first pass. (As a proper summary). 

TL;DR - Can forgo removing the editor first-pass, but the summary needs to add overall value.

Tbaut:
If we do this step by step with reviews in between, I'm ok to go on from here.
IMO The overview from this PR  adds value on in the fact that it replaces the staking>account whose purpose is different (that's my next target :) ).

I get your point and it will hopefully not take me weeks to have something.

Tbaut:
Need a tip for the styling.
How things look right now:
image

- red: `display: flex`
- orange and yellow: `flex:1`
- blue is a [dirty hack](https://github.com/polkadot-js/apps/pull/1163/files#diff-9b1b7c2ebb802f5c7b83a67c287e2032R46) for now.

Question:
The only way I know to do this nicely is to add a spacer to each row, even if the row has nothing on the right to show, this would result in:
image

Any better idea?

jacogr:
Since it is tabular, you may as well use a table or css-grid. (Tables are great for tabular layouts.)

Tbaut:
I read the edited post now that I've "learned" and implemented something with a simple `grid`!

Tbaut:
Here is how things look like now. I'd say it's not the prettiest. I'll cleaning things a little and it'll be ready for review soon.
image

It's rather close to the [mockup above](https://github.com/polkadot-js/apps/pull/1163#issuecomment-493042937). The transfer button is missing, I'll add it. Anything you're thinking about to make it look better? @nczhu @yjkimjunior @jacogr

nczhu:
Looks great! let me know if you want to do any `?` icons to explain what the terms mean, e.g.: 
- `redeemable`: these are your recently unlocked funds from ... that can be redeemed back into your account as spendable balance
- `locked`: these are funds that are temporarily locked as a result of your voting/governance activities

Can get you copy like that ^... but you know... phrased better :)

Tbaut:
Totally, I was also thinking about that. It'll look super ugly in the end with all those (?) but let's get useful things first and then we make it nice, maybe :)

jacogr:
Just one small UX comment on the editing - we should have something that indicated the fields are editable, as it stands, going to have the same "move to mouse to discover" issues, and well, getting tired of those.... :)

Tbaut:
Agreed, we can have those edit button be always visible. Maybe smaller. I'll see if I can find something that works.

Tbaut:
Edit buttons on tags and names are now always visible (although with opacity 0.7 -> 1 on hover). Could be totally hidden as long as the mouse isn't on the card itself maybe.

I'll open issues to add the Transfer button and (?) helps.

Tbaut:
- copy button on the right
- grid is zig-zag (I messed this up at the last minute it seems)
- label 0.7, balance 1 opacity, no more bold
- took #1082 in, no more show / hide on hover
- mini bar fixed, it used to be a lot smaller indeed.

>  Not as easy to get 100%, but when clicking edit the UI jumps down. Move the name slightly down and then just play with the margin/passing until it overlays what is there. (Tags also jump the screen down)

Taking a look at that now.

Tbaut:
The jump should be mitigated, crossing my fingers that this is cross-screen compatible.. I had to go into the .05em to pretend not seeing the micro jumps.

Tbaut:
Almost giving up on the tag jump, the jump will depends on the number of tags (or say their length). I tried to put the dropdown in a `position: absolute` and let the tags display underneath but that's super hacky..

jacogr:
I must admit, it took me 3 tries before I saw the tags jump. It is not as obvious. Fix for a single row, otherwise just let it jump. That solves 99% of cases without hacks.

Tbaut:
The UI should not jump anymore with no tag or tags on 1 line. But since what matters is the place tags take in the dropdown not when they are rendered, you can have tags on 1 line that actually end up being on 2 lines in the dropdown.. and the UI jumps. I don't think we can do better than taking this Dropdown out of the flow with a position absolute. But it's annoying, I'll give it another shot tomorrow.

jacogr:
I think it looks fine as it is - and certainly the gray-scale is helping a lot, imho.

My last UI-relate comment is just what seems like wasted space in the layout, compare these 2 (staking seems to use the avilable space, the accounts do not) -

Polkadot:Substrate Portal 2019-05-21 21-20-17
Polkadot:Substrate Portal 2019-05-21 21-20-03

jacogr:
... and yes, the damn CopyButton also screws up since the name/address are not properly aligned - I broke it, I'll fix that...

Tbaut:
- Agreed that toning things down helps.
- I've tried something by hiding the edit buttons when not on the card. I looks cleaner IMO, lmk if you still find this irritating (overview below) :)
- For the copy button that prevents the address to be centerd, I've used a hack [here](https://github.com/polkadot-js/apps/pull/1163/files#diff-ac54e2de59aa4a9d39d402ceaa550154R123) to achieve something similar for the edit buttons. I'd be glad to learn how to do this in a better way.
- I docked the buttons (and the rest) to the top right and avoid their jump. There's still space depending on the setup though, I can't show 4 accounts side by side on say my laptop screen because some rows tend to be large (e.g "redeemable" + amoount + button):
image

jacogr:
Hiding edit while not on the card is fine. They are in-obtrusive enough.

Will take a peek. Anxious to get this in :)

Tbaut:
- Hack on the copy button to get addresses centered
- Align things left (name/ balance) for AddressRow (mostly used in modal)
- Fix wrong size alignment for small name for nominators AddressMini

Issue #1161, opened 2019-05-14
Rework Unlocking modal in Toolbox > Sign


Tbaut:
The modal visible in Toolbox > Sign > Unlock account is not consistent with other modals we have in Apps. Similar to the work done here https://github.com/polkadot-js/apps/pull/1143

This modal probably needs to present the account on the left (with names etc..) and the description text on the right.

![image](https://user-images.githubusercontent.com/33178835/57682983-2359ad00-7633-11e9-8dfa-0d32c6cda866.png)

jacogr:
And, boy, o, boy, is that one ugly.

Issue #1162, opened 2019-05-14
Enter key submits the form


Tbaut:
Goal:
Make sure modals and common forms can be submitted with `enter` keys.

Background:
In many cases users that needed to type something as the last parameter of a form, e.g an amount or password, their hand are on the keyboard and they hit the `enter` / `return` key to submit the form to save a couple seconds.

Actions:
- Staking > Bond / BondExtra / Unbond / Set Session 
- Transfer
- Account > Edit / Create / Restore
- Address book > Edit / Create
- Chain state
- Extrinsics
- Toolbox > Unlock

I probably forgot many, but that'd be a start

jacogr:
This could be useful for some cases - https://github.com/polkadot-js/apps/blob/master/packages/ui-app/src/Input.tsx#L31 (Added it recently for the submission of the block query on the explorer)

PR #1166, opened 2019-05-15
Unlock modal update

jacogr:
- Closes https://github.com/polkadot-js/apps/issues/1161

Polkadot:Substrate Portal 2019-05-15 08-03-56

Does not do enter-on-submit from #1162 (Didn't want to modify password - or even look if it has the `onEnter`). However, at least password has `autoFocus` now, which helps.

### DIFF
--- a/packages/app-toolbox/src/Unlock.tsx
+++ b/packages/app-toolbox/src/Unlock.tsx
@@ -6,9 +6,7 @@ import { I18nProps } from '@polkadot/ui-app/types';
 import { KeyringPair } from '@polkadot/keyring/types';
 
 import React from 'react';
-import styled from 'styled-components';
-import { Trans } from 'react-i18next';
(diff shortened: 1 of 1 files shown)


packages/app-toolbox/src/Unlock.tsx
```diff
@@ -75,72 +52,61 @@ class Unlock extends React.PureComponent<Props, State> {
...
+          value={address}
+        >
+          <p>{t('You are about to unlock your account to allow for the signing of messages. Once active the signature will be generated based on the content provided.')}</p>
+          <div>
             <Password
-              className='medium'
+              autoFocus
               isError={!!unlockError}
               help={t('The account\'s password specified at the creation of this account.')}
               label={t('unlock account using')}
```
Tbaut: With the additional text, I'd say it's clear that the account will be unlocked.

```suggestion
              label={t('password')}
```

PR #1170, opened 2019-05-15
Button/TxButton submit on enter

### LINKED ISSUE #1162: Enter key submits the form

kwingram25:
Closes #1162 

Add submitOnEnter prop to trigger enter-submit for Button and TxButton and activate on most forms

Needs testing to ensure no event conflicts or side effects, seems safe so far


packages/app-accounts/src/Vanity/index.tsx
```diff
@@ -98,6 +98,7 @@ class VanityApp extends React.PureComponent<Props, State> {
               ? t('Stop generation')
               : t('Start generation')
           }
+          submitOnEnter={!isRunning}
```
axelchalon: Button doesn't listen to prop changes so I think only the initial value of the prop will be taken into account here

kwingram25:
Testing an approach using refs to trigger buttons/txButtons from Input.onEnter() - currently on Transfer.tsx and Vanity.tsx only

This feels like a lot less elegant in my opinion with a lot more almost-repeating code to add - also needs a lot more work on each possible input component, for example text inputs on InputExtrinsic

kwingram25:
Add TxComponent to link Button/TxButton ref (this.button) with enter events on text inputs (this.submit / this.sendTx)

jacogr:
Some conflicts snuck in.

jacogr:
@Tbaut Since you logged the issue, would like you to play with it as well. (Hence have not merged as-is - it feels fine to me)

kwingram25:
Had to move the ref from class method to constructor, strange bug where it was null on modals

Issue #800, opened 2019-03-08
Support multiple queries in a single withCall


jacogr:
This has 2 parts to it, the apps `ui-app` part and the API part (the latter should be done first and logged seperately from this, this is more the meta issue)

Apps -

We want to be able to make calls to retrieve multiple values in one, for instance if we currently want to query the balances for `Alice` and `Bob`, we would currently do

```js
withCalls(
  ['query.balances.freeBalance', Alice],
  ['query.balances.freeBalance', Bob]
)
```

At the very least this is problematic since we cannot really pull in multiple params and adjust the number of queries based on the input. Additionally it just explodes code-wise. So we want to be able to do stuff (syntax needs tweaking, example only)

```js
withCalls(
  ['query.balances.freeBalance', Alice, Bob, Dave, ...]
)
```

The second part that touches the API is the optimisation of the calls. 

The `state_storage` subs can support multiple keys per subscription. Currently we are only doing a single subscription for each query even when we know from the use side that we need to do multiples in a row. Doing multiples in one-go can reduce and optimise the load/number of subs greatly on both sides. 

Even if at first we only implement this via the same type, maybe even `api.query.balances.freeBalance.multi(Alice, Bob, Dave, ...)` it will drop a lot of the use and can be used to make thing more optimal, for instances in the case of `api.derive.*`

cc @siman

siman:
This is a really needed enhancement! For example, this will make possible loading a list of balances by account ids and then sort that list of accounts by their balance. Right now I don't even know how to do this with a current approach when I need to get balance in a sub-component (a Row) but how parent component (a List) can sort its rows? Only a Row knows its balance.

amaurymartiny:
I'm only commenting on the 2nd part. The `.multi` sounds good, it's clear on what it does (principle of least surprise), so I'm okay with this addition.

But as you said, it's only on one storage key, so I won't help in most derive.* where the storage keys are different, e.g. I can't do a multi-subscription to:
```
api.query.balances.freeBalance(Alice)
api.query.system.accountNonce(Alice)
```
both in one subscription.

I'd see something like:
```javascript
api.query.multi(
  api.query.balances.freeBalance(Alice),
  api.query.system.accountNonce(Alice)
).then/subscribe()
```

The freeBalance example above would be slightly more verbose:
```javascript
api.query.multi(
  api.query.balances.freeBalance(Alice),
  api.query.balances.freeBalance(Bob),
  api.query.balances.freeBalance(Dave),
)
[...]

PR #1173, opened 2019-05-16
Convert to new api.derive (multi-query) functions

jacogr:
Part of #800

### DIFF
--- a/packages/app-staking/src/Account/index.tsx
+++ b/packages/app-staking/src/Account/index.tsx
@@ -28,9 +28,6 @@ type Props = ApiProps & I18nProps & {
   recentlyOffline: RecentlyOfflineMap,
   freeBalance?: Balance,
   staking_info?: DerivedStaking,
-  staking_stakers?: Exposure,
-  staking_nominators?: [Array<AccountId>],
-  staking_validators?: [ValidatorPrefs],
   stashOptions: Array<KeyringSectionOption>
 };
 
@@ -45,9 +42,12 @@ type State = {
   isSessionKeyOpen: boolean,
   isValidatingOpen: boolean,
   isUnbondOpen: boolean,
+  nominators?: Array<AccountId>,
   sessionId: string | null,
+  stakers?: Exposure,
   stakingLedger?: StakingLedger,
-  stashId: string | null
+  stashId: string | null,
+  validatorPrefs?: ValidatorPrefs
 };
 
 function toIdString (id?: AccountId | null): string | null {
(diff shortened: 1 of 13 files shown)


jacogr:
@Tbaut This is still WIP, but I converted the staking -> Accounts page to use the new derive call (which is much less load on the RPC as well). I would appreciate is you could, while I'm far from done, just give that pages a play-through as it stands.

packages/app-staking/src/Account/index.tsx
```diff
@@ -274,50 +311,42 @@ class Account extends React.PureComponent<Props, State> {
...
+            label={t('Unbond')}
+          />
+        );
+      }
+    } else if (isActiveController) {
       const isNominating = !!staking_nominators && !!staking_nominators[0] && staking_nominators[0].length;
       const isValidating = !!staking_validators && !!staking_validators[0] && !(staking_validators[0].isEmpty);
 
+      // if we are validating/nominating show stop
       if (isValidating || isNominating) {
```
Tbaut: This is always false (`isValidating` is not derived from props) so that the "stop validation" on Alice doesn't show

Tbaut:
Other than "Stop Validating" not showing up, the rest looks good.
Tested:
- Bond > BondExtra 
- Unbond a part
- Unbond everything (Unbond button doesn't show up anymore)
- SessionKey

Everything was working well except the "Stop Validating" (just tested the validating but the logic is similar for nominating)

jacogr:
Grrr. Stop should show. Ok. Will re-check the damn nested logic again.

packages/app-staking/src/Account/index.tsx
jacogr: Ahhh.  Wrong block. Going to split these anyway. Dangit. A complete mess there as it stands, no wonder.

jacogr:
@Tbaut Seems to do the right thing for me now... Hopefully...

image

Overall, dropped the repeated queries by a helluva lot (and the ones there are multi), so it should improve at least the RPC load a lot.

Would appreciate knowing if it works for you?

Tbaut:
I got this error whit a freshly reloaded page, whatever I do (tried `unbond`, `set session key`, nomminate)
image

^ didn't appear after and also didn't seem to prevent any action. Other than that I'm not sure the nomination works as expected (need to try on master). clicking on `nomminate` seems to send the extrinsic but I never appear as a nominator anywhere even after a session or an era. But again, I need to triple check this.

jacogr:
Nomination does work for me (at least on Alex), but will do master as well.

jacogr:
The Button thing was not caused here, but I've also seen it very sporadically - will need to log that one. Somewhere in TxButton we are doing something not quite right in some cases - whatever those may be.

Tbaut:
:+1: Ok I can confirm everything works, incl nominating, now sure wth happened.

jacogr:
Cool. Step 1. A start. Every bit helps.

Issue #1168, opened 2019-05-15
Rework addresses with card layout from #1044


jacogr:
Goal:

Consistency

Background:

The accounts pages are being reworked as part of #1044 , pull in the same layout into addresses. (Probably would have been simpler to start with addresses since there are less moving parts)

Actions:

- Complete one, do the other

Tbaut:
Doing this one in parallel with https://github.com/polkadot-js/apps/issues/1044

First getting the overview screen with Name edition.

PR #1174, opened 2019-05-17
Address book overview and name editing

Tbaut:
- addresses part of https://github.com/polkadot-js/apps/issues/1168


Tbaut:
Too many things changed

Issue #1185, opened 2019-05-21
Convert all query.balances.free to derive.balances.all


jacogr:
Goal:

Consistency, speed & caching.

Background:

Before multi, calling the indv. queries was less expensive since the all variant made multiple calls. And then when calling free the query was cached. However now with multi the indv. query is not actually cached anymore - we are no making multiple subscriptions to retrieve the same values.

Actions:

- Search for all query.balances.*
- Convert these to derive.balances.all - extracting the value we are interested in

Overall this will reduce the number of subscriptions.

PR #1186, opened 2019-05-22
Convert balances.freeBalance to derive.balances.all

Tbaut:
- closes https://github.com/polkadot-js/apps/issues/1185
- reorder alphabetically on the fly


packages/ui-reactive/src/Unlocking.tsx
```diff
@@ -13,7 +13,6 @@ import { TxButton } from '@polkadot/ui-app';
 import { withCalls } from '@polkadot/ui-api';
 
 type Props = BareProps & CallProps & I18nProps & {
-  balances_freeBalance?: BN,
```
Tbaut: This wasn't used, my bad I guess since I've mainly touched this component.

jacogr: It happens. I've done that a couple of times as well.

PR #1191, opened 2019-05-22
Redirect correctly when user has no account

Tbaut:
- fix link and hide overview tab when user has no account
- make button primary (blue in substrate template)


jacogr:
Some conflicts with https://github.com/polkadot-js/apps/pull/1192 - i.e. I'm using Button with Tooltip

Tbaut:
I must have been drunk when testing. This should be fixed now.

PR #1192, opened 2019-05-22
Hide edit on injected accounts, Button w/ tooltip

jacogr:


jacogr:
erk. Missed that.

jacogr:
Ok, something to not show "add tags" when there are not and not editable

jacogr:
Agreed. Need to play with this a bit though - I need to see it over some time. Make no mistake, it is great, but I have some nigglies I cannot quite put my finger on.

jacogr:
... then again, I have "some nigglies I cannot put my finger on" everywhere :)

PR #1194, opened 2019-05-22
Make enable() call to extensions

jacogr:
- Allows the use of `enable()` for extensions - this allows the extension to approve/reject
- Needs https://github.com/polkadot-js/extension/pull/20
- Demo https://youtu.be/Tio_Tl31M44

### DIFF
--- a/packages/apps/src/index.tsx
+++ b/packages/apps/src/index.tsx
@@ -11,7 +11,7 @@ import ReactDOM from 'react-dom';
 import { HashRouter } from 'react-router-dom';
 import store from 'store';
 import { getTypeRegistry } from '@polkadot/types';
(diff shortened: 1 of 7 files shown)


jacogr:
Agreed. I think however need to re-think these pop-overs - between this one, the connected event and the "you have no accounts", being at the bottom is not great. I'm almost leaning towards somewhere like we have for the modals. Right in your face.

On the specific auth thing as well - the extension with just the lil bubble is not prominent enough - so there is some work there to do something so it is in your face. (@amaurymartiny  also commented on this one - I'm not sure how to do it, but certainly for both this and the "action signing now" - the little label is not enough). TL;DR - for the extension, will need to figure something out... learning as a I go.

Issue #1181, opened 2019-05-21
Add (?) helper for balances


Tbaut:
As per https://github.com/polkadot-js/apps/pull/1163#issuecomment-494320676 it would make sense to add (?) in `Account`>`Overview` to help users understand what they are looking at. Something along the line:

- total: these are the total amount of vested funds that can be bonded
- available: these are the funds that are available to transfer or bond
- redeemable: these are your recently unlocked funds that can be redeemed back into your account as spendable balance
- locked: these are funds that are temporarily locked as a result of your staking/governance activities

Tbaut:
done in https://github.com/polkadot-js/apps/pull/1216

PR #1198, opened 2019-05-24
Helper for balance overview

Tbaut:
- closes https://github.com/polkadot-js/apps/issues/1181
- need to refactor the `Unlocking` component (todo).

looks like this, (?) could be toned down a little:
image

### DIFF
--- a/packages/app-accounts/src/Account.tsx
+++ b/packages/app-accounts/src/Account.tsx
@@ -69,6 +69,9 @@ const Wrapper = styled.article`
   .label-redeemable {
     grid-column:  1;
     text-align: right;
+    > i.help {
+      margin-left: .3em
+    }
   }
 
   .result-available,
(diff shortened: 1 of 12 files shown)


jacogr:
Can't we add it to the label (consistency with what we have everywhere else?) It would also remove the zig-zag on those.

Tbaut:
I tried different one and didn't find it particularly nice:
image

jacogr:
Just needs a bit of padding to the left of the icon. Yes, think it could be a bit much, however that is mostly due to the amount of stuff we have there.

- It aligns visually
- when actually exploring, you can just move the mouse down - so it is faster to get to grips with the actual meaning of all these

Tbaut:
Sounds good, I'll get it back there and see if we can make it a little lighter visually with opacity (apply this generally, I don't think it can hurt too much as long as they are visible)

Tbaut:
image

Tbaut:
closing as pulled by  #1210

PR #1207, opened 2019-05-24
Pull in @polkadot/extension-dapp

jacogr:
- Use extension helper functions from https://github.com/polkadot-js/extension/tree/master/packages/extension-dapp - remove code moved there
- Small AddressSummary display fixes (tags, name)


jacogr:
The AccountSummary needs some rework - it is a monster now.

PR #1209, opened 2019-05-24
Transfer fees calc cleanups

jacogr:
Reworked the fee calculations to one-shot promises - instead of subscriptions. This needs to be applied everywhere where there are calculations.

(It potentially just keeps on creating the original nested versions each time there are changes - do it explicitly when asked)


jacogr:
I just took a look while working on this - however, need to apply the same type of logic to elsewhere as well - I just have no clue where all elsewehere is... :) (off the top of my head)

Issue #1208, opened 2019-05-24
Move transfer modal action to the default


jacogr:
Goal:

Remove the functionality duplication between transfer actions. 

Background:

The transfer modal was recently introduced as part of the work on accounts. While the addresses work are still outstanding, swapping the default now leads to no loss in functionality, with extensions for the future.

Actions:

- Add summary (only available) balances for to and from on the transfer modal
- Swap `app-transfer` to hidden

PR #1210, opened 2019-05-24
Add available balances to transfer modal

jacogr:
Sprawled completely out of control -

- fixed max calculations to not kill RPC (both modal & transfer app)
- fixed nonce display in extrinsics (cased by the point below)
- reworked ui-reactive to original form (just simple wrappable classes)
- add `Label` & `LabelHelp` components (see PR closed below)
- Hide/show all clickables on card non-hover to make things cleaner
- Fix onSearch bug in `AddressInput` (was not working)
- Adjusted `AddressRow` layout... swapped to it everywhere for cards
- Added `AddressInfo` for the grid-y balance display (staking & accounts)
- ... show available balances on transfer
- Pull in and closes https://github.com/polkadot-js/apps/pull/1198 (Labels completely reworked)
- Part of  https://github.com/polkadot-js/apps/issues/1208

Only "part of" since probably need to be a bit less cowboy and only remove the transfer app when we can deposit into an address anywhere. The first thing that will happen is that in the "hide universe" (without it everywhere), is, well, questions since people will get confused.

... so, boiling the frog slowly

### DIFF
--- a/packages/app-accounts/src/modals/Transfer.tsx
+++ b/packages/app-accounts/src/modals/Transfer.tsx
@@ -48,9 +48,8 @@ const Wrapper = styled.div`
     text-align: right;
     padding-right: 1rem;
 
-    .label-available {
+    .label {
       opacity: 0.7;
-      margin-right: 0.75rem;
     }
   }
 
@@ -161,7 +160,7 @@ class Transfer extends React.PureComponent<Props> {
   private renderContent () {
     const { address, t } = this.props;
     const { extrinsic, hasAvailable, maxBalance, recipientId, senderId } = this.state;
-    const available = t('available balance');
+    const available = <span className='label'>{t('available ')}</span>;
 
     return (
       <Modal.Content className='app--account-Backup-content'>
--- a/packages/app-accounts/src/modals/Transfer.tsx
+++ b/packages/app-accounts/src/modals/Transfer.tsx
@@ -56,8 +56,8 @@ const Wrapper = styled.div`
 
   ${media.DESKTOP`
     article.padded {
-      margin: .75rem 15rem;
-      padding: 0;
+      margin: .75rem 0 0.75rem 15rem;
+      padding: 0.25rem 1rem;
     }
   `}
 
--- a/packages/app-accounts/src/modals/Transfer.tsx
+++ b/packages/app-accounts/src/modals/Transfer.tsx
@@ -235,10 +235,7 @@ class Transfer extends React.PureComponent<Props> {
 
     while (!prevMax.eq(maxBalance)) {
       prevMax = maxBalance;
-
-      extrinsic = address && recipientId
-        ? api.tx.balances.transfer(recipientId, prevMax)
-        : null;
+      extrinsic = api.tx.balances.transfer(recipientId, prevMax);
 
       const txLength = calcSignatureLength(extrinsic, accountNonce);
       const fees = transactionBaseFee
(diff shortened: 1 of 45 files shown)


jacogr:
All the millions if extra info, only displays on hover -

image

.. and finally, transfer, is well, balance-y

image

jacogr:
Current...

Polkadot:Substrate Portal 2019-05-25 18-21-23
Polkadot:Substrate Portal 2019-05-25 18-21-02

jacogr:
And staking now has balances like Accounts  -

image

Ok, enough is enough. Done. (For now)

packages/app-accounts/src/modals/Transfer.tsx
```diff
@@ -136,46 +159,43 @@ class Transfer extends React.PureComponent<Props> {
...
-            <Checks
-              accountId={address}
-              extrinsic={extrinsic}
-              isSendable
-              onChange={this.onChangeFees}
-            />
-          </Wrapper>
-        </Modal.Content>
-      </>
+      <Modal.Content className='app--account-Backup-content'>
```
Tbaut: "Transfer" instead of "Backup"?

Tbaut:
Wow you have played a lot! I find things a little overwhelming, particularly with mouse over.
I think we should be scared to space things a little and if it takes too much place, we can still have an overview and an expanded view at the click of a button.
image

Little nit, I found the address (and copy button) to overlap with the "delete" button.

packages/ui-app/src/LabelHelp.tsx
```diff
@@ -0,0 +1,65 @@
...
+};
+
+const Wrapper = styled.div`
+  cursor: help;
+  display: inline-block;
+  line-height: 1rem;
+  margin: 0 0 0 0.25rem;
+`;
+
+export default class Labelled extends React.PureComponent<Props, State> {
```
Tbaut: Should be `LabbelHelp` I guess. I wonder how this can work without warning as it.

jacogr:
Yes, agreed. So couple of things -

- One major issue is the (?)'s - we should move those to the help and have the explanation there. After you know what it means, it is just noise. Maybe then with keeping the help on hover, but over the label?

- Then the copy and edit buttons... there is a reason why the damn copy was removed in the first place. (I keep on calling it "damn copy"...) Need some ideas around those where it is obvious, but not too much.

packages/ui-app/src/LabelHelp.tsx
jacogr: Ahh, doesn't matter what the class is named for use. So we do

`export { default as LabelHelp } from './LabelHelp'` - so it is actually called "default" is export.

But yes, so it works whatever it is called, but should be changed. Copy error there.

Tbaut:
> After you know what it means, it is just noise.

Agreed, with a greeting welcome message that talks about the (?) in the top right hand corner, this help overlay could become more used. We should reduce its content to the bare minimum though.

> Maybe then with keeping the help on hover, but over the label?

Either it shows a tooltip instantly (1), and it can surely become a little annoying as the balances are all over the place and those tooltip will popup as the mouse flies above. Or we add a delay before showing (2) them and users may totally miss the functionality altogether. 

Not sure if the (1) would really be this annoying in practice, we should try.

> Need some ideas around those where it is obvious, but not too much.

Let me insist on the fact that this type of behaviour isn't new to most users, and they naturally go with the mouse on the address if they want to see it fully (we can still do it on the icon as well it can't hurt). We should try something along the line of what I proposed in https://github.com/polkadot-js/apps/issues/1176 or something like 
image

jacogr:
So want this in as-is, then the follow-ups as well -

- tiny copy (consistent with edit as actions)
- move of balance info to help on accounts (next PR)
- all AddressSummary - AddressRow (eg. creates, next PR)

PR #1211, opened 2019-05-26
Tiny copy button (same as edit, full-row action)

jacogr:
As per https://github.com/polkadot-js/apps/pull/1210#issuecomment-496028122


Tbaut:
I'd suggest to keep the same cursor with the + on the copy button, for consistency.

jacogr:
Yes, it is the same, the ‘copy’ cursor

PR #1212, opened 2019-05-27
Swap AddressSummary for AddressRow

jacogr:
- As mentioned in https://github.com/polkadot-js/apps/pull/1210/#issuecomment-496079805
- Next step would be to remove Summary altogether and pull the styles into AddressRow (finally making it clean)


jacogr:
The transfer app is problematic, the sooner it goes, the better, i.e. https://github.com/polkadot-js/apps/issues/1208

jacogr:
Yes. It is _very_ weird on the left as well. So this is weird, but looks slightly better. ("slightly" being the operative word here)

PR #1224, opened 2019-05-27
Swap Addresses to grid layout (as in accounts)

jacogr:
Closes https://github.com/polkadot-js/apps/issues/1168


jacogr:
As cleanups, followups -

- AddressGrid that has the outer styles
- AddressCard that has the inner styles

(as shared between Account & Address components)

Issue #1237, opened 2019-05-29
Rework Locked, Unlocking and Redeemable


Tbaut:
After https://github.com/polkadot-js/api/pull/959 gets in, we should be able to separate cleanly  redeemable and unlocking out of `Unlocking` component

We need to have a labels when unlocking, now it's missing:
![image](https://user-images.githubusercontent.com/33178835/58557817-96f8de00-821f-11e9-81e5-7e80fcd661ca.png)

and prevent this bug:
![image](https://user-images.githubusercontent.com/33178835/58557845-a710bd80-821f-11e9-9d52-0feafbe800bd.png)

Issue #1232, opened 2019-05-29
Combine AddressSummary/Row


jacogr:
Goal:

Less code to maintain, cleanup of existing code.

Background:

Historically we had AddressRow & AddressSummary to display addresses. Recently these have all be moved to AddressRow. Additionally, for consistent Card-like styling, we have Card & CardGrid.

We are no finally at the point where we have a good understanding of the various elements that are needed for these Cards, be it for -

- Modals - here we are using the simple AddressRow without any additional information, ie. only icon, name & address/index
- Staking overview - Basic AddressRow with simple bonded balance, with children for the actual content
- Accounts/Addresses/Contracts/Staking - AddressRow with buttons, children (AddressInfo + custom per view) as well as external links, e.g. explorer

Actions:

Just typing these out, no particular order -

- [x] AddressRow to render balances via AddressInfo, i.e. non-grid (basically, remove the custom rendering logic for these, adjust with styles in AddressInfo to get the required as-current effect), https://github.com/polkadot-js/apps/pull/1259
- [x] Combine AddressRow & AddressSummary - we only need 1 component, logic from Summary, styles from Row, https://github.com/polkadot-js/apps/pull/1260
- [x] Move styles here away from components.css - basically from a layout perspective, it is move effective just to create new component, copy style in - and importantly, check overrides in any apps, https://github.com/polkadot-js/apps/pull/1260
- [x] Specialized AddressCard that just cleans up slightly where things are used, i.e. pass <Card><AddressRow {...props} /></Card> (We need the specialized small-view, i.e. for modals, so cannot just have a card - this is where the link display comes into play with positioning), https://github.com/polkadot-js/apps/pull/1269
- [x] Get rid of the absolute positioning for the button bar and well as links. These need to stay top/bottom as existing. (No matter what the size is). This also addresses the current overlaps we have in resizing (links done in #1269, header in #1270)
- [x] Ensure RecentlyOffline display works as intended - this also uses absolute positioning, this can be combined into IdentityIcon with a overlay prop, instead of the global absolute style it has atm, done in https://github.com/polkadot-js/apps/pull/1276

Not going to do this, since there are no better ideas, 

- Editing needs to use more "standard" components, at least for name - it is not perfect with the shrunken sizes, all input boxes need to be a standard size. Basically, after the flex changes, at least the input is a sane size (taking up the space), trying to make the full height may just be more effort than it is worth. (For now...)

Next-up after the contract re-work is in:

[...]

PR #1239, opened 2019-05-29
Use derive staking info

Tbaut:
- Closes https://github.com/polkadot-js/apps/issues/1237
- Use the new derive.staking.info
- Remove Unlocking component
- Move the redeem button styling to `AddressInfo`
- Duplicate `Bond` (using derive.staking.info). Too deeply integrated with divers Address* to be removed here. Can be part of https://github.com/polkadot-js/apps/issues/1232


packages/ui-app/src/AddressInfo.tsx
```diff
@@ -172,7 +172,7 @@ class AddressInfo extends React.PureComponent<Props> {
             data-for={`controlled-trigger${index}`}
           />
           <Tooltip
-            text={`${remainingBlocks} blocks left`}
+            text={t(`${remainingBlocks} blocks left`)}
```
jacogr: Should be `t('{{remainingBlocks}} blocks left', { replace: { remainingBlocks } })` - yes, I know copied, however the ui-reactive stuff doesn't have translation.

packages/ui-app/src/AddressInfo.tsx
```diff
@@ -79,19 +79,28 @@ class AddressInfo extends React.PureComponent<Props> {
         {balanceDisplay.bonded && (
           <>
             <Label label={t('bonded')} />
-            <Bonded
-              className='result'
-              params={value}
-            />
+            <div className='result'>
+                {staking_info && staking_info.stakingLedger
+                ? formatBalance(staking_info.stakingLedger.active)
```
jacogr: Funky indent here, extra 2 spaces before `?` and `:`

packages/ui-app/src/AddressInfo.tsx
```diff
@@ -130,6 +139,46 @@ class AddressInfo extends React.PureComponent<Props> {
       </div>
     );
   }
+
+  private renderRedeemButton () {
+    const { staking_info, t } = this.props;
+
+    return (staking_info && staking_info.controllerId && <TxButton
```
jacogr: Would read better as 

```js
return (staking_info && staking_info.controllerId && (
  <TxButton
    accountId={staking_info.controllerId.toString()}
    className='iconButton'
    icon='lock'
    size='small'
    isPrimary
    key='unlock'
    params={[]}
    tooltip={t('Redeem these funds')}
    tx='staking.withdrawUnbonded'
  />
));
```

packages/ui-app/src/AddressInfo.tsx
```diff
@@ -130,6 +139,46 @@ class AddressInfo extends React.PureComponent<Props> {
...
+    return (
+      staking_info &&
+      staking_info.unlocking &&
+      staking_info.unlocking.map(({ remainingBlocks, value }, index) => (
+        <div key={index}>
+          {formatBalance(value)}
+          <Icon
+            name='info circle'
+            data-tip
+            data-for={`controlled-trigger${index}`}
```
jacogr: Maybe just `unlocking-trigger-${index}` (I believe we use `controlled-trigger-` in places, so this would just be slightly safer?

packages/ui-app/src/AddressInfo.tsx
Tbaut: Oops, missed that one :)

PR #1247, opened 2019-05-30
Display extrinsic meta documentation on hover

### LINKED ISSUE #969: Display extrinsic meta documentation on hover
Goal:

To make specific extrinsic information available at the fingertips of the user, as opposed to referencing more external documentation.

Background:

Currently the metadata returns in-code documentation that expands on the methods for both the extrinsics and chain state. This information is used to display a very condensed summary in the extrinsic selection, i.e.

image

We wish to expand this to actually show all information.

Actions:

- Expand the `submit the following extrinsic` label (or any label passed for an Extrinsic/Proposal to contain `help` with the actual documentation.
- As the method selection changes, the updated documentation becomes available as part of the hover. (This is actually better suited to the dropdown itself, and _may_ work as a tooltip - but these things are large)

Bonus:

Also apply to the chain state app

pbteja1998:
Resolves #969 

---

### Extrinsic

Screenshot from 2019-05-31 00-31-37 (1) (1)

Screenshot from 2019-05-31 14-22-56 (1)

---

### Chain State
Screenshot from 2019-05-31 00-40-49 (1)

---

### DIFF
--- a/packages/app-storage/src/Selection/Modules.tsx
+++ b/packages/app-storage/src/Selection/Modules.tsx
@@ -46,7 +46,7 @@ class Modules extends TxComponent<Props, State> {
 
   render () {
     const { t } = this.props;
-    const { isValid, key: { method, section }, params } = this.state;
+    const { isValid, key: { method, section, meta }, params } = this.state;
 
     return (
       <section className='storage--actionrow'>
@@ -55,6 +55,11 @@ class Modules extends TxComponent<Props, State> {
             defaultValue={this.defaultValue}
             label={t('selected state query')}
             onChange={this.onChangeKey}
+            help={
+              t(meta && meta.documentation
+                ? meta.documentation.join(' ') : ''
+              )
+            }
           />
           <Params
             key={`${section}.${method}:params` /* force re-render on change */}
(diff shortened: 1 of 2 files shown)


pbteja1998:
@jacogr Other than these two, any thing else needs to be changed?

packages/app-storage/src/Selection/Modules.tsx
```diff
@@ -55,6 +55,11 @@ class Modules extends TxComponent<Props, State> {
             defaultValue={this.defaultValue}
             label={t('selected state query')}
             onChange={this.onChangeKey}
+            help={
+              t(meta && meta.documentation
```
jacogr: We don't need to add the `t` wrapper around here (these come directly from the node, so is not translatable), and would suggest simply -

`help={meta && meta.documentation && meta.documentation.join(' ')}`

packages/ui-app/src/Extrinsic.tsx
```diff
@@ -61,6 +61,11 @@ class ExtrinsicDisplay extends React.PureComponent<Props, State> {
           label={label}
           onChange={this.onChangeMethod}
           withLabel={withLabel}
+          help={
+            meta && meta.documentation
```
jacogr: As per above comment around just using `&&`

pbteja1998:
@jacogr Made the requested changes.

Issue #1245, opened 2019-05-30
FF has incorrect trimmed explorer headers


jacogr:
![image](https://user-images.githubusercontent.com/1424473/58645232-38a52b80-8303-11e9-8052-03f0891b2471.png)

PR #1248, opened 2019-05-31
Fix ellipsis bug with FF

Tbaut:
closes https://github.com/polkadot-js/apps/issues/1245
div -> span does the trick.

edit: new screenshot with align-middle
image

regarding closing an issue == opening a new one, the research by block field is in the way :)


jacogr:
Yes, not sure wtf to do with the upper stuff - it uses the nasty absolute again, so probably just hide it on small screens.

Issue #1246, opened 2019-05-30
Used availableBalance in fee calculations (not free)


jacogr:
The available shows what is actually useful for fee subtraction, i.e. spendable

PR #1258, opened 2019-05-31
Use balances.all when possible and show existential deposit

Tbaut:
closes https://github.com/polkadot-js/apps/issues/1246

Alexander, recipient too low
image

Substrate node, dropping account
image

### DIFF
--- a/packages/app-accounts/src/modals/Transfer.tsx
+++ b/packages/app-accounts/src/modals/Transfer.tsx
@@ -21,7 +21,6 @@ import translate from '../translate';
 
 type Props = ApiProps & I18nProps & {
   balances_fees?: DerivedFees,
-  balances_votingBalance?: DerivedBalances,
   onClose: () => void,
   recipientId?: string,
   senderId?: string,
--- a/packages/app-accounts/src/modals/Transfer.tsx
+++ b/packages/app-accounts/src/modals/Transfer.tsx
@@ -3,7 +3,7 @@
 // of the Apache-2.0 license. See the LICENSE file for details.
 
 import { ApiProps } from '@polkadot/ui-api/types';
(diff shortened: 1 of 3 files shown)


packages/app-accounts/src/modals/Transfer.tsx
```diff
@@ -21,7 +21,6 @@ import translate from '../translate';
 
 type Props = ApiProps & I18nProps & {
   balances_fees?: DerivedFees,
-  balances_votingBalance?: DerivedBalances,
```
Tbaut: This wasn't used it seems

packages/ui-signer/src/Checks/Transfer.tsx
```diff
@@ -38,16 +38,16 @@ export class Transfer extends React.PureComponent<Props, State> {
...
 
-    if (balances_votingBalance.votingBalance.isZero()) {
+    if (balances_all.votingBalance.isZero()) {
       extraFees = extraFees.add(fees.creationFee);
     }
 
     const extraAmount = amount instanceof Compact ? amount.toBn() : new BN(amount);
-    const isCreation = balances_votingBalance.votingBalance.isZero() && fees.creationFee.gtn(0);
-    const isNoEffect = extraAmount.add(balances_votingBalance.votingBalance).lte(fees.existentialDeposit);
+    const isCreation = balances_all.votingBalance.isZero() && fees.creationFee.gtn(0);
```
jacogr: Can we use `balances.all.availableBalance`

jacogr: No, this is correct, my bad. Sorry, r-e-c-i-p-i-e-n-t...

PR #1259, opened 2019-05-31
Render AddressRow balances from AddressInfo

jacogr:
A small part of https://github.com/polkadot-js/apps/issues/1232

Polkadot:Substrate Portal 2019-05-31 18-01-53

### DIFF
--- a/packages/ui-app/src/AddressInfo.tsx
+++ b/packages/ui-app/src/AddressInfo.tsx
@@ -5,6 +5,7 @@
 import { DerivedBalances, DerivedStaking } from '@polkadot/api-derive/types';
 import { BareProps, I18nProps } from './types';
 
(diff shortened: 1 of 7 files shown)


packages/ui-app/src/AddressInfo.tsx
```diff
@@ -103,6 +105,19 @@ class AddressInfo extends React.PureComponent<Props> {
     );
   }
 
+  private renderBonded (bonded: true | Array<BN>, stakingLedger: StakingLedger) {
```
Tbaut: I'm intrigued by this function. In which use case would we receive an `Array<BN>`? The next `if` also made me ask "what does `bonded` really mean here, if it's true"

It'd be great to have a couple comments inline

Issue #1263, opened 2019-06-02
Allow direct link to account creation (Account overview + creation modal)


Tbaut:
As per https://github.com/polkadot-js/apps/pull/1261#pullrequestreview-244627827
It would be handy to link to the account overview and trigger the account modal if needed so that a new user with no account could be linked directly to the account creation modal

PR #1266, opened 2019-06-02
Direct link to account creation

Tbaut:
- closes https://github.com/polkadot-js/apps/issues/1263


packages/ui-app/src/types.ts
```diff
@@ -5,6 +5,7 @@
 import { WithTranslation } from 'react-i18next';
 import { RpcRxInterface } from '@polkadot/rpc-rx/types';
 import { ActionStatus } from '@polkadot/ui-app/Status/types';
+import { LocationDescriptorObject } from 'history';
```
Tbaut: This looks totally wrong. I wonder how to cleanly allow the `location` prop to have a `state`

packages/app-accounts/src/Overview.tsx
```diff
@@ -26,10 +26,16 @@ type State = {
 };
 
 class Overview extends React.PureComponent<Props, State> {
-  state: State = {
-    isCreateOpen: false,
-    isImportOpen: false
-  };
+  constructor (props: Props) {
+    super(props);
+    const { state = {} } = this.props.location;
```
jacogr: (nit) Blank line before.

Then, can do this as `const { state: { isCreateOpen } = {} } = props;`

(Removing the line below)

packages/app-accounts/src/Overview.tsx
```diff
@@ -26,10 +26,16 @@ type State = {
...
+    const { isCreateOpen } = state;
+
+    this.state = {
+      isCreateOpen: isCreateOpen,
```
jacogr: isCreateOpen may be undefined, so should rather be `isCreateOpen: !!isCreateOpen`

packages/ui-app/src/types.ts
jacogr: Not sure either. (And couldn't find something by a quick 2 min search). Maybe just comment, we are going to wonder about this 2 days from now...

packages/app-accounts/src/Overview.tsx
Tbaut: I guess `const { state : { isCreateOpen = false } = {} } = this.props.location;` works (and avoids the need to `check isCreateOpen` for undefined. Your suggestion bothered TS because `isCreateOpen` wasn't initialized.

jacogr: The below comment would have fixed that, i.e. `!!isCreateOpen`, however the `= false` also works.

Tbaut: TS was complaining at this exact line `const { state: { isCreateOpen } = {} } = props` so that I had to set a default here.

Issue #958, opened 2019-04-10
Staking process and overview ideas


Tbaut:
I thought about a step by step process that would let users validate/nominate following a simple process, with guidance thanks to all the (?) along the way, and with a limited amount of signing modals. The goal is to :
1. Setup all you need to validate/nominate in a couple steps and sign 1 extrinsic (allow me to dream :drooling_face: ).
1. Present in the `Account Actions` only the accounts involved, and group them.
1. Have a few call to actions to start validating/nominating right away.

There are limitations currently on the API side e.g the inability to set the controller, session key, bounded amount, reward destination and commission, all in 1 extrinsic). ping @kianenigma
There might be other technical limitations, let me know.

I want to first see what you think, know if that's something we could eventually do, before getting out and validate the ideas or even get some more by talking to the validators.

Few call to actions to start with and an empty page:
![image](https://user-images.githubusercontent.com/33178835/55909948-7fb55100-5bdd-11e9-83d8-21d92b33d15e.png)

**Validation setup step 1/3:**
Start with the `Stash`, the bounding amount and the `Session` account.
![image](https://user-images.githubusercontent.com/33178835/55910033-b1c6b300-5bdd-11e9-8b0e-61a180b04f7e.png)
**Validation setup step 2/3:**
Setup the `Controller` and all the params attached to it. Note that renaming the "payment preferences" into "commission" and expressing it in percentage (with a slider) seems to me easier to grasp.
![image](https://user-images.githubusercontent.com/33178835/55910102-dd499d80-5bdd-11e9-845a-5c9d15f9b7c3.png)

**Validation example step 3/3:**
Overview, no more (?) as everything should have been covered before.
![image](https://user-images.githubusercontent.com/33178835/55910128-edfa1380-5bdd-11e9-91ec-f05aa3d9bbc3.png)

**Accounts Actions overview:**
Group the accounts together without size difference. Show all info.
Not represented on this screen, users could edit any param by clicking on an edit button that would appear a mouse over next to each parameter or account. e.g change controller/session/bounded/commission... Depending on the parameter to edit, the user would have to sign with the `stash` or `controller` account.
![image](https://user-images.githubusercontent.com/33178835/55910848-affdef00-5bdf-11e9-8292-8f6ed6635c88.png)

Additional ideas:
- On the validation/nomination setup, show at the bottom of the modal an indication that the user is at step x/3
[...]

PR #1268, opened 2019-06-03
Redo staking process and overview

Tbaut:
- closes https://github.com/polkadot-js/apps/issues/958

- [x] removes filters on Staking>Actions (probably fits better in Accounts>Overview now)
- [x] adds "New Stake" button to select a stash and controller (ex `Bonded` modal)
- [x] moves the "max balance" logic to a dedicated `InputBalanceBonded` component. it's used in 2 modals so far.
- [x] edit bond button allow users to set a new bonded amount. Depending on if it's positive or negative we will `unbond` or `bondExtra`.
- [x] working "validate" button/modal
- [x] working "nominate" button/modal
- [x] explain to users what will happen when changing bonded funds
- [x] show validator prefs
- [x] edit validator prefs
- [x] show nominee nicely
- [x] allow session change
- [x] allow nominee change
- [x] allow controller change


Tbaut:
how things look like so far, here is an example with both `bondExtra` and `unbond`:
bondExtra

Tbaut:
If you have some time to check @jacogr at this point it's almost on paar feature-wise with what we have.
Styling the nominee is missing, the rest (explain to users what will happen when changing bonded funds show, validator prefs, edit validator prefs) can be added in separate PRs.

jacogr:
Cool, will take a peek.

jacogr:
Overall comments -

- The little edits for withdraw/bond additional is really not obvious - at all. Funds are normally not editable, so having these showing as "editable" aligns with what is done everywhere, but feels wrong and is unexpected
- I suppose edit for the preferences are ok, since it it more aligned with an actual edit.
- Overall all these little button just gets lost since they sit at the right. Even as "clickable links", they would work better. (Buttons would have other issues, the icons have the current issues)

Not quite sold on this -

image

Even after reading, I'm still confused as to what I'm actually doing here. I would probably suggest being really explicit -

- I want to with draw x (this happens from the bonded)
- I want to add x (this happens from the available)

Tbaut:
>There is nothing to show me progress, there is nothing to show me the next step.

I'll make it more look like a process

> The new stake button is very weird - not sure if it is the size or the lack of spacing, but something looks very off compared to accounts/addresses

Agreed, the spacing is off, it's also bigger.

> The little edits for withdraw/bond additional is really not obvious

Agreed. Would you see an "edit" clickable link then? That'd look bad IMO and also be weird compared to the edit icon shown elsewhere.

There's also the question of: how do we allow to change the controller, the nominees.. This could also have ended up being shown next to "controller" / "nominee" but would be off as well. I wonder if we shouldn't have a general dropdown to simply list all the big editables such as:
- editBonding
- changeController
- changeSession
- changeNomminees
- change Validator preferences

Tbaut:
would end up looking something like: 
image

What do you think?

jacogr:
The above edit would work. Just not sure if it should be "edit icon" vs "some kinds of settings icon" - but that is a single word change.

Tbaut:
> Which I know you seem to like... so
      withdraw bonded
      add to bonded (naming for both these are off)

Hehe I'm actually not too attached to it, I could feel how weird it was and looking for something more natural with what I had, but this setting/edit menu seems to be the right thing to get rid of it :)

jacogr:
I understand that, these things are tough to get right. But at the very last each iteration makes it better and moves it forward.

Tbaut:
I should be better regarding:
- AddressMini being off
- the Validating process should feel more like a process with "Step" and "next".. well with signing in the middle :)
- the setting button is doing its job pretty well I think
- New stake button has the default spacing and size.. but it's not ideal IMO. We can change that once the changes from Keith are in master.

Tbaut:
I've failed to be able to make the defaultValue work for the multi imputs in `Nominate`. I didn't manage to modify `InputAddress` to make it work well. I've left it in a working state without defaultValue to be effectively preselected.

Tbaut:
stil a couple `console.log` left here and there to help debug [the problem](https://github.com/polkadot-js/apps/pull/1268#issuecomment-500322117) that I couldn't fix yet. Other than that the functionalities should be there and working without being too ugly hopefully..

jacogr:
Small comment - the available is quite important, for both the controller and stash. (And quite horrible to click and then realise you need funds.) So would probably lean towards adding available for each of the one on the left and removing it from the right. (On the right not really sure, at first glance, what it relates to - the bonded and withdrawable ones are fine there, since they are "actionable")

packages/app-accounts/src/Account.tsx
```diff
@@ -60,9 +60,9 @@ class Account extends React.PureComponent<Props> {
       >
         {this.renderModals()}
         <AddressInfo
+          accountId={address}
```
jacogr: This is probably `address`, rather than `accountId`?

packages/app-staking/src/Actions/Account/SetSessionAccount.tsx
```diff
@@ -44,61 +45,61 @@ class Key extends React.PureComponent<Props, State> {
...
-            isDisabled={!sessionId || !!sessionError}
-            isPrimary
-            label={t('Set Session Key')}
-            onClick={onClose}
-            params={[sessionId]}
-            tx='session.setKey'
-          />
-        </Button.Group>
-      </Modal.Actions>
+          <Button.Group>
```
jacogr: I have tried to get the linter to complain about these - maybe it will go better once we are on eslint...

packages/app-staking/src/Actions/Accounts.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+  render () {
+    const { recentlyOffline, t } = this.props;
+    const accounts = keyring.getAccounts();
+    const stashOptions = this.getStashOptions();
+
+    return (
+      <Wrapper
+        buttons={
+            <Button
```
jacogr: eagle-eyed spot of 4 spaces as opposed to 2 here

packages/ui-app/src/AddressInfo.tsx
```diff
@@ -30,19 +30,26 @@ export type CryptoActiveType = {
...
+  withRewardDestination?: boolean,
+  withExtended?: boolean | CryptoActiveType,
+  withValidatorPrefs?: boolean | ValidatorPrefsType
 };
 
 // <AddressInfo
 //   withBalance // default
 //   withExtended={true} // optional
-//   value={address}
+//   accountId={address}
```
jacogr: Suppose we can remove all these commented-out code?

packages/app-staking/src/Actions/Accounts.tsx
Tbaut: you really have eagle eyes!

Tbaut:
It should be good again now, I pulled in https://github.com/polkadot-js/apps/pull/1291

PR #1280, opened 2019-06-05
Fix AddressRow cut

Tbaut:
There's no such thing as a too small PR.
Fixes that (visible in modals):
image


jacogr:
Ummm.... I am sure I checked those, obviously now. Thanks.

Tbaut:
Right, looks good thanks to the input + label that don't shrink but it'd be better at a lower level just in case:

image

PR #1284, opened 2019-06-06
Adjust AddressRow-buttons right margin

jacogr:
First item fixed, second one the old version. Just make stuff consistent.

image


Tbaut:
A little bit too "at the edge" to my taste, I'd rather have something like `margin: 0 -.5rem 0 0;` to have them aligned with the address and the polkascan link:
image

PR #1287, opened 2019-06-08
Bump api

Tbaut:
Pull in the reward destination from `derive.staking.info`


Tbaut:
I missed the fact that it's in https://github.com/polkadot-js/apps/pull/1286/files

PR #1292, opened 2019-06-13
Fix nits from previous PRs

Tbaut:
- Show names on staking overview used to show `VALIDATOR (STASH)` even for known accounts.
- Account address in lowercase for "account or names"
- fix account cards off
- Selection wasn't showing in changeRewardDestination modal.
- avoid the edit or copy icon to get wrapped alone eg (before/after) :
image

image


jacogr:
Not sure about the first one at all -

image

Tbaut:
Do you mean the second address of your screen shot (5DqAz..) is actually part of your address book or accounts? I doubled checked and it shows as expected for me also on Alex. e.g a known address, my account, and finally an unknown one

image

jacogr:
The first one (this is on Alex) is a validator I added to my addressbook. But now that you mention it, it does not show the value from accounts, i.e. my `First validator` shows up, however `Alice_stash` does not.

packages/ui-app/src/Row.tsx
```diff
@@ -109,14 +109,18 @@ export const styles = `
...
     box-sizing: border-box;
     margin: 0;
     padding: 0;
     overflow: hidden;
     text-overflow: ellipsis;
     text-transform: uppercase;
     white-space: normal;
+
+    .withAddr {
+      text-transform: lowercase;
```
jacogr: This is not correct - addresses have uppercase letters as well.

jacogr: For AddressMini, this is what was done - https://github.com/polkadot-js/apps/blob/master/packages/ui-app/src/AddressMini.tsx#L137

Apply the text-transform to `withName` only, so `withAddr` has just the default.

Tbaut:
Thanks, I missed the one with explorer link. Had a look around and found similar problems in council where the name would have been displayed only for address book contact but not your own accounts.

jacogr:
Thank you, wanted to check those all - think we went a bit heavy-handed on applying the type.

Tbaut:
I'll stop here for this PR to keep it small

packages/ui-app/src/CopyButton.tsx
```diff
@@ -85,4 +87,8 @@ export default styled(CopyButton)`
   button.ui.mini.icon.primary.button.iconButton {
     cursor: copy;
   }
+
+  .copySpan {
```
jacogr: Does this really make a difference? The span is basically around a single element? (A button with only an icon, which should not be breakable)

packages/ui-app/src/Row.tsx
```diff
@@ -41,6 +41,9 @@ export const styles = `
     z-index: 1;
   }
 
+  .editSpan {
+    white-space: nowrap;  }
```
jacogr: (nit) misplaced `}` at the end.

packages/ui-app/src/CopyButton.tsx
Tbaut: This is black magic.. Found the trick *on the internet* and it does work on both chrome and FF :man_shrugging:

PR #1299, opened 2019-06-16
Add checks for contract fees

kwingram25:
Still unclear how fees are calculated, reached out to Sergei for help!

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,6 @@
   ],
   "resolutions": {
     "@polkadot/api": "^0.82.0-beta.6",
-    "@polkadot/api-contract": "^0.82.0-beta.6",
     "@polkadot/keyring": "^0.94.0-beta.1",
     "@polkadot/types": "^0.82.0-beta.6",
     "@polkadot/util": "^0.94.0-beta.1",
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
   ],
   "resolutions": {
     "@polkadot/api": "^0.82.0-beta.6",
+    "@polkadot/api-contract": "^0.82.0-beta.6",
     "@polkadot/keyring": "^0.94.0-beta.1",
     "@polkadot/types": "^0.82.0-beta.6",
     "@polkadot/util": "^0.94.0-beta.1",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -11,7 +11,6 @@
   ],
   "resolutions": {
     "@polkadot/api": "^0.82.0-beta.6",
-    "@polkadot/api-contract": "^0.82.0-beta.6",
```
jacogr: We should just re-add this.

packages/ui-signer/src/Checks/Call.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+type Props = I18nProps & {
+  endowment: BN | Compact,
+  fees: DerivedContractFees,
+  onChange: (fees: ExtraFees) => void
+};
+
+type State = ExtraFees & {};
+
+export class Call extends React.PureComponent<Props, State> {
```
jacogr: Would probably make more sense as ContractCall? (Call can also mean, well, `call` :))

packages/ui-signer/src/Checks/Call.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+    return update;
+  }
+
+  render () {
+    return null;
+  }
+}
+
+export default translate(Call);
```
jacogr: We can remove the translate wrapper, no text here.

packages/ui-signer/src/Checks/Deploy.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+type Props = I18nProps & {
+  endowment: BN | Compact,
+  fees: DerivedContractFees,
+  onChange: (fees: ExtraFees) => void
+};
+
+type State = ExtraFees & {};
+
+export class Deploy extends React.PureComponent<Props, State> {
```
jacogr: ContractDeploy? (Just to align with Call)

packages/ui-signer/src/Checks/Deploy.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+    return update;
+  }
+
+  render () {
+    return null;
+  }
+}
+
+export default translate(Deploy);
```
jacogr: As above, redundant.

package.json
kwingram25: It's not included yet as part of /api?

kwingram25:
Fees are correct afaik!

PR #1304, opened 2019-06-18
Use empty props from CardGrid

Tbaut:
- [x]  address book
- [x]  accounts
- [x]  staking actions

image
image
image


packages/ui-app/src/CardGrid.tsx
```diff
@@ -13,14 +13,16 @@ type Props = I18nProps & {
...
+  isEmpty?: boolean
 };
 
 class CardGrid extends React.PureComponent<Props> {
   render () {
-    const { buttons, children, className } = this.props;
+    const { buttons, children, className, isEmpty = false } = this.props;
 
-    if (!children || (children as Array<any>).length <= 0) {
+    console.log('children',children);
```
Tbaut: For staking<actions the children is never empty as long as we have account, and the filtering whether something is shown or not is done after. 

The only way I thought of to make this work would be to filter the shown accounts upstream, this means bloating a little `/app-staking/src/index.tsx` with `derive.staking`. Any better idea?

jacogr: Can filter and always show, ie. If none pass empty array or null? Have not looked in detail.

Tbaut: yes that's what I meant with "filtering upstream". I'll do that then.

Tbaut:
I had to do convoluted things in `Actions/Accounts.tsx` . Mapping all accounts (props) and call `api-staking.bonded` to pass the info in the `state`. My brain could have been burnt in the process and I wonder if you can think about something better.

I forgot to use the `wip` flag. This PR is now ready to review.

packages/app-staking/src/Actions/Accounts.tsx
```diff
@@ -31,13 +33,40 @@ const Wrapper = styled(CardGrid) `
 
 class Accounts extends React.PureComponent<Props,State> {
   state: State = {
-    isNewStakeOpen: false
+    isNewStakeOpen: false,
+    myStashes: []
   };
 
+  async componentWillReceiveProps ({ allAccounts, api }: Props, { isNewStakeOpen }: State) {
```
jacogr: Yes, I hate it. The issue here is that -

- we can certainly simplify
- this will be done each time it is called (put a console.log in here, it will probably be ugly with the number of times called)

So short-term -

- you could store the allAccounts keys in state as well, checking first if it has changed before calling
- def don't make this async, rather in the `.then` set the state
- at the very least, needs a multi call instead of single calls

The correct approach probably would be -

- add something like a `pickProps(props: Props)` function alongside the `transform` in api
- use this to transform the keys
- make a multi-call

Tbaut:
Should look a little better now :)

kwingram25:
Nice job! Was meaning to do this ha

packages/app-staking/src/Actions/Accounts.tsx
```diff
@@ -2,36 +2,73 @@
...
-    isNewStakeOpen: false
+    allAccounts: undefined,
+    isNewStakeOpen: false,
+    myStashes: []
   };
 
+  componentWillReceiveProps ({ allAccounts, api }: Props) {
+    const previousState = this.state;
+
+    if (allAccounts && allAccounts !== previousState.allAccounts) {
```
jacogr: This will always evaluate to true - as soon as saved in state, the reference is not the same anymore.

packages/app-address-book/src/Overview.tsx
```diff
@@ -43,6 +44,8 @@ class Overview extends React.PureComponent<Props, State> {
             />
           </Button.Group>
         }
+        isEmpty={ emptyScreen }
```
jacogr: (nit) Remove spaces inside

packages/app-staking/src/Actions/Accounts.tsx
```diff
@@ -2,36 +2,73 @@
...
+        return result;
+      })
+      .then((myStashes) => {
+        this.setState({
+          isNewStakeOpen: previousState.isNewStakeOpen,
+          allAccounts,
+          myStashes
+        });
+      })
+      .catch((e) => {
```
jacogr: Can just be `catch(console.error)`

packages/app-staking/src/Actions/Accounts.tsx
```diff
@@ -2,36 +2,73 @@
...
+    allAccounts: undefined,
+    isNewStakeOpen: false,
+    myStashes: []
   };
 
+  componentWillReceiveProps ({ allAccounts, api }: Props) {
+    const previousState = this.state;
+
+    if (allAccounts && allAccounts !== previousState.allAccounts) {
+      api.query.staking.bonded.multi(Object.keys(allAccounts))
```
jacogr: Some weird indentation slipped in here.

packages/app-staking/src/Actions/Accounts.tsx
```diff
@@ -2,36 +2,73 @@
...
+
+          if (value.toString() !== '') {
+            result.push(Object.keys(allAccounts)[index]);
+          }
+        });
+
+        return result;
+      })
+      .then((myStashes) => {
+        this.setState({
```
jacogr: This could be problematic and make React throw a hissy fit - and sadly we have a coup[le of those that slipped in lately. So basically, imagine -

- component is loaded
- query is made
- component is unloaded
- result comes back, then fires and it tries to set the state... on a non-existent component 
- boom!

packages/app-staking/src/Actions/Accounts.tsx
jacogr: Incorrect, sorry. Scratch that.

kwingram25:
I have just extended both CardGrid and Column to use empty, header, etc. props so this may need to be revisited soon - #1324

Issue #1305, opened 2019-06-18
Use 2-column layout for democracy


jacogr:
Referendums (current) at the left, proposals (future) at the right. This aligns with the rest of the views as well. The only issue for both of these is that the actions would need to be moved, would also suggest making this into a modal, once again, aligning with the rest of the app. (Needs to be logged additionally)

Issue #1306, opened 2019-06-18
Move voting & seconding to modals


jacogr:
Since we are using the modal action approach everywhere now, align democracy.

For proposals - button click opens up seconding modal.
For referendums - button click opens up voting modal.

(Split from #1305)

PR #1309, opened 2019-06-18
Democracy layout changes

jacogr:
- Closes https://github.com/polkadot-js/apps/issues/1305
- It is not great yet, with the actions moved it will de-clutter, i.e. https://github.com/polkadot-js/apps/issues/1306

Polkadot:Substrate Portal 2019-06-18 14-25-42


jacogr:
Ummm. That is a tougher one. Think it is actually existing in the block explorer as well then. (Where params are used to display- the sizing of labels with the boxes are an issue). 

Will have to figure out a way to get Labelled to take the container size (not screen size) into account.

jacogr:
Yes, that specific issue will need to be kicked down the road. With at least #1306 in place, can actually adjust the rest of the items (since they are static only), so it doesn't happen. With the dropdown there though, currently not possible.

So going to see what it ends up like after #1306.

(the alternative is my long-standing dream of reworking the labels, sizes and placements, but that is not for now)

jacogr:
Yeap, as expected, explorer as well - 

image

jacogr:
And when I speak about label layout changes, this is what I have in the back of my mind (quick single-item play-through.) -

image

The only issue is that there will be playing to be done on the actual places where Params with Arrays, Tuples, Structs are used. So the simple case is 3 lines of CSS (and the removal of a bunch of others), the more-involved case, dunno know.

The takes up only slightly more space, but as a bonus gets rid of all the `min-width` settings all-over.

PR #1310, opened 2019-06-18
Small rework of styled (to bottom)

jacogr:
Where these things are simple, remove the `const Wrapper = styled.div'...styles...'` and move to the the end of the files.

- remove `Wrapper` element
- add className to element (div, section, main, etc...)
- add the styles to the end of the file

There are still some left, not major issue, some of these probably need to be split into multiple components, others need some extra classes. But the bulk of the simple-Wrapper-only ones are taken care of here. Just adding some more consistency to the codebase.


jacogr:
@Tbaut Thank you. Did the same in addition to an own review of the code to ensure I didn't miss anything. _Seems_ ok.

PR #1324, opened 2019-06-20
Treasury app

kwingram25:
Starting point for treasury - proposals, settings

I can't reliably set the pot or settings or approve/reject because of bad origins yet so testing is limited. Also need assistance with help messages


packages/app-treasury/src/Overview/Approve.tsx
```diff
@@ -0,0 +1,80 @@
...
+      <div className='democarcy--Proposal-second'>
+        <InputAddress
+          help={t('Select the account you wish to approve or reject the proposal with.')}
+          label={t('using my account')}
+          onChange={this.onChangeAccount}
+          type='account'
+          withLabel
+          hideAddress
+        />
+        <Button.Group>
```
jacogr: Can you look at the new democracy stuff and align with that? (Which is turn aligns with the way the cards look elsewhere).

I don't believe the card layout there is rock-solid, i.e. it is very fit-for-purpose (i.e. it _should_ be a component in ui-app, it is not - basically: header, buttons, content - more a function of the way it was structured than anything else)

packages/app-treasury/src/Overview/index.tsx
```diff
@@ -0,0 +1,26 @@
...
+import Proposals from './Proposals';
+
+type Props = AppProps & BareProps & I18nProps;
+
+export default class Overview extends React.PureComponent<Props> {
+  render () {
+    return (
+      <>
+        <Summary />
+        {/*<Referendums />*/}
```
jacogr: I?sn't there the concept of "stuff to be voted on now"?

If so, once again, would look at the new-ish democracy, it uses a 2 Column layout to do these

packages/app-treasury/src/index.css
```diff
@@ -0,0 +1,79 @@
+/* Copyright 2017-2019 @polkadot/app-democracy authors & contributors
+/* This software may be modified and distributed under the terms
+/* of the Apache-2.0 license. See the LICENSE file for details. */
+
+.treasury--Item-header {
```
jacogr: Would prefer now having new css files added (I understand why, was easiest at that time since democracy is _very_ similar)

packages/app-treasury/src/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+          isRoot: true,
+          name: 'overview',
+          text: t('Treasury overview')
+        },
+        {
+          name: 'propose',
+          text: t('Submit proposal')
+        },
+        {
+          name: 'settings',
```
jacogr: Could just be a button top right? (With modal)

packages/app-treasury/src/Overview/index.tsx
kwingram25: I don't know, it seems like things can be proposed and then approved or rejected so no second phase I don't think?

kwingram25:
Implemented fixes, extended CardGrid, and also added a TxModal component since a lot of this logic is constantly reused for modals w/TxButtons

packages/app-treasury/src/Overview/Approve.tsx
```diff
@@ -0,0 +1,86 @@
...
+  proposalId: string
+};
+
+type State = TxModalState & {
+  isApproving: boolean | null
+};
+
+class Approve extends TxModal<Props, State> {
+
+  private approveOptions = [
```
jacogr: These should be `t` wrapped.

packages/app-treasury/src/Overview/Item.tsx
```diff
@@ -0,0 +1,61 @@
...
+  proposalId: string
+};
+
+class Item extends React.PureComponent<Props> {
+  render () {
+    const { children, className, proposalExtra, proposalId } = this.props;
+
+    return (
+      <Card className={className}>
+        <div className='ui--Row'>
```
jacogr: Yes, this is horrible. However, let's log it as an issue - we need to have some standard container we can use here and in democracy instead of doing what we are doing here (or what we are doing in democracy). Not really re-usable.

jacogr: Logged as https://github.com/polkadot-js/apps/issues/1336

kwingram25:
Addressed changes and added Collection comp to manage both Column and CardGrid since they have similar purpose

PR #1328, opened 2019-06-21
Beauty is in the eye of the beerholder

jacogr:
Experimental.

Do yourself a favor and don't look at the code to start, it is just mostly a bunch of CSS tweaks. 

First pull down, and click around in the app, dialogs, extrinsics, signer, any and all forms, then look at the code.


kwingram25:
Cool, I really like this!

- Some minor problems arranging with contract ABIs
Screenshot from 2019-06-24 16-14-54
- Noticed the label disappears for dropdowns but it looks like a z-index fix
Screenshot from 2019-06-24 16-17-19
- label overlaps when there are two inputs side by side
Screenshot from 2019-06-24 16-18-38

In general i think the label could be a little more distinct in appearance, maybe all uppercase or a distinct color? Now it seems to run together with the text some of the time

jacogr:
The overrun on the linked inputs I saw, not 100% sure what to do there since they are 2, but actually tested as one. 

The contract one will need to look at. I missed a input type adjustment there. 

The drop down was not too worried about when open, but will take a peek at the zindex.

jacogr:
Cleanup up the messages overlaps (and adjusted InputFile somewhat alongside).

Tbaut:
I really like the move from the label inside the input.
Found an overlap with the filter in staking>overview
image

When an input is disabled, the little arrow is hidden and looks a little weird:
image

Although I like the way it's displayed, at first glance I thought the icon being offset was a bug. I got used to it in the mean time, but I think it could help to center the icon (before/after).
image



we could do something to integrate it. An idea from a previous mockup that I found elegant.
image

jacogr:
I played with the integration. The labels not starting at the same place is really hurting my eyes. ;)

jacogr:
centered

image

jacogr:
There are 2 bugs remaining - settings JSON (which is just an f-up) and then the linked lists. (e.g. extrinsic with section & method).

Issue #1336, opened 2019-06-22
Add container for Treasury/Democracy to DRY-up


jacogr:
As mentioned in https://github.com/polkadot-js/apps/pull/1324#discussion_r296439472 and also found in https://github.com/polkadot-js/apps/pull/1338

I think democracy is an issue here, the Item->something indirection makes it difficult to create a card and pass -

- the header
- the buttons
- the children

(Which is what the container should do - it just gets those props and renders it in a consistent fashion)

jacogr:
@kwingram25 I believe we can close this now?

kwingram25:
Added with #1358

PR #1338, opened 2019-06-22
Basic parachains app

jacogr:
- Closes #359

From Alex -

image

There is some more info in 2.x that can be pulled (the queue here is already for 2.x, just displays nothing on 1.x). However, a lot of these can be large. 

This is a start. 

@kwingram25 There may be conflicts with yours since we stuck the new apps in the same position. However, yours probably makes more sense on top of this one, so if both are in, just the ordering of your first? (Not sure, it is starting to be a lot)

And yes, steps into the same holes as democracy and treasury now with the cards - lots and lot of duplication. https://github.com/polkadot-js/apps/issues/1336


jacogr:
Needs https://github.com/polkadot-js/api/issues/1048

jacogr:
Ok, there is a lot of work left since the polkadot master has a very large number of additions (and some cool stuff). So getting this in as a base so we can start logging/adding next-gen stuff.

PR #1346, opened 2019-06-25
Fix upload and deploy contract(s) after rename

ascjones:
In substrate, the `contract` module was renamed to `contracts`. This looks up the record on success by the correct name, should be backwards compatible.

### DIFF
--- a/packages/app-contracts/src/Deploy.tsx
+++ b/packages/app-contracts/src/Deploy.tsx
@@ -11,7 +11,7 @@ import React from 'react';
 import { RouteComponentProps } from 'react-router';
 import { withRouter } from 'react-router-dom';
 import { Abi } from '@polkadot/api-contract';
-import { api, withApi, withMulti } from '@polkadot/ui-api';
+import { api as apiUi, withApi, withMulti } from '@polkadot/ui-api';
 import keyring from '@polkadot/ui-keyring';
 import { Button, Dropdown, InputBalance, TxButton } from '@polkadot/ui-app';
 import { AccountId, getTypeDef } from '@polkadot/types';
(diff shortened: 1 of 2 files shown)


packages/app-contracts/src/Deploy.tsx
```diff
@@ -248,9 +248,10 @@ class Deploy extends ContractModal<Props, State> {
   }
 
   private onSuccess = async (result: SubmittableResult) => {
-    const { history } = this.props;
+    const { api, history } = this.props;
```
Stefie: `api` is already being imported in line 14, that's why the build is failing. I don't know if it was an oversight or on purpose that `api` was not taken from the `props`, but if it's on purpose you can solve it by just renaming the import `import { api as apiUi, withApi, withMulti } from '@polkadot/ui-api';` and then changing these two lines to `apiUi`.

https://github.com/polkadot-js/apps/blob/7dbf43e108f08d017014895d8a6fb86b6ed4227d/packages/app-contracts/src/Deploy.tsx#L259

https://github.com/polkadot-js/apps/blob/7dbf43e108f08d017014895d8a6fb86b6ed4227d/packages/app-contracts/src/Deploy.tsx#L270

ascjones: I've implemented your suggestion, since I don't know whether it was on purpose or not. Is the `api` instance from props the same as the imported one?

Stefie: hmm, should be. Maybe @kwingram25 can clarify, he wrote the contracts app

jacogr: It is a copy gone wrong - it is used in the store since it doesn't have a context/props. https://github.com/polkadot-js/apps/pull/1347

PR #1347, opened 2019-06-25
Remove api imports (only on store, non-context)

jacogr:
cc @kwingram25 Would like your eyes on this.

### DIFF
--- a/packages/app-contracts/src/Deploy.tsx
+++ b/packages/app-contracts/src/Deploy.tsx
@@ -11,7 +11,7 @@ import React from 'react';
 import { RouteComponentProps } from 'react-router';
 import { withRouter } from 'react-router-dom';
 import { Abi } from '@polkadot/api-contract';
-import { api as apiUi, withApi, withMulti } from '@polkadot/ui-api';
+import { withApi, withMulti } from '@polkadot/ui-api';
 import keyring from '@polkadot/ui-keyring';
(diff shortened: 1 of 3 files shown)


packages/app-contracts/src/Deploy.tsx
```diff
@@ -256,8 +256,6 @@ class Deploy extends ContractModal<Props, State> {
     if (record) {
       const address = record.event.data[1] as any as AccountId;
 
-      await apiUi.isReady;
```
jacogr: No need, components are only rendered when API is ready

packages/app-contracts/src/store.ts
```diff
@@ -30,8 +30,6 @@ class Store extends EventEmitter {
   }
 
   async saveCode (codeHash: string | Hash, partial: Partial<CodeJson>) {
-    await api.isReady;
```
jacogr: Not needed here, only needed in loadAll which can happen before we are ready.

packages/app-contracts/src/Deploy.tsx
kwingram25: Got it!

jacogr:
Yes. As soon as the api is ready it gets added. And since the app waits for the api before rendering (well apart from settings), it will be there.

jacogr:
Just the store needs some magic, sometimes - at least in load. Just to make 100% sure there since it is not gated in the same way, so technically can be used from anywhere.

Issue #1332, opened 2019-06-22
Remove constructor info display from deployment


jacogr:
![image](https://user-images.githubusercontent.com/1424473/59961921-32a80200-94df-11e9-875e-32192568a7f0.png)

The screen is quite full already, and it doesn't really add any extra information. Additionally, we could forgo the tags as well (on this screen), I'm willing to bet that in 99.9% of the cases the devs just want to get their code up and will ignore it anyway.

PR #1349, opened 2019-06-25
Remove constructor name display

jacogr:
Closes https://github.com/polkadot-js/apps/issues/1332

However, would like @kwingram25 to weigh in here - inmost cases this is just extra unactionable info on-screen, however where a constructor has params, the "free floating" params now may be out of context? (Which is why the constructor was displayed in that form initially)

TL;DR May be heavy handed, but with the right "clean up, only display relevant info" intent?

### DIFF
--- a/packages/app-contracts/src/Deploy.tsx
+++ b/packages/app-contracts/src/Deploy.tsx
@@ -22,15 +22,12 @@ import Params from './Params';
 import store from './store';
 import translate from './translate';
 
-type ConstructOptions = Array<{key: string, text: string, value: string}>;
-
 type Props = ContractModalProps & ApiProps & I18nProps & RouteComponentProps & {
   codeHash?: string
 };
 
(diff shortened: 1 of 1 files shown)


kwingram25:
Actually yes it is needed some of the time, for example the incrementer needs an initValue param passed

Screenshot from 2019-06-25 15-19-01

jacogr:
Yes 100% - so the question there is if the display is good enough without the extra row or if we need some adjustments. (Params will still be shown)

kwingram25:
How bout show deploy() only when there are arguments to set?

jacogr:
That can work. Still not convinced on display deploy - I added that initially only since I had no better ideas... however still don’t. ;) (But certainly too much on no-args deploy)

jacogr:
Closing, can circle-back if needed. Don't have the gap to look at enhance either.

PR #1350, opened 2019-06-25
Method: don't use globals

axelchalon:
Use the pure function `findByCallIndex` introduced in https://github.com/polkadot-js/api/pull/1047 instead of `findFunction` relying on injected globals.

(Need to update the dependency once the api PR is merged)


jacogr:
Closing. The API interface has been changed around this by passing the registry around, and using the registry for lookups.

PR #1351, opened 2019-06-26
ActionItem component for responding to proposals

kwingram25:
Use new shared component for democracy and treasury proposals - next up for council motions as well


packages/ui-app/src/ActionItem.tsx
```diff
@@ -0,0 +1,90 @@
+// Copyright 2017-2019 @polkadot/app-democracy authors & contributors
+// This software may be modified and distributed under the terms
+// of the Apache-2.0 license. See the LICENSE file for details.
+
+import BN from 'bn.js';
+import React from 'react';
+import styled from 'styled-components';
+import { Method, Proposal } from '@polkadot/types';
```
Tbaut: (nit) types should be placed first and separated by an empty line from external deps.

PR #1354, opened 2019-07-02
Upgrade api to use MetadataV6

amaurymartiny:
Pulls in https://github.com/polkadot-js/api/pull/1059

Works with latest master right after Basti's PR got merged in.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,10 +10,10 @@
     "packages/*"
   ],
   "resolutions": {
-    "@polkadot/api": "^0.82.0-beta.18",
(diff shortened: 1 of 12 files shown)


yjkimjunior:
note on the CI hissy fit - `StorageFunction` was renamed to `StorageEntry` in Basti's PR and hence also renamed in metadata v6.

amaurymartiny:
Yeah, I'll fix that.

@yjkimjunior Btw could you add a BREAKING CHANGE entry in the Changelog in api? I think this is worth mentioning for users

PR #1364, opened 2019-07-05
Add wrong password indication

### LINKED ISSUE #1363: it's not clear that youIf entered a wrong password it would be better to have a mention; "wrong password"
When used one wrong password the UI stay waiting but not mention; "wrong password"

it shouldn't be a big deal to add it. Thanks

In this case after used wrong password just stay in this way a long time.
Screenshot from 2019-07-05 14-27-55

Screenshot from 2019-07-05 14-33-27

Tbaut:
- closes #1363

I didn't want to use the bells and whistles from a class, however I couldn't manage the have the styled component work with `withMulti`.
```js
export default withMulti(
  styled(PasswordCheck)`
    margin-left: 15em;
    color: #9f3a38;
  `,
  translate
);
```
The style wasn't taken into account.. any hint welcome

This is how it looks:
image

### DIFF
--- a/packages/ui-signer/src/Modal.tsx
+++ b/packages/ui-signer/src/Modal.tsx
@@ -177,7 +177,9 @@ class Signer extends React.PureComponent<Props, State> {
           value={currentItem.accountId}
           tabIndex={1}
         />
-        <PasswordCheck unlockError= {unlockError} />
+        <PasswordCheck
+          unlockError={unlockError}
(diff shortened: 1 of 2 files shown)


jacogr:
The reason it doesn’t work - the top-level element needs to pass/add the classMame prop

Tbaut:
Thanks, it's in now :)

PR #1371, opened 2019-07-09
Bonded and available as labelExtra in Staking > bond more funds

### LINKED ISSUE #1368: Swap bonder to labelExtra
Pictures -

This one

image

Should look like this one for the extra labels

image

Tbaut:
closes #1368 

this is how it looks, I find it quite hard to read:
image

### DIFF
--- a/packages/app-staking/src/Actions/Account/BondExtra.tsx
+++ b/packages/app-staking/src/Actions/Account/BondExtra.tsx
@@ -9,7 +9,7 @@ import { CalculateBalanceProps } from '../../types';
 import BN from 'bn.js';
 import React from 'react';
 import styled from 'styled-components';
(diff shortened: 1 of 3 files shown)


jacogr:
Hard to read -

- too much info?
- colors?
- all of the above and more?

Tbaut:
Both the color and the amount of info account for it IMO. Having the length almost match the address also doesn't help since the address is hard to read, but barely needs to be read anyway.
Played a bit, I couldn't get anything better than this (only one balance shown and font 10& less transparent). 
image

jacogr:
Happy with the label color adjustments, it was a bit random. (Just as long as it applied to both the label, the small one and the labelExtra, all good)

Tbaut:
Looks like this now:
image

jacogr:
Well, I like it, CI doesn't. CI seems to win.

PR #1377, opened 2019-07-10
Cleanup linked inputs

jacogr:


jacogr:
@Tbaut Since you are working on the consts, just check the transform here - moved it to class-level, so would be good to align with that in your new components

PR #1379, opened 2019-07-10
Fix for democracy overflow and signing

Tbaut:
Changes look much bigger than they really are, I haven't added any css to the signing component, just alphabetical ordering, usage of `withMulti` and passing `className` as props instead of state because it wasn't taken into account this way.

Bring back this `min-width: 0` to prevent overflow of ActionItem in treasury, democracy and motions.

before:
image
image

after:
image
image
image


jacogr:
Excellent! Thanks a million for this.

packages/app-toolbox/src/Sign.tsx
```diff
@@ -2,19 +2,24 @@
...
 import { KeyringPair } from '@polkadot/keyring/types';
-import { hexToU8a, isHex, stringToU8a, u8aToHex } from '@polkadot/util';
-import React from 'react';
 
+import React from 'react';
 import styled from 'styled-components';
+import { Button , Input, InputAddress, Output, Static } from '@polkadot/ui-app';
+import { hexToU8a, isHex, stringToU8a, u8aToHex } from '@polkadot/util';
+import keyring from '@polkadot/ui-keyring';
+import { withMulti } from '@polkadot/ui-api';
```
jacogr: (nit) ui-api moved up the chain

Issue #1234, opened 2019-05-29
UI mishap due to positioning on extrinsics


jacogr:
![image](https://user-images.githubusercontent.com/1424473/58550612-6c525980-820e-11e9-90f0-d5a339120b63.png)

PR #1393, opened 2019-07-12
Mortal/immortal extrinsics

kwingram25:
Shown on block explorer only (initially)

### DIFF
--- a/packages/app-explorer/src/BlockInfo/Extrinsics.tsx
+++ b/packages/app-explorer/src/BlockInfo/Extrinsics.tsx
@@ -8,11 +8,12 @@ import React from 'react';
 import styled from 'styled-components';
 import { AddressMini, Call, Column, LinkPolkascan } from '@polkadot/ui-app';
 import { formatNumber } from '@polkadot/util';
-import { Extrinsic, Method } from '@polkadot/types';
+import { BlockNumber, Extrinsic, Method } from '@polkadot/types';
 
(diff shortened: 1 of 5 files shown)


jacogr:
yes, yes and yes, brilliant :)

packages/app-explorer/src/BlockInfo/Extrinsics.tsx
```diff
@@ -40,8 +41,14 @@ class Extrinsics extends React.PureComponent<Props> {
 
   // FIXME This is _very_ similar to what we have in democracy/Item
   private renderExtrinsic = (extrinsic: Extrinsic, index: number) => {
+    const { blockNumber, t } = this.props;
     const { meta, method, section } = Method.findFunction(extrinsic.callIndex);
 
+    let eraEnd;
+    if (extrinsic.signature.era.isMortalEra) {
```
jacogr: Not convinced. So the block it shows, it the block it has been included at. So basically, you can send at `#1234` it gets included at `#1239` and now we will show `#1239`, instead of the initial value.

Decode it, and then use the period and phase - https://github.com/polkadot-js/api/blob/master/packages/types/src/type/ExtrinsicEra.ts#L192

jacogr: I was wrong. Looked with @kwingram25 through the code, it seems to be doing the right thing.

Issue #1378, opened 2019-07-10
substrate 2.x, no nextKeyFor


jacogr:
![image](https://user-images.githubusercontent.com/1424473/61003057-479be680-a363-11e9-81bc-b5f63cec70c7.png)

PR #1397, opened 2019-07-13
Staking compatibility with v2

Tbaut:
- one part of https://github.com/polkadot-js/apps/issues/1378 the rest should be sorted with `derive.staking.info` (https://github.com/polkadot-js/api/issues/1113)
- cleaning the balance props that was carried around and never used
- a lot of renaming to make things clearer (IMHO), e.g `stashes` --> `allStashes` (includes the ones currently **not** selected in the validator set) or the monster `currentValidatorsControllersV1OrStashesV2` that represents either the stash (v2) or the controllers (v1) from the validators currently in the validator set. lmk if you think about something better.
- working well on Alex, and almost nothing is shown on v2 (because of the derive)

### DIFF
--- a/packages/app-staking/src/Actions/Account/index.tsx
+++ b/packages/app-staking/src/Actions/Account/index.tsx
@@ -30,7 +30,7 @@ type Props = ApiProps & I18nProps & {
   recentlyOffline: RecentlyOfflineMap,
   balances_all?: DerivedBalances,
   staking_info?: DerivedStaking,
-  stashOptions: Array<KeyringSectionOption>
+  stashOptions: KeyringSectionOption[]
 };
 
 type State = {
@@ -47,7 +47,7 @@ type State = {
   isStashValidating: boolean,
   isUnbondOpen: boolean,
   isValidateOpen: boolean,
-  nominees?: Array<string>,
+  nominees?: string[],
   sessionId: string | null,
   stakers?: Exposure,
   stakingLedger?: StakingLedger,
--- a/packages/app-staking/src/Overview/index.tsx
+++ b/packages/app-staking/src/Overview/index.tsx
@@ -2,7 +2,7 @@
 // This software may be modified and distributed under the terms
 // of the Apache-2.0 license. See the LICENSE file for details.
 
-import { Balance } from '@polkadot/types';
+// import { Balance } from '@polkadot/types';
 import { BareProps } from '@polkadot/ui-app/types';
 import { ComponentProps } from '../types';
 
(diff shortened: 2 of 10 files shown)


packages/app-staking/src/Overview/index.tsx
```diff
@@ -13,22 +12,29 @@ import { formatNumber } from '@polkadot/util';
...
-    const nextSorted = this.sortByBalance(
-      controllers.filter((address) =>
-        !validators.includes(address)
-      )
-    );
-    const validatorsSorted = this.sortByBalance(validators);
+    const { chain_subscribeNewHead, allControllers, allStashes, recentlyOffline, currentValidatorsControllersV1OrStashesV2 } = this.props;
+    let nextSorted: Array<string>;
+
+    if (Object.keys(api.consts).length) {
```
jacogr: I'm wondering if we should add a boolean to `ui-api` for this - so we can just go `api.isSubstrate2` (or something like that - just worried we do this in a number of places)

packages/app-staking/src/types.ts
```diff
@@ -14,11 +14,10 @@ export type Nominators = {
 
 export type ComponentProps = {
   allAccounts?: SubjectInfo,
-  balances?: DerivedBalancesMap,
-  controllers: Array<string>,
-  recentlyOffline: RecentlyOfflineMap,
-  stashes: Array<string>,
-  validators: Array<string>
+  allControllers: Array<string>,
```
jacogr: Just a comment, when we swap to eslint, the typescript-recommended rules (which we use as-is, less bike-shedding all-around), wants Arrays as `string[]` - would be good if we start using that in new code already and get used to it. (Plus, also less to adjust with the eslint PR that is ongoing)

packages/app-staking/src/index.tsx
```diff
@@ -25,18 +24,17 @@ import translate from './translate';
 
 type Props = AppProps & ApiProps & I18nProps & {
   allAccounts?: SubjectInfo,
-  balances?: DerivedBalancesMap,
-  session_validators?: Array<AccountId>,
-  staking_controllers?: [Array<AccountId>, Array<Option<AccountId>>],
+  allStashesAndControllers?: [Array<AccountId>, Array<Option<AccountId>>],
```
jacogr: See the `Array<T>` comment (not a nit, just something we should probably do, but if not, no trainsmash either, it will get done)

packages/app-staking/src/Overview/index.tsx
Tbaut: :+1:

packages/app-staking/src/Actions/Account/index.tsx
```diff
@@ -30,7 +30,7 @@ type Props = ApiProps & I18nProps & {
   recentlyOffline: RecentlyOfflineMap,
   balances_all?: DerivedBalances,
   staking_info?: DerivedStaking,
-  stashOptions: Array<KeyringSectionOption>
+  stashOptions: KeyringSectionOption[]
```
jacogr: +1, thanks :)

PR #1414, opened 2019-07-17
Show banner to prompt extension install

### LINKED ISSUE #1413: Direct account creation to compatible extensions
On the accounts page, add

- banner that explains that externtal is more portable, and recommended
- Link to compatible extensions for the browser

jacogr:
Closes #1413 

Works, but is pretty much stone-cold ugly atm...

Chrome -

image

Firefox -

image


jacogr:
image

jacogr:
@Tbaut Would appreciate a peek - it only shows up when you have no extensions.

packages/app-accounts/src/Banner.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+EXTENSIONS.forEach(({ browsers, desc, name }): void => {
+  Object.entries(browsers).forEach(([browser, link]): void => {
+    available[browser as Browser].push({ link, desc, name });
+  });
+});
+
+const browserInfo = detect();
+const browserName: Browser | null = (browserInfo && (browserInfo.name as Browser)) || null;
+const isSupported = browserName && ['chrome', 'firefox'].includes(browserName);
```
Tbaut: Can't we reuse the `Browser` type here?

packages/app-accounts/src/Banner.tsx
```diff
@@ -0,0 +1,106 @@
...
+    const { className, t } = this.props;
+
+    if (isWeb3Injected || !isSupported || !browserName) {
+      return null;
+    }
+
+    return (
+      <div className={className}>
+        <div className='box'>
+          <p>{t('It is recommended that you store and create the accounts that you care about, externally from the app. On {{yourBrowser}} the following browser signer extensions are available for use -', { replace: {
```
Tbaut: I would be a little more straight forward and talk simply (e.g without "signer") and using words like "secure/insecure"
```suggestion
          <p>{t('Creating and managing accounts from a website is not secure. It is recommended that you store and create accounts using an extension. For {{yourBrowser}} you can use-', { replace: {
```

packages/app-accounts/src/Banner.tsx
```diff
@@ -0,0 +1,106 @@
...
+                href={link}
+                rel='noopener noreferrer'
+                target='_blank'
+              >
+                {name}
+              </a> ({desc})
+            </li>
+          ))
+          }</ul>
+          <p>{t('The list is evolving and is updated as more extensions and external signers become available and it supported by the extension-dapp bridge. ')}<a
```
Tbaut: I'd keep it simple here as well. Extension devs will reach out and normal users don't really need more info
```suggestion
          <p>{t('The list is evolving and is updated as more extensions and external signers become available. ')}<a
```

packages/app-accounts/src/Banner.tsx
jacogr: Type vs value, so cannot quite use browser. 

However, your are right in another way, can pull those from the available keys.

jacogr:
The issue with the large banner is large screens. It looks horrible. So agree that the small is not perfect, but think it is certainly better.

For the accounts screen itself, keeping it non-closable. So it is always there until actioned.

Tbaut:
:+1:

jacogr:
Made a compromise between big and small, i.e. text is max 50rem, banner take up all the horizontal space. It still needs work, but probably ok for now, I'm just bike-shedding here atm...

Issue #1415, opened 2019-07-17
substrate 2.x Democracy & Treasury `Submit proposal` broken


Stefie:
With metadata v6  some API methods in democracy and treasury were replaced by `consts`  / `parameter_types`, which breaks the submit proposal functionality.

When trying to submit a proposal it keeps spinning indefinitely.
<img width="514" alt="Screenshot 2019-07-17 at 16 46 58" src="https://user-images.githubusercontent.com/125398/61385501-cc7e8700-a8b2-11e9-968c-117707ac02b6.png">

`main.1401b119.js:27 query.democracy.minimumDeposit :: s: Unable to find api.query.democracy.minimumDeposit`

PR #1416, opened 2019-07-17
Add endpoint fallbacks option and handling of consts to ui-api

Stefie:
Solves https://github.com/polkadot-js/apps/issues/1415

I added an option to pass fallback endpoints to the `ui-api` to provide an easy way to make the components backwards compatible.

I changed the ui-api so that it loops through the endpoint and fallback endpoints, if provided.
This way, we can just pass deprecated endpoints to `withCalls`.

If no fallbacks are provided `withCalls` will behave like before.

For 1 fallback call:
`['consts.democracy.launchPeriod', { fallback: ['query.democracy.launchPeriod'] }]`

For multiple fallback calls:  
`['consts.contracts.contractFee', { fallback: ['query.contracts.contractFee', 'query.contract.contractFee'] }]`

### DIFF
--- a/packages/app-contracts/src/Codes/ValidateCode.tsx
+++ b/packages/app-contracts/src/Codes/ValidateCode.tsx
@@ -16,7 +16,7 @@ import translate from '../translate';
 
 type Props = ApiProps & I18nProps & {
   codeHash?: string | null;
-  contract_codeStorage?: Option<PrefabWasmModule>;
+  contracts_codeStorage?: Option<PrefabWasmModule>;
   onChange: (isValid: boolean) => void;
 };
 
@@ -33,9 +33,9 @@ class ValidateCode extends React.PureComponent<Props> {
     isValid: false
   };
 
-  public static getDerivedStateFromProps ({ codeHash, contract_codeStorage, onChange }: Props): State {
+  public static getDerivedStateFromProps ({ codeHash, contracts_codeStorage, onChange }: Props): State {
     const isValidHex = !!codeHash && isHex(codeHash) && codeHash.length === 66;
-    const isStored = !!contract_codeStorage && contract_codeStorage.isSome;
+    const isStored = !!contracts_codeStorage && contracts_codeStorage.isSome;
     const isValid = isValidHex && isStored;
 
     // FIXME Really not convinced this is the correct place to do this type of callback?
--- a/packages/app-contracts/src/Codes/ValidateCode.tsx
+++ b/packages/app-contracts/src/Codes/ValidateCode.tsx
@@ -70,7 +70,6 @@ class ValidateCode extends React.PureComponent<Props> {
 
 export default translate(
   withCalls<Props>(
-    ['query.contracts.codeStorage', { paramName: 'codeHash' }], // 2.x
-    ['query.contract.codeStorage', { paramName: 'codeHash' }] // 1.x
+    ['query.contracts.codeStorage', { fallback: 'query.contract.codeStorage', paramName: 'codeHash' }]
   )(ValidateCode)
 );
--- a/packages/app-contracts/src/Codes/ValidateCode.tsx
+++ b/packages/app-contracts/src/Codes/ValidateCode.tsx
@@ -70,6 +70,6 @@ class ValidateCode extends React.PureComponent<Props> {
 
 export default translate(
   withCalls<Props>(
-    ['query.contracts.codeStorage', { fallback: 'query.contract.codeStorage', paramName: 'codeHash' }]
+    ['query.contracts.codeStorage', { fallbacks: ['query.contract.codeStorage'], paramName: 'codeHash' }]
   )(ValidateCode)
 );
--- a/packages/app-js/src/snippets/storage-examples.ts
+++ b/packages/app-js/src/snippets/storage-examples.ts
@@ -11,11 +11,12 @@ export const storageGetInfo: Snippet = {
   code: `// Get chain state information
 // Make our basic chain state/storage queries, all in one go
 
-const [minimumPeriod, validators, transferFee] = await Promise.all([
+const [minimumPeriod, validators] = await Promise.all([
   api.query.timestamp.minimumPeriod(),
-  api.query.session.validators(),
-  api.query.balances.transferFee()
+  api.query.session.validators()
 ]);
+
+const transferFee = api.consts.balances.transferFee;
 
 console.log('minimum period between blocks: ' + minimumPeriod);
 console.log('transfer fee: ', transferFee);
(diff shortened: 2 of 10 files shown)


packages/app-contracts/src/Codes/ValidateCode.tsx
```diff
@@ -70,7 +70,6 @@ class ValidateCode extends React.PureComponent<Props> {
 
 export default translate(
   withCalls<Props>(
-    ['query.contracts.codeStorage', { paramName: 'codeHash' }], // 2.x
-    ['query.contract.codeStorage', { paramName: 'codeHash' }] // 1.x
+    ['query.contracts.codeStorage', { fallback: 'query.contract.codeStorage', paramName: 'codeHash' }]
```
jacogr: I like this.

packages/app-js/src/snippets/storage-examples.ts
```diff
@@ -11,12 +11,13 @@ export const storageGetInfo: Snippet = {
   code: `// Get chain state information
 // Make our basic chain state/storage queries, all in one go
 
-const [minimumPeriod, validators, transferFee] = await Promise.all([
+const [minimumPeriod, validators] = await Promise.all([
```
jacogr: minimumPeriod is also gone :(

Stefie: I replaced it wit `blockPeriod ` in the example

packages/app-contracts/src/Codes/ValidateCode.tsx
Stefie: Thanks! :) The missing method calls were throwing error messages that I couldn't get rid of so this seemed like the more practical way

packages/app-js/src/snippets/storage-examples.ts
```diff
@@ -11,13 +11,14 @@ export const storageGetInfo: Snippet = {
   code: `// Get chain state information
 // Make our basic chain state/storage queries, all in one go
 
-const [minimumPeriod, validators, transferFee] = await Promise.all([
-  api.query.timestamp.minimumPeriod(),
-  api.query.session.validators(),
-  api.query.balances.transferFee()
+const [blockPeriod, validators] = await Promise.all([
+  api.query.timestamp.blockPeriod(),
```
jacogr: Sorry, dang tough because we can't visually inspect an actual newly-compiled node, in this case I'm guessing you looked at a week-old metadata,. updated here - https://github.com/polkadot-js/api/pull/1142/files#diff-a60854aee74a1ede6dc8edf6544d4240R349

packages/app-contracts/src/Codes/ValidateCode.tsx
jacogr: Yes, it is actually funny - originally it wasn't displayed, we added the logging of that quite recently (it certainly helps in finding exactly what goes wrong).

packages/app-js/src/snippets/storage-examples.ts
Stefie: Yeah, I thought docker-compose would pull the latest versions of the containers automatically, I just found out that you need to do that manually.

packages/ui-api/src/with/call.tsx
```diff
@@ -122,30 +136,54 @@ export default function withCall<P extends ApiProps> (endpoint: string, { at, at
...
 
-        assert(area.length && section.length && method.length && others.length === 0, `Invalid API format, expected <area>.<section>.<method>, found ${endpoint}`);
-        assert(['rpc', 'query', 'derive'].includes(area), `Unknown api.${area}, expected rpc, query or derive`);
-        assert(!at || area === 'query', `Only able to do an 'at' query on the api.query interface`);
+        const endpoints = [endpoint].concat(fallbacks);
+        let apiSection, area, section, method;
+        let i = 0;
 
-        const apiSection = (api as any)[area][section];
+        while (!apiSection && i < endpoints.length) {
```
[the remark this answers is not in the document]
Stefie: changed to `.map()`

packages/ui-api/src/with/call.tsx
```diff
@@ -166,18 +166,16 @@ export default function withCall<P extends ApiProps> (
...
-        const endpoints = [endpoint].concat(fallbacks || []);
+        const endpoints: string[] = [endpoint].concat(fallbacks || []);
         let apiSection, area, section, method;
-        let i = 0;
 
-        while (!apiSection && i < endpoints.length) {
-          [endpoint, apiSection, area, section, method] = this.constructApiSection(endpoints[i]);
-          i++;
-        }
+        [endpoint, apiSection, area, section, method] = endpoints
```
jacogr: Could remove 176 and just `const [...]` here?

packages/ui-api/src/with/call.tsx
```diff
@@ -219,18 +217,18 @@ export default function withCall<P extends ApiProps> (
...
-        const [apiMethod, params, isSubscription, isConst] = info;
+        const [apiMethod, params, isSubscription, area] = info;
         const updateCb = (value?: any): void =>
           this.triggerUpdate(this.props, value);
 
         await this.unsubscribe();
 
         try {
-          if (isSubscription) {
+          if (isSubscription || area === 'derive' || (area === 'query' && (!at && !atProp))) {
```
jacogr: Yeap, probably cleaner here now since we pass the area back. (Could actually remove `isSubscription` in this case as well, and just make the area for that `subscribe` - since we pass empty anyway, may as well combine the 2)

Stefie: we also have https://github.com/polkadot-js/apps/pull/1416/files#diff-e2040f6b00b230fed0e435eb6386b9a3R161

packages/ui-api/src/with/call.tsx
Stefie: I'm overwriting the passed endpoint here (https://github.com/polkadot-js/apps/pull/1416/files#diff-e2040f6b00b230fed0e435eb6386b9a3R162), so const doesn't work because it's declared already.

jacogr: Ahh yes, cool. But I also don't see us actually it, i.e. the `endpoint` as overwritten downwards? So if could even be as simple as `const [, apiSection, area, section, method] = endpoints`

(Or even not returning it, since it is not actually needed atm)

packages/ui-api/src/with/call.tsx
jacogr: Yes, I saw that, which is what lead me to the question. So what I'm suggesting,

- line 162 : return `[fn, params, 'subscribe']` (combine `true` and the `''`)
- line 192: return `[apiSection[method], newParams, method.startsWith('subscribe') ? 'subscribe' : area]`
- line 227 (here): if (['derive', 'subscribe']includes(area) || (area === 'query' && ...`

PR #1422, opened 2019-07-19
Switch to fallbacks

Stefie:
+ drop pre 1.0 support


packages/app-accounts/src/Banner.tsx
```diff
@@ -0,0 +1,105 @@
+// Copyright 2017-2019 @polkadot/app-accounts authors & contributors
```
jacogr: Not sure wht this was added here? Looks like a merge gone wrong. This file is in master.

Stefie:
Ok, I think you're right and it's really cleaner do add a second PR instead of this merging mess.

PR #1427, opened 2019-07-20
Vote for council candidates

kwingram25:
- Uses new derive.elections types
- Unclear behavior when votes are cast but more candidates are added - should it be possible to remove votes and redo?


packages/app-council/src/Motions/Propose.tsx
```diff
@@ -14,17 +14,17 @@ import { withApi, withCalls, withMulti } from '@polkadot/ui-api';
 
 import translate from '../translate';
 
-type Props = TxModalProps & ApiProps & {
+interface Props extends TxModalProps, ApiProps {
```
jacogr: Nice! Ok, we should probably apply this everywhere and only have interfaces (unless it is a simple `type Props = BareProps`)

Not for here, but something to keep in mind as we touch stuff.

jacogr:
Looking at the last one, yes, eslint is much more picky... However, does also pick up more gremlins along the way and at the very least aligns with the TS community out there, i.e. the TS team thjemselves are using the same rulesets. (We only have 2 vs 4 spaces overrides.)

It does get easier though. Would just suggest setting up VSCode to show these instead of waiting for CI, i.e. add the following to your settings.json

```json
    "eslint.validate": [
        "javascript",
        "javascriptreact",
        {
            "language": "typescript",
            "autoFix": true
        },
        {
            "language": "typescriptreact",
            "autoFix": true
        }
    ],
```

packages/app-council/src/Overview/Vote.tsx
```diff
@@ -0,0 +1,283 @@
...
+    }
+
+    const { [index]: bool } = approvals;
+
+    return (
+      <Toggle
+        label={
+          bool
+            ? (
+              <b style={{ color: 'green' }}>
```
jacogr: Can we extract this one and the red one? (Basically when we have `{ color: 'green' )` as a constant, we skip the creation of the new object each time.

packages/app-council/src/Overview/Vote.tsx
```diff
@@ -0,0 +1,283 @@
...
+    const { electionsInfo: { candidates }, t } = this.props;
+    const { approvals, oldApprovals } = this.state;
+
+    return (
+      <>
+        {
+          oldApprovals && (
+            <article
+              className='warning padded'
+              style={{
```
jacogr: Extract the style as a constant.

packages/app-council/src/Overview/Vote.tsx
```diff
@@ -0,0 +1,283 @@
...
+                display: 'flex',
+                alignItems: 'center',
+                margin: '0.5rem 0'
+              }}
+            >
+              <div style={{ flex: '1 1' }}>
```
jacogr: Extract as a constant.

kwingram25:
>Do we need to explicitly vote for all candidates in a round, currently we force votes for each one? If this is needed, all-ok, just wondering if this is a Substrate-thing or a UI-thing.

I assume it is all at once since the only method for this is elections.setApprovals which requires an array of yes/no in canddiate order. But since this locks the entire account balance(!) I don't understand why it's possible for candidates to keep submitting themselves, since the approvals then are out of sync and no one can vote twice

kwingram25:
Retract voter now works but I was able to submit despite a signer warning of insufficient funds...

package.json
```diff
@@ -10,10 +10,10 @@
     "packages/*"
   ],
   "resolutions": {
-    "@polkadot/api": "^0.82.0-beta.87",
-    "@polkadot/api-contract": "^0.82.0-beta.87",
+    "@polkadot/api": "^0.82.0-beta.94",
```
jacogr: Will bump the api now as well as soon as the metadata v7 PR passes. (Which will fix collective queries as well)

Issue #1433, opened 2019-07-21
Update API for new v2 extrinsics (0x82) as available


jacogr:
Normally upgrades are not worth a log, however in this case since it is a breaking change through the chain, need to properly test against -

- Alex
- Substrate 2.x (either pre or post merge - if pre, both master and v2-tx branches)

Introduced in https://github.com/polkadot-js/api/pull/1150

Issue #1432, opened 2019-07-21
Adjust signers to match new signPayload interface


jacogr:
As introduced in https://github.com/polkadot-js/api/pull/1152

This applies to -

- `ui-signer` - should be relatively close, but worth a check
- `app-js` - for the transaction signer inside the console

Needs #1433 first

PR #1435, opened 2019-07-22
Update for new Extrinsic v1/v2 support

jacogr:
- Closes https://github.com/polkadot-js/apps/issues/1433
- Closes https://github.com/polkadot-js/apps/issues/1432
- Pull in the new API with v2 extrinsic support
- Fix BlockInfo display to comply with new interface

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,10 +10,10 @@
     "packages/*"
   ],
   "resolutions": {
-    "@polkadot/api": "^0.82.0-beta.78",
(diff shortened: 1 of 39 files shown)


jacogr:
Tested e2e, all ok. May leavw the signer for a follow-up ( as logged), this means that submitting tx's via the js console is non-operational atm as this PR stands. (For now the signer is stubbed, will take a look and make a call on that one)

Issue #1445, opened 2019-07-25
Accounts displayed as if already validating - recent v2


Tbaut:
How to reproduce
in Staking > Actions
- create a new stake w/ Bob_stash and Bob and enough to stake, 100Peta
- look at the results, it's displayed as if Bob was validating already

![image](https://user-images.githubusercontent.com/33178835/61881806-8bf8bc00-aef7-11e9-9606-b8e81c2f0464.png)

jacogr:
Just need to figure out if the derive is broken or if we do something weird now in the UI.

Either way, the info is not consistent between old & new :(

Tbaut:
> if we do something weird now in the UI

it's this one :) I had a very dodgy
```js
const isStashValidating = !!validatorPrefs && !validatorPrefs.isEmpty && !isStashNominating;
```
Supposing that the `validatorPrefs` were empty when not validating, which they probably were back in the days. I'll fix that.

Issue #1408, opened 2019-07-16
substrate 2.x, can't see the chosen session account before validating


jacogr:
Reported by @Tbaut

Tbaut:
This needs additional exposed info https://github.com/paritytech/substrate/issues/3148

jacogr:
I'm going to take this one over to close out staking.

jacogr:
Needs https://github.com/polkadot-js/api/pull/1266 for proper nextSessionId extraction

Issue #1343, opened 2019-06-23
Staking overview is unbearably slow, non-responsive


jacogr:
On Alex.

Same with the actions (to an extent). The delayed loading is fine, however it seems like the UI is freezing while waiting. Either we are doing something stupid in `withCall` or just something stupid overall.

One things to check, alongside ops -

https://github.com/polkadot-js/apps/blob/master/packages/apps/src/Content/index.tsx#L103

If the above is re-enabled (basically loading the nominators/validators up-front - a very very expensive experience, is there impact on the load to the node? i.e. Are we keeping things in-check?)

Additionally... well, need to look, something is up here and it is not getting better as the numbers increase.

Tbaut:
In my tests, I found that `derive.staking.controllers` takes extremely long to answer on Alexander (average >8s) and is the main source of lag.

See example [here](https://polkadot.js.org/apps/#/js/share/qgGIYXBpLnJwYy5hdXRob3IucGVuZGluZ0V4dHJpbnNpY3MoKGUVDCgpID0+IHsKICBpZh0WNC5sZW5ndGggPT09IDApAR9wICBjb25zb2xlLmxvZygnTm8gcGVuZGluZyBleHQNWDQnKTsKICAgIHJldHVybgEMBH0KNjgAAGUVLBQpOwp9KTs=)

with the code:
```js
const timer = Date.now();

const data = await api.derive.staking.controllers();
console.log('data',data);

const timeTaken = Date.now() - timer
console.log('timeTaken (ms)',timeTaken);
```

jacogr:
Yes.

The issue is that those are linked lists, so there is no way around getting each one, one at a time.

We need to play to re-enable this - https://github.com/polkadot-js/apps/blob/master/packages/apps/src/Content/index.tsx#L103

I actually believe it won't add more RPC connections, since the WS connection should be shared. So we just need to check that.

(I'm almost of the opinion of just enabling that, testing and putting it up).

jacogr:
Closing. Rework done on UI and API layers.

PR #1446, opened 2019-07-26
Fix account wrongly seen as current validator

Tbaut:
- closes https://github.com/polkadot-js/apps/issues/1445
- remove useless `isActiveStash` (I remember telling this to myself earlier.. but hadn't had the chance yet)
- nits alphabetical ordering

Because of https://github.com/polkadot-js/apps/issues/1408 (I'll get to it soon) it's not easy to play with, you need to send `staking.validate` manually.

### DIFF
--- a/packages/app-staking/src/Actions/Account/index.tsx
+++ b/packages/app-staking/src/Actions/Account/index.tsx
@@ -27,9 +27,10 @@ import Validate from './Validate';
 
 type Props = ApiProps & I18nProps & {
   accountId: string;
+  allStashes?: string[];
+  balances_all?: DerivedBalances;
   className?: string;
   recentlyOffline: RecentlyOfflineMap;
-  balances_all?: DerivedBalances;
   staking_info?: DerivedStaking;
   stashOptions: KeyringSectionOption[];
 };
@@ -37,7 +38,6 @@ type Props = ApiProps & I18nProps & {
 interface State {
   controllerId: string | null;
   destination: number;
-  isActiveStash: boolean;
   isBondExtraOpen: boolean;
   isNominateOpen: boolean;
   isSetControllerAccountOpen: boolean;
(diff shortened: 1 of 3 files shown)


Tbaut:
working on Alexander.. but unacceptably wrong the first 10s, because it derives from `derive.staking.controllers` aka the_query_that_takes_ages cf https://github.com/polkadot-js/apps/issues/1343#issuecomment-511114873 

stare at this for 15s, it ends up being right:
slooooooow

crsCR:
> working on Alexander.. but unacceptably wrong the first 10s, because it derives from `derive.staking.controllers` aka the_query_that_takes_ages cf [#1343 (comment)](https://github.com/polkadot-js/apps/issues/1343#issuecomment-511114873)
> 
> stare at this for 15s, it ends up being right:
> slooooooow

Yes, I'm familiar... And what happens when you press to validate when you're actually validating? Slashed? Anything else? I think doing this this is how I ended up with ghost nodes (double spend frighteners) on telemetry: https://github.com/paritytech/substrate-telemetry/issues/166

Tbaut:
Repeating what I said in the chat, this is a lag, a bad one, but the problem you see on telemetry is different IMHO.

Clicking on Validate whereas you are already validating will most probably do nothing. it will tell the system that you want to validate.. which you were already, so it'll be ignored.

crsCR:
I'm pretty certain the process I describe in the linked to issue that featured this lag caused my ghost nodes (observed by others).... Without knowing more on how telemetry gets its data I can't offer any suggestions - and I can't try recreate as now in unbond limbo. I do worry though that if the cause for these ghost nodes isn't found then it might reappear later and cause some serious double spend worry, at the least. Maybe the issue can't happen again already as the telemetry server has had changes done to specifically address, I don't have details...Perhaps some telemetry /app conferring should happen.

jacogr:
Closing this for conversation - this is a merged PR. If visibility is needed, this is not the right place.

PR #1450, opened 2019-07-29
Fixed InputNumber.tsx bnToInputValue convert bug

WoeOm:
Get mod length before convert to bignumber

In function bnToInputValue , 0.001 and 0.1 will convert to the same value.


jacogr:
Thank you very much, will take a play-though. On first glance it looks spot-on.

PR #1452, opened 2019-07-30
GenericAsset Transfer app

### LINKED ISSUE #1451: New GenericAsset app
New GenericAsset app for [srml-generic-asset module](https://github.com/paritytech/substrate/blob/master/srml/generic-asset)

Similar to transfer one, but add new field for user to set asset id.

Depends on https://github.com/polkadot-js/api/issues/1228

xlc:
Part of #1451
Depends on https://github.com/polkadot-js/api/pull/1229

Needs some more thoughts about the UI/UX. Currently only able to make transfer call.

Can be tested with this branch: https://github.com/cennznet/substrate/tree/node-with-generic-asset

### DIFF
--- a/packages/app-generic-asset/LICENSE
+++ b/packages/app-generic-asset/LICENSE
@@ -1,0 +1,201 @@
+                              Apache License
+                        Version 2.0, January 2004
+                    http://www.apache.org/licenses/
+
+TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+1. Definitions.
+
+  "License" shall mean the terms and conditions for use, reproduction,
+  and distribution as defined by Sections 1 through 9 of this document.
+
+  "Licensor" shall mean the copyright owner or entity authorized by
+  the copyright owner that is granting the License.
+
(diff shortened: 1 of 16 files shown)


jacogr:
@xlc The api is current in master now, so the deps should be sorted.

xlc:
Only did a MVP. Later will create a tab to create new token and another one to mint/burn/update permission.

Ideally this should work for a chain without balances module but it requires all the logics about balances module are abstracted which is not easy to implement it cleanly.

jacogr:
Sorry, have not had a gap yet for in-depth. Took a look first thing this morning, and generally the code looks solid (I have some minor nit-picks here and there, but not something that I'll waste effort with).

I just want to give a play-through and got stuck on other things today. So will give my best attempt for tomorrow.

jacogr:
Sorry, seems I introduced conflicts. Was on the way to merging it now that things are more-or-less sane again.

jacogr:
I completely missed the "fix the conflicts" update - this really has hung around for way too long.

So please accept my apologies on this, it certainly was not the intent and nor was it quite a "normal" time. I really appreciate all the work that . has gone into this.

PR #1454, opened 2019-07-30
Support query double map value

### LINKED ISSUE #1453: Chain state app support query double map value

xlc:
Closes #1453


kwingram25:
Thank you for this! pulled into #1457 :)

PR #1464, opened 2019-08-01
Fix isEmpty for CardGrid

xlc:
Found this issue while working on #1452

The prop `isEmpty` was ignored. You can observe this bug in Address book with no addresses.

Also `Collection.isEmpty` gives false if children is `[false]`. I am not sure what is the desired behavior as do we want to ignore any non-renderable children or just strictly check children is empty array?

### DIFF
--- a/packages/ui-app/src/Collection.tsx
+++ b/packages/ui-app/src/Collection.tsx
@@ -53,9 +53,9 @@ export default class Collection<P extends CollectionProps, S extends CollectionS
     return !children || (Array.isArray(children) && children.length === 0);
   }
 
-  public static getDerivedStateFromProps ({ children }: CollectionProps): Pick<any, never> {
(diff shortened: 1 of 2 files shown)


packages/ui-app/src/Collection.tsx
```diff
@@ -53,9 +53,9 @@ export default class Collection<P extends CollectionProps, S extends CollectionS
     return !children || (Array.isArray(children) && children.length === 0);
   }
 
-  public static getDerivedStateFromProps ({ children }: CollectionProps): Pick<any, never> {
+  public static getDerivedStateFromProps ({ isEmpty, children }: CollectionProps): CollectionState {
```
xlc: what exactly is `Pick<any, never>`?

jacogr: It should be `Pick<State, never>` - which is not pretty, but matches up with the actual underlying React definitions. (`Partial<State>` does not work in these cases where you only return some of the keys)

PR #1469, opened 2019-08-02
Adjusted article padding and corrected Row's buttons position (1465)

Lowhearth:
Hello :) 

Looking good in My Accounts and My Contacts:

### DIFF
--- a/packages/ui-app/src/styles/app.css
+++ b/packages/ui-app/src/styles/app.css
@@ -73,7 +73,7 @@ article {
   border-radius: 0.25rem;
   box-sizing: border-box;
   margin: 0.25rem;
-  padding: 1.25rem 1.5rem;
+  padding: 1.25rem;
   position: relative;
   text-align: left;
 
--- a/packages/app-staking/src/Overview/Address.tsx
+++ b/packages/app-staking/src/Overview/Address.tsx
@@ -232,7 +232,7 @@ export default withMulti(
     .blockNumber {
       background: #3f3f3f;
       border-radius: 0.25rem;
-      bottom: 0.75rem;
+      top: 0rem;
       box-shadow: 0 3px 3px rgba(0,0,0,.2);
       color: #eee;
       font-size: 1.5rem;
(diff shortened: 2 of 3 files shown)


jacogr:
Thank you, it certainly looks good based on the screenshots. The code changes look innocent enough, will take a play-though since the same layout is shared everywhere there are cards.

packages/ui-app/src/styles/app.css
```diff
@@ -73,7 +73,7 @@ article {
   border-radius: 0.25rem;
   box-sizing: border-box;
   margin: 0.25rem;
-  padding: 1.25rem 1.5rem;
+  padding: 1.25rem;
```
jacogr: Not convinced about this one, however does not seem to detract visually.

Lowhearth: Had to put a negative margin instead of a negative "top"/"right" to correclty displace the entire div, seems pretty consistent now, I would appreciate a second check tho :


Also added the mentioned ellipsis :

jacogr: Will take a peek thanks. (And thanks a million for the eliipsis as well, unexpected to see it in here, but make my account management much easier...)

jacogr:
There is something very, very weird going on here -

image

- the "without nominator" now looks correct
- look at the above, the spacing is different between the 2 - the one with 2 nominators has more space between.

Really no idea what causes it.

For the "side-tracking very-long-addresses", thing that we seem to have slipped in here (my fault) I think need a no-wrap in there as well, i.e.

image

Not sure what that does when you make the screen smaller though, it may create other issues.

Lowhearth:
I really appreciate that you are taking the time to review my changes.  Watching you and your team working that hard, on weekends, is really inspiring. I will revise the code as many times as is needed.

Alright so, setting the div buttons container to a height of  0 should keep the bottom div at the same distance as if it were the old absolute positioned div.
To make It work with the block number tag I also had to fixed it to the top (I've observed that currently iterates between two positions, at the top when there is only Controller account, at the middle when there is a Controller account and a Session, so I guess that keeping it docked to the top left will be alright)







WIll wait for your feedback :)

jacogr:
Cool, yes... make no mistake, I do very well know the absolute was a hack. The above looks really good, happy with the movements, eg. last block produced - it is just as good as the original placement considering what we have to work with. 

I'll take an in-depth look first thing in the morning.

PR #1475, opened 2019-08-05
Swap session inputs to address (matches rotate RPC for 2.x)

jacogr:


Tbaut:
`InputValidationSessionSr25519.tsx` and `InputValidationSessionEd25519.tsx` are very similar. I just had a super quick look, should these 2 be combined?

jacogr:
Yes, they should - I'm leaving them as of now, just waiting for the 3rd key (imOnline) be up and running before cleaning up. (So probably will combine #1477 in here as well)

Issue #1421, opened 2019-07-19
Is there any plan on i18n?


exrobbie:
Hi, this is definex.io from China. We appreciate your excellent works and have a question on i18n.

We saw you have already coded locales files but it doesn't have any contents, do you have any plan on i18n? We are glad to help if you need our assistant to translate Chinese.

jacogr:
Yes, the plan has always been to support i18.l All through the UI all the strings are already using a i18n layer (in our case https://www.i18next.com/) everywhere in the interface. So technically the answer is a resounding "yes".

When it comes to actually doing it, things are slightly more tricky, just because we have not had translations as of yet - the first one is always more involved while we get the process smooth. 

As a first step, we need to exact the actual contents of the strings for translation - I'll take a look through the `i18next-*` utilities as how to best do this. Once done it really is just a matter of providing translations of those strings. (Which is no-doubt a lot of effort, as well as keeping it updated).

exrobbie:
We can provide Chinese translations and keep them updated for you. Do you have a schedule to do the i18n?

jacogr:
There are no . reasons to now start at this point - the only stumbling block atm is just making the string available, but there are tools for that, so just need to find the right one to extract it into a usable format.

So just give me a couple of days to catch breath and will take a look and we can get the ball rolling.

exrobbie:
That would be great, thank you so much.

jacogr:
My humble apologies for the very long delay in getting back to this - In #1478 has run the i18next tool against the codebase to extract. (The locales to extract for is configure in the config file, for now just extracting it to a default "en" to see the effect, so we would add "cn" here as well)

The good news is that all the strings are extracted without issues. What i don't like about this is that it is mostly uncategorised, so it is just a bundle of data without context. So some playing is still required to get some categories going, even . if that means having to specify each section seperately.

exrobbie:
Thank you for you excellent work, please don't hesitate to contact us if there's anything we could help.

jacogr:
And we actually have properly extracted strings now after #1710 . So we can actually get this done.

jacogr:
Closing, translations in after #2086

PR #1478, opened 2019-08-06
Extract i18next strings

jacogr:
Part of #1421


jacogr:
Better approach in #1710 - this one is being replaced.

PR #1479, opened 2019-08-06
Move ui-api to react-api-promise

jacogr:


jacogr:
To be replaced, there has been a lot of auto-replace here that went completely wrong.

PR #1482, opened 2019-08-08
Fix: Unalble to determine top level customise type

wooparadog:
Problem:

We have a type alias of `pub type TokenSymbol = Vec<u8>;`, and it's a parameter in our substrate open endpoint. 

when we try to use polkadot-js/apps to send a request, the console says:

```
Unable to determine default type for {"info":5,"type":"TokenSymbol"}
```

and the input is grayed out:

image

This request will let plain type get a raw default value.

### DIFF
--- a/packages/ui-params/src/initValue.ts
+++ b/packages/ui-params/src/initValue.ts
@@ -104,7 +104,7 @@ export default function getInitValue (def: TypeDef): RawParamValue | RawParamVal
 
         if (instance instanceof BN) {
           return new BN(0);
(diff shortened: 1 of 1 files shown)


jacogr:
Go to settings -> developer, add the custom type. (The help overlay has info on . how to do so)

The metadata does not show the reflected types the node has, only the names. It is up to the developer to let the API know which types are available.

wooparadog:
Er.. I've forgot to mention that we did put the type description in settings -> developer...

PR #1484, opened 2019-08-08
 Import session keys for author_rotateKeys

jacogr:


packages/ui-app/src/InputAddressSimple.tsx
```diff
@@ -0,0 +1,73 @@
+// Copyright 2017-2019 @polkadot/ui-app authors & contributors
```
jacogr: Well, this was needed - then not anymore.

Left it in.

PR #1496, opened 2019-08-12
Display Data/Keys params as hex

jacogr:


packages/react-params/src/valueToText.tsx
```diff
@@ -34,7 +34,7 @@ function valueToText (type: string, value: any, swallowError: boolean = true, co
     ? unknown
     : div(
       {},
-      ['Bytes', 'Data'].includes(type)
+      ['Bytes', 'Data', 'DoubleMap<Keys>', 'Keys'].includes(type)
```
xlc: `DoubleMap<Keys>` is not really a proper type. whatever code produces this needs fix IMO

jacogr: Well aware of that. However, this needs to get done ASAP, so it is not being fixed here. (The api does it properly, the UI doesn't)

jacogr: Added a FIXME there as well. (The whole thing is very obvious when you actually use the storage app - so just needs the same check as implemented in the API doc generation)

PR #1497, opened 2019-08-12
Render session hex

### LINKED ISSUE #1487: substrate 2.x, display multiple session icons
Because rotateKeys come into play, the actual display for a single key as we have it is non-sensical. Either display the icons/short, or just drop (3 in the case of substrate, 4 in the case of polkadot)

### LINKED ISSUE #1500: Hide validator prefs when not validating
double bug here, Validators prefs shouldn't show up if not set, or not a validator intention. The "Change validator preferences" Menu is `0`:
image

Saw that on Polkadot master.

jacogr:
- closes #1500 
- closes #1487

### DIFF
--- a/packages/app-staking/src/Actions/Account/index.tsx
+++ b/packages/app-staking/src/Actions/Account/index.tsx
@@ -424,13 +424,11 @@ class Account extends React.PureComponent<Props, State> {
 
   private renderPopupMenu (): React.ReactNode {
     const { balances_all, isSubstrateV2, t } = this.props;
-    const { stashId, hexSessionId, isStashNominating, isStashValidating, sessionIds } = this.state;
+    const { hexSessionId, isStashNominating, isStashValidating, sessionIds } = this.state;
 
     // only show a "Bond Additional" button if this stash account actually doesn't bond everything already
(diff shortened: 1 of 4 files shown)


Tbaut:
added some changes to show the hex session in the `Action` tab:
image

I think it wouldn't be good to reduce the amount of bytes to show, like the address 3 first and last bytes instead of 4 right now. This would also help to keep those cards smaller.

For the overview I'm almost at the point of questioning the utility of having this info here. This is not very useful to know other's session key, and our own can be seen in the next tab :man_shrugging:

jacogr:
Happy to drop it from the actual overview - we can re-considder if it becomes an issue.

jacogr:
We can do 3..3, i.e. `48` bitlength

Tbaut:
Tested on Substrate, Polkadot and Alexander
How it looks like on a v2:
image
image

Tbaut:
Removed the absolute for Blocknumber in V2.
I have created a dedicated `className` for V1 and V2 to make the cleaning easier in the future.
that's a gif:
authoring

Validator prefs aren't displayed if not validator
image

PR #1514, opened 2019-08-20
Allow metadata refresh to adjust routes

jacogr:
UI adjustments to cater for https://github.com/polkadot-js/api/pull/1326


jacogr:
So outdated, not worth rescuing.

PR #1541, opened 2019-08-26
Allows claiming to an account in address book

### LINKED ISSUE #1539: Allow claiming to "view" addresses from the address book in the Claims app
There is no reason one cannot make a claim to a view-only address as long as the signed message contains the public key for this address.

Please assign me.

lsaether:
fixes #1539

### DIFF
--- a/packages/app-claims/src/index.tsx
+++ b/packages/app-claims/src/index.tsx
@@ -129,7 +129,7 @@ class App extends TxModal<Props, State> {
                 help={t('The account you want to claim to.')}
                 label={t('claim to account')}
                 onChange={this.onChangeAccount}
-                type='allPlus'
+                type='all'
               />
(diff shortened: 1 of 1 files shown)


packages/app-claims/src/index.tsx
```diff
@@ -124,7 +124,13 @@ class App extends TxModal<Props, State> {
                   chain: system_chain.toString()
                 }
               })}</h3>
-              {this.renderInputAccount()}
+              <InputAddress
+                defaultValue={this.state.accountId}
+                help={t('The account you want to claim to.')}
+                label={t('claim to account')}
+                onChange={this.onChangeAccount}
+                type='allPlus'
```
jacogr: Rather just `all`, `allPlus` has contracts as well.

```suggestion
                type='all'
```

PR #1542, opened 2019-08-27
Fix address row name overflow

### LINKED ISSUE #1502: Overflow on staking pages
image

kwingram25:
Closes #1502 

Don't seem to be any issues with text overflow normal instead of ellipsis (unless I'm forgetting)?

### DIFF
--- a/packages/react-components/src/Row.tsx
+++ b/packages/react-components/src/Row.tsx
@@ -141,7 +141,7 @@ export const styles = `
     white-space: normal;
 
     .withName {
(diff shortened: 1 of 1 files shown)


jacogr:
It does work, but really would like the same view as we have in accounts, i.e. after this we end up with -

image

While we have the following in accounts

image

And actually not sure how to wangle the CSS here to get it to cooperate :(

jacogr:
Replaced by #1593

PR #1564, opened 2019-08-30
Functional components (apps, react-components)

kwingram25:
Continued


packages/apps/src/SideBar/NodeInfo.tsx
```diff
@@ -15,34 +15,30 @@ interface Props extends ApiProps, BareProps {}
...
+    </div>
+  );
+}
 
-    if (!isApiReady) {
-      return null;
-    }
+function NodeInfo (props: Props): React.ReactElement<Props> {
+  const { api, className } = props;
+  const uiInfo = `apps v${pkgJson.version}`;
```
jacogr: Can actually move this out to file-level

packages/apps/src/SideBar/NodeInfo.tsx
```diff
@@ -15,34 +15,30 @@ interface Props extends ApiProps, BareProps {}
...
+      <NodeName />&nbsp;
+      <NodeVersion label='v' />
+    </div>
+  );
+}
 
-    if (!isApiReady) {
-      return null;
-    }
+function NodeInfo (props: Props): React.ReactElement<Props> {
```
jacogr: Why not ({ Api, className }: props)

packages/apps/src/overlays/Connecting.tsx
```diff
@@ -17,59 +17,57 @@ type Props = I18nProps & ApiProps;
...
+    return null;
   }
 
-  private renderConnecting (): React.ReactNode {
-    const { className, isApiConnected, t } = this.props;
-
-    if (isApiConnected) {
-      return null;
-    }
+  const wsUrl = settings.apiUrl;
```
jacogr: This stuff we can keep file level. It doesn’t need to be re-evaluated.

packages/apps/src/overlays/Connecting.tsx
```diff
@@ -17,59 +17,57 @@ type Props = I18nProps & ApiProps;
...
+  return (
+    <BaseOverlay
+      className={className}
+      icon='globe'
+    >
+      <div>{t('You are not connected to a node. Ensure that your node is running and that the Websocket endpoint is reachable.')}</div>
+      {
+        isFirefox && isWs
+          ? <div>{t('With the Firefox browser connecting to insecure WebSockets ({{wsUrl}}) will fail due to the browser not allowing localhost access from a secure site.', { replace: { wsUrl } })}</div>
+          : undefined
```
jacogr: Completely throw-away comment - I always wonder about these where we mix null and undefined. Doesn’t make a difference either way, but we are inconsistent. 

Don’t do anything, just saw it

packages/apps/src/SideBar/NodeInfo.tsx
kwingram25: Because I pass props to renderNode later :)

packages/apps/src/SideBar/NodeInfo.tsx
kwingram25: How do you mean?

jacogr: So uiInfo will never change. So the constant can and should be defined outside of the function, ie. File-level.

PR #1593, opened 2019-09-08
Avoid overflow on staking pages #1502

Lowhearth:
Avoid overflow on staking pages while keeping consistent with the address ellipsis that we have on accounts app.

That's what I've got so far:

Wasn't able to detect any side effects by the css changes, I'd appreciate a close look on those.

Thanks


jacogr:
Brilliant! Will take a deep-dive tomorrow :)

PR #1602, opened 2019-09-12
UI support for v2 ABI

kwingram25:
- Untested for contract fns with args
- Needs api dependency update (dep contents of https://github.com/polkadot-js/api/pull/1320)


kwingram25:
Ready to go I think unless any more issues you found?

PR #1607, opened 2019-09-13
Reset themes, single palette

jacogr:
Basically this is a "wipe and start over". 

- Remove polkadot theme & substrate themes
- Only have one single theme (theme selector applies to icon only)
- Bring in `darkorange` highlights

@Tbaut will kill me for this... 

... but as it is we are way off from the original "polkadot" and the "substrate" one has not been getting attention. So really this is a reset and trying to do as little as possible, use a single color palette while still making it pleasing.

... As-is, still WIP.


jacogr:
As-is

image

image

Tbaut:
> @Tbaut will kill me for this...

Haha, because you replace things with orange, I'm actually all for it, I love orange!.
Although I'm not entirely convinced yet by the combination with the blue.

I had a play through and ended up with a [greyish](https://coolors.co/4d4e4f-f19135-93b7be-fafafa-fafafa).. maybe a little too light to be perfectly readable.

64843736-2d51e380-d606-11e9-8cfe-6f5eeb66ceb3

jacogr:
@Tbaut Try with the current - it is now actually the orginal polkadot blue. Maybe the orange could use some small toning down to (not quite darkorange). So the other way around, tweaking the orange to be slightly more "washed out". (... slightly being the . operative word)

Tbaut:
I like this blue more with that orange :)
The one thing (that I remember was bothering me a bit in the past) is the highlight for validators. Orange being orange, it doesn't fit well with everything. This makes the whole staking overview for instance quite unpleasant to my eyes.

image

I wonder if something neutral like black would allow to distinguish them enough from normal accounts? This would at least pass to the "bunt" from the identicons.

jacogr:
So reasonably happy - happy need to check democracy - those colors are still to be updated.

jacogr:
Not perfect, good enough.

PR #1620, opened 2019-09-17
Ledger support

jacogr:
- Requires https://github.com/polkadot-js/ui/pulls

image

Doesn't sign yet, however should be hookable now....


jacogr:
Round #1 - it should still be tied to a specific chain and be available generally (currently hidden behind a flag). However detection is in place (adding the account), along with . a signer that signs on the actual device.

jacogr:
image

Yes!

Issue #1621, opened 2019-09-18
Address Prefix (should be called ss58) not honored


jacogr:
![image](https://user-images.githubusercontent.com/1424473/65135266-ee58cf00-da05-11e9-8c0d-ac4348861d21.png)

It doesn't actually apply the global override, but rather defaults to the "connect node" at all times.

PR #1630, opened 2019-09-18
ss58Format overrides

jacogr:
- Will require `@polkadot/util` 1.5.1 when https://github.com/polkadot-js/common/pull/472 is in
- Closes https://github.com/polkadot-js/apps/issues/1621
- Prefix change is a reload (this is not great)

### DIFF
--- a/packages/react-api/src/Api.tsx
+++ b/packages/react-api/src/Api.tsx
@@ -158,6 +158,7 @@ export default class Api extends React.PureComponent<Props, State> {
 
     // finally load the keyring
     keyring.loadAll({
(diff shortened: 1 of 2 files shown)


packages/react-api/src/Api.tsx
```diff
@@ -158,6 +158,7 @@ export default class Api extends React.PureComponent<Props, State> {
 
     // finally load the keyring
     keyring.loadAll({
+      addressPrefix: ss58Format,
```
jacogr: Until we have 1.5.1

PR #1649, opened 2019-09-23
More func. components, Button labelIcon -> icon (req.)

jacogr:


jacogr:
Yes, I am purposefully not touching contracts. (Even if you don't convert, that will just end up in merge hell for somebody)

Issue #1609, opened 2019-09-13
Notifications close & close-all


jacogr:
Background & Goals:

Status messages, e.g. indicating that actions has taken place (such as address copied) or the results from a transaction appear on the UI and disappear after some time. The time here is short enough so it doesn't "hang around forever", but also long enough that it does get in the way when the user just wants to move on with work.

Simple events:

![image](https://user-images.githubusercontent.com/1424473/64849849-92acd100-d614-11e9-9c84-529c19369294.png)
![image](https://user-images.githubusercontent.com/1424473/64849874-9e989300-d614-11e9-8b6b-fce9b65ec579.png)

Result from a transfer to a new account:

![image](https://user-images.githubusercontent.com/1424473/64849977-d273b880-d614-11e9-8262-522c2a660c95.png)

We would like to improve the use experience here, keeping the auto-close as-is, but also allowing the ability to dismiss either a single item or all of them.

Actions:

Inside react-components/Status add the ability to dismiss completed items. It is important that this only applies to "completed", i.e. the same status checks that is done when it auto-closes should be used to make it manually closable.

Add -

- to each closable/completed item a close button (icon . only) to the top right
- add a close-all that appears when there are closable items in the status overlay box (outside notifications, also top right)

(Split from https://github.com/polkadot-js/apps/issues/1544)

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 150.0 DAI (150.0 USD @ $1.0/DAI)  attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/polkadot-js/apps/1609/3455).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/polkadot-js/apps/1609/3455).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
* $150,287.40 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Workers have applied to start work__.


These users each claimed they can complete the work by 1 month from now.
Please review their action plans below:


[...]

PR #1672, opened 2019-09-27
Added dismiss notifications feature

eswarasai:
### Description
 - [x] Added close icon for each notification
 - [x] Added `Dismiss all` button for all the notifications

Fixes: #1609

### DIFF
--- a/packages/react-components/src/Button/Button.tsx
+++ b/packages/react-components/src/Button/Button.tsx
@@ -13,7 +13,7 @@ import Tooltip from '../Tooltip';
 
 let idCounter = 0;
 
-export default function Button ({ children, className, floated, fluid, icon, isBasic = false, isCircular = false, isDisabled = false, isLoading = false, isNegative = false, isPositive = false, isPrimary = false, label, labelPosition, onClick, size, style, tabIndex, tooltip }: ButtonProps): React.ReactElement<ButtonProps> {
+export default function Button ({ children, className, floated, icon, isBasic = false, isCircular = false, isDisabled = false, isFluid = false, isLoading = false, isNegative = false, isPositive = false, isPrimary = false, label, labelPosition, onClick, size, style, tabIndex, tooltip }: ButtonProps): React.ReactElement<ButtonProps> {
   const [triggerId] = useState(`button-${++idCounter}`);
   const props = {
     basic: isBasic,
@@ -23,7 +23,7 @@ export default function Button ({ children, className, floated, fluid, icon, isB
     'data-for': triggerId,
     disabled: isDisabled,
     floated,
-    fluid,
+    fluid: isFluid,
     labelPosition,
     loading: isLoading,
     negative: isNegative,
--- a/packages/react-components/src/Button/Button.tsx
+++ b/packages/react-components/src/Button/Button.tsx
@@ -13,7 +13,7 @@ import Tooltip from '../Tooltip';
 
 let idCounter = 0;
 
-export default function Button ({ children, className, floated, icon, isBasic = false, isCircular = false, isDisabled = false, isLoading = false, isNegative = false, isPositive = false, isPrimary = false, label, labelPosition, onClick, size, style, tabIndex, tooltip }: ButtonProps): React.ReactElement<ButtonProps> {
+export default function Button ({ children, className, floated, fluid, icon, isBasic = false, isCircular = false, isDisabled = false, isLoading = false, isNegative = false, isPositive = false, isPrimary = false, label, labelPosition, onClick, size, style, tabIndex, tooltip }: ButtonProps): React.ReactElement<ButtonProps> {
   const [triggerId] = useState(`button-${++idCounter}`);
   const props = {
     basic: isBasic,
@@ -23,6 +23,7 @@ export default function Button ({ children, className, floated, icon, isBasic = 
     'data-for': triggerId,
     disabled: isDisabled,
     floated,
+    fluid,
     labelPosition,
     loading: isLoading,
     negative: isNegative,
--- a/packages/react-components/src/Status/index.tsx
+++ b/packages/react-components/src/Status/index.tsx
@@ -10,14 +10,22 @@ import styled from 'styled-components';
 import { GenericCall } from '@polkadot/types';
 
 import AddressMini from '../AddressMini';
+import Button from '../Button';
 import Icon from '../Icon';
-import { classes } from '../util';
 import translate from '../translate';
+import { classes } from '../util';
 
 interface Props extends I18nProps {
   stqueue?: QueueStatus[];
   txqueue?: QueueTx[];
 }
+
+const STATUS_COMPLETE: QueueTxStatus[] = [
+  // status from subscription
+  'finalized', 'usurped', 'dropped', 'invalid',
+  // normal completion
+  'cancelled', 'error', 'sent'
+];
 
 function iconName (status: string): any {
   switch (status) {
@@ -69,11 +77,17 @@ function renderStatus ({ account, action, id, message, removeItem, status }: Que
   return (
     <div
       className={classes('item', status)}
-      onClick={removeItem}
       key={id}
     >
       <div className='wrapper'>
         <div className='container'>
+          <Icon
+            name='close'
+            onClick={removeItem}
+          />
+          <div className='short'>
+            <Icon name={iconName(status)} />
+          </div>
           <div className='desc'>
             <div className='header'>
               {action}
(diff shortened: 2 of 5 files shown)


eswarasai:
@jacogr - So I've handled close and close all functionality for the notifications. Let me know your thoughts on the UI/UX and also any further modifications needed. I'll try to put up the fixes ASAP. Thanks!

jacogr:
Thank you very much! I took a brief look at the code and it looks sane (not 100% in-depth, just an initial pass-through). Will go through in detail and test/play ASAP.

packages/react-components/src/Status/index.tsx
```diff
@@ -144,8 +145,24 @@ function Status ({ className, stqueue = [], txqueue = [] }: Props): React.ReactE
     return null;
   }
 
+  const _onDismiss = (): void => {
+    allst.map(s => s.removeItem());
```
jacogr: (nitpick) Prefer the following form 

```js
allst.map((s): void => s.removeItem())`
```

1. no naked params, i.e. `(s)`
2. it is strange, but linting should complain here - the `: void` should be picked up as missing

packages/react-components/src/Status/index.tsx
```diff
@@ -144,8 +145,24 @@ function Status ({ className, stqueue = [], txqueue = [] }: Props): React.ReactE
...
+    alltx.map(t => t.removeItem());
```
jacogr: (nitpick) same comment as the preceding

packages/react-components/src/Status/index.tsx
```diff
@@ -144,8 +145,24 @@ function Status ({ className, stqueue = [], txqueue = [] }: Props): React.ReactE
...
+  };
+
   return (
     <div className={`ui--Status ${className}`}>
+      {allst.length + alltx.length > 1 && (
```
jacogr: Just be a tiny . bit more explicit here, i.e. prefer the following form -

```js
(allst.length + alltx.length) > 1
```

No functional difference, but when reading the code late at night, it is just immediately obvious
```

File: packages/react-components/src/Status/index.tsx [hunk 4]
```diff
@@ -144,8 +145,24 @@ function Status ({ className, stqueue = [], txqueue = [] }: Props): React.ReactE
...
+        <div className='dismiss'>
+          <Button
+            isNegative
+            onClick={_onDismiss}
+            label={t('Dismiss all')}
+            icon=''
```
jacogr: Add the `cancel` as an icon. (It is left aligned which is a bit unfortunately, however, it is fine in this case)

packages/react-components/src/Status/index.tsx
```diff
@@ -144,8 +145,24 @@ function Status ({ className, stqueue = [], txqueue = [] }: Props): React.ReactE
...
+  };
+
   return (
     <div className={`ui--Status ${className}`}>
+      {allst.length + alltx.length > 1 && (
+        <div className='dismiss'>
+          <Button
+            isNegative
+            onClick={_onDismiss}
+            label={t('Dismiss all')}
```
jacogr: Would prefer `Dismiss all notifications` (probably moreso with the the icon change indicated.)

packages/react-components/src/Status/index.tsx
```diff
@@ -144,8 +145,24 @@ function Status ({ className, stqueue = [], txqueue = [] }: Props): React.ReactE
...
 
+  const _onDismiss = (): void => {
+    allst.map(s => s.removeItem());
+    alltx.map(t => t.removeItem());
+  };
+
   return (
     <div className={`ui--Status ${className}`}>
+      {allst.length + alltx.length > 1 && (
+        <div className='dismiss'>
```
jacogr: We probably need to split the notifications, i.e. with opacity from the button so the button itself has no opacity. (It is slightly harder to read)

packages/react-components/src/Status/index.tsx
jacogr: Would also suggest having the button stretch across the full notification width.

jacogr:
This actually looks quite good. Some small code comments. Additionally, we want to limit the "close" to only those that have been completed. See this example -

1. I send a transaction



2. Not I close the popup
3. I get back to this screen 



Basically since signer uses the same queue, it loses track when it disappears from the queue.

The alternative would be to just introduce a hidden state, i.e. when you click "x" it hides it instead of removing it.

packages/react-components/src/Status/index.tsx
eswarasai: Probably because of the missing parentheses 🤔

eswarasai:
@jacogr - Okay, I was a bit confused with this `completed` status. Because we filter the notifications and won't display the one's that are `completed`. Moreover, earlier, user could have clicked on a notification and removed that similar to the existing close behaviour.

So if I change this, I'm not really sure when would it be removed exactly? I'll need some help here to understand the technical implementation behind this. Thanks 🙂

packages/react-components/src/Status/index.tsx
eswarasai: I'm not sure I understand this comment. Can you maybe share a screenshot or something so that I understand it better? Thanks!

jacogr:
So I believe the simplest would be to do the following -

- only display the `x` when we are in a completed status
- filter your "Dismiss all notifications" to only take "completed" into account as well (instead of simply `alltx.length | allst.length`)

Adding "hidden" could have some unintended effects, i.e. the status could update and it would popup again. (I know the whole status handling is nowhere close to perfect here... hence it may just create chaos)

Alternatively, if we keep it to all -

- add an additional flag for `isHidden` (or `isClosed`)
- instead of closing when calling remove, set the this flag
- just don't display those

That is also quite simple overall with a small extension. (And there are benefits being able to hide anything)

eswarasai:
@jacogr - I think what I need to do here is add the additional complete status check only for `alltx` queue and display the dismiss icon only for them. But in this case, the pending transactions notification won't be removed from the UI until the transaction is successful. Would that be a problem?

jacogr:
No, it is not an issue. Perfect for the initial version.

eswarasai:
@jacogr - I've just pushed the above changes. Except for one comment I left unresolved above and a code duplication between `STATUS_COMPLETE` constant, I think everything should work as expected. Let me know what else is pending so that we can merge this one soon 🙂

packages/react-components/src/Button/Button.tsx
```diff
@@ -23,6 +23,7 @@ export default function Button ({ children, className, floated, icon, isBasic =
     'data-for': triggerId,
     disabled: isDisabled,
     floated,
+    fluid,
```
jacogr: I like this. 

Small comment - can we make the prop`isFluid` (aligning with the other booleans) and then just `fluid: isFluid` here

packages/react-components/src/Status/index.tsx
jacogr: Will re-check the latest version - it may just be a non-issue with extending over the full width. (The initial version was quite hard to read since it is opaque.) However, let's see what the full version + icon looks like, it may just solve itself.

jacogr:
@eswarasai Thank you. checking now. Just one small request - don't force push, it just makes testing . hell since I had to resolve a couple of conflicts due to it. (Since we squash merge all PRs, it ends up being one commit anyway, so it doesn't pollute at all)

packages/react-components/src/Status/index.tsx
jacogr: Yes, it is fine with the current full width.

packages/react-components/src/Status/index.tsx
```diff
@@ -10,15 +10,23 @@ import styled from 'styled-components';
...
-import { classes } from '../util';
 import translate from '../translate';
+import { classes } from '../util';
 
 interface Props extends I18nProps {
   stqueue?: QueueStatus[];
   txqueue?: QueueTx[];
 }
 
+const STATUS_COMPLETE: QueueTxStatus[] = [
```
jacogr: One last request - rather just move this into a `constants.ts` file and import it in both `index.tsx` and `Queue.tsx`. (That we we have no duplication)

eswarasai: Sorry. Shouldn't have bugged you for adding `constants.ts`. It was pretty obvious 😅

eswarasai:
@jacogr - Just pushed the changes. Sorry about the force push. Will keep that in mind going forward. Thanks!

Issue #1325, opened 2019-06-21
Upgrade react-dropzone to 10.x version


jacogr:
**Background:**

The [react-dropzone](https://www.npmjs.com/package/react-dropzone) component currently in use in the `react-components/InputFile` is at version 7.x while the actual package is already at 10.1.x. 

Since 8.x the interfaces has changed significantly, the `Dropzone` component now uses async promises. Since we really want to track latest versions of components (we want to take advantage of features and fixes), we are desperately in need of an upgrade here.

**Actions:**

For https://github.com/polkadot-js/apps/blob/master/packages/react-components/src/InputFile.tsx -

- update the package in https://github.com/polkadot-js/apps/blob/master/packages/react-components/package.json#L36
- change the interfaces to support the latest async way of doing
- test changes... the easiest is probably via the extrinsics app (system -> setCode uses a file input)

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 60.0 DAI (60.0 USD @ $1.0/DAI)  attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/polkadot-js/apps/1325/3517).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/polkadot-js/apps/1325/3517).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
* $147,589.10 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Workers have applied to start work__.


These users each claimed they can complete the work by 1 month from now.
Please review their action plans below:


**1) [fr4ctalbit](https://gitcoin.co/profile/fr4ctalbit) has applied to start work _(Funders only: [approve worker](https://gitcoin.co/issue/polkadot-js/apps/1325/3517?mutate_worker_action=approve&worker=fr4ctalbit) | [reject worker](https://gitcoin.co/issue/polkadot-js/apps/1325/3517?mutate_worker_action=reject&worker=fr4ctalbit))_.**

1) update the package
2) upgrade interfaces
3) test

[...]

PR #1682, opened 2019-10-01
updated react-dropzone to 10.1.9

shyamks:
@jacogr - This is the update required for #1325. Works similar to the previous version after the updates. Let me know what you think. 😃


shyamks:
Was'nt able to locate the page Extrinsics on my sidebar. But i guess I have figured out why this is happening. Hope this checks out.

PR #1683, opened 2019-10-01
Add insert key functionality

jacogr:
image

image

cc @shawntabrizi


shawntabrizi:
Looks awesome. Will see if we can pop this right into our tutorials!

@JoshOrndorff @hcastano

JoshOrndorff:
Question: Can this be used without the staking module installed?

I'm working on updating the [private network tutorial](https://substrate.dev/docs/en/tutorials/start-a-private-network-with-substrate) to Substrate master. That tutorial uses the node template which does not have the staking module installed.

One solution would be to put this feature on the accounts tab, although that doesn't make a ton of sense for staked networks like polkadot.

jacogr:
For non-staking-module networks, no idea where to put this tbh. It really is advanced, so understand the need to "tweak", but certainly not on accounts as it stands.

PR #1691, opened 2019-10-02
Allow ss58 input on Bytes fields

jacogr:
This allows ss58 inputs anywhere we have Bytes types. 

(This is the reverse of existing functionality where we already allow 0x... publicKey inputs in any address field - in both the existing and new case, we just convert transparently)

image

cc @shawntabrizi cc @JoshOrndorff - this may help somewhat


shawntabrizi:
Nice!!

PR #1697, opened 2019-10-06
Align Sig Validity Indicator (#1660)

### LINKED ISSUE #1660: Align sig validity indicator
**Background:**

In the toolbox -> verify section we have the following display:

image

The icon indicates red when invalid and green when valid. At this point . it is completely mis-aligned with the way we display indicators, e.g. see the address icon in the dropdown just above. In all cases the indicators/fields are aligned such that the icon is half in-out the box and the text never moves.

(The same can be seen on the settings page)

**Actions:**

1. Move the display icon to align with the icons on the left in other inputs - very similar to what we have in the actual settings where the icon is to the left of the box, half-overlapping. For inspiration we can take a look at the way it is done there.

2. Adjust the size of these icons, once moved, to be the same size as the other half in-out indicators. (same size as the icons in settings, same size as the icon for accounts)

touhonoob:
resolve #1660

### DIFF
--- a/packages/app-toolbox/src/Verify.tsx
+++ b/packages/app-toolbox/src/Verify.tsx
@@ -36,7 +36,7 @@ interface State {
 }
 
 const AlignedIcon = styled(Icon)`
-  &&&.ui--AlignedIcon::before {
+  &&&::before {
     position: relative;
     left: 0.88rem;
     top: 1rem; 
@@ -45,6 +45,10 @@ const AlignedIcon = styled(Icon)`
     font-size: 32px;
     background: white !important;
     border-radius: 50%;
+  }
+  
+  &&&.big.icon {
+    font-size: 32px;
   }
 `;
 
@@ -105,7 +109,6 @@ class Verify extends React.PureComponent<Props, State> {
         <div className='ui--row'>
           <div className="ui--AlignedIconContainer" style={{ position: 'absolute', zIndex: 1 }}>
             <AlignedIcon
-              className="ui--AlignedIcon"
               color={isValid ? 'green' : (isValidSignature ? 'red' : undefined)}
               name={isValid ? 'check circle' : (isValidSignature ? 'exclamation circle' : 'help circle')}
               size="big"
--- a/packages/app-toolbox/src/Verify.tsx
+++ b/packages/app-toolbox/src/Verify.tsx
@@ -35,6 +35,19 @@ interface State {
   signature: string;
 }
 
+const AlignedIcon = styled(Icon)`
+  &&&.ui--AlignedIcon::before {
+    position: relative;
+    left: 0.88rem;
+    top: 1rem; 
+    width: 32px;
+    height: 32px;
+    font-size: 32px;
+    background: white !important;
+    border-radius: 50%;
+  }
+`;
+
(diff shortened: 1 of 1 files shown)


packages/app-toolbox/src/Verify.tsx
```diff
@@ -64,6 +65,18 @@ class Verify extends React.PureComponent<Props, State> {
   public render (): React.ReactNode {
     const { t } = this.props;
     const { cryptoOptions, cryptoType, data, defaultPublicKey, isHexData, isValid, isValidAddress, isValidSignature, signature } = this.state;
+    const AlignedIcon = styled(Icon)`
```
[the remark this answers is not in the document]
touhonoob: Hi @jacogr 
Thanks for the review:
1. moved to file-level
2. it's already adjusted to 32px by setting font-size as 24pt
Screenshot from 2019-10-07 22-29-15

packages/app-toolbox/src/Verify.tsx
```diff
@@ -206,4 +208,17 @@ class Verify extends React.PureComponent<Props, State> {
...
+        position: relative;
+        left: 0.75rem;
+        top: 1rem; 
+        width: 32px;
+        height: 32px;
+        font-size: 24pt;
+        background: white !important;
+        border-radius: 50%;
+      }
+    `;
```
jacogr: Could we indent to 2 speces max, i.e.

```js
const AlignedIcon = styled(Icon)`
  &&&.ui--AlignedIcon::before {
  ...
  }
`;
```
```

File: packages/app-toolbox/src/Verify.tsx [hunk 3]
```diff
@@ -206,4 +208,17 @@ class Verify extends React.PureComponent<Props, State> {
...
 }
 
+const AlignedIcon = styled(Icon)`
+      &&&.ui--AlignedIcon::before {
+        position: relative;
+        left: 0.75rem;
+        top: 1rem; 
+        width: 32px;
+        height: 32px;
+        font-size: 24pt;
```
jacogr: Can we stick to `32px` here (i.e. no need to convert-in-the-head between the 2 measurements)

packages/app-toolbox/src/Verify.tsx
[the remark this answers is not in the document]
touhonoob: I moved the icon a bit to the right side. It should look better now.
Selection_162

packages/app-toolbox/src/Verify.tsx
touhonoob: fixed

packages/app-toolbox/src/Verify.tsx
touhonoob: fixed

touhonoob:
@jacogr `&&&.ui--AlignedIcon i.big.icon` doesn't work as `&&&` here refers to the `i` element

jacogr:
Good point, but we still need to apply it to the actual `i.big.icon` and not just the `::before` of the element.

touhonoob:
@jacogr Got it. This should be fixed via 64c3f3eb3b47173145cf1f567d40e39c5007bf96

Issue #1676, opened 2019-09-27
React i18n, with output usable for translators


jacogr:
**Background:**

Since the start this repo was developed with i18n in mind. It uses [react-i18next](https://react.i18next.com/) as a framework, with 99% of all strings wrapped in a translatable form, e.g https://github.com/polkadot-js/apps/blob/master/packages/app-accounts/src/Overview.tsx#L57

(In the above we are indicating 99%, since there are some floating string we def. missed - they should really be included as well, however not in the scope here)

Having strings translatable, without being extracted however does not quite help. In the same vein having strings extracted but without proper categorization to help along translators, also doesn't help. And this is the scope of this issue...

In short: we want to have the strings extracted (preferably via toolset) and ordered and made available in such a way that translators can actually perform their tasks without friction.

**Prior work:**

For background, this was the original issue logged - https://github.com/polkadot-js/apps/issues/1421 (We split this one out to give clarity to the requirements and attach a bounty to it)

Some initial work has been done in https://github.com/polkadot-js/apps/pull/1478

The end-result is promising, but we are sitting at step 1 - we have the data, but it is overwhelming. Rather instead of a list of a bunch of strings https://github.com/polkadot-js/apps/pull/1478/files#diff-b9def2b395fcf7945c9622f2afb43520R8 we would like something better.

**Actions:**

- With the i18next or react-i18next toolsets, exact the translation strings (like attempted above)
- Unlike the above output, at the very least we would like this -
  - (a) containing all the strings (the above does that)
  - (b) categorized at the very least on a per-app basis (it could be more granular, but not up to component level, since translators would typically look at a single screen to see where things fit)

There really is no specific tools that "must be used" here. Rather, anything that meets the goals can be used - be it the tools used in the above attempt, any additional i18next extraction tools (discarding the above attempt) or a combination thereof with scripts. As long as the output is manageable for translation and can be used inside the UI.

Any additional comments/suggestions as to how to make it easy for the "end-user" (being the translator) would also be appreciated.

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 300.0 DAI (300.0 USD @ $1.0/DAI)  attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/polkadot-js/apps/1676/3520).
[...]

PR #1710, opened 2019-10-10
Added parser to extract i18n strings

eswarasai:
### Description
- [x] Added extractor for i18n strings
- [x] Categorized i18n strings on a per-app basis
- [x] Loading i18n `locales` from the public folder

Fixes: #1676


eswarasai:
@jacogr - I've used `i18next-scanner` to extract the i18n strings as it provides custom handlers which we can leverage to replace the strings in the files to include namespaces.

All I need for us is to decide on the namespace separator we're going to use in the codebase. All the packages names are namespaces at the moment. It can be an issue while extracting if we use namespaces like semi-colon or question mark as the strings can contain them. I was wondering to have a bit more structured one's like double hyphen, for example, `app-accounts--Account forgotten`. Let me know your thoughts on this. Thanks!

jacogr:
Yes! This is exactly on the right track. Each app is nicely organized. This is now actually workable since a translator can work through each.

So how does the namespace come into play? Where do these changes need to be made? I would be happy with a `--`, assuming everything is just transparently handled for us.

eswarasai:
@jacogr - Cool, glad to know I'm on the right track. Thanks for pointing me to `withTranslations` usage. I'm trying to leverage that with these namespace translations I've generated. Can probably avoid the namespace separator `--` altogether.

I'll update you here once everything is in place 🙂

eswarasai:
@jacogr - Okay, so I had to move these `locales` into `packages/apps/public` folder for them to load via the `i18n` config via `Backend` module and leverage the `withTranslations` appropriately for each and every package.

While generating these locales, by default, all the keys are being saved in `ui.json` and into their respective `{{namespace}}.json`. Can you please test the `withTranslation` in `translate.ts` by removing the `ui` in the namespace array and see if everything works as expected? Thanks!

eswarasai:
@jacogr - No worries on the review front 😅 

In order for you to test, the locale's being loaded appropriately, you can check the Network tab for `apps.json`, `ui.json` etc. being loaded. By default, it is English language locale that's being fetched.

For supporting other languages, I'm not sure if using `i18next-scanner` we can generate the locale files without removing the value for the existing keys or else, it'll be difficult to actually do the translations.

I'll check both of these and update you first thing tomorrow 🙂

eswarasai:
@jacogr - I've removed the `ui` namespace from being loaded/used across all the packages. While generating the locale files using `i18next-scanner`, the default namespace is `ui` meaning all the i18n strings will be saved by default in `locales/{{lng}}/ui.json`. But this file is never loaded on the UI as I've removed it from the actual `react-i18n` namespaces configuration.

Also, I've tested with generating locale for different languages via the script as well as changing the language programatically using `i18n.changeLanguage()` and can see the locale files of that language being loaded in the Network tab and rendered as well.

PS: Re-running the `yarn i18next-scanner` command after generating the locale files won't update the value of the key in the namespace in case it's changed. These scenarios will be logged onto the console which will be very helpful for debugging purposes.

Let me know if there's anything else needs to be taken care of before merging this one 🙂

jacogr:
@eswarasai This sounds really good. So does this mean that the `ui` namespace is not translated? (i.e. the strings in `react-components`?)

I've actually just cloned when I saw your updated, so starting to play now.

eswarasai:
@jacogr - `react-components` itself is a namespace, so all it's translations are within this. `ui.json` is just a default locale that's being generated by the `i18next-scanner` which can be ignored honestly.

jacogr:
Checking in as well, it works for react-components -

image

eswarasai:
Yep. Each namespace belongs to it's own package name.

eswarasai:
@jacogr - Just merged master and resolved the conflict. Let me know if there's anything else. Thanks!

PR #1726, opened 2019-10-14
Enum, Struct & Tuple to func. components

jacogr:


jacogr:
This one looks beyond saving - it creates absolute havoc. Will retry with smaller chunks.

Issue #1795, opened 2019-10-28
JS console is not working


jacogr:
i.e. this doe not execute -

```js
const events = await api.query.system.events(); // .at('0x3db7fb0c4b6a18a29b53bb177d5b897fd636a35397c45867cf2f490bffdde535');
console.log(events)
```

jacogr:
... rather... only running/non-running not showing

CaiYiLiang:
Hi @jacogr 

I think this command works. 

![image](https://user-images.githubusercontent.com/14360297/67729550-64654280-fa56-11e9-9bfb-4590b282ee20.png)

jacogr:
Yes. It started working. I had issues around the same time with the hosted node. 

However the run/stop display button display def. does not work, but that is minor/cosmetic.

jacogr:
Closed in #1804 - Once again, thanks an absolute million!

PR #1804, opened 2019-10-29
Js console

CaiYiLiang:
Hi @jacogr 
I think the issue #1795 is caused by the `fn _stop()`, set the running `true` then switch to `false`.
I have modified a bit to enable the display of stop/play button.

While I think the best solution is to switch the  isRunning-state according to exec result.


jacogr:
Thanks a million for this.

Yes, I agree, the best would be to swap to the execution result - currently what happens is that something does complete (i.e. a single entries & log) and then it still shows running. So it really is not close to optimal.

PR #1861, opened 2019-11-09
Use most recent contract from result block

kwingram25:
Needs https://github.com/polkadot-js/api/pull/1559


packages/app-contracts/src/Deploy.tsx
```diff
@@ -292,7 +292,7 @@ class Deploy extends ContractModal<Props, State> {
     const { api, history } = this.props;
 
     const section = api.tx.contracts ? 'contracts' : 'contract';
-    const record = result.findRecord(section, 'Instantiated');
+    const record = result.findRecord(section, 'Instantiated', true);
```
jacogr: So looking at this and the API version - we are being naive. We should have a `filterRecords` with an Array returned.

PR #1892, opened 2019-11-15
Map non-account stashes

jacogr:
image

- Unknown, non-account stash now shows.
- there are gaps -
  - when you don't have the controller, the controller options will also show in the menu
  - when you don't have the stash, the stash options will also show in the menu

... but it gets it going.


TriplEight:
:+1:

PR #1924, opened 2019-11-22
Fix Vec<u8> issue

kwingram25:
-Fix Vec<u8> ABI display
-Fix issue with Struct param display


packages/react-params/src/Param/Struct.tsx
```diff
@@ -18,8 +18,13 @@ export default function StructParam (props: Props): React.ReactElement<Props> {
   const [params, setParams] = useState<ParamDef[]>([]);
 
   useEffect((): void => {
-    const rawType = createType(registry, type.type as any).toRawType();
-    const typeDef = getTypeDef(rawType);
+    let typeDef;
+    try {
+      const rawType = createType(type.type as any).toRawType();
```
jacogr: This needs to pass the registry, i.e. `createType(registry, type.type)`

packages/react-params/src/Param/Struct.tsx
```diff
@@ -18,8 +18,13 @@ export default function StructParam (props: Props): React.ReactElement<Props> {
   const [params, setParams] = useState<ParamDef[]>([]);
 
   useEffect((): void => {
-    const rawType = createType(registry, type.type as any).toRawType();
-    const typeDef = getTypeDef(rawType);
+    let typeDef;
+    try {
+      const rawType = createType(type.type as any, registry).toRawType();
```
jacogr:
```suggestion
      const rawType = createType(registry,  type.type as an).toRawType();
```

PR #1927, opened 2019-11-22
Refactor Backup component with hooks

LukeSugiura:


LukeSugiura:
Thank you for sharing the background.

I thought refactoring helps me understand the source code a bit better. Let me know if reviewing this kind of PR is getting your way too much.

P.S It may be worth considering form libraries like Formik since common logic (e.g. touched, errors, validation, submit) can be repetitive.

jacogr:
As always, thank you very much! It does exactly what it says on the tin.

PR #1930, opened 2019-11-23
Fix Dropdown component

LukeSugiura:
We should eventually convert all components that pass "defaultValue" prop to Dropdown component with "value" prop

### DIFF
--- a/packages/app-settings/src/General.tsx
+++ b/packages/app-settings/src/General.tsx
@@ -79,48 +79,48 @@ function General ({ className, isModalContent, onClose, t }: Props): React.React
         <>
           <div className='ui--row'>
             <Dropdown
-              defaultValue={prefix}
               help={t('Override the default ss58 prefix for address generation')}
               label={t('address prefix')}
               onChange={_handleChange('prefix')}
               options={prefixOptions}
+              value={prefix}
             />
           </div>
           <div className='ui--row'>
             <Dropdown
-              defaultValue={icon}
               help={t('Override the default identity icon display with a specific theme')}
               label={t('default icon theme')}
               onChange={_handleChange('icon')}
               options={iconOptions}
+              value={icon}
             />
           </div>
           <div className='ui--row'>
             <Dropdown
-              defaultValue={uiMode}
               help={t('Adjust the mode from basic (with a limited number of beginner-user-friendly apps) to full (with all basic & advanced apps available)')}
               label={t('interface operation mode')}
               onChange={_handleChange('uiMode')}
               options={uiSettings.availableUIModes}
+              value={uiMode}
             />
           </div>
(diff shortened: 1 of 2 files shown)


LukeSugiura:
Odd behavior I noticed is that when navigating through accounts options with arrow keys, It focuses section title such as "ACCOUNTS" or "RECENT".

LukeSugiura:
- [ ] remove headers from Dropdown options and use Dropdown.Header component instead.
(https://react.semantic-ui.com/modules/dropdown/#content-header)

jacogr:
When trying to make a transfer -

image

Closed above, opened again via transfer on right side, entered a value - the submit doesn't become active -

image

In the cases where it is used to display values (in this case case block 778476 on Kusama, it does work exactly right (disabled, defaultValue being passed) -

image

LukeSugiura:
The reason why the submit doesn't become active is that initial senderId state on Transfer component remains undefined even after InputAddress component figures out the account address (_defaultValue).

One quick and dirty solution is to call onChange prop right after assigning _defaultValue.
https://github.com/polkadot-js/apps/blob/master/packages/react-components/src/InputAddress/index.tsx#L162

jacogr:
So not sure in this case that it will be only an address thing - for instance in tge Params components where Dropdowns are used for Enums. These are really tricky since the "form creator" really does not know what the full initial value should look like, Having to do so would mean that he needs to understand the full metadata structure populated. So rather, trhis is delegated to the individual components and they (in combination), _always_ fire the initial results - so this all bubbles up and gives the caller the actual initial value.

The components are dual use -

- for "normal" forms, like Transfer where the caller actually knows what he is busy with
- for "metadata" forms, like . the Extrinsic app, where, well, it can be anything as defined by the chain - the elephant is eaten inn small chunks with each component knowing how to treat "structures" of a specific type

So here, just fixing "InputAddress", but breaking other dropdowns, have consequences all-over the UI, especially in the cases where forms are not fixed, but rather dynamically generated from structures, not values.

LukeSugiura:
Thank you for the in-depth explanation!

> for "metadata" forms, like . the Extrinsic app, where, well, it can be anything as defined by the chain - the elephant is eaten inn small chunks with each component knowing how to treat "structures" of a specific type

I honestly wasn't aware of this use case. And now it started to make sense why this bottom-up approach has been deployed.

I should take time to read code around Extrinsic component and play around with democracy app to fully understand the architecture.

jacogr:
You can see the extrinsics app in action just by clicking on the left - as you select the different sections + methods, the details are filled in. (It is far from perfect, struct, especially custom and Vectors are still sort-of usable, but need some TLC)

LukeSugiura:
move to: #1937

PR #1937, opened 2019-11-24
Allow paste + enter on Dropdown

### LINKED ISSUE #1786: Allow paste + enter to select pasted address
Currently you need to select via mouse

LukeSugiura:
Close #1786 

I think I nailed it this time while respecting current form architecture.

Old RP: #1930 

Known issue: https://github.com/polkadot-js/apps/pull/1930#issuecomment-557818851
I'll try to fix this in separate PR.

### DIFF
--- a/packages/react-components/src/Dropdown.tsx
+++ b/packages/react-components/src/Dropdown.tsx
@@ -4,7 +4,7 @@
 
 import { BareProps } from './types';
 
-import React, { useEffect, useState } from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import styled from 'styled-components';
 import SUIButton from 'semantic-ui-react/dist/commonjs/elements/Button/Button';
 import SUIDropdown, { DropdownProps } from 'semantic-ui-react/dist/commonjs/modules/Dropdown/Dropdown';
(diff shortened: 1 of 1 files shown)


jacogr:
Thanks a million for the persistence - I’ll be able to give it a run-thought tomorrow.

packages/react-components/src/Dropdown.tsx
```diff
@@ -40,31 +40,37 @@ interface Props<Option> extends BareProps {
...
+      return;
+    }
+    handleOnChange(value);
+  };
 
-    // only update parent if we have had something changed
-    if (JSON.stringify({ v: newValue }) !== JSON.stringify({ v: stateValue })) {
-      _onChange(null, { value: newValue });
+  useEffect(() => {
+    if (JSON.stringify({ v: valueProp }) === JSON.stringify({ v: stateValue })) {
```
jacogr: Not 100% sure on this one - since we use the stateValue internally, it may be “captured”. So if we don’t want to fire the effect of n it, and not need it on re-render it will probably make sense to move “stateValue” to as useRef.

jacogr:
This works as expected. I am just not convinced about this `value={valueRef.current}` - it means it actually doesn't force a re-render when the ref changes. I have not seens anything weird from it, however leaves me uncomfortable. 

So it is a rock and a hard place - you don't want to use the only stateValue in the effect (it needs to be in params if used), but at the same time not sure using the bare ref as a value is 100% either.

LukeSugiura:
> it is a rock and a hard place...

Yea, I agree. Both solutions have some shortcomings.
I have some time today so I'm going to explore other approaches. Hopely find a better solution.

jacogr:
Thank you for the fix. It actually showed the underlying issues here - snuck in after the conversion to functional. (And still not perfect, but certainly better)

I really appreciate your work on this.

PR #1951, opened 2019-11-27
Replace TxComponent and TxModal (base)

kwingram25:


packages/react-components/src/Button/Button.tsx
```diff
@@ -13,50 +13,65 @@ import Tooltip from '../Tooltip';
...
-          trigger={triggerId}
-        />
-      )}
-    </>
-  );
+interface State {
+  triggerId: string;
+}
+
+export default class Button extends React.PureComponent<Props, State> {
```
jacogr: I really don't want to go back to a class-based component here.

kwingram25:
New version:

- useTx hook to manage generating extrinsic from different possible sources (identity, method string and params, etc.), setting accountId, busy state
- TxButton removed
- TxModal a "dumb" component that receives TxState generated by useTx and rendering props

Issue #1958, opened 2019-11-28
Use multi-variant enum as an extrinsic parameter


anakornk:
I'm having a problem with using multi-variant enum as an extrinsic parameter. In this example, my enum is ValueType and i'm trying to call set_enum which has ValueType enum as one of its parameters through polkadot js api. When the enum only has one variant (any of the 4), all works fine. I could also combine uint32 and uint64. However, the other combinations will report an error. For instance, when i tried having all 4 variants together, I could submit an extrinsic with uint32 and uint64, but using either Bool or Char32 will give me an error.

Enum
```
#[derive(Clone, Encode, Decode, PartialEq, Eq, RuntimeDebug)]
pub enum ValueType {
	Char32([u8; 32]),	
	Uint32(u32),
	Uint64(u64),
	Bool(bool),
}
```
```
		fn set_enum(origin, value: ValueType) -> Result {
			// Check if signed
			let sender = ensure_signed(origin)?;
			Ok(())
		}
```

Call set_enum from polkadotjs api
```
api.tx.content
    .setEnum({
      char32: "0x12345678123456781234567812345678"
    })
    .signAndSend(alicePair, { nonce }, ({ events = [], status }) => { ... }
```

Error
```
2019-11-28 12:26:15        RPC-CORE: submitAndWatchExtrinsic(extrinsic: Extrinsic): ExtrinsicStatus:: 1002: Verification Error: Execution(ApiError("Could not convert parameter `tx` between node and runtime: Error decoding field Call :: Content.0")): RuntimeApi("Execution(ApiError(\"
2019-11-28 12:26:15             DRR: Error: submitAndWatchExtrinsic(extrinsic: Extrinsic): ExtrinsicStatus:: 1002: Verification Error: Execution(ApiError("Could not convert parameter `tx` between node and runtime: Error decoding field Call :: Content.0")): RuntimeApi("Execution(ApiError(\"
    at errorHandler (/Users/anakornk/Code/graph4w3/graph-js-sdk/node_modules/@polkadot/rpc-core/index.js:259:26)
    at /Users/anakornk/Code/graph4w3/graph-js-sdk/node_modules/@polkadot/rpc-core/index.js:233:9
    at processTicksAndRejections (internal/process/task_queues.js:82:5)
```

Seems like there is something wrong with the encoding/decoding part. Am i missing anything?

jacogr:
Transferred, since there seems to be an Enum issue with params.

[...]

PR #1964, opened 2019-11-30
Rework dropdown changes

### LINKED ISSUE #1786: Allow paste + enter to select pasted address

jacogr:
- Closes #1786 
- This should also have a fix for #1958 
- Draws inspiration from #1937 

cc @LukeSugiura - I really would appreciate your eyes on this one as well.


LukeSugiura:
LGTM 🎉
That's neat to store lastUpdate ref as a string so that it is distinguishable with an undefined prop on the second time running _setStore.

What I haven't understood yet is why there are both defaultValue and value props. I looks like both props have the same behavior.

PR #1996, opened 2019-12-03
Propose tech committee motions

kwingram25:
- Add modal for proposing technical committee motions
- extend useTx to track isSubmittable


packages/app-tech-comm/src/Proposals/Propose.tsx
```diff
@@ -0,0 +1,98 @@
...
+    !!method && setMethod(method);
+  };
+
+  useEffect(
+    (): void => setThreshold([threshold, _hasThreshold(threshold)]),
+    [memberCount]
+  );
+
+  const { apiDefaultTxSudo } = useApi();
+  const txState = useTx(
```
jacogr: This whole things is just magic, no description of what is what. We will need to update this to be either an object of not having to pass `{...txState}` (more magic) to the Modal. Contrast this to TxButton (old), where it is obvious what you are passing.

packages/app-tech-comm/src/Proposals/Propose.tsx
```diff
@@ -0,0 +1,98 @@
...
+    }
+  );
+
+  return (
+    <TxModal
+      isOpen={isOpen}
+      onClose={onClose}
+      {...txState}
+      header={t('Propose a committee motion')}
+      content={
```
jacogr: Context here can just be `children` of the component. (Also removed the need for the fragment)

PR #2089, opened 2020-01-03
Bump to latest api

amaurymartiny:
Wait for Kusama 1033


amaurymartiny:
Works for me locally 🚀  cc @kwingram25

PR #2092, opened 2020-01-06
Kulupu

jacogr:
cc @sorpaas


packages/react-api/src/overrides/spec/kulupu.ts
```diff
@@ -0,0 +1,17 @@
...
+  DifficultyAndTimestamp: {
+    difficulty: 'Difficulty',
+    timestamp: 'Moment'
+  },
+  DispatchError: 'DispatchErrorTo198',
+  DispatchResult: 'DispatchResultTo198',
+  DispatchInfo: {
+    weight: 'Weight',
+    class: 'DispatchClass'
+  }
```
jacogr: cc @sorpaas - There can be removed once bumped to latest Substrate master

sorpaas: Just a small note, the bump will probably take several months due to the infrequent fork schedule. (It currently uses a [hard fork module](https://github.com/kulupu/kulupu/blob/master/anyupgrade/src/lib.rs) and has not yet transitioned to on-chain voting.)

jacogr: No issues, I saw that. No issues, the types are there, nothing will break in the meantime, we should just keep it somewhere in the memory banks to update once there are changes.

Issue #2055, opened 2019-12-15
Uncaught Error: Unable to retrieve keypair


q9f:
```
react.62389b3d.js:17 Uncaught Error: Unable to retrieve keypair '0xC...'
    at t.default (main.69fe13ec.js:21)
    at t.default.get (main.69fe13ec.js:88)
    at t.default.getPair (main.69fe13ec.js:88)
    at m.getPair (main.69fe13ec.js:88)
    at x (main.69fe13ec.js:100)
    at S.renderUnlock (main.69fe13ec.js:100)
    at S.renderContent (main.69fe13ec.js:100)
    at S.render (main.69fe13ec.js:100)
    at Ji (react.62389b3d.js:17)
    at Zi (react.62389b3d.js:17)
```

steps to reproduce:

- have distinct controller `0xC` and stash `0x5`
- bond funds for `0x5` and start validating. 
- forget the controller account `0xC`
- go the staking actions and try to change any settings for the validator `0x5`
- white page of doom

jacogr:
Yes, it currently does not try to see if you actually have the account before trying to use it. (The reverse is _slightly_ better, i.e. having the controller and not the stash. As it blatantly obvious here, was was done assuming you have everything.)

q9f:
yeah, it's an edge case. i agree usually, you should have the controller but not the stash.

jacogr:
Still needs a fix :) Thanks for reporting, the holes need to be closed.

PR #2111, opened 2020-01-13
Own controller check on staking options

jacogr:
Closes https://github.com/polkadot-js/apps/issues/2055


bobgalicia:
@jacogr Staking with a hardware wallet and the web wallet as controller, if you try to bond more funds without the hardware wallet attached the response is a white screen, not a caught error, maybe related with polkadot-js#2055

Issue #2114, opened 2020-01-14
Minified React error: after refresh, some pages show briefly and are replaced by a white screen


chevdor:
I saw this first yesterday evening.
"@sentientrue | KUSAMA Generic" seems to be running into the same or similar issues.

I am using Brave, latest version. Here are the UI versions:
```
parity-polkadot v0.7.16
api v0.100.1
apps v0.39.0-beta.47
```

The issue comes mostly (I did not check all pages/apps) on the staking page, it happened to me especially when clicking on the "account actions".

The page loads, briefly shows and everything disappear, showing a white screen.

Force refresh does not help.
The issue also shows now for me on https://polkadot.js.org/apps/#/staking

The issues triggers after an error pops up in the dev console:
```
react.de211c3b.js:17 Error: Minified React error #310; visit https://reactjs.org/docs/error-decoder.html?invariant=310 for the full message or use the non-minified dev environment for full errors and additional helpful warnings.
    at pi (react.de211c3b.js:17)
    at mi (react.de211c3b.js:17)
    at Object.vi [as useState] (react.de211c3b.js:17)
    at useState (react.de211c3b.js:9)
    at X (react.de211c3b.js:32)
    at t.useTranslation (main.f645795e.js:21)
    at main.f645795e.js:57
    at main.f645795e.js:57
    at ui (react.de211c3b.js:17)
    at Gi (react.de211c3b.js:17)
```

jacogr:
Staking page - `Rendered more hooks than during the previous render.` (check overview and account rendering here)

chevdor:
I just tested on Chrome and do NOT get the issue.

chevdor:
Even on Brave, I see the issue on a single machine...

chevdor:
This is spreading :)
Now happening on https://polkadot.js.org/apps/#/accounts as well

chevdor:
Summary of where the issue shows up: red X = issue):
- ✅ https://polkadot.js.org/apps/#/explorer
- ❌ https://polkadot.js.org/apps/#/accounts
- ❌ https://polkadot.js.org/apps/#/staking

chevdor:
I am seeing also a warning that may be related:
```
main.f645795e.js:93 Cannot find component for Extrinsic, defaulting to Unknown
e @ main.f645795e.js:93
```

jacogr:
The last one is harmless (hence a warning).

Issue probably caused by https://github.com/polkadot-js/apps/pull/2109

[...]

PR #2130, opened 2020-01-15
Connecting overlay rendering order

jacogr:


jacogr:
Same as in https://github.com/polkadot-js/apps/issues/2114 (ensure order when translation is used)

PR #2201, opened 2020-01-26
Collective voting components refactor

kwingram25:
- TxButton to functional
- Reuse components for council/tech committee


packages/react-components/src/Collective/Slashing.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+import { useTranslation } from '../translate';
+import useAvailableSlashes from './useAvailableSlashes';
+
+interface Option {
+  text: string;
+  value: number;
+}
+
+export default function Slashing ({ className, isMember, members = [] }: Props): React.ReactElement<Props> {
```
jacogr: This is def not collective - it is specific to council.

packages/react-components/src/Collective/Proposals.tsx
```diff
@@ -0,0 +1,69 @@
...
+    <div className={className}>
+      <Button.Group>
+        <Propose
+          collective={collective}
+          isMember={isMember}
+          members={members}
+          prompt={proposePrompt}
+          thresholdHelp={thresholdHelp}
+        />
+        {isCouncil && (
```
jacogr: We really don't want this in here - we want the logic for council and the way it displays, in council. As soon as we start treating both the same, we are going to run into a mess here.

jacogr:
Fo some reason the proposals are now disabled for me.

image

jacogr:
Also cannot vote on the members. Existing vs new -

image

image

kwingram25:
Hmm looks related to isMember check, I've always had it on since I only tried on testnet...

jacogr:
Closing. We only have 2 `extends TxModal` left -

- claims app (not sure why this is a modal)
- inside contracts

So all the areas here are covered.

PR #2203, opened 2020-01-27
Expose signed transactions

mzolkiewski:
@jacogr This is meant to be similar to what I've already done here: https://github.com/polkadot-js/tools/pull/79
Basically it lets you expose the signed extrinsic for later use.
Hope this looks good!


jacogr:
I took a quick look - the retrieval of all information is def. not something I would put in here. The issue atm is that the `signAsync` is not fully on feature-par with `signAndSend`, e.g. https://github.com/polkadot-js/api/blob/master/packages/api/src/submittable/createClass.ts#L85 - it however should retrieve all the nonce, era, etc. info.

mzolkiewski:
Nice, I saw your PR that addresses the issue you raised above: https://github.com/polkadot-js/api/pull/1753
It looks great and I agree that it makes much more sense to have that logic sit in there instead of the UI.
Do you need any help with fixing the tests? If not, then how soon will I be able to use this? I'm eager to update this PR with the new code :)

jacogr:
@mzolkiewski Sorry, didn't see your question since ... well, was busy fixing that test. :) Will try to get it in ASAP.

mzolkiewski:
@jacogr Wow, really appreciate the quick turnaround!
I've updated my PR so please give it another look.

mzolkiewski:
Hi @jacogr just wondering if you got a chance to take another look?

jacogr:
It is on the list, sorry for the delay. Have not played yet, just took a brief look to get a feel earlier today. 

(Atm working through the other PR waiting, this one is next up)

packages/app-accounts/src/modals/Transfer.tsx
```diff
@@ -127,6 +127,18 @@ function Transfer ({ className, onClose, recipientId: propRecipientId, senderId:
...
       </Modal.Content>
       <Modal.Actions onCancel={onClose}>
+        <TxButton
+          accountId={senderId}
+          extrinsic={extrinsic}
+          icon='sign-in'
+          isBasic
+          isDisabled={!hasAvailable}
+          isSign
+          label={t('Sign Transaction')}
```
jacogr: I am actually a bit worried putting this directly only the transfer modal - the next would be to add it to all modals.

 It just seems saner to have it specifically on the actual signer. So instead of pushing through `isSign` on the specific functions (which is _everywhere_), if we put it on the signer Modal (which, is a mess) it can either go and submit or just sign instead of submitting.

packages/app-extrinsics/src/Selection.tsx
```diff
@@ -63,6 +63,16 @@ class Selection extends TxComponent<Props, State> {
             withSpinner
           />
           <Button.Or />
+          <TxButton
+            accountId={accountId}
+            isBasic
+            isDisabled={!isValid}
+            isSign
```
jacogr: Same comment as transfer - we have a lot of modals that all do submission, here we cover only a fraction of them.

packages/react-components/src/Modal.tsx
```diff
@@ -19,7 +19,9 @@ interface ModalProps extends BareProps {
 }
 
 interface ActionsProps extends BareProps {
+  cancelLabel?: string;
```
jacogr: `cancelLabel?: React.ReactNode`

Having said that, I am not sure I would like this override, prefer the simplicity of consistency.

packages/react-signer/src/Modal.tsx
```diff
@@ -40,25 +45,31 @@ interface Props extends I18nProps, ApiProps, BaseProps {
 }
 
 interface State {
+  accountNonce?: string;
```
jacogr: Not 100% sure here why we have `accountNonce` and then `nonce` later on. It could be a hold-over from somewhere.

packages/react-signer/src/Modal.tsx
```diff
@@ -146,12 +167,16 @@ class Signer extends React.PureComponent<Props, State> {
...
 
-  public async componentDidUpdate (): Promise<void> {
+  public async componentDidUpdate (_: Props, prevState: State): Promise<void> {
     const { currentItem } = this.state;
 
     if (currentItem && currentItem.status === 'queued' && !(currentItem.extrinsic || currentItem.payload)) {
       return this.sendRpc(currentItem);
     }
+
+    if (currentItem?.isSign && currentItem?.accountId && currentItem.accountId !== prevState.currentItem?.accountId) {
```
jacogr: Would probably not do this. Rather just rely on the `api` (which also needs to re-enable this one https://github.com/polkadot-js/api/pull/1525)

packages/react-signer/src/Modal.tsx
```diff
@@ -183,25 +206,15 @@ class Signer extends React.PureComponent<Props, State> {
...
     return (
-      <Modal.Actions
-        onCancel={
-          isQrVisible
-            ? this.onCancelQr
-            : this.onCancel
-        }
-      >
-        {!isRenderError && (!isQrVisible || !isQrScanning) && (
+      <Modal.Actions cancelLabel={signedTx ? t('Close') : undefined} withOr={!signedTx} onCancel={isQrVisible ? this.onCancelQr : signedTx ? this.onCancelSign : this.onCancel}>
```
jacogr: Ahhh... ignore the previous comment on `Cancel`, I see you have cleaned this on up. All ok.

packages/react-signer/src/Modal.tsx
```diff
@@ -183,25 +206,15 @@ class Signer extends React.PureComponent<Props, State> {
...
+        {!isRenderError && (!isQrVisible || !isQrScanning) && !signedTx && (
           <>
             <Button.Or />
             <Button
-              className='ui--signer-Signer-Submit'
+              className="ui--signer-Signer-Submit"
```
jacogr:
```suggestion
              className='ui--signer-Signer-Submit'
```

actually not sure why eslint doesn't complain about " vs '

packages/react-signer/src/Modal.tsx
```diff
@@ -212,17 +225,11 @@ class Signer extends React.PureComponent<Props, State> {
...
-              }
-              icon={
-                isQrVisible
-                  ? 'qrcode'
-                  : currentItem.isUnsigned
-                    ? 'sign-in'
-                    : isExternal
-                      ? 'qrcode'
-                      : 'sign-in'
+                        : currentItem.isSign
```
jacogr: So here would have the "sign (no submission)" button, along with the others - instead of sending this through via the queues. So you can basically "sign & send" or "only sign" on any tx.

packages/react-signer/src/Modal.tsx
```diff
@@ -414,7 +457,19 @@ class Signer extends React.PureComponent<Props, State> {
     }
 
     return this.sendExtrinsic(currentItem, password);
-  }
+  };
+
+  private updateNonce = async (): Promise<void> => {
```
jacogr: If we insist on doing this, the nextIndex is actually the right thing to use.

packages/app-accounts/src/modals/Transfer.tsx
mzolkiewski: Yeah I wanted to do it like that at first but the problem is that the signer needs to show extra fields for nonce and blocks when you want to just sign so there would have to be an extra step, like:
- click 'sign' in the signer
- this puts the signer in 'sign' mode and shows extra fields
- now you click 'sign' again and this actually starts the signing process

But from what you're saying this is probably the lesser evil so I'm gonna do something like that.

packages/react-signer/src/Modal.tsx
mzolkiewski: This is for the hint in the nonce input. It shows the user the current nonce of the account and is the default value for `nonce` but you can change `nonce` to whatever you want

packages/react-signer/src/Modal.tsx
mzolkiewski: Actually, the ability to edit the nonce is very useful - if you e.g. sign 3 transactions one after another then they need to have different nonces for each of them to work once submitted, so I wouldn't just automatically use the current nonce from the api.

packages/react-signer/src/Modal.tsx
mzolkiewski: I remember there's an exception to the `'` vs `"` which is this exact case - plain strings as classNames. I'll see if I can force my eslint to use single quotes everywhere.

packages/app-accounts/src/modals/Transfer.tsx
mzolkiewski: Makes sense - moved the logic inside the signer Modal.

mzolkiewski:
@jacogr All right, that should do it. Please check it out again.

jacogr:
Will play through.

jacogr:
Some comments as playing through (will edit, just need a place to jot stuff down) -

1. I am getting this when trying to sign an account from an extension -

image

2. When on the signing page, I have no way of getting back to the original. (I actually recommend a toggle rather - bottom left, that toggles between the 2)

3. The button looks very out of place (this can be fixed with 2 above)

... back to playing

mzolkiewski:
Sorry about that, I was testing with extensions only.
Good idea with the toggle - UX is much better now I think.

Hopefully it's all good now.

jacogr:
Thank you, I'll grab this first thing in the morning :)

mzolkiewski:
That's great to hear. Thanks for all the feedback and looking forward to seeing this merged :)

PR #2225, opened 2020-02-02
Change dockerfile to use local version as opposed to cloning polkadot…

philipstanislaus:
…-js/apps

The current behavior does not allow the dockerfile to be used with forked repositories.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,7 +8,7 @@ RUN curl -sL https://deb.nodesource.com/setup_10.x | bash -
 RUN apt-get install -y nodejs
 
 WORKDIR /app
-RUN git clone https://github.com/polkadot-js/apps
+COPY . .
 
 WORKDIR /app/apps
 RUN npm install yarn -g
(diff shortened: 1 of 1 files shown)


philipstanislaus:
For some reason, the copy step in the second stage of the docker container fails. Any idea why that might be? 

```bash
Step 13/15 : COPY --from=builder /app/apps/packages/apps/build /var/www/html
COPY failed: stat /var/lib/docker/overlay2/a8e25cee2025ed979e38854c9603aeafe705c0e6ea5be6b3931c11fcedfac3bd/merged/app/apps/packages/apps/build: no such file or directory
```

jacogr:
@philipstanislaus No idea what causes it, honestly have not played with this for quite some time now. Will take a look though, it should be something small

jacogr:
Ok, so believe it is this line - `COPY --from=builder /app/apps/packages/apps/build /var/www/html`

When cloning, it clones it into the `apps` directory, so adjusting the above to the following should work (I believe) - `COPY --from=builder /app/packages/apps/build /var/www/html` (Replacing `/app/apps` with `/app`)

jacogr:
Thanks a million! (I just applied my small updates to the branch)

philipstanislaus:
Awesome, thanks @jacogr for the fast turnaround. Just tested this locally, works like a charm.

PR #2231, opened 2020-02-03
Build only apps component via docker

jacogr:


jacogr:
@philipstanislaus This one should speed up the build by a bit, since it only compiled the actual webpack stuff. (However for a clean build most time is probably spent elsewhere... but every bit helps)

philipstanislaus:
It implies though that you run `yarn build` before `docker build .`, right?

jacogr:
No. The issue is that the yarn build builds all the packages for publishing in addition to the apps component (which is via webpack). 

Effectively it doubles up since the webpack build does all the compilation for the stuff it needs. 

Here it just builds what is needed. (Running yarn clean in root before this will work)

philipstanislaus:
Sounds good!

Issue #2243, opened 2020-02-06
First endpoint in `packages/app-settings/src/available/endpoints.ts` should be the default node


philipstanislaus:
We have a custom list of endpoints in https://github.com/centrifuge/polkadot-js-apps/blob/master/packages/app-settings/src/available/endpoints.ts , but the Portal still loads flaming fir as the default node. 

It would be great if the default node would be the first entry in `packages/app-settings/src/available/endpoints.ts` instead. 

You can see this behaviour on https://portal.chain.centrifuge.io/

jacogr:
Build with `NODE_ENV=production WS_URL=wss://<my url> yarn build` and it should pull in the default as specified as part of the `WS_URL` environment.

Need a better solution there.

jacogr:
The above probably won't work, just took a peek - it _will_ work, but seem to always default to that URL then. Which, well, it not optimal, doesn't seem to be change-able.

philipstanislaus:
@jacogr thanks, I meant actually something like this: https://github.com/polkadot-js/apps/pull/2250

jacogr:
Should be fixed with this - https://github.com/polkadot-js/apps/blob/master/packages/apps/src/initSettings.ts#L21

jacogr:
@philipstanislaus Could you please give it a play-through to see if it indeed does what it is supposed to? Hopefully got it right.

philipstanislaus:
Just verified – works as expected. Many thanks!

PR #2244, opened 2020-02-06
Apply WS_URL consistently

jacogr:
Don't clobber any user-definitions with WS_URL (as per https://github.com/polkadot-js/apps/issues/2243#issuecomment-582832220)


packages/apps/src/index.tsx
```diff
@@ -27,7 +26,7 @@ const rootElement = document.getElementById(rootId);
 //  - http://localhost:3000/?rpc=wss://substrate-rpc.parity.io/#/explorer
 //  - http://localhost:3000/#/explorer?rpc=wss://substrate-rpc.parity.io
 const urlOptions = queryString.parse(location.href.split('?')[1]);
-const _wsEndpoint = urlOptions.rpc || process.env.WS_URL || settings.apiUrl;
+const _wsEndpoint = urlOptions.rpc || settings.apiUrl;
```
philipstanislaus: Wouldn't it make more sense to use `process.env.WS_URL` at the end of the list, i. e. 

```ts
const _wsEndpoint = urlOptions.rpc || settings.apiUrl || process.env.WS_URL;	
```

jacogr:
It is already applied as part of settings

PR #2250, opened 2020-02-06
Add first available endpoint as default endpoint

philipstanislaus:


jacogr:
This should solve the issue - https://github.com/polkadot-js/apps/pull/2247/files#diff-7cc16d38cb134f88622649168306b728R21

(The issue is that the ui-settings already apply WS_URL and tries defaults - with the above it check for WS_URL or if not set, check if something is stored - if not, it will fallback to the first in the list)

philipstanislaus:
Sounds good, thanks for diving into this!

Issue #2285, opened 2020-02-18
Address prefix setting is no longer respected 


joshua-mir:
Likely due to #2282 - changing the address prefix away from the default no longer appears to have any effect - the node default is always used regardless of the setting.

jacogr:
Ahhh, not display, copy. That bloody thing went in went with half-way since the deps were bumped by mistake. (The above is actually a fix for 1 issue, another one needs to be applied).

PR #2286, opened 2020-02-18
Set registry ss58Format based on overrides

jacogr:
Closes https://github.com/polkadot-js/apps/issues/2285


packages/react-api/src/Api.tsx
```diff
@@ -75,6 +75,9 @@ async function loadOnReady (api: ApiPromise): Promise<State> {
 
   console.log('api: found chain', systemChain, JSON.stringify(properties));
 
+  // explicitly override the ss58Format as specified
+  registry.setChainProperties(createType(registry, 'ChainProperties', { ...properties, ss58Format }));
```
jacogr: Actual fix - since it is not populated by the chain properties by the API

PR #2311, opened 2020-02-28
Bump deps

jacogr:


antonmedv:
Hi, I see you use codeflask. Have you seen [CodeJar](https://medv.io/codejar/)? It has much better browser support on mobile devices.

PR #2324, opened 2020-03-02
Add extra Input types to the UI

### LINKED ISSUE #2380: Add input option for fixed point decimals
Some pallets use fixed point values like `Perbill` and `Fixed64`. It would be good to be able to enter these values as float values in the UI.

apopiak:
Hacky way of adding a new input type `Float` that allows entering float values with integer and fraction parts separated by `,` or `.` and uses it to allow entering `Perbill` and `Fixed64`.

(Mis-)Uses the SI facilities to multiply the values by 1 billion. The SI multipliers are thus used differently for float versus integer value entry. It is unclear to me why the integer version works.

Also adds support for `VecDeque`.

Closes  #2380


apopiak:
I don't understand the CI errors. `formatBalance` is used in the file in question. o.O

packages/react-params/src/Param/findComponent.ts
```diff
@@ -60,7 +62,7 @@ const componentDef: TypeToComponent[] = [
   { c: Text, t: ['String', 'Text'] },
   { c: Struct, t: ['Struct'] },
   { c: Tuple, t: ['Tuple'] },
-  { c: Vector, t: ['Vec'] },
+  { c: Vector, t: ['Vec', 'VecDeque'] },
```
jacogr: Would not add this, the API does not handle these at all (so there is no proper TypeDef for these for actual proper parsing) - just for the simple reason that it is not in Substrate master.

apopiak: ok

jacogr:
Can you please merge in master - after the Kusama update I cannot just run through testing there. (And neither on substrate master). Sorry, should have shouted 1t 6pm when I tried)

PR #2337, opened 2020-03-06
Reordering in InputAddressMulti

### LINKED ISSUE #2330: Phragmen ordering
With the phragment selection for both -

- council candidates
- validators to nominate

The order is becoming important. In addition to selection, the selection needs to be able to be ordered before submission.

kwingram25:
Closes #2330

- Add react-beautiful-dnd
- Allow reordering of addresses for InputAddressMulti component


jacogr:
Don't believe the toggle and "marking inactive" should be there, so will be adjusting that. (Since we have 2 lists, the toggle doesn't make much sense here)

Issue #2347, opened 2020-03-08
Doesn't work on Firefox


xlc:
Firefox 73.0.1 (64-bit)

```
TypeError: u.startsWith is not a functionmain.9edf9b4b.js:85:106806
    <anonymous> https://polkadot.js.org/apps/main.9edf9b4b.js:85
    f https://polkadot.js.org/apps/runtime.25bdf284.js:1
    <anonymous> https://polkadot.js.org/apps/main.9edf9b4b.js:57
    f https://polkadot.js.org/apps/runtime.25bdf284.js:1
    <anonymous> https://polkadot.js.org/apps/main.9edf9b4b.js:57
    f https://polkadot.js.org/apps/runtime.25bdf284.js:1
    <anonymous> https://polkadot.js.org/apps/main.9edf9b4b.js:57
    f https://polkadot.js.org/apps/runtime.25bdf284.js:1
    r https://polkadot.js.org/apps/runtime.25bdf284.js:1
    t https://polkadot.js.org/apps/runtime.25bdf284.js:1
    <anonymous> https://polkadot.js.org/apps/main.9edf9b4b.js:1
```

PR #2348, opened 2020-03-09
fix  could not connect after clearing the cache

sakutaro220:
Fixed a problem where the connection destination URL could not be obtained after clearing the cache or when connecting for the first time.

Maybe the cause of #2347


jacogr:
Thanks for this, even through not used. You really helped to zoom in exactly where it was. I really do appreciate it.

sakutaro220:
You're welcome!

PR #2360, opened 2020-03-10
feat: add ability to export network specifications from network setting screen

hanwencheng:
It would be nice if we could have an additional button to export the current network specifications to QR code, which could be used for Signer or otther places.

It rely on the PR on Polkadot.js/UI for display the networkSpecs QR code : https://github.com/polkadot-js/ui/pull/299

### How it works
After node connected, user could click the button, and the current network specs will be shown.

User could set the color with a 6-digit hex text input, or use a `generate random` button.

### How to Test

Sidebar Select "Settings" Button -> Select "Metadata" Tab

Screen Shot 2020-03-26 at 18 20 41


packages/apps/public/locales/zh/app-settings.json
```diff
@@ -16,6 +16,7 @@
   "default interface language": "默认接口语言",
   "Cancel": "取消",
   "Save & Reload": "保存并重载",
+  "Export Network Specs": "导出网络配置",
```
jacogr: Thanks an absolute million for this! The cn translation is lagging atm, after the initial push, there has been no updates. :(

hanwencheng:
yeah, it is work in progress, I will update it in the weekend.

packages/apps/public/locales/zh/app-settings.json
hanwencheng: I will open another PR to add some Chinese Translations, and keep this PR focus.

packages/page-settings/src/modals/NetworkSpecs.tsx
```diff
@@ -0,0 +1,194 @@
...
+      <Input
+        className='full'
+        help={t('Decimals decides the smallest unit of the token, which is 1/10^decimals')}
+        isDisabled
+        label={t('Decimals')}
+        value={networkSpecs.decimals}
+      />
+      <QrNetworkSpecs className='settings--networkSpecs-qr' networkSpecs={debouncedQrData}/>
+    </Modal.Content>
+    <Modal.Actions onCancel={onClose} withOr={false}>
```
hanwencheng: One thing I am not sure is that how do I use `<Modal.Actions>` with only a "cancel" button, lint gives warning that there is no children

hanwencheng:
The idea is if user is using their own node,  and want to export the network information to Signer (related [issue](https://github.com/paritytech/parity-signer/issues/302) )  they will need to get the networrk info. It is convienient and trusted to directly use Polkadot.js to get these info.

Network info mainly can not be changed, whereas user is still capable to change name, path id and color.

jacogr:
Ok, the issues is that all that info is available in chain properties, ss58, decimals, etc. - yet we expect the user to input these? Does not make any sense to me.

hanwencheng:
`Gensis Hash`, `Prefix`, and `Decimals` are disabled Input Component, user cannot input them. I used Input component as convinieent and though it has same styling with the previous component like `name`, `path Id` and `color`.

jacogr:
Ok, so what is color meant to be? PathId? 

(Colors are aligned with Polkascan based on chain only in apps, pathId has no meaning)

hanwencheng:
I did not realize that there is already `ui-highlight-border` class with chain color, now I use it as default color if it is a known network with 451fd12 . Color is the same as it is using here.  User can generate custom color for it in case it is a custom node that not unknown to us.

Path ID is special for Signer, when user derive a network for an account, the account will automatically prefixed with its `pathId`, e.g In Kusama network, path id is `kusama`, so every account derived will be prefixed with `//kusama`, like `//kusama//funding`, `//kusama//staking`, etc.

jacogr:
The "special to signer" stuff I would remove here. Atm it has not context in the Apps UI (paths are on the list somewhere here, as in the extension, but it really is app-specific. Passing stuff along is fine, but it needs to be known on both sides)

PS: Sorry fo the stoopid questions earlier, I missed the initial screenshot - and then (mistake #2), also missed the `isDisabled` in the code. So comedy of errors on my side.

hanwencheng:
Removed the pathId in commit  71effda , building test cannot pass because polkadot-js/ui#299 is not merged.

packages/apps-config/src/ui/general/index.ts
```diff
@@ -37,6 +37,14 @@ const nodeColors: Record<string, any> = [
   [node.toLowerCase().replace(/-/g, ' ')]: color
 }), {});
 
+function sanitize (value?: string): string {
+  return value?.toLowerCase().replace('-', ' ') || '';
+}
+
+export const getSystemChainColor = (systemChain: string, systemName: string): string | undefined => {
```
jacogr: 1. Move to ui/index.ts (alongside other utility)
2. use `export function (...): string | undefined { ... }`

packages/page-settings/src/General.tsx
```diff
@@ -13,6 +14,7 @@ import uiSettings, { SettingsStruct } from '@polkadot/ui-settings';
 import { useTranslation } from './translate';
 import { createIdenticon, createOption, save, saveAndReload } from './util';
 import SelectUrl from './SelectUrl';
+import NetworkSpecs from './modals/NetworkSpecs';
```
jacogr: Probably not in here. But under Settings we now that the `Extensions` tab which deals with metadata injection. (Wrongly named for this use-case as well) We can just drop the modal completely and always display the QR there.

packages/page-settings/src/General.tsx
```diff
@@ -68,6 +72,8 @@ export default function General ({ className, isModalContent, onClose }: Props):
     },
     [settings]
   );
+  const _showNetworkSpecsQr = (): void => setQrOpen(true);
```
jacogr: Can just the the `useToggle` hook from `@polkadot/react-hooks`

packages/page-settings/src/modals/NetworkSpecs.tsx
```diff
@@ -0,0 +1,180 @@
...
+  let color = '#';
+  for (let i = 0; i < 6; i++) {
+    color += letters[Math.floor(Math.random() * 16)];
+  }
+  return color;
+}
+
+const buildNetworkSpecs = async (api: ApiPromise, systemChain: string, systemName: string): Promise<Partial<NetworkSpecsStruct>> => {
+  const [properties, blockHash] = await Promise.all([
+    api.rpc.system.properties(),
```
jacogr: This is already available on the registry, `registry.ss58Format`, etc.

packages/page-settings/src/modals/NetworkSpecs.tsx
```diff
@@ -0,0 +1,180 @@
...
+    api.rpc.chain.getBlockHash(0)
```
jacogr: Already on the api as `api.genesisHash`. `const { api } = useApi(); console.log(api.genesisHash.toHex());`

packages/page-settings/src/modals/NetworkSpecs.tsx
```diff
@@ -0,0 +1,180 @@
...
+  ]);
+  const DEFAULT_DECIMALS = createType(registry, 'u32', 12);
+  const DEFAULT_SS58 = createType(registry, 'u32', addressDefaults.prefix);
```
jacogr: Don't need to create these. Use from registry, if not available override via `||`

packages/page-settings/src/modals/NetworkSpecs.tsx
```diff
@@ -0,0 +1,180 @@
...
+      ...state,
+      ...delta
+    };
+    setQrData(newState);
+    return newState;
+  };
+  const [networkSpecs, setNetworkSpecs] = useReducer(reducer, initialState);
+
+  const { api, systemChain, systemName, isApiReady } = useApi();
+  if (!isApiReady) return null;
```
jacogr:
```suggestion
  const { api, systemChain, systemName, isApiReady } = useApi();
  
  if (!isApiReady) {
    return null;
  }
```

Spaceing and no inline returns. Additionally, all returns needs to be below all effects, states and callbacks.

packages/page-settings/src/modals/NetworkSpecs.tsx
```diff
@@ -0,0 +1,180 @@
...
+    getNetworkSpec();
+  }, [api]);
+
+  type inputListener = (v: string) => void;
+  const _onChangeValue = (k: keyof NetworkSpecsStruct): inputListener => (v: string): void => setNetworkSpecs({ [k]: v });
+  const _onSetRandomColor = (): void => setNetworkSpecs({ color: getRandomColor() });
+  const _checkColorValid = (): boolean => /^#[\da-fA-F]{6}|#[\da-fA-F]{3}$/.test(networkSpecs.color);
+
+  return <Modal
+    className={className}
```
jacogr:
```suggestion
  return (
    <Modal
      className={className}
```

packages/page-settings/src/modals/NetworkSpecs.tsx
jacogr: Was not really to be meant to be used without. For now just add an explicit close as well (to work around this), since we will combine this into the "external sources" thing and drop the modal.

packages/page-settings/src/modals/NetworkSpecs.tsx
```diff
@@ -0,0 +1,180 @@
...
+    </Modal.Content>
+    <Modal.Actions onCancel={onClose} withOr={false}>
+    </Modal.Actions>
+  </Modal>;
+}
+
+interface ColoredComponentProps extends BareProps {
+  color: string;
+}
+const ColoredComponent = React.memo(styled(({ color, className }: ColoredComponentProps) => (<div color={color} className={className}/>))`
```
jacogr: Prefer not doing this. Just add a class on main.

packages/apps-config/src/ui/general/index.ts
hanwencheng: which `ui/index.ts` file, could you please name the full path? in `apps`?

packages/page-settings/src/modals/NetworkSpecs.tsx
hanwencheng: what do you mean `on main`?

packages/apps-config/src/ui/general/index.ts
jacogr: Here, there is a util in there already - https://github.com/polkadot-js/apps/blob/master/packages/apps-config/src/ui/index.ts#L7

packages/page-settings/src/modals/NetworkSpecs.tsx
jacogr: Exposed the stuff I use for the extension as a hook - you can just pull the fields from there, then we don't have duplication.

https://github.com/polkadot-js/apps/pull/2460

packages/page-settings/src/modals/NetworkSpecs.tsx
jacogr: The styled components end up being less than optiomal. Have to deal with that in the extension, here have managed to keep it (somewhat) into control. So there is nothing wrong with -

```
<div className={className}>
  ...
  <div className='sub'>...</div>
</div>

export React.memo(styled(Main)`
...
.sub { ... }
```

packages/apps-config/src/ui/general/index.ts
hanwencheng: thanks, I see it, just updated with `master`

packages/page-settings/src/modals/NetworkSpecs.tsx
jacogr: We just need to add color in there, that is missing for you.

packages/page-settings/src/modals/NetworkSpecs.tsx
hanwencheng: I mean by `on main` where to put this component? what is the legacy way to do that?

Also NetworkSpecs.tsx need to become a component now. 

 My idea to create an `extension` folder, and put this `ColoredComponent.tsx` together with `NetworkSpecs.tsx` into it, and split `Extension.tsx` into  two parts : `index.tsx` and `ExtensionInjection`.

hanwencheng:
Refactor Modal into "Metadata" tab, thanks for adding `useChainInfo` hooker!

packages/page-settings/src/Metadata/NetworkSpecs.tsx
```diff
@@ -0,0 +1,146 @@
...
+  if (!isApiReady) {
+    return null;
+  }
+
+  type inputListener = (v: string) => void;
+  const _onChangeValue = (k: keyof NetworkSpecsStruct): inputListener => (v: string): void => setNetworkSpecs({ [k]: v });
+  const _onSetRandomColor = (): void => setNetworkSpecs({ color: getRandomColor() });
+  const _checkColorValid = (): boolean => /^#[\da-fA-F]{6}|#[\da-fA-F]{3}$/.test(networkSpecs.color);
+
+  return <div
```
jacogr: Can we just adapt this form to be -

```js
return (
  <div className={className}>
...
```

packages/page-settings/src/Metadata/index.tsx
```diff
@@ -0,0 +1,27 @@
+import { useTranslation } from '@polkadot/app-settings/translate';
```
jacogr:
```suggestion
// Copyright 2017-2020 @polkadot/app-settings authors & contributors
// This software may be modified and distributed under the terms
// of the Apache-2.0 license. See the LICENSE file for details.

import { useTranslation } from '@polkadot/app-settings/translate';
```

PR #2421, opened 2020-03-19
Improve support for WS_URL

chevdor:
This PR brings a slightly better support for WS_URL and especially allows using WS_URL even in cases where the user already saved another network. This is especially useful on docker/k8s infrastructures where the containers may need to link dynamically.

In that case, passing `WS_URL` accordingly will make the endpoint show up in the list and even make it the default if the user did not define anything yet for this (UI) url.

If `WS_URL` is not provided, things work just like today and nothing special shows up:
image

However, if `WS_URL` is provided, it will show up in the endpoints list:
image

I put the new ENV section at the top of the list for 2 reasons:
- if WS_URL is not specified (which is likely the majority...) the list looks as it does today
- it however WS_URL is provided, it will be selected as the default option on a browser where the user did not previously saved a value.

### DIFF
--- a/packages/apps-config/src/settings/endpoints.ts
+++ b/packages/apps-config/src/settings/endpoints.ts
@@ -84,14 +84,14 @@ let endpoints = [
 ];
 
 if (ENV.length > 0) {
-  endpoints = endpoints.concat([
+  endpoints = [
     {
       isHeader: true,
-      text: 'ENV',
+      text: 'Custom ENV',
       value: ''
     },
     ...ENV
-  ]);
+  ].concat(endpoints);
 }
 
 // The available endpoints that will show in the dropdown. For the most part (with the exception of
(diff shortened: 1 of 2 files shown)


packages/apps-config/src/settings/endpoints.ts
```diff
@@ -12,6 +12,15 @@ const DEV: Option[] = [
   }
 ];
 
+const ENV: Option[] = [];
+if (process.env.WS_URL) {
```
jacogr:
```suggestion
const ENV: Option[] = [];

if (process.env.WS_URL) {
```

chevdor:
I don't mind adapting the `WS_URL: <url>`. I was between `WS_URL: <url>` and `URL: <url>`.
I just fear that using on `<url>` will bring the question from user: "where is this URL coming from?".

PR #2422, opened 2020-03-19
WIP: Switch the docker image to the webpack server

chevdor:
This PR changes a few things.

First, it adds entries to the .dockerignore file to reduce the amount of files sent to build the image (450MB before the change, we are now around 100MB). I would like to remove also the .yarn folder but ran into some issues so I left it.

Second, it changes the docker image. Instead of a 2 stages image building and serving a static site, it switches to a single stage where webpack is serving the site. That allows using ENV variables without extra trickery.

To test, the docker build can be built as:
`docker build -t chevdor/polkadot-ui .`

and ran with:
`docker run --rm -it -e WS_URL=ws://localhost:9946 -p 80:80 chevdor/polkadot-ui`

The `WS_URL` part only making sense once #2421 has been merged.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,1 +1,3 @@
 node_modules
+.git
+.github
--- a/.github/workflows/pr-any.yml
+++ b/.github/workflows/pr-any.yml
@@ -14,6 +14,8 @@ jobs:
       with:
         node-version: '12.x'
     - name: ${{ matrix.step }}
+      env:
+        NODE_ENV: development
       run: |
         yarn install --immutable | grep -v 'YN0013'
         yarn ${{ matrix.step }}
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,25 +1,13 @@
-FROM ubuntu:18.04 as builder
+FROM node:alpine
 
-# Install any needed packages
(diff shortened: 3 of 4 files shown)


jacogr:
Honestly not sure about this at all - these have been known to be used in prod, static + nginx is really more sane.

chevdor:
There are ways to keep the previous static build and using nginx AND use ENV vars but it is a bit more convoluted. If you are not convinced, we may keep the current Dockerfile as it is, add the new one I introduce here and look into the 'nginx' option. I agree that would be better to stick with a static version but supporting ENV is IMO also very important.

chevdor:
For reference, here is the convoluted version allowing to keep the static generated version.
https://github.com/polkadot-js/apps/issues/2444

jacogr:
Convoluted here is certainly better than running the webpack server in prod.

chevdor:
I would not disagree. We could leave this PR on hold and I will check if I can do something on the other topic.

chevdor:
This PR becomes obsolete once #2450 is merged.

chevdor:
Closing this PR since #2450 is now ready and IMO much better.

PR #2450, opened 2020-03-24
Add ENV support to the generated site

### LINKED ISSUE #2444: Add support for ENV while using the generated static version of the UI
By default, a static syte generated by webpack and served over nginx cannot use ENV variables.
That makes it inconvenient when using the UI with Docker for instance.

There are ways however.

Here is a un a nutshell what can be done. The user provides a '.env' file.

In docker, the regular `CMD` is replaced by:
```
CMD ["/bin/bash", "-c", "/usr/share/nginx/html/env.sh && nginx -g \"daemon off;\""]
```

The magic `env.sh` script looks like:

```shell
#!/bin/bash

# Recreate config file
rm -rf ./env-config.js
touch ./env-config.js

# Add assignment 
echo "window.process_env = {" >> ./env-config.js

# Read each line in .env file
# Each line represents key=value pairs
while read -r line || [[ -n "$line" ]];
do
  # Split env variables by character `=`
  if printf '%s\n' "$line" | grep -q -e '='; then
    varname=$(printf '%s\n' "$line" | sed -e 's/=.*//')
    varvalue=$(printf '%s\n' "$line" | sed -e 's/^[^=]*=//')
  fi

  # Read value of current variable if exists as Environment variable
  value=$(printf '%s\n' "${!varname}")
  # Otherwise use value from .env file
  [[ -z $value ]] && value=${varvalue}
  
  # Append configuration property to JS file
  echo "  $varname: \"$value\"," >> ./env-config.js
done < .env

echo "}" >> ./env-config.js
````

It needs to be brought into the image with some:
`COPY env.sh .`

And basically generate our ENV as a JS file.
The config needs to be loaded in the App with some kind of:
`<script type="text/js" src="%PUBLIC_URL%/env-config.js">`
or a more 'react' version of is.

At that point, the ENV is available as expected as `process.env.WS_URL` for instance.

chevdor:
This PR fixes #2444.

The resulting image is around 30MB uncompressed. It contains a generated static version of the site.
The ENV we are interested in is loaded by `env.sh`. For now, the list of supported ENV variables is limited (to `WS_URL`). If we want to support more, `env.sh` should be updated or a pattern should be defined (such as `POLKADOT_UI_WS_URL`, ...).

A variable such as WS_URL then may come from 2 sources.
When using webpack, WS_URL is available at `process.env.WS_URL`. When running the static static however, this variable is now available as `window.process_env.WS_URL`.

## case 1: no ENV
`docker run --rm -it -p 80:80 chevdor/polkadot-ui`
image

## case 2: pass each ENV:
`docker run --rm -it -e WS_URL=ws://myurl:9876 -p 80:80 chevdor/polkadot-ui`
image

## case 3: use an ENV file:
`docker run --rm -it --env-file .env-example -p 80:80 chevdor/polkadot-ui`
image


chevdor:
I pushed an image generated based on this PR. You can use it as:
`docker run --rm -it -e WS_URL=ws://myurl:9876 -p 80:80 chevdor/polkadot-ui:0.40.0-beta179`

PR #2460, opened 2020-03-26
Expose useChainInfo (re-use of some info for signer)

jacogr:
Data can be pulled from chainInfo (not everything needed) in #2360

### DIFF
--- a/packages/page-settings/src/useChainInfo.ts
+++ b/packages/page-settings/src/useChainInfo.ts
@@ -1,0 +1,34 @@
+/* eslint-disable @typescript-eslint/no-empty-interface */
+// Copyright 2017-2020 @polkadot/app-settings authors & contributors
+// This software may be modified and distributed under the terms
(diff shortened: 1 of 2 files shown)


packages/page-settings/src/useChainInfo.ts
```diff
@@ -0,0 +1,34 @@
...
+// of the Apache-2.0 license. See the LICENSE file for details.
+
+import { MetadataDef } from '@polkadot/extension-inject/types';
+
+import { useEffect, useState } from 'react';
+import { getChainTypes } from '@polkadot/apps-config/api';
+import { getSystemIcon } from '@polkadot/apps-config/ui';
+import { useApi } from '@polkadot/react-hooks';
+
+interface ChainInfo extends MetadataDef {}
```
jacogr: Probably want to add `color` in here, it seems like the only thing missing for the signer.

PR #2466, opened 2020-03-27
Don't crash if there is no tip field in extrinsics

Sushisource:
Hello - this line was causing us some pain on our network which does not include tips in extrinsics. Thanks!


jacogr:
Thanks for this.

Issue #2248, opened 2020-02-06
Apps only includes Alice et al accounts on "Local Testnet"


JoshOrndorff:
When using a network named "Local Testnet" or "Development", the accounts for Alice, Bob, etc are pre-populated in the UI. This is great for demos and for workshops.

However, if the network name is changed (for example in the chainspec), then the Alice accounts no longer show up. This is a problem because changing the network name allows one to have a dedicated page in [telemetry](https://telemetry.polkadot.io).

Concretely, If I host a workshop in Denver, I want to send my participants to https://telemetry.polkadot.io/#list/Denver%20Testnet, but I still want them to have the Alice accounts pre-populated.

jacogr:
the development flag is purposefully limited to, well, development networks. For instance we actually did have people using Alice as their accounts on Krumme Lanka...

You can certainly tweak it without own instances, but doubt that I am willing to open a door to known accounts for random networks here.

JoshOrndorff:
One thought I had was to detect the network based on the `id` rather than the network `name`. Would that alleviate your concern?

For example, my chainspec might look like:
```json
{
  "name": "My Specific Testnet",
  "id": "local_testnet",
  ...
}
```

By still giving it `id: local_testnet` I indicate that it's safe to display the dev accounts, and by giving it a custom name I still get its own telemetry page.

Feel free to close if this is still out of scope.

jacogr:
So you don’t receive the id via RPC (unless I am missing a trick, which is possible)

JoshOrndorff:
Ah, you're correct.

My final thought is whether it makes sense to expose a setting on the settings page. Something like "Force inclusion of development accounts". It could be accompanied by an appropriately scary warning about how to accounts aren't at all secret when it is enabled.

PR #2559, opened 2020-04-12
Use system_chainType

jacogr:
Closes https://github.com/polkadot-js/apps/issues/2248

The `system_chainType` was specifically added to enable this behavior, with this PR the UI now checks the RPC as a first step. (On older versions it will fallback to "string" detection, as was the case up till now)


JoshOrndorff:
Cross referencing the Substrate pr where this rpc went in. https://github.com/paritytech/substrate/pull/5576

Looking forward to using this at next workshops!

PR #2563, opened 2020-04-13
AddressMenu popup when clicking on address name

kwingram25:
- Add popup showing address identity & actions and editing of name and tags
- Remove inline identity management


packages/page-accounts/src/Contacts/Address.tsx
```diff
@@ -129,48 +105,32 @@ function Address ({ address, className, filter, isFavorite, toggleFavorite }: Pr
...
-      }
-    },
-    [address, tags, toggleEditTags]
-  );
-
   if (!isVisible) {
     return null;
   }
 
+  const _toggleForget = (): void => setIsForgetOpen(!isForgetOpen);
```
jacogr, line 112: These need to be `useCallback`. Same with the functions below. We don't want to re-create them each render. (The removed instances were `useCallback`). The issue is if not, then all the child components will be re-done each and everytime there is a trigger.

packages/react-components/src/AccountName.tsx
```diff
@@ -130,113 +110,99 @@ function extractIdentity (address: string, identity: DeriveAccountRegistration,
...
-
-      setName(nickname);
-    } else {
-      setName(defaultOrAddr(defaultName, cacheAddr, accountIndex));
-    }
-  }, [api, defaultName, info, isRegistrar, t, toggle, toggleJudgement, value]);
+    _setName();
+  }, [_setName]);
+
+  const onUpdateName = (): void => {
```
jacogr, line 165: useCallback

packages/react-components/src/Input.tsx
```diff
@@ -116,7 +116,7 @@ function Input ({ autoFocus = false, children, className, defaultValue, help, ic
 
       if (onEnter && event.keyCode === 13) {
         (event.target as any).blur();
-        onEnter();
+        typeof onEnter === 'function' && onEnter();
```
jacogr, line 119: `isFunction`

packages/react-hooks/src/useAccountInfo.ts
```diff
@@ -0,0 +1,194 @@
...
+      isEditable: (!identity?.display && (isInContacts || (accountOrAddress && !(accountOrAddress.meta.isInjected || accountOrAddress.meta.isHardware)))) || false,
+      isExternal: accountOrAddress?.meta.isExternal || false,
+      isInContacts,
+      isOwned
+    }));
+    setTags(accountOrAddress?.meta.tags ? accountOrAddress.meta.tags.sort() : []);
+    setName(accountOrAddress?.meta.name || '');
+  }, [identity, isAccount, isAddress, setTags, value]);
+
+  const onSaveName = (): void => {
```
jacogr, line 116: useCallback

packages/react-hooks/src/useAccountInfo.ts
```diff
@@ -0,0 +1,194 @@
...
+        const currentKeyring = keyring.getPair(value);
+
+        currentKeyring && keyring.saveAccountMeta(currentKeyring, meta);
+      } catch (error) {
+        keyring.saveAddress(value, meta);
+      }
+    }
+  };
+
+  const onSaveTags = (): void => {
```
jacogr, line 134: useCallback

jacogr:
I think we should have a sidebar, i.e. on click slide that out. Atm, we are adding extra content (it seems) to the actual name, which has performance implications. Additionally, this happens -

image

I clicked on something, however now where it hovers it is completely out - the content below it is changing and it is not pointing to the right address anymore.

jacogr:
Use a context, that injects a "toggle(address)" - that tiggers the sidebar.

packages/page-accounts/src/Contacts/Address.tsx
```diff
@@ -129,48 +105,41 @@ function Address ({ address, className, filter, isFavorite, toggleFavorite }: Pr
...
-          currentKeyring && keyring.saveAccountMeta(currentKeyring, meta);
-        } catch (error) {
-          keyring.saveAddress(address, meta);
-        }
-      }
-    },
-    [accName, address, toggleEditName]
+  const _toggleForget = useCallback(
+    (): void => setIsForgetOpen(!isForgetOpen),
+    [isForgetOpen]
```
jacogr, line 110: These can just all be `useToggle`

packages/page-accounts/src/Sidebar/Sidebar.tsx
```diff
@@ -0,0 +1,606 @@
...
+  const [isTransferOpen, toggleIsTransferOpen] = useToggle();
+  const [isJudgementOpen, toggleIsJudgementOpen] = useToggle();
+
+  if (!accountInfo) {
+    return null;
+  }
+
+  const useIdentity = !!api.query.identity?.identityOf;
+
+  const _toggleTransfer = (): void => {
```
jacogr, line 50: useCallback or useToggle

packages/page-accounts/src/Sidebar/Sidebar.tsx
```diff
@@ -0,0 +1,606 @@
...
+                    <li>
+                      <IconLink
+                        icon='address card'
+                        label={t('Add identity judgment')}
+                        onClick={_toggleJudgement}
+                      />
+                    </li>
+                  )}
+                  <li>
+                    <IconLink
```
jacogr, line 438: We do have a `LinkExternal` component - that caters for all (Subscan, Polkadscan)

packages/react-components/src/AddressMini.tsx
```diff
@@ -67,6 +68,7 @@ function AddressMini ({ balance, bonded, children, className, iconInfo, isPadded
                 <AccountName
                   noLookup={noLookup}
                   value={value}
+                  withSidebar={withSidebar}
```
jacogr, line 71: It is probably cool if these display an underline (maybe dotted) on hover - will just aid discoverability at least at the start.

packages/page-accounts/src/Sidebar/Sidebar.tsx
```diff
@@ -0,0 +1,606 @@
...
+  );
+
+  const buttonProps = {
+    onMouseEnter: toggleIsHoveringButton,
+    onMouseLeave: toggleIsHoveringButton
+  };
+
+  return (
+    <>
+      <Screen onClick={onClose} />
```
jacogr, line 99: This actually "feels" wrong to me - so I click on an account, then click on the next and well, it doesn't "take". So I need to click-click to get the next one.

Add an explicit button to remove on the popup.

(Which really should take up the full sidebar, it is lost at the bottom atm.)

For instance, this works...

image

packages/page-accounts/src/Sidebar/Sidebar.tsx
kwingram25: Seems to be just an icon with tooltip though unless I'm mistaken?

jacogr: No, it is either an icon on a text link - https://github.com/polkadot-js/apps/blob/master/packages/react-components/src/LinkExternal.tsx#L48

(Would love to use the named version in tables, but it is just unwieldily, think it is `withShort` everywhere now, except in the explorer - just because of spacing)

jacogr: image

^^^ inside the explore block view

PR #2565, opened 2020-04-13
Sidebar context with popup

jacogr:
Not to merge, sample


jacogr:
Closing, only a sample of approach.

PR #2577, opened 2020-04-16
Update node-template types to include Weight `u32`

shawntabrizi:
Chains building with https://github.com/substrate-developer-hub/substrate-node-template/ will currently be using `alpha-v6` which has u32 weights. For the time being, I believe it makes sense for us to keep the default setting as Weight = u32 until the node template moves to a Substrate version that uses u64 weights.

### DIFF
--- a/packages/apps-config/src/api/spec/node-template.ts
+++ b/packages/apps-config/src/api/spec/node-template.ts
@@ -7,5 +7,7 @@
 
 export default {
   Address: 'AccountId',
-  LookupSource: 'AccountId'
+  LookupSource: 'AccountId',
+  // Remove when https://github.com/substrate-developer-hub/substrate-node-template/ moves to alpha-v7
+  Weight: 'u32'
 };


packages/apps-config/src/api/spec/node-template.ts
```diff
@@ -7,5 +7,7 @@
 
 export default {
   Address: 'AccountId',
-  LookupSource: 'AccountId'
+  LookupSource: 'AccountId',
+  // Remove when https://github.com/substrate-developer-hub/substrate-node-template/ moves to alpha-v7
```
JoshOrndorff, line 11: I have a tentative plan to start maintaining these types on a per-node-template-version basis. That way older versions of the node template will still be usable after newer versions are released. Concretely, node template alpha.6 will still be usable after alpha.7 is released and neither will require custom type input.

Have to talk it over with devhub team still.

JoshOrndorff:
@jacogr would you mind having a look at this soonish? It will restore our tutorials to working order.

jacogr:
I honestly don't see why this would need to go in -

- it either breaks on master (without), or
- it breaks on older (with, as set atm)

So either way, the question will continue to come up, this does not solve it. It just shifts it from one side to another, somebody is always unhappy and will have to log something somewhere.

Thus far it has come up once (for node-template) that I've seen in support channels, so either the new node-template banner with FAQ entry links gives people the info they need or it is not an issue. Not sure which way it leans atm.

JoshOrndorff:
We discussed this at the devhub team meeting today, and I believe we have a plan that will solve this moving forward. Feeback on the plan is very welcome.

**Short term plan**
Merge this PR so all our tutorials, which explicitly reference alpha.6, start working again.

**Long term plan**
Start including proper node-template-runtime versions and add the proper types for each released node template version to Apps. So whenever the next type-incompatible change goes into the node template it's spec version will actually increase to 2. Then when alpha.7 is tagged and the next node template is released, we merge another PR here adding the types for that node template runtime.

JoshOrndorff:
I guess it's worth emphasizing that my ultimate goal is not to simply get this PR merged. My goal is to have multiple relatively recent versions of the node template work with Apps so that users following our tutorials and using our templates can use Apps and only encounter injecting their custom types at an appropriate point in their learning experience, which IMO is not when they are running a `--dev` chain for the first time.

@jacogr if there are other techniques to achieving that goal that we haven't considered, I'm open to hearing them.

jacogr:
The issue is exactly that - the template is completely un-spec-versioned, so they are all treated exactly the same. So there is (atm) no way of doing anything specific, like can be done for other spec-types.

Will merge, but as it stands, won't have any impact on support - it will break, somewhere.

JoshOrndorff:
Thank you for merging this.

> the template is completely un-spec-versioned

Yes, this will change. The next node template release will have a proper spec version. Before that node template release is cut, we'll have a companion PR here that adds the correct types for the new spec version.

Dumb question: Where would I add the logic that says "If it is node-template-runtime spec version 1 use these types, but if it's node-template-runtime spec version 2 use other types"?

jacogr:
So for the spec-specific versions it actually gets managed in the api repo. (So it is applicable to the API and UIs - much less overhead).

 The caveat there is that it does not deal with custom types at all, ie. As you would have on a custom chain. (Not an issue for node-template since it uses all the base Substrate types)

api/packages/types-known

PR #2578, opened 2020-04-16
Support for optional usage of pallet_balance

Voxelot:
The Polkadot UI currently expects the balance module to be enabled and has breaking Javascript errors if it's not included in the runtime. Since our chain uses pallet_generic_asset instead of pallet_balance, these changes are needed for the UI to tolerate the balance module being optionally available.

### DIFF
--- a/packages/page-123code/src/SummaryBar.tsx
+++ b/packages/page-123code/src/SummaryBar.tsx
@@ -17,7 +17,7 @@ function SummaryBar (): React.ReactElement {
   const { api, systemChain, systemName, systemVersion } = useApi();
   const bestNumber = useCall<BlockNumber>(api.derive.chain.bestNumber, []);
   const bestNumberLag = useCall<BlockNumber>(api.derive.chain.bestNumberLag, []);
-  const totalInsurance = useCall<Balance>(api.query.balances.totalIssuance, []);
+  const totalIssuance = useCall<Balance>(api.query.balances?.totalIssuance, []);
   const validators = useCall<DeriveStakingValidators>(api.derive.staking.validators, []);
 
   return (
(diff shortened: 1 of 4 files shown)


Sushisource:
Similar to https://github.com/polkadot-js/apps/pull/2466 - I'd suggest turning on TS strict mode to avoid more of these, though I understand that can be a huge undertaking :smile:

h4x3rotab:
LGTM

jacogr:
strict mode won't help - since the base types has the balances module. Tests are done against Kusama, Westend and Substrate master (which is more-or-less what I have bandwidth for). Anything else may need tweaks.

packages/page-123code/src/SummaryBar.tsx
```diff
@@ -42,9 +42,9 @@ function SummaryBar (): React.ReactElement {
             ))
           }</Bubble>
         )}
-        <Bubble icon='circle' label='total tokens'>
-          {formatBalance(totalInsurance)}
-        </Bubble>
+        {totalIssuance && (<Bubble icon='circle' label='total tokens'>
+          {formatBalance(totalIssuance)}
+        </Bubble>)}
```
jacogr, line 47:
```suggestion
        {totalIssuance && (
          <Bubble icon='circle' label='total tokens'>
            {formatBalance(totalIssuance)}
          </Bubble>
        )}
```

packages/page-explorer/src/Summary.tsx
```diff
@@ -27,12 +29,12 @@ function Summary (): React.ReactElement<{}> {
...
           <BlockToTime blocks={ONE_BLOCK} />
         </CardSummary>
-        <CardSummary
+        {api.query.balances && (<CardSummary
           className='ui--media-small'
           label={t('total issuance')}
         >
           <TotalIssuance />
-        </CardSummary>
+        </CardSummary>)}
```
jacogr, line 37: Same tweak as above (Sorry cannot make a suggestion here :()

Voxelot:
@jacogr think it's ready for another look. Although these formatting issues might apply to other areas I haven't been modifying as well. Might need to come up with some linting/prettier rules for this ;)

jacogr:
The linting rules are actually there and have removed a lot of manual overhead. So the one in here is quite interesting, and seems to be a bug in the rules itself, it allowed multi params in a single line. (I'll need to replicate that one and report upstream).

For the `{blah && (<Xyz` - have looked, don't quite have something that fits yet. (And indeed there are some inconsistencies, due to that)

PR #2639, opened 2020-04-27
add nodle arcadia endpoint

ETeissonniere:
## Context
Add nodle arcadia endpoint and logo to the list of default networks.


ETeissonniere:
Oops, fixed the linting errors! For now we don't have custom types, but we will soon have. When we push upgrades we will make sure to push them here?

jacogr:
Perfect, that is fine - I am quite fast (well, as fast as humanly possible) in merging those. (Since they won't break anything else, so fee free, no pressure, but if it makes your life easier, happy to click "merge")

PR #2652, opened 2020-04-28
Update Kulupu type definitions

sorpaas:
An update ("era change") is [planned](https://github.com/kulupu/kulupu/tree/era/1) on block number 320,000, approximately 8 days from now, on 6th May. This completely updates to Substrate master by exporting the current Kulupu state and regenerate a new genesis block. As a result, we don't need any of those custom type definitions from that point on.

It uses `Lookup = Indices` so I believe I don't need the custom type definitions as in `substrate-node-template`.

This indeed means that in the next ~8 days (or whenever this PR is merged to 6th May), the Kulupu interface on polkadot.js.org is broken, but I'd rather it breaks before the hard fork rather than after the hard fork.

### DIFF
--- a/packages/apps-config/src/api/spec/kulupu.ts
+++ b/packages/apps-config/src/api/spec/kulupu.ts
@@ -6,21 +6,9 @@
 /* eslint-disable sort-keys */
 
 export default {
-  Address: 'GenericAddress',
   Difficulty: 'U256',
   DifficultyAndTimestamp: {
     difficulty: 'Difficulty',
     timestamp: 'Moment'
-  },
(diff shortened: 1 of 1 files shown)


jacogr:
Thanks a million - will defer merging this a bit.

And yes, you are 100% spot-on, with `Indices` lookups, nothing needed.

sorpaas:
@jacogr Appreciate if we can get this merged. Hard fork happening in 48 hours!

jacogr:
I have been watching the chain... and waiting for it to get closer :) But yes, will merge as to not chase up your blood pressure :)

PR #2670, opened 2020-05-01
Add russian translation

gregzaitsev:


jacogr:
Speechless. Thanks an absolute million for this. This really helps and avoids issues such as what we have seen recently - the whole world does not speak English.

PR #2690, opened 2020-05-05
Updates Edgeware network types

drewstone:


drewstone:
This may not cover the changes to Berlin yet and we're working to check those too. At least these are more up to date.

PR #2700, opened 2020-05-07
kulupu: add type definition of Era

sorpaas:
This adds the type definition of `Era` to Kulupu. This is a simple structure that stores pre-squash-fork genesis block hash, final block hash and state root, just in case they need to be used later.


jacogr:
Thank you!

PR #2709, opened 2020-05-08
Allow council motion with non-existing end (old)

jacogr:


packages/page-council/src/Motions/Motion.tsx
```diff
@@ -32,6 +32,8 @@ function Motion ({ className, isMember, members, motion: { hash, proposal, votes
     return null;
   }
 
+  console.error(votes);
+
```
jacogr, line 36:
```suggestion
```

PR #2734, opened 2020-05-13
Add attestation to claims

Tbaut:
Let DOT token holders claim their token by:
- signing a statement and a claim with their Ethereum key if their tokens are not pre-claimed
- signing a statement only if the tokens where pre-claimed before launch


packages/apps/src/overlays/Attest.tsx
```diff
@@ -0,0 +1,48 @@
...
+    return null;
+  }
+
+  return (
+    <BaseOverlay className={className}
+      icon='warning sign'
+      type='error'>
+      <div>
+        {t(
+          '{{number}} of your accounts need to sign the attest before receiving DOTs. Please sign the attest ',
```
amaurymartiny, line 39: should we list the accounts that need to sign the attest? Should be okay pretty if users have 1-2 accounts.

packages/page-claims/src/index.tsx
```diff
@@ -73,180 +67,181 @@ const Signature = styled.textarea`
...
-      claim: null,
-      didCopy: false,
-      ethereumAddress: null,
-      signature: null,
-      step: 0
-    };
-    this.state = this.defaultState;
-  }
+const ClaimsApp = (props: Props) => {
+  // constructor (props: Props) {
```
amaurymartiny, line 71: @jacogr we're refactoring this one into a FC. Is there an example of another component using TxModalNew, that we can take inspiration from?

amaurymartiny: (shows outdated, but my question is still actual)

jacogr: That should not be a modal. I would not suggest using the TxModal, just a page with TxButton does 100%.

And yes, it should be a FC. There are 3 major areas left which has not been converted -

- claims (only done because of the testing)
- contracts (in progress)
- signer (it needs quiet before it can be done)

There are some bits and piece left in various places, so either done as a "touch this, convert" or (for the most part) "grab one or two, come back next week".

amaurymartiny: Perfect, this simplifies things

packages/page-claims/src/index.tsx
```diff
@@ -73,180 +67,194 @@ const Signature = styled.textarea`
...
+  const [signature, setSignature] = useState<EcdsaSignature | null>(null);
+  const [step, setStep] = useState<Step>(Step.Account);
+  const [accountId, setAccountId] = useState<string | null>(null);
+  const { api, isApiReady, systemChain } = useApi();
+  const { t } = useTranslation();
+  const [isPreclaimed, setIsPreclaimed] = useState(false);
+
+  useEffect(() => {
+    if (isApiReady && typeof api.query.claims.claimAttest !== undefined) {
+      setIsOldClaimProcess(false);
```
amaurymartiny, line 84: Is there a guarantee that isApiReady will be `true` before rendering this component? Like somewhere higher in the components tree?

If yes, we could simplify this useEffect&useState into just
```ts
const isOldClaimProcess = !!api.query.claims.claimAttest;
```

packages/page-claims/src/index.tsx
```diff
@@ -74,8 +74,9 @@ const ClaimsApp = (props: Props): React.ReactElement => {
   const [signature, setSignature] = useState<EcdsaSignature | null>(null);
   const [step, setStep] = useState<Step>(Step.Account);
   const [accountId, setAccountId] = useState<string | null>(null);
-  const { api, systemChain } = useApi();
+  const { api, isApiReady, systemChain } = useApi();
   const { t } = useTranslation();
+  const [isPreclaimed, setIsPreclaimed] = useState(false);
```
jacogr, line 80:
```suggestion
  const isPreclaimed = useCall<boolean>(api.query.claims.preclaims, [accountId], {
    transform: (option: Option<EthereumAddress>) => option.isSome
  });
```

1. isApiReady is (mostly) never applicable since components aren't rendered before it is (some exceptions on outer layers)
2. prefer never using the APIs like below, we have `useCall` (once again, there are some exceptions, especially in custom hooks and with stuff like `queryMulti` which is not great)

packages/page-claims/src/index.tsx
```diff
@@ -85,6 +86,16 @@ const ClaimsApp = (props: Props): React.ReactElement => {
...
+  useEffect(() => {
+    if (isApiReady && accountId) {
+      api.query.claims
+        .preclaims<Option<EthereumAddress>>(accountId)
+        .then((opt): void => {
+          setIsPreclaimed(opt.isSome);
+        });
+    }
+  }, [accountId, api.query.claims, isApiReady]);
+
```
jacogr, line 107:
```suggestion
```

Delete with above in mind.

packages/page-claims/src/index.tsx
jacogr: Indeed. As long as the route has `needsApi` set - https://github.com/polkadot-js/apps/blob/master/packages/apps/src/Content/index.tsx#L50

jacogr: ... which it does - https://github.com/polkadot-js/apps/blob/master/packages/apps-routing/src/claims.ts#L13

jacogr: The reason to never use the promises directly also means the boilerplate is reduced and you also don't need to resort to `useMountedRef` (so unmounts won't spam the console, state updates are only done _while_ mounted)

packages/apps/src/overlays/Attest.tsx
```diff
@@ -0,0 +1,49 @@
+// Copyright 2017-2020 @polkadot/apps authors & contributors
```
jacogr, line 1: Not the right way to do it to fit in with existing process. The overlays are absolutely horrible. (Cannot get rid of them all atm) 

Add a counter alert to the menu, example -

- https://github.com/polkadot-js/apps/blob/master/packages/page-settings/src/useCounter.ts
- https://github.com/polkadot-js/apps/blob/master/packages/apps-routing/src/settings.ts#L16

packages/page-claims/src/index.tsx
```diff
@@ -73,183 +62,155 @@ const Signature = styled.textarea`
...
-      ethereumAddress: null,
-      signature: null,
-      step: 0
-    };
-    this.state = this.defaultState;
-  }
-
-  public componentDidUpdate (): void {
-    if (this.state.didCopy) {
+const ClaimsApp = (): React.ReactElement => {
```
jacogr, line 65:
```suggestion
function ClaimsApp (): React.ReactElement {
```

packages/page-claims/src/index.tsx
```diff
@@ -73,183 +62,155 @@ const Signature = styled.textarea`
...
-  withApi
-);
+          )}
+        </Column>
+      </Columar>
+    </main>
+  );
+};
+
+export default ClaimsApp;
```
jacogr, line 216:
```suggestion
export default React.memo(ClaimsApp);
```

Does not make a difference here, but consistency everywhere.

packages/apps/src/overlays/Attest.tsx
```diff
@@ -0,0 +1,49 @@
...
+function Attest ({ className }: Props): React.ReactElement<Props> | null {
+  const { t } = useTranslation();
+  const { allAccounts } = useAccounts();
+  const { api } = useApi();
+
+  // Find accounts that need attest. They are accounts that
+  // - already preclaimed,
+  // - didn't sign the attest yet.
+  // `claims.preclaims` returns Some() for these accounts.
+  const needAttest = useCall<[Option<EthereumAddress>]>(api.query?.claims?.preclaims?.multi, [allAccounts])?.filter((opt) => opt.isSome);
```
jacogr, line 28:
```suggestion
  const needAttest = useCall<[Option<EthereumAddress>]>(isApiReady && api.query.claims?.preclaims?.multi, [allAccounts])?.filter((opt) => opt.isSome);
```

Consistency with elsewhere. Since this (or a preferred counter alignment), is outside of content it needs the ready check - `useCall` does handle `false` correctly in this case, doing nothing.

packages/apps/src/overlays/Attest.tsx
amaurymartiny: I would counter-argue that the red notifications on the menu bar are not strong enough for the message we're trying to deliver.

For now, the red notifications are indications that something is happening, as an "info"-level alert (e.g. referendum/proposal are happening right now). It's fine if the user dismisses them (and sometimes will).

On the other hand, attest is a mandatory step: users can't even receive their tokens before attesting. A "warning"/"error"-level alert is needed.

While their UX is indeed not perfect, I do believe that they are better suited, and show immediately the action the users need to take (via the explicit text). "Cannot get rid of them all atm" -> I do see a small "x" button to hide the top bar notification.

jacogr: Sorry, it is just the way it is in the apps UI. Consistency, adjust, please.

packages/apps/src/overlays/Attest.tsx
Tbaut: This can be added in a second iteration eventually.

packages/page-claims/src/Warning.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+  preclaimsArray.forEach((opt, index) => {
+    if (opt.isSome) {
+      needAttestArray.push(allAccounts[index]);
+    }
+  });
+
+  if (!needAttestArray || !needAttestArray.length) {
+    return null;
+  }
```
jacogr, line 43:
```suggestion
  const needsAttestArray = useCall<string[]>(api.query.claims.preclaims?.multi, [allAccounts], {
     transform: (preclaims: Option<EthereumAddress>[]) =>
       preclaims
         .map((opt, index) => [opt.isSome, allAccounts[index]])
         .filter(([isSome]) => isSome)
         .map(([, address)) => address) 
  });
  
  if (!needsAttestArray || !needsAttestArray.length) {
    return null;
  }
```

Only transform and create once.

packages/page-claims/src/Warning.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+  return (
+    <Card isError>
+      <div className={className}>
+        {
+          t(
+            'You need to sign an attestation for the following account(s): {{needAttestArray}}',
+            { replace: { needAttestArray } }
+          )
+        }
```
jacogr, line 53:
```suggestion
        {t('You need to sign an attestation for the following account(s): {{needAttestArray}}', {
          replace: { needAttestArray }
        })}
```

jacogr: This actually needs to be something visibly aligning with the rest, like `AddressMini`

packages/page-claims/src/index.tsx
```diff
@@ -2,42 +2,33 @@
...
+import { EcdsaSignature, EthereumAddress, StatementKind } from '@polkadot/types/interfaces';
 
-import React from 'react';
+import React, { useState, useCallback, useEffect } from 'react';
 import { Trans } from 'react-i18next';
 import styled from 'styled-components';
 import CopyToClipboard from 'react-copy-to-clipboard';
-import { withApi, withMulti } from '@polkadot/react-api/hoc';
-import { Button, Card, Columar, Column, InputAddress, Tooltip } from '@polkadot/react-components';
+import { Button, Card, Columar, Column, InputAddress, Tooltip, Input } from '@polkadot/react-components';
```
jacogr, line 12:
```suggestion
import { Button, Card, Columar, Column, Input, InputAddress, Tooltip } from '@polkadot/react-components';
```

packages/page-claims/src/index.tsx
```diff
@@ -2,42 +2,33 @@
...
 import { TokenUnit } from '@polkadot/react-components/InputNumber';
-import TxModal, { TxModalState, TxModalProps } from '@polkadot/react-components/TxModal';
 import { u8aToHex, u8aToString } from '@polkadot/util';
 import { decodeAddress } from '@polkadot/util-crypto';
 
+import AttestDisplay from './Attest';
 import ClaimDisplay from './Claim';
 import { recoverFromJSON } from './util';
+import { useTranslation } from './translate';
+import Warning from './Warning';
```
jacogr, line 22:
```suggestion
import { useTranslation } from './translate';
import { recoverFromJSON } from './util';
import AttestDisplay from './Attest';
import ClaimDisplay from './Claim';
import Warning from './Warning';
```

packages/page-claims/src/index.tsx
```diff
@@ -73,75 +64,134 @@ const Signature = styled.textarea`
...
+  const [ethereumAddress, setEthereumAddress] = useState<EthereumAddress | null>(null);
+  const [signature, setSignature] = useState<EcdsaSignature | null>(null);
+  const [step, setStep] = useState<Step>(Step.Account);
+  const [accountId, setAccountId] = useState<string | null>(null);
+  const { api, systemChain } = useApi();
+  const { t } = useTranslation();
+  const isPreclaimed = useCall<boolean>(api.query.claims.preclaims, [accountId], {
+    transform: (option: Option<EthereumAddress>) => {
+      if (option.isSome) {
+        setEthereumAddress(option.unwrap());
```
jacogr, line 78: Not convinced about setting the state inside the transform.

packages/page-claims/src/index.tsx
```diff
@@ -73,75 +64,134 @@ const Signature = styled.textarea`
...
+    setDidCopy(true);
+  }, []);
+
+  const statementKind = useCall<StatementKind | undefined>(ethereumAddress && api.query.claims.signing, [ethereumAddress], {
+    transform: (option: Option<StatementKind>) => option.unwrapOr(undefined)
+  }) || '';
+
+  const prefix = u8aToString(api.consts.claims.prefix.toU8a(true));
+  const payload = accountId
+    ? `${prefix}${u8aToHex(decodeAddress(accountId), -1, false)}${statementKind}`
```
jacogr, line 143:
```suggestion
  const statementKind = useCall<string>(ethereumAddress && api.query.claims.signing, [ethereumAddress], {
    transform: (option: Option<StatementKind>) => option.unwrapOr('').toString()
  });

  const prefix = u8aToString(api.consts.claims.prefix.toU8a(true));
  const payload = accountId
    ? `${prefix}${u8aToHex(decodeAddress(accountId), -1, false)}${statementKind || ''}`
```

packages/page-claims/src/index.tsx
```diff
@@ -73,75 +64,134 @@ const Signature = styled.textarea`
...
+                // We only need to know the Ethereum address for the new process
+                // to know the StatementKind for users to sign
+                !isOldClaimProcess && <Input
+                  autoFocus
+                  className='full'
+                  help={t('The the Ethereum address you used during the pre-sale (starting by "0x")')}
+                  label={t('Pre-sale ethereum address')}
+                  onChange={onChangeEthereumAddress}
+                  value={ethereumAddress || ''}
+                />}
```
jacogr, line 191:
```suggestion
                !isOldClaimProcess && (
                  <Input
                    autoFocus
                    className='full'
                    help={t('The the Ethereum address you used during the pre-sale (starting by "0x")')}
                    label={t('Pre-sale ethereum address')}
                    onChange={onChangeEthereumAddress}
                    value={ethereumAddress || ''}
                  />
                )}
```

packages/page-claims/src/index.tsx
```diff
@@ -73,75 +64,134 @@ const Signature = styled.textarea`
...
+              {(!!ethereumAddress || isOldClaimProcess) && <>
```
jacogr, line 192: Same as above, when you do the shortcut,

```
condition && (
  <Element>...</Element>
)

**Comment by Tbaut** [comment]
@jacogr it should be better now.
We still have 2 problems when testing with [Gav's branch](https://github.com/paritytech/polkadot/pull/1091), most probably linked to the rust, but you might have an idea:
- whatever we put as statement as parameter to the `claims.attest` function, the TX is successful, it is explained [here in details](https://github.com/paritytech/polkadot/pull/1091#pullrequestreview-411600478), our test is made from the extrinsics tab.
- when we submit a `claims.claimAttest` from the extrinsics tab as an unsigned TX (because the origin should be "none"). We get "invalid transaction call". Our Payload is `(configured prefix string)(address)(statement)` which looks right now like "Pay DOTs to the Polkadot account:d424d067be59e542d3e44ad620f2d928d4cf44ebaa1dc8236171a717b2e0b53dAlternative".

File: packages/page-claims/src/useCounter.ts [hunk 18]
```diff
@@ -0,0 +1,21 @@
...
+  const { allAccounts } = useAccounts();
+  const { api, isApiReady } = useApi();
+
+  // Find accounts that need attest. They are accounts that
+  // - already preclaimed,
+  // - didn't sign the attest yet.
+  // `claims.preclaims` returns Some() for these accounts.
+  const needAttest = useCall<[Option<EthereumAddress>]>(isApiReady && api.query.claims?.preclaims?.multi, [allAccounts])?.filter((opt) => opt.isSome);
+
+  return needAttest ? needAttest.length : 0;
```
jacogr, line 20:
```suggestion
  const counter = useCall<number>(isApiReady && api.query.claims?.preclaims?.multi, [allAccounts], {
    transform: (preclaims: Option<EthereumAddress>[]) =>
      preclaims.filter((opt) => opt.isSome).length
  });

  return counter || 0;
```

Technically since we do the exact same query in 2 places, we should have a hook that just extracts the accounts into an array (as above) and this one should just return the length of that hook.

packages/page-claims/src/Warning.tsx
Tbaut: Agreed, this still needs some work.

jacogr: Yes, not complaining - it would actually be nice to show the Polkadot with the Ethereum address as linked.

packages/page-claims/src/index.tsx
Tbaut: I moved it out.

packages/page-claims/src/index.tsx
```diff
@@ -73,183 +64,196 @@ const Signature = styled.textarea`
...
+  }, []);
 
-    return [
-      accountId ? accountId.toString() : null,
-      signature || null
-    ];
-  }
+  const onChangeEthereumAddress = useCallback((value: string) => {
+    // FIXME We surely need a better check than just a trim
+    const trimmedAddress = value.trim() as unknown as EthereumAddress;
```
amaurymartiny, line 128: `value.trim()` is a string, I think we can either change the state's type to string, or create an actual EthereumAddress here.

packages/page-claims/src/useCounter.ts
Tbaut: I extracted it in a `usePreclaimPolkadotAddresses` hook used locally. Is that what you were thinking about, or did you think about `@polkadot/react-hooks`?

packages/page-claims/src/Warning.tsx
```diff
@@ -22,21 +22,21 @@ export interface Props{
...
-     transform: (preclaims: Option<EthereumAddress>[]) =>
-       preclaims
-         .map((opt, index) => [opt.isSome, allAccounts[index]])
-         .filter(([isSome]) => isSome)
-         .map(([, address)) => address) 
+    transform: (preclaims: Option<EthereumAddress>[]) =>
+      preclaims
+        .map((opt, index) => [opt.isSome, allAccounts[index]])
+        .filter(([isSome]) => isSome)
+        .map(([, address]) => address as string)
```
jacogr: This also works (the first one defining the types) -

```suggestion
        .map((opt, index): [boolean, string] => [opt.isSome, allAccounts[index]])
        .filter(([isSome]) => isSome)
        .map(([, address]) => address)
```

I sometimes forget that TS is pretty good and then "not-so-good".

Tbaut: ah nice thanks, this is much nicer

packages/page-claims/src/useCounter.ts
jacogr: Locally is 100%. (Since it is only used here - sadly there are some local-only ones that slipped into react-hooks, just have not move the 1 or 2 yet)

packages/page-claims/src/Warning.tsx
```diff
@@ -31,10 +31,18 @@ function Warning ({ className }: Props): React.ReactElement<Props> | null {
       <div className={className}>
         {
           t(
-            'You need to sign an attestation for the following account(s): {{needsAttestArray}}',
-            { replace: { needsAttestArray } }
+            'You need to sign an attestation for the following {{account}}:',
+            { replace: { account: needsAttestArray.length > 1 ? 'accounts' : 'account' } }
```
Tbaut, line 35: not 100% sure about how clever this is for translation. Can I use a `t('')` in a translation's `replace`?

Tbaut:
It now looks like:
image

I didn't spend time to show the linked ETH address because I don't think this would help users. In case an attest is needed, there's no need for the user to enter or even know the ETH address. ll they need to do is select this account, and we'll retrieve it for them, they'll just have to click the "Attest" button.

packages/page-claims/src/Warning.tsx
jacogr: Not a good idea - you assume all languages has an english-like construct. The sanest approach (since it is a whole sentence), would be to have 2 full strings.

packages/page-claims/src/Warning.tsx
```diff
@@ -31,10 +31,18 @@ function Warning ({ className }: Props): React.ReactElement<Props> | null {
...
           )
         }
+        {
+          needsAttestArray.map((address) => {
+            return <AddressMini
+              key={address}
+              value={address}
+            />;
+          })
```
jacogr, line 44:
```suggestion
          needsAttestArray.map((address) => (
            <AddressMini
              key={address}
              value={address}
            />
          ))
```

packages/page-claims/src/Warning.tsx
Tbaut: Got it thanks, I made the changes.

packages/page-claims/src/Attest.tsx
```diff
@@ -11,6 +11,7 @@ import { BalanceOf, EthereumAddress, StatementKind } from '@polkadot/types/inter
 
 import { useTranslation } from './translate';
 import { useApi } from '@polkadot/react-hooks';
+import { getStatement } from './util';
```
jacogr, line 14: Import above this should be moved.

packages/page-claims/src/Attest.tsx
```diff
@@ -65,7 +66,7 @@ function Attest ({ accountId, className, ethereumAddress, statementKind }: Props
             icon='send'
             isPrimary
             label={t('Attest')}
-            params={[statementKind]} // Replace with actual
+            params={[getStatement(statementKind)]}
```
jacogr, line 69: Probably want to memo/state this param since it gets called each render. 

(As it stands the arrays we pass everywhere is horrible anyway since it will change when re-rendered, but have not gotten to that cleanup, but at least we probably don't want to make a "the call" here)

packages/page-claims/src/Claim.tsx
```diff
@@ -26,12 +26,16 @@ interface Props {
 
 // Depending on isOldClaimProcess, construct the correct tx.
 function constructTx (accountId: string, ethereumSignature: string | null, kind: StatementKind | undefined, isOldClaimProcess: boolean): {
-  params: any[];
-  tx: string;
+  params?: any[];
+  tx?: string;
```
jacogr, line 30: If this type is big enough to split over multiple lines (severely decreasing readability), extract it into an interface.

packages/page-claims/src/Warning.tsx
```diff
@@ -0,0 +1,61 @@
+// Copyright 2017-2020 @polkadot/app-claims authors & contributors
+// This software may be modified and distributed under the terms
+// of the Apache-2.0 license. See the LICENSE file for details.
+
+// Copyright 2017-2020 @polkadot/app-settings authors & contributors
+// This software may be modified and distributed under the terms
+// of the Apache-2.0 license. See the LICENSE file for details.
```
jacogr, line 7:
```suggestion
```

packages/page-claims/src/Claim.tsx
```diff
@@ -3,24 +3,44 @@
...
+
+// Depending on isOldClaimProcess, construct the correct tx.
+function constructTx (accountId: string, ethereumSignature: string | null, kind: StatementKind | undefined, isOldClaimProcess: boolean): ConstructTx {
+  if (!ethereumSignature || !kind) {
+    return {};
+  }
+
+  return isOldClaimProcess
+    ? { params: [accountId, ethereumSignature], tx: 'claims.claim' }
+    : { params: [accountId, ethereumSignature, getStatement(kind)], tx: 'claims.claimAttest' };
```
jacogr, line 40: As an aside, we could just directly construct the tx here and use `const extrinsic = constructTx(...)` and then in TxButton use `extrinsic={extrinsic}` as opposed to the `tx` and `params`. (Which is also more efficient, since we don't need to re-do that stuff unless something changes) 

Here is an example as done for payouts

https://github.com/polkadot-js/apps/blob/master/packages/page-staking/src/Payouts/PayButton.tsx#L111

Nothing wrong here, just suggesting another option which may fit better for the use-case or thinking. 

(Longer term thinking of moving more stuff in that direction, since atm the construction params are effectively untyped and there are too many ways of doing the same thing)

Swader:
Given that Apps recently added IPFS deployments and we're Web3, I humbly suggest dogfooding a little more and:

- including the statement on IPFS instead of the web3 or polkadot website
- signing the link, which should include a gateway like ipfs.io

This will:

- let web2 people access it with the same level of convenience as usual
- let web3 people and those of us who run IPFS nodes access and store the statement locally and pin it

Thereby:

- demonstrating commitment to the web 3.0 ethos
- keeping the statement immutable post-signing

packages/page-claims/src/Statement.tsx
```diff
@@ -0,0 +1,38 @@
...
+      <ReactMd
+        className='statement'
+        escapeHtml={false}
+        source={getStatementMd(kind)}
+      />
+    </div>
+  );
+}
+
+export default React.memo(styled(Statement)`
```
amaurymartiny, line 36: no need for styled

packages/page-claims/src/index.tsx
```diff
@@ -164,19 +160,15 @@ function ClaimsApp (): React.ReactElement {
...
+  const _statementKind = useCall<string>(!isPreclaimed && !isOldClaimProcess && ethereumAddress && api.query.claims.signing, [ethereumAddress], {
     transform: (option: Option<StatementKind>) => option.unwrapOr('').toString()
   });
-
-  const statement = statementKind
-    ? statementKind === ALTERNATIVE_STATEMENT
-      ? statementAlternative
-      : statementRegular
-    : null;
+  const statementKind = api.createType('StatementKind', _statementKind);
```
amaurymartiny, line 166: Would recommend, in the transform, to do `option.unwrapOr(null)`. Then you won't need this last line, and `statementKind` will have type `StatementKind | null`

Tbaut:
Any clue on the test failing @jacogr ? it's not happy about the md for some reasong.

packages/page-claims/src/index.tsx
Tbaut: Thanks, that works. Because of the `useCall` it's `StatementKind | undefined | null` though.

jacogr:
Yes, add it as empty for jest - https://github.com/polkadot-js/apps/blob/master/jest.config.js#L20

packages/page-claims/src/Statement.tsx
```diff
@@ -0,0 +1,48 @@
...
+}
+
+function Statement ({ className, kind }: Props): React.ReactElement<Props> | null {
+  const { t } = useTranslation();
+  const statementUrl = getStatementUrl(kind);
+
+  return (
+    <div className={className}>
+      {t('Please read the following statement carefully, it is also available at ')}
+      <a href={statementUrl}>{statementUrl}</a>
```
amaurymartiny, line 26:
```suggestion
      <a href={statementUrl} target="_blank">{statementUrl}</a>
```

so that they won't need to choose account & paste eth address when coming back to apps

amaurymartiny:
@jacogr We're just waiting for the Rust PR to be ready to do the final e2e test. In the meantime, would appreciate another round of review.

jest.config.js
```diff
@@ -18,7 +18,7 @@ module.exports = Object.assign({}, config, {
   moduleNameMapper: {
     ...internalModules,
     '\\.(css|less)$': 'empty/object',
-    '\\.(jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$': 'empty/object'
+    '\\.(jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga|md)$': 'empty/object'
```
jacogr, line 21: Would have preferred an explicit entry so this doesn't get lost inside "media types".

packages/page-claims/src/Attest.tsx
```diff
@@ -0,0 +1,109 @@
...
+  const statementSentence = getStatementSentence(statementKind);
+
+  return (
+    <Card isSuccess>
+      <div className={className}>
+        <Statement
+          kind={statementKind}
+        />
+        {hasClaim &&
+          <h3>{t('Attestation for: ')} <FormatBalance value={claimValue} /></h3>}
```
jacogr, line 66: The ` ` at the end is very problematic for i18n. FormatBalance however has a label.

packages/page-claims/src/Attest.tsx
```diff
@@ -0,0 +1,109 @@
...
+            params={[statementSentence]}
+            tx='claims.attest'
+          />
+        </Button.Group>
+      </div>
+    </Card>
+  );
+}
+
+// FIXME Same styles as ./Claim.tsx
```
jacogr, line 83: Export the styles there, import here.

packages/page-claims/src/Statement.tsx
```diff
@@ -0,0 +1,49 @@
...
+  kind: StatementKind;
+}
+
+function Statement ({ className, kind }: Props): React.ReactElement<Props> | null {
+  const { t } = useTranslation();
+  const statementUrl = getStatementUrl(kind);
+
+  return (
+    <div className={className}>
+      {t('Please read these terms and conditions carefully. By submitting this statement, you are deemed to have accepted these Terms and Conditions. If you do not agree to these terms, please refrain from accessing or proceeding. You can also find them at: ')}
```
jacogr, line 26: Spacing. Remove the end-space, if you need a space, rather use a style to add a margin.

packages/page-claims/src/Statement.tsx
```diff
@@ -0,0 +1,49 @@
...
+      <a href={statementUrl}
+        rel='noopener noreferrer'
+        target='_blank'>{statementUrl}</a>
+      <div className='statement'>
+        {kind.isDefault ? <RegularStatement /> : <SaftStatement />}
```
jacogr, line 31:
```suggestion
        {kind.isDefault
          ? <RegularStatement />
          : <SaftStatement />
        }
```

packages/page-claims/src/index.tsx
```diff
@@ -73,183 +68,242 @@ const Signature = styled.textarea`
...
+
+    if (!api.query.claims || !api.query.claims.preclaims) {
+      return setPreclaimEthereumAddress(null);
+    }
+
+    api.query.claims
+      .preclaims<Option<EthereumAddress>>(accountId)
+      .then((preclaim): void => {
+        setEthereumAddress(preclaim.unwrapOr(null)?.toString());
+        setPreclaimEthereumAddress(preclaim.unwrapOr(null)?.toString());
```
jacogr, line 106: Double unwrap. Extract into const.

packages/page-claims/src/index.tsx
```diff
@@ -73,183 +68,242 @@ const Signature = styled.textarea`
...
+  const onCopy = useCallback(() => {
+    setDidCopy(true);
+  }, []);
 
-  protected onChangeSignature = (event: React.SyntheticEvent<Element>): void => {
-    const { value: signatureJson } = event.target as HTMLInputElement;
+  // If it's 1/ not preclaimed and 2/ not the old claiming process, fetch the
+  // statement kind to sign.
+  const _ethereumAddress = ethereumAddress ? api.createType('EthereumAddress', ethereumAddress) : null;
+  const statementKind = useCall<StatementKind | null>(!isPreclaimed && !isOldClaimProcess && _ethereumAddress && api.query.claims.signing, [_ethereumAddress], {
```
jacogr, line 163: You don't want to pass a value that changes every render into `useCall` and params. (Well, it does work, but is smelly)

packages/page-claims/src/statements/regular.tsx
```diff
@@ -0,0 +1,144 @@
...
+// This software may be modified and distributed under the terms
+// of the Apache-2.0 license. See the LICENSE file for details.
+
+import React from 'react';
+
+// Copy-pasted from https://statement.polkadot.network/regular.html
+function RegularStatement (): React.ReactElement {
+  return (
+    <>
+      <p className='bold_text'>Statement</p>
```
jacogr, line 11: This really is messy like this. Should be md.

packages/page-claims/src/Attest.tsx
```diff
@@ -0,0 +1,109 @@
...
+}
+
+function Attest ({ accountId, className, ethereumAddress, statementKind }: Props): React.ReactElement<Props> | null {
+  const { t } = useTranslation();
+  const { api } = useApi();
+  const [claimValue, setClaimValue] = useState<BalanceOf | null>(null);
+  const [claimAddress, setClaimAddress] = useState<EthereumAddress | null>(null);
+  const [isBusy, setIsBusy] = useState(false);
+
+  const _fetchClaim = useCallback(
```
jacogr, line 31: Not sure why this is a callback. Extract the function, pass the api, address, flag setter in from where it is called. (Or keep inside effect)

packages/page-claims/src/index.tsx
```diff
@@ -73,183 +68,242 @@ const Signature = styled.textarea`
...
+                        icon='sign-in'
+                        isDisabled={!accountId || !signature}
+                        label={t('Confirm claim')}
+                        onClick={goToStepClaim}
+                      />
+                    </Button.Group>
+                  )}
+                </>
+                : <div>
+                    The Ethereum address you submitted could not be found in the genesis.
```
jacogr, line 282: ?

packages/page-claims/src/index.tsx
```diff
@@ -73,183 +68,242 @@ const Signature = styled.textarea`
...
+                      onClick={goToStepSign}
+                    />
+                  </Button.Group>
+                )}
+              </Card>
+            )}
+          {(step >= Step.Sign && !isPreclaimed) && (
+            <Card>
+              {(isOldClaimProcess || !!statementKind)
+                ? <>
```
jacogr, line 239: Because we don't use `()` around a multi-line component structure, the intent is completely lost. Improve for readability.

packages/page-claims/src/index.tsx
```diff
@@ -73,183 +68,242 @@ const Signature = styled.textarea`
...
+                  <h3>{t('{{step}}. Sign with you ETH address',
+                    { replace: { step: isOldClaimProcess ? '2' : '3' } })}</h3>
+                  {!isOldClaimProcess && (<Statement
+                    kind={statementKind}
+                  />) }
+                  <div>
+                    {t('Copy the following string and sign it with the Ethreum account you used during the pre-sale in the wallet of your choice, using the string as the payload, and then paste the transaction signature object below')}
+                    :
```
jacogr, line 247: No sure that this `:` is here - is it part of the text, the unreadable `?:` or something else?

packages/page-claims/src/index.tsx
jacogr: Also, since there is one element (which is the fixed `t('...')`), keep it alongside the div, no need to break the line. (As above)

packages/page-claims/src/index.tsx
jacogr: Use `<div>{t('...')}</div>` since it is a single element

packages/page-claims/src/statements/regular.tsx
jacogr: (And needs to be under statements/polkadot - will make a comment around that is the final overall later)

packages/page-claims/src/util.ts
```diff
@@ -20,6 +20,11 @@ interface SignatureParts {
   signature: Buffer;
 }
 
+const DEFAULT_STATEMENT_URL = 'https://statement.polkadot.network/regular.html';
+const ALTERNATIVE_STATEMENT_URL = 'https://statement.polkadot.network/saft.html';
```
jacogr, line 24: Chain-specific, treat as such.

packages/page-claims/src/util.ts
```diff
@@ -108,3 +113,31 @@ export function recoverFromJSON (signatureJson: string | null): RecoveredSignatu
     };
   }
 }
+
+export function getStatementUrl (kind?: StatementKind | null): string {
```
jacogr, line 117: Pass network type alongside.

packages/page-claims/src/util.ts
```diff
@@ -108,3 +113,31 @@ export function recoverFromJSON (signatureJson: string | null): RecoveredSignatu
...
+    return DEFAULT_STATEMENT_URL;
+  } else if (kind?.isAlternative) {
+    return ALTERNATIVE_STATEMENT_URL;
+  } else {
+    return '';
+  }
+}
+
+// From a StatementKind, get the hardcoded actual statement to sign
+export function getStatementSentence (kind?: StatementKind | null): string {
```
jacogr, line 128: Pass network type.

packages/page-claims/src/statements/regular.tsx
amaurymartiny: The W3F decided to use a html instead of a md. We don't actually have the md anymore. Would rather stay as close as what the W3F gives us.

packages/page-claims/src/Attest.tsx
amaurymartiny: Actually copied from https://github.com/polkadot-js/apps/blob/master/packages/page-claims/src/Claim.tsx#L30, thought that was the way you did it in apps.

packages/page-claims/src/index.tsx
```diff
@@ -73,183 +68,250 @@ const Signature = styled.textarea`
...
+                  </Button.Group>
+                )}
+              </Card>
+            )}
+          {(step >= Step.Sign && !isPreclaimed) && (
+            <Card>
+              {(isOldClaimProcess || !!statementKind)
+                ? (
+                  <>
+                    <h3>{t<string>('{{step}}. Sign with you ETH address',
```
amaurymartiny, line 246:
```suggestion
                    <h3>{t<string>('{{step}}. Sign with your ETH address',
```

packages/page-claims/src/Statement.tsx
```diff
@@ -0,0 +1,73 @@
...
+  .statement{
+    border: 1px solid #c2c2c2;
+    background: #f2f2f2;
+    height: 15rem;
+    overflow-y: scroll;
+    padding: 1rem;
+    width: 100%;
+    margin: 1rem 0;
+    white-space: normal;
+    word-break: break-all;
```
lsaether, line 71:
```suggestion
    word-break: break-word;
```

Tbaut: ah, saw this too late, I think it achieves the same to just remove it, also `break-word` is deprecated according to https://developer.mozilla.org/en-US/docs/Web/CSS/word-break

packages/page-claims/src/Statement.tsx
```diff
@@ -0,0 +1,72 @@
...
+  }
+}
+
+function Statement ({ className, kind, systemChain }: Props): React.ReactElement<Props> | null {
+  const { t } = useTranslation();
+  const statementUrl = getStatement(systemChain, kind)?.url;
+
+  return (
+    <div className={className}>
+      {t('Please read these terms and conditions carefully. By submitting this statement, you are deemed to have accepted these Terms and Conditions. If you do not agree to these terms, please refrain from accessing or proceeding. You can also find them at:')}
```
lsaether, line 47:
```suggestion
      {t('Please read these terms and conditions carefully. By submitting this statement, you are deemed to have accepted these Terms and Conditions. If you do not agree to these terms, please refrain from accessing or proceeding. You can also find them at: ')}
```

A space here prevents the "at" from being attached to the link like this:

image

Putting the space fixes it:

image

packages/page-claims/src/statements/polkadot/regular.tsx
```diff
@@ -0,0 +1,144 @@
+// Copyright 2017-2020 @polkadot/app-claims authors & contributors
```
jacogr, line 1: Remove, rather than duplicate - point to ipfs (we have the hash, default) and website the website. (Better than duplicating the html and even md for it)

packages/page-claims/src/util.ts
```diff
@@ -108,3 +108,34 @@ export function recoverFromJSON (signatureJson: string | null): RecoveredSignatu
...
   }
 }
+
+export interface Statement {
+  hash: string;
+  sentence: string;
+  url: string;
+}
+
+export function getStatement (network: string, kind?: StatementKind | null): Statement | undefined {
```
jacogr, line 118: +1

packages/page-claims/src/statements/polkadot/regular.tsx
Tbaut: I think we legally have the obligation to show the statement on the page where the attestation is made. Can you confirm @lsaether ?

lsaether: Yes the statements need to appear on the page directly.

Tbaut: We had a little discussion with Amaury to day we implemented that. We chose to duplicate it rather than fetching it from the website, because:
- it won't change
- it removes a dependency to this website

But it's ultimately up to you @jacogr if you prefer the content of the statement to be fetched.

jacogr: I indicated before that this whole React component for it is just horrible. We are duplicating and adding bloat for no reason, at the very least it is not a rework of the text like we have. 

Since we sign the hash, the preferred approach is always going to be o "point to the hash". If there is a legal reason not to, well, the hash of what we sign and what we display is not the same since we reworked. They need to match.

amaurymartiny: @jacogr there's no silver bullet. Please tell us which one you want between 1, 2, 3 or 4:
1. Keep as-is. As you said, it's not matching.
2. Use md. Cleaner, but it's also not matching. Plus we need to do html->md conversion.
3. Use `fetch()`. We will get back a HTML file, there will be some ugly code to get the innerHTML of the `body`, and probably some `dangerouslySetInnerHTML`. Plus latency for user.
4. Use iframe. We lose control over styling.

We're happy with any solution.

jacogr: Well, we are out of time :) I would prefer 4. It is the original, no tweaks.

amaurymartiny: iframe done.

> Well, we are out of time

How would you do it if we had time?

jacogr: Simple answer - without too many interruptions from me. :)

1. Skip local non-matching
2. Allow opening in with target="_blank"` ipfs (with matching hash) and polkadot (cannot check hash 100%)

Not going to fight this, since it needs to go.

amaurymartiny:
Thanks everyone who participated in this PR 🎉

PR #2738, opened 2020-05-13
fix flow between claim and attest

amaurymartiny:
to be merged into tbaut's branch

### DIFF
--- a/packages/page-claims/src/index.tsx
+++ b/packages/page-claims/src/index.tsx
@@ -3,7 +3,7 @@
 // of the Apache-2.0 license. See the LICENSE file for details.
 
 import { Option } from '@polkadot/types';
-import { Balance, EcdsaSignature, EthereumAddress } from '@polkadot/types/interfaces';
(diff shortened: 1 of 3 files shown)


packages/page-claims/src/index.tsx
```diff
@@ -80,7 +79,7 @@ const ClaimsApp = (props: Props): React.ReactElement => {
   const [isPreclaimed, setIsPreclaimed] = useState(false);
 
   useEffect(() => {
-    if (isApiReady && typeof api.query.claims.claimAttest !== undefined) {
+    if (isApiReady && typeof api.query.claims.claimAttest !== 'undefined') {
```
amaurymartiny, line 82: Is there a guarantee that isApiReady will be `true` before rendering this component? Like somewhere higher in the components tree?

If yes, we could simplify this useEffect&useState into just
```ts
const isOldClaimProcess = !!api.query.claims.claimAttest;
```

PR #2742, opened 2020-05-13
Add condition for the warning banner

Tbaut:
this is meant to be merged into `tbaut-asset`

### DIFF
--- a/packages/apps/src/overlays/Attest.tsx
+++ b/packages/apps/src/overlays/Attest.tsx
@@ -24,7 +24,7 @@ function Attest ({ className }: Props): React.ReactElement<Props> | null {
   const [needAttest, setNeedAttest] = useState<Option<EthereumAddress>[]>([]);
 
   useEffect(() => {
-    if (typeof api.query.claims.preclaims !== undefined) {
+    if (typeof api.query?.claims?.preclaims !== undefined) {
       setIsOldClaimProcess(false);
     }
   }, [api]);
--- a/packages/apps/src/overlays/Attest.tsx
+++ b/packages/apps/src/overlays/Attest.tsx
@@ -2,10 +2,11 @@
 // This software may be modified and distributed under the terms
 // of the Apache-2.0 license. See the LICENSE file for details.
 
-import React from 'react';
-import { useApi, useAccounts, useCall } from '@polkadot/react-hooks';
+import React, { useEffect, useState } from 'react';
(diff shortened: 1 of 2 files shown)


packages/apps/src/overlays/Attest.tsx
```diff
@@ -18,13 +19,28 @@ interface Props {
...
-  const { api } = useApi();
+  const { api, isApiReady } = useApi();
+  const [isOldClaimProcess, setIsOldClaimProcess] = useState(true);
+  const [needAttest, setNeedAttest] = useState<Option<EthereumAddress>[]>([]);
+
+  useEffect(() => {
+    if (isApiReady && typeof api.query?.claims?.preclaims !== 'undefined') {
+      setIsOldClaimProcess(false);
+    }
+  }, [api, isApiReady]);
```
Tbaut, line 30: VSCode would like to add `api.query.claims.preclaims` to the depandancy array, but this makes `apps` to crash because it's not available when being loaded.

packages/apps/src/overlays/Attest.tsx
```diff
@@ -18,13 +19,28 @@ interface Props {
...
 
   // Find accounts that need attest. They are accounts that
   // - already preclaimed,
   // - didn't sign the attest yet.
   // `claims.preclaims` returns Some() for these accounts.
-  const needAttest = useCall<[Option<EthereumAddress>]>(api.query.claims?.preclaims.multi, [allAccounts])?.filter((opt) => opt.isSome);
+  useEffect(() => {
+    if (!isOldClaimProcess) {
+      api.query.claims.preclaims.multi<Option<EthereumAddress>>(allAccounts)
```
Tbaut, line 38: using `api` directly and not `useCall` because it needs to be conditionned.

amaurymartiny: Actually, would it be enough to just replace this line with (I added some `?`):

```ts
const needAttest = useCall<[Option<EthereumAddress>]>(api.query?.claims?.preclaims?.multi, [allAccounts])?.filter((opt) => opt.isSome);
```

if api is not ready, or the chain doesn't have the claims pallet, or the claims module doesn't have preclaims, then `needAttest` would be just undefined. Which won't show the banner.

Tbaut: yes, that works indeed!

Tbaut:
Much smaller now

Issue #2604, opened 2020-04-23
Specify gas costs in mega gas


athei:
There are upcoming changes for the contracts pallet which effectively unify gas and weight (https://github.com/paritytech/substrate/pull/5712). What this means for contract developers is that the gas costs are _way_ higher now as they are specified in weight (we still call it gas for now when communicating with contract developers).

Previously the cost for one wasm instruction was 1 gas. Now it is 500_000 weight = gas where every weight represent 1 pico second of execution on our reference machine. Currently, the contract UI only allows for 9 numbers to be entered which does not allow for any meaningful calculation.

I propose to change the field so that you specify **mega** gas which means that we have a granularity of micro second execution time there. Maybe even add a dynamically displayed calculation that shows how much execution time of the whole block the entered gas limit constitutes with a nice progress bar.

Example:
We enter 1_000_000 mega gas. Then we could display something like this:
1 second execution time, 50% of Block, [Progress Bar]

PR #2743, opened 2020-05-13
Set higher default for contracts maximum gas allowed

Stefie:
I read in one of the smart contract channels, that @kwingram25 is in the process of bigger contracts refactor. So this is just a a really basic quick fix for https://github.com/polkadot-js/apps/issues/2604 to enable people to instantiate their contracts through the UI until there's time for a more involved approach.

I just set higher default gas limit, and switched the bitlength for the gasLimit input field from u32 to u128.

### DIFF
--- a/packages/page-contracts/src/Contracts/Call.tsx
+++ b/packages/page-contracts/src/Contracts/Call.tsx
@@ -172,6 +172,7 @@ function Call (props: Props): React.ReactElement<Props> | null {
               value={endowment}
             />
             <InputNumber
+              bitLength={128}
               defaultValue={DEFAULT_GAS_LIMIT}
               help={t('The maximum amount of gas that can be used by this call. If the code requires more, the call will fail.')}
               isDisabled={isBusy}
--- a/packages/page-contracts/src/Modal.tsx
+++ b/packages/page-contracts/src/Modal.tsx
@@ -146,6 +146,7 @@ class ContractModal<P extends ContractModalProps, S extends ContractModalState> 
 
     return (
       <InputNumber
+        bitLength={128}
         help={t('The maximum amount of gas that can be used by this deployment, if the code requires more, the deployment will fail.')}
         isDisabled={isBusy}
         isError={!isGasValid}
--- a/packages/page-contracts/src/constants.ts
+++ b/packages/page-contracts/src/constants.ts
@@ -4,7 +4,7 @@
 
 export const ENDOWMENT = 1000;
 
-export const DEFAULT_GAS_LIMIT = '500000';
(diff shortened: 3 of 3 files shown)


Stefie:
I noticed that the api is not getting the correct parameters or api-contracts itself is not handling the parameters correctly. It's using the gasLimit as the `at: BlockHash`. I was looking into it myself  but couldn't pin it down, because the error could be both in apps or the api. I can create an issue, if you could point me to the repo where the bug is most likely located.

Screenshot 2020-05-13 at 20 14 48

jacogr:
Feel free to log, but I won't look at it for quite some time - these things are just impossible for me to play with or test and will always take the backseat to modules used on live chains where iteration is currently rapid.

But even the `Balance` parsing is broken in that RPC -

`{"id":51,"jsonrpc":"2.0","method":"contracts_call","params":[{"origin":"5CiPPseXPECbkjWCa6MnjNokrgYjMqmKndv2rSnekmSK2DjL","dest":"5CiPPseXPECbkjWCa6MnjNokrgYjMqmKndv2rSnekmSK2DjL","value":"0x0000000000000000002bdbb64bc09000","gasLimit":1000000,"inputData":"0x00"}]}`

fails with

`-32602: Invalid params: invalid type: string "0x0000000000000000002bdbb64bc09000", expected u128.`

That is properly constructed and fails. You cannot pass a `u128` over as a number in the JSON and it doesn't parse hex. so that call will never work with any non-zero balance. 

```rust
#[derive(Serialize, Deserialize)]
#[serde(rename_all = "camelCase")]
#[serde(deny_unknown_fields)]
pub struct CallRequest<AccountId, Balance> {
	origin: AccountId,
	dest: AccountId,
	value: Balance,
	gas_limit: number::NumberOrHex<u64>,
	input_data: Bytes,
}
```

TL;DR That RPC is currently unusable, despite the above error. Construction is correct based on schema.

Stefie:
> That RPC is currently unusable, despite the above error. Construction is correct based on schema.

Is that a general problem with RPC implementation or does it only apply for the contract rpc methods?

jacogr:
So couple of issue there -

1. Unlike the gas, it doesn't allow hex - you cannot represent numbers > 2^53-1 in JS JSON, so number generally should always allow for hex. (Which is the way it is being outputted)

2. We don't use this in other RPCs, generally everything is SCALE-encoded.

3. We had serious issues with paymentInfo to get the partial fee serialized (it was finally adjusted to only output strings). But this is out, not in.

Anyway, until fixed, nothing that can be done. The serialization is correct, the parsing thereof is not.

Stefie:
Okay, thanks for looking into it!

PR #2747, opened 2020-05-14
Show the amount of token that will be claimed

Tbaut:
To be merged in `tbaut-attest`

image

### DIFF
--- a/packages/page-claims/src/Attest.tsx
+++ b/packages/page-claims/src/Attest.tsx
@@ -2,33 +2,70 @@
 // This software may be modified and distributed under the terms
 // of the Apache-2.0 license. See the LICENSE file for details.
 
-import React from 'react';
+import React, { useEffect, useCallback, useState } from 'react';
 import styled from 'styled-components';
 import { Button, Card, TxButton } from '@polkadot/react-components';
+import { FormatBalance } from '@polkadot/react-query';
+import { Option } from '@polkadot/types';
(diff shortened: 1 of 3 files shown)


packages/page-claims/src/Attest.tsx
```diff
@@ -2,33 +2,70 @@
...
+        {hasClaim &&
+          <h3>{t('Attestation for: ')} <FormatBalance value={claimValue} /></h3>}
         <Button.Group>
           <TxButton
             accountId={accountId}
             icon='send'
             isPrimary
             label={t('Attest')}
-            params={['dsafdsalkfj']} // Replace with actual
+            params={[statementKind]} // Replace with actual
```
amaurymartiny, line 68: iiuc, we shouldn't attest the statementKind, but the actual statement in english: https://github.com/paritytech/polkadot/pull/1091/files#diff-6713ee0e82ee03f026aa21e5e99fb055R66

i.e. `statementKind` is an Enum, but we should put here the string matching the enum

amaurymartiny: We can leave as-is for now, as anything will work, we it should probably be replaced at some point with hardcoded strings like:
```tsx
statementKind.siDefault() ? 'statement1' : 'statement2'
```

Tbaut: yup absolutely, that's my understanding as well. Although right now the `statement === statemendKind`

PR #2759, opened 2020-05-17
Add block number to events on explorer page

danforbes:
image

#2756


danforbes:
Thank you for doing all the heavy lifting! I know this will make it a lot easier for people like Joe and I when we test out upgrades and stuff :rocket:

PR #2765, opened 2020-05-18
Basic Electron App setup

krzysztof-jelski:


packages/react-api/src/Api.tsx
```diff
@@ -22,6 +22,7 @@ import addressDefaults from '@polkadot/util-crypto/address/defaults';
 
 import ApiContext from './ApiContext';
 import registry from './typeRegistry';
+import initSettings from '@polkadot/apps/initSettings';
```
jacogr, line 25: Would prefer the API not import from `@polkadot/apps` (The ends up being a circular dep when all is said and done)

README.md
```diff
@@ -17,6 +17,7 @@ This can be accessed as a hosted application via [https://polkadot.js.org/apps/]
 The repo is split into a number of packages, each representing an application. These are -
 
 - [apps](packages/apps/) This is the main entry point. It handles the selection sidebar and routing to the specific application being displayed.
+- [apps-electron](packages/apps-electron/) Desktop app which is build based on [apps](packages/apps-electron/) package.
```
krzysztof-jelski, line 20:
```suggestion
- [apps-electron](packages/apps-electron/) Desktop app running [apps](packages/apps/).
```

package.json
```diff
@@ -1,12 +1,28 @@
 {
+  "name": "polkadot",
```
krzysztof-jelski, line 2:
```suggestion
  "name": "polkadot-apps",
```

Jozwiaczek: Earlier I left it here because that field is related to the electron app name but for now, I also add a field in electron build config for that, so field 'name' is no longer related to the electron build process.

packages/apps/public/locales/en/apps-electron.json
```diff
@@ -0,0 +1 @@
+{}
```
krzysztof-jelski, line 1: Does the lack of this file break anything?

packages/apps-electron/webpack.renderer.config.js
```diff
@@ -0,0 +1,46 @@
...
+const HtmlWebpackPlugin = require('html-webpack-plugin');
+const findPackages = require('../../scripts/findPackages');
+
+const ENV = process.env.NODE_ENV || 'development';
+const isProd = ENV === 'production';
+const context = __dirname;
+
+module.exports = merge(
+  baseConfig({
+    alias: findPackages().reduce((alias, { dir, name }) => {
```
krzysztof-jelski, line 20: Would it be possible to pull up the `alias` to the `base` config as well? It's the same here and in the webapp.

packages/react-api/src/Api.tsx
```diff
@@ -23,6 +23,7 @@ import addressDefaults from '@polkadot/util-crypto/address/defaults';
 
 import ApiContext from './ApiContext';
 import registry from './typeRegistry';
+import './initSettings';
```
jacogr, line 26: Not 100% sure this is 100% the right place, but let's keep it here, don't have better ideas atm.

tsconfig.json
```diff
@@ -5,8 +5,11 @@
     "**/build/**/*"
   ],
   "compilerOptions": {
+    "jsx": "react",
```
jacogr, line 8: This actually should have no impact at all since we build via Babel.

krzysztof-jelski: I think I added it just to make my IDE happy (WebStorm/Idea). Probably can find a way to do without.

packages/apps-electron/src/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+                <HashRouter>
+                  <Apps />
+                </HashRouter>
+              </AccountSidebar>
+            </Events>
+          </BlockAuthors>
+        </Api>
+      </Queue>
+    </ThemeProvider>
+  </Suspense>,
```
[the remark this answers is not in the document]
krzysztof-jelski: I believe they need to diverge soon, as we will use different services/providers in both apps. Still, the UI structure will likely stay the same. If you're OK with that I'd prefer to wait with extracting a shared component until we have a better idea of what is different and what is the same.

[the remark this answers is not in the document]
jacogr: Yes, happy to have it later.

tsconfig.json
jacogr: No, it is fine, doesn't do any harm. (The original intent is just lost, since it effectively has no impact :))

PR #2780, opened 2020-05-20
Publish electron apps on release

vanruch:
This PR should add electron executives (.dmg, .exe and linux binary) to release metadata.

Could not test it properly on a forked repo because forks are [readonly](https://github.com/actions/first-interaction/issues/10#issuecomment-546628432) so we'll have to check it doing a release on this repository.


jacogr:
That is fine. These things always require some tweaks, even when we are happy...

package.json
```diff
@@ -48,7 +48,7 @@
     "build:electronRenderer": "cd packages/apps-electron && NODE_ENV=production webpack --config webpack.renderer.config.js",
     "build:devElectronMain": "cd packages/apps-electron && webpack --config webpack.main.config.js",
     "build:electronMain": "cd packages/apps-electron && NODE_ENV=production webpack --config webpack.main.config.js",
-    "prepackElectron": "yarn build:electron && yarn postinstall:electron",
+    "prepackElectron": "yarn build && yarn build:electron && yarn postinstall:electron",
```
jacogr, line 51: Maybe rename this one to `build:release:electron` (For the sake of consistency). at the same time would need to remove the step for that in the workflow yml, otherwise it will be fired twice

.github/workflows/push-master.yml
```diff
@@ -27,6 +27,28 @@ jobs:
       run: |
         yarn install --immutable | grep -v 'YN0013'
         yarn ${{ matrix.step }}
+  release:
```
jacogr, line 30: For the time being, can we make it run each merge? Just until we are happy. (Otherwise it will be a bit of a nightmare atm to trigger via release)

vanruch: It will be run, just not released (e.g. when doing release, execs should be added to assets while on a normal merge, we don't have anywhere to put the apps)

PR #2794, opened 2020-05-22
Code signing and Notarization

Jozwiaczek:
## How to sign code
1. Open the Keychain Access app or the Apple Developer Portal. Export all certificates related to your app into a single file (e.g. certs.p12) and set a strong password
2. Base64-encode your certificates using the following command: base64 -i certs.p12 -o encoded.txt
3. In your project's GitHub repository, go to Settings → Secrets and add the following two variables:
    - mac_certs: Your encoded certificates, i.e. the content of the encoded.txt file you created before
    - mac_certs_password: The password you set when exporting the certificates

_The same goes for Windows code signing (windows_certs and windows_certs_password secrets)._

### Configure  App Store Connect API
1. Sign in to App Store Connect and open the API key page.
2. Create a new API key by clicking on the “+” symbol. Set access permissions to “App 
 Manager”
3. Download the key file and move it into the ~/private_keys/ directory, where altool will be able to find it
4. Copy your Issuer ID (1) and Key ID (2):
### Define the following secrets in your repository's settings on GitHub:
- api_key: Content of the API key file (with the p8 file extension)
- api_key_id: Key ID found on App Store Connect
- api_key_issuer_id: Issuer ID found on App Store Connect


.github/workflows/push-master.yml
```diff
@@ -40,16 +40,36 @@ jobs:
...
+          # Import Apple API key for app notarization on macOS
+          run: |
+            mkdir -p ~/private_keys/
+            echo '${{ secrets.api_key }}' > ~/private_keys/AuthKey_${{ secrets.api_key_id }}.p8
+
       - name: Build/release Electron app
         uses: samuelmeuli/action-electron-builder@v1
         with:
+          # Base64-encoded code signing certificate for macOS
+          mac_certs: ${{ secrets.mac_certs }}
```
jacogr, line 55: Not 100% if actions are clever enough to check the case, but it didn't allow me to add lowercase names, so the names are correct, but it is exposed as `secrets.MAC_CETRS` etc. May be worth-while to just adjust the `secrets.*` to uppercase, just in-case. (Even if it does work and converts)

jacogr:
Let's see...

PR #2797, opened 2020-05-22
Enhance UX for claim process

Tbaut:
- remove the `Statement` component for old process (that was a bug)
- show a friendly message when the ethereum address did corresponds to neither a Default or Alternative statement kind.
- add the (non final) md

### DIFF
--- a/packages/page-claims/src/index.tsx
+++ b/packages/page-claims/src/index.tsx
@@ -235,7 +235,7 @@ function ClaimsApp (): React.ReactElement {
             )}
           {(step >= Step.Sign && !isPreclaimed) && (
             <Card>
-              {(isOldClaimProcess || (statementKind?.isAlternative || statementKind?.isDefault))
+              {(isOldClaimProcess || !!statementKind)
                 ? <>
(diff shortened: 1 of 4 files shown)


packages/page-claims/src/index.tsx
```diff
@@ -235,46 +235,52 @@ function ClaimsApp (): React.ReactElement {
...
-                rows={10}
-              />
-              {(step === Step.Sign) && (
-                <Button.Group>
-                  <Button
-                    icon='sign-in'
-                    isDisabled={!accountId || !signature}
-                    label={t('Confirm claim')}
-                    onClick={goToStepClaim}
+              {(isOldClaimProcess || (statementKind?.isAlternative || statementKind?.isDefault))
```
amaurymartiny, line 238:
```suggestion
              {(isOldClaimProcess || !!statementKind)
```

If i'm not mistaken, that should work too, because statementKind will be `null | undefined` in that case.

Tbaut: yes this works, thanks

PR #2802, opened 2020-05-24
Bumps w/ eslint

jacogr:


jacogr:
Not perfect, but already better.

PR #2817, opened 2020-05-26
Electron updater

Jozwiaczek:
It will not work without properly setup code-signing.


package.json
```diff
@@ -104,6 +104,7 @@
     "terser-webpack-plugin": "^3.0.1",
     "webpack": "^4.43.0",
     "webpack-cli": "^3.3.11",
+    "webpack-dev-server": "^3.11.0",
```
jacogr, line 107: Can't we just use the plugin? We effectively have 2 things doing the same.

Jozwiaczek: Yes, I will change that in sperate PR.

krzysztof-jelski, line 107: This is removed in master, so probably a mistake in merge.

packages/apps-electron/src/electron.ts
```diff
@@ -2,10 +2,12 @@
 // This software may be modified and distributed under the terms
 // of the Apache-2.0 license. See the LICENSE file for details.
 
-import { BrowserWindow, app, screen } from 'electron';
+import { BrowserWindow, app, screen, dialog } from 'electron';
 import path from 'path';
+import { autoUpdater } from 'electron-updater';
```
jacogr, line 7:
```suggestion
import { BrowserWindow, app, dialog, screen } from 'electron';
import { autoUpdater } from 'electron-updater';
import path from 'path';
```

packages/apps-electron/src/featureToggles.ts
```diff
@@ -0,0 +1,7 @@
+function getBoolEnv(env: string | undefined, fallback: boolean) {
```
jacogr, line 1:
```suggestion
// Copyright 2017-2020 @polkadot/apps authors & contributors
// This software may be modified and distributed under the terms
// of the Apache-2.0 license. See the LICENSE file for details.

function getBoolEnv(env: string | undefined, fallback: boolean) {
```

packages/apps-electron/src/electron.ts
```diff
@@ -29,4 +32,29 @@ function createWindow (): Promise<unknown> {
   return win.loadFile(mainFilePath);
 }
 
-app.whenReady().then(createWindow).catch(console.error);
+const onReady = async () => {
+  await createWindow();
+  if (features.autoUpdater) {
```
jacogr, line 37:
```suggestion
  await createWindow();
  
  if (features.autoUpdater) {
```

packages/apps-electron/src/electron.ts
```diff
@@ -44,15 +44,15 @@ const onReady = async () => {
 app.whenReady().then(onReady).catch(console.error);
 
 if (features.autoUpdater) {
-  autoUpdater.on('update-not-available', async () => {
-    await dialog.showMessageBox({
+  autoUpdater.on('update-not-available', () => {
+    // eslint-disable-next-line no-void
+    void dialog.showMessageBox({
```
jacogr, line 49: Doing `.catch(console.error)` on line 52 is probably more like what we use elsewhere. So swallow, but at least let the dev know something happenned. Or even `.catch(() => undefined)` if we really don't care at all.

jacogr: EDIT: Nothing wrong with the override, I just feel like I added way too much especially after the bump to typescript-eslint 3 (which is a bigger pain in the behind), so would like to avoid some even if we just swallow.

krzysztof-jelski: Done.

Issue #2827, opened 2020-05-27
Bug in Kusama Claims


lsaether:
The new changes for Polkadot claims introduced a bug in Kusama claims.

![image](https://user-images.githubusercontent.com/25539605/82984522-9fcaee00-9ff2-11ea-9ed2-cd4d0767f845.png)

I believe the bug is [here](https://github.com/polkadot-js/apps/blob/0091fc6e57f10e2ed10a6cbb2ea052f5faa6a0c7/packages/page-claims/src/Claim.tsx#L43)

```ts
// Depending on isOldClaimProcess, construct the correct tx.
function constructTx (
  systemChain: string,
  accountId: string,
  ethereumSignature: string | null,
  kind: StatementKind | undefined,
  isOldClaimProcess: boolean
): ConstructTx {
  if (!ethereumSignature || !kind) {
    return {};
  }

  return isOldClaimProcess
    ? { params: [accountId, ethereumSignature], tx: 'claims.claim' }
    : { params: [accountId, ethereumSignature, getStatement(systemChain, kind)?.sentence], tx: 'claims.claimAttest' };
}
```

It should only check `kind` if it's on the Polkadot chain.

PR #2828, opened 2020-05-27
Kusama claims fix

jacogr:
- Introduced in https://github.com/polkadot-js/apps/pull/2734
- Closes https://github.com/polkadot-js/apps/issues/2827

@lsaether would like a check from you to make sure it seems sound, since you specifically found the cause

### DIFF
--- a/packages/page-claims/src/Claim.tsx
+++ b/packages/page-claims/src/Claim.tsx
@@ -40,13 +40,15 @@ function constructTx (
   kind: StatementKind | undefined,
   isOldClaimProcess: boolean
 ): ConstructTx {
-  if (!ethereumSignature || !kind) {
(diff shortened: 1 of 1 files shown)


lsaether:
Yep give me a few minutes to test it on Kusama dev chain to make sure that it's the cause.

lsaether:
Took some tinkering around but I was able to find the right runtime. I got some `system.event` errors when making the claim but I guess this was due to not having the right metadata for my Kusama dev chain (?). Anyway, the tokens showed up in my account.

Here are the system errors (in case they're relevant):

image

jacogr:
Yes, the above is some type mismatch - we adjusted Polkadot in the last day. Thanks for this! Hopefully we didn't get too many people bumping into the issue

Issue #2829, opened 2020-05-27
DOT claim, selecting address fails with "Unable to retrieve keypair"


ascjones:
I mistakenly selected a contact address (rather than an account) from the "Select your Polkadot CC1 account" dropdown on `Claims`. Should this be allowed at all if there is no keypair for signing the claim? I think @0x7cfe may have reported something similar in Riot. At the very least it should not give the Whte Screen of Death.

Repro:

1. Add address as a contact in `Accounts > My contacts`
2. In `Claims`, select the address from the dropdown
3. The balance to be claimed together with the Ts&Cs are displayed
4. Click `I agree` and then the white screen of death. Error from the console: `Error: Unable to retrieve keypair '0x...`

jacogr:
Didn't see any other issues logged here, so can't comment about the cause for other issues.

But yes, based on  #2734 the `type='account'` is missing in the dropdown.

0x7CFE:
> I think @0x7CFE may have reported something similar in Riot.

Yep, it was exactly that. I tried pasting claim address without adding it to the accounts. I was in the middle of writing the issue but you did it first :)

PR #2830, opened 2020-05-27
Only allow actual accounts to claim

jacogr:
Closes https://github.com/polkadot-js/apps/issues/2829

### DIFF
--- a/packages/page-claims/src/index.tsx
+++ b/packages/page-claims/src/index.tsx
@@ -196,7 +196,7 @@ function ClaimsApp (): React.ReactElement {
               help={t<string>('The account you want to claim to.')}
               label={t<string>('claim to account')}
               onChange={setAccountId}
-              type='all'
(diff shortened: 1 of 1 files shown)


lsaether:
Won't this be a problem for people who don't want to expose their accounts to Apps? I'm thinking in the situation of cold storage. Would this still allow for pasting in an address that's not plugged in as an account?

jacogr:
You can't sign with it, so it breaks completely.

lsaether:
That's true for the attestations but the component is also be used for old style Claims using the Ethereum signature (correct me if I'm wrong). So someone could keep their Polkadot keys in cold storage but still use their Eth keys to make the claim. There was at least one person who did it this way during Kusama claims.

jacogr:
Agreed, which is what it was `type='all'` - it should be a check based on type.

PR #2862, opened 2020-05-29
Fix electron auto release

Jozwiaczek:
Set releasing flag as 'always', extract author name and email into a separate field for auto-releasing, set Linux config to publish only on GitHub


.github/workflows/push-master.yml
```diff
@@ -49,6 +49,7 @@ jobs:
           # (No need to define this secret in the repo settings)
           github_token: ${{ secrets.github_token }}
           build_script_name: build:release:electron
+          release: true
```
jacogr, line 52: Not 100% sure I want to switch this on - it means we made one of each merge, and there are a lot.

PR #2865, opened 2020-06-01
Metamask snap as web3 provider

mpetrunic:
First of all, I don't think we should merge this just yet because:
- there is still no official metamask release with support for snap
- there are some outstanding ux issues with metamask snaps

This is more poc of how we could use metamask for interacting with Polkadot and it's part of our grant milestone.

To test it out, you can install our release of metamask with snap support (latest official develop branch): https://github.com/nodefactoryio/metamask-snap-polkadot#metamask

Implementation reasoning:
- we aren't injecting Metamask into providers by default because it will start shooting prompts to install plugin so if user clicks to add metamask, it will prompt to install plugin and inject account using your standard methods and providers
- we are lacking some better way to detect if user's metamask supports plugins(snaps), once we figure that out we can display some warning or even hide the button

We would love to hear feedback and ideas how to integrate this better :slightly_smiling_face:


jacogr:
This is brilliant! I am really excited. I agree with the approach here, rather start playing sooner and let's get ready.

Will give it a play through and start going through the code.

jacogr:
Before I forget (and I will), once there is an official release and along with this PR, we just need to add a link to the extension/snap in here as well - https://github.com/polkadot-js/apps/blob/master/packages/page-accounts/src/Accounts/BannerExtension.tsx

joelamouche:
@mpetrunic Hi, what is the status of this? Is that still something you want to implement?

mpetrunic:
> @mpetrunic Hi, what is the status of this? Is that still something you want to implement?

We are still waiting for metamask to release snaps. Feel free to close PR. I will give it another go once ready.

jacogr:
No issues. I have been checking in on them from time to time as well :)

joelamouche:
FYI allowing users to sign transactions with MetaMask is top priority for my team (moonbeam.network) and we want to do anything in our power to push this.

I made this PR in December to inject web3 into the app but in order to sign transactions we need to access the sign function `eth_sign` that isn't prepended with a message as specified in https://github.com/ethereum/go-ethereum/pull/2940 but this function has been removed because deemed too dangerous.
https://github.com/polkadot-js/extension/pull/566

We are advocating for a revert of this change where `eth_sign` would be accessible with a proper warning.
Unfortunately Im not very optimistic about this change being passed and in the mean time we are looking at SNAPS as a possible solution. I'm not super clear on wether or not `eth_sign` will be available in SNAPS

jacogr:
Going to close this one. Not because it is not important, but rather since it (atm) doesn't see much action. Happy to see it return when snaps do make an appearance.

joelamouche:
yes indeed: for now we are focusing on getting MetaMask compatibility in the apps, and we will be postponing the creation of a snap until it's stable

PR #2873, opened 2020-06-03
Rewrite & update to page-contracts

### LINKED ISSUE #2604: Specify gas costs in mega gas

kwingram25:
Closes #2604 

- Migrate classes to functional
- Rewrite Row and related
- Extra hooks


packages/page-contracts/src/Contracts/Call.tsx
```diff
@@ -35,10 +36,10 @@ function Call (props: Props): React.ReactElement<Props> | null {
...
   const callMessage = callContract?.getMessage(isNull(callMessageIndex) ? undefined : callMessageIndex);
 
-  const [accountId, setAccountId] = useState<StringOrNull>(null);
-  const [endowment, setEndowment] = useState<BN>(new BN(0));
-  const [gasLimit, setGasLimit] = useState<BN>(new BN(DEFAULT_GAS_LIMIT));
-  const [isBusy, setIsBusy] = useState(false);
+  const [accountId, setAccountId] = useAccountId();
+  const [endowment, isEndowmentValid, setEndowment] = useFormField<BN>(new BN(0));
+  const [gasLimit, isGasLimitValid, setGasLimit] = useNonZeroBn(new BN(GAS_LIMIT));
+  const [isBusy, , setIsBusy] = useToggle();
```
erler, line 42:
```suggestion
  const [isBusy, setIsBusy] = useToggle();
```

jacogr: This is not quite correct - `[isBusy, toggleIsBusy, setIsBusy]`, so the usage here is correct.

erler: Ah sorry, my bad

jacogr: No issues, extra eyeballs are _always_ good.

packages/react-components/src/Tags.tsx
```diff
@@ -46,6 +46,7 @@ function Tags ({ children, className, color = 'grey', isEditable, isEditing, onC
 
   const _onSave = useCallback(
     (): void => {
+      console.log('_onSave');
```
jacogr, line 49:
```suggestion
```

PR #2880, opened 2020-06-04
Electron app: save accounts to FS, do not download backup, hide warning banner

vanruch:


packages/page-accounts/src/Accounts/modals/Create.tsx
```diff
@@ -147,7 +147,11 @@ function createAccount (suri: string, pairType: KeypairType, { genesisHash, name
     status.status = 'success';
     status.message = success;
 
-    downloadAccount(result);
+    InputAddress.setLastValue('account', address);
+
+    if (getEnvironment() === 'web') {
```
jacogr, line 152: Didn't even think of this, good catch!

Issue #2916, opened 2020-06-10
Uncaught Error: Extrinsic staking.withdrawUnbonded


jiangfuyao:
```
<https://polkadot.js.org/apps/#/staking/actions>
Kusama Network
Node:  Kusama (Polkadot Canary, hosted by Web3 Foundation)
Parity Polkadot v0.8.4-8121b60-x86_64-linux-gnu
api v1.18.1
apps v0.44.0-beta.68
```
When finished unbond locking time.
Click Lock icon(`Withdraw these unbonded funds`), got error:

```
Uncaught Error: Extrinsic staking.withdrawUnbonded expects 1 arguments, got 0.
    at e.default (polkadotjs.bc1f9cfd.js:1)
    at I (polkadotjs.bc1f9cfd.js:1)
    at Object.withdrawUnbonded (polkadotjs.bc1f9cfd.js:1)
    at Object.onClick (main.51284b77.js:23)
    at t.exports (vendor.76ce7a47.js:8)
    at t.exports (vendor.76ce7a47.js:8)
    at t.exports (vendor.76ce7a47.js:8)
    at vendor.76ce7a47.js:8
    at react.707119cd.js:32
    at Object.l (react.707119cd.js:25)
```
![image](https://user-images.githubusercontent.com/34789555/84245384-89b34680-ab37-11ea-934b-aaf06c799d32.png)

jacogr:
*sigh* Was changed - https://github.com/paritytech/substrate/blob/master/frame/staking/src/lib.rs#L1594

Ref, no needs slashing spans - https://github.com/paritytech/substrate/pull/5964

PR #2918, opened 2020-06-10
Withdraw with slashing spans

jacogr:
Closes https://github.com/polkadot-js/apps/issues/2916


jacogr:
Will get in, need a release out - at the very least it will allow non-slashed people to withdraw. Can re-visit if wrong, aka the non-slashed path should be happy, on slashed ymmv :)

Issue #2939, opened 2020-06-11
Broken claims page on Kusama? 


joshua-mir:
The claims page on Kusama appears to claim that accounts do not have valid claims when according to chainstate, they do.

(verified with a couple of random addresses from https://gist.github.com/joshua-mir/b9c78bd8524cc9db2b1e36d4042aab78), so I assume this applies to all claims.

jacogr:
The runtime upgrade broke it. The claims page detects it as "invalid" and since there is no statement, doesn't show the claims. So indeed, the logic is broken - had the same issue with somebody on CC1 which didn't have a statement. (He got it working through this branch https://github.com/polkadot-js/apps/tree/jg-no-statementKind)

jacogr:
Closed in #2941

PR #2941, opened 2020-06-12
Handle no statementKind

jacogr:
Seems we need this for #2939 and it is not a once-off, Kusama is broken.

### DIFF
--- a/packages/page-claims/src/Claim.tsx
+++ b/packages/page-claims/src/Claim.tsx
@@ -33,22 +33,14 @@ interface ConstructTx {
 }
 
 // Depending on isOldClaimProcess, construct the correct tx.
-function constructTx (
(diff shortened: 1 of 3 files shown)


jacogr:
cc @lsaether Is it possible that you can just check Kusama & CC1 for me - the first without a statement, the second with. (Originally actually started this for somebody without a statement on CC1, since the upgrade Kusama is also broken since the new module went in)

Have tested that it does handle the no statement now, so it goes to the next step on Kusama, but would like to ensure it can actually be claimed...

lsaether:
Yep, sorry just getting to this. Let me test and report back here in a few minutes...

jacogr:
@lsaether I owe you are case of beers. Thank you!

PR #2944, opened 2020-06-12
feat: support import with mini secret key

### LINKED ISSUE #2885: Enable import derived account from Parity Signer
It would be good to import the derived account from signer to polkadot.js app 

Related issue on Parity Signer [here](https://github.com/paritytech/parity-signer/issues/640)

Pokadot.js app has several ways for importing the keypairs from Keyrings of Polkadot.js/common , and there is [addFromSeed](https://github.com/polkadot-js/common/blob/master/packages/keyring/src/keyring.ts#L121-L135) function, but I am not sure if it use the SecretKey or MiniSecretKey, since in our case only SecretKey would fit, while the MiniSecretKey is used to generate for the root account (non-derivated account).

P.S I did some research on Substrate and I find [from_seed_slice](https://github.com/paritytech/substrate/blob/master/primitives/core/src/sr25519.rs#L486) could use SecretKey to generate Keypairs but it is not included in the Subkey tool AFAIK. The subkey tools can not `inspect` a SecretKey to keypairs.  The [seed](https://github.com/paritytech/substrate/blob/master/bin/utils/subkey/src/main.rs#L93) here actually refer to MiniSecretKey for generate the non-derivation account.

In addition to that, we could use `Add Via QR` button, if we find it is a QR representing a SecretKey hex , then add it to keyring and regard it as an internal account.

hanwencheng:
closes #2885

based on UI's PR: https://github.com/polkadot-js/ui/pull/330#issue-433737179

This PR add support for importing `mini_secret_key` from Parity Signer.

It still use the `Add via QR` button, but if the QR code represent a secret (or call it seed), it can import the account into keyring instead of regarding it as an external account.
 
Screen Shot 2020-06-12 at 17 11 31


packages/page-accounts/src/Accounts/modals/Qr.tsx
```diff
@@ -36,7 +36,7 @@ function QrModal ({ className = '', onClose, onStatusChange }: Props): React.Rea
   const password2StateHooks = useState(defaultPasswordState);
   const [{ isPassValid: isPass2Valid }] = password2StateHooks;
 
-  const isValid = !!address && isNameValid && isPassValid && isPass2Valid;
+  const isValid = isAddress ? (!!address && isNameValid) : (!!address && isNameValid && isPassValid && isPass2Valid);
```
jacogr, line 39:
```suggestion
  const isValid = isAddress
    ? (!!address && isNameValid)
    : (!!address && isNameValid && isPassValid && isPass2Valid);
```

Getting a bit too much to read on a single line (should probably also remove the duplication there now)

packages/page-accounts/src/Accounts/modals/Qr.tsx
```diff
@@ -26,6 +28,15 @@ function QrModal ({ className = '', onClose, onStatusChange }: Props): React.Rea
...
   const [scanned, setScanned] = useState<Scanned | null>(null);
+  const [isAddress, setIsAddress] = useState<boolean>(false);
+  const [address, setAddress] = useState<string>('');
+  const defaultPasswordState = { isPassValid: false, password: '' };
+  const passwordStateHooks = useState(defaultPasswordState);
+  const [{ isPassValid, password }] = passwordStateHooks;
+  const password2StateHooks = useState(defaultPasswordState);
+  const [{ isPassValid: isPass2Valid }] = password2StateHooks;
+
+  const isValid = isAddress ? (!!address && isNameValid) : (!!address && isNameValid && isPassValid && isPass2Valid);
```
jacogr, line 39: Alternative without supes...

```suggestion
  const isValid = !!address && isNameValid && (isAddress || (isPassValid && isPass2Valid));
```

packages/page-accounts/src/Accounts/modals/Qr.tsx
```diff
@@ -101,6 +127,8 @@ function QrModal ({ className = '', onClose, onStatusChange }: Props): React.Rea
                   <p>{t<string>('The local name for this account. Changing this does not affect your on-line identity, so this is only used to indicate the name of the account locally.')}</p>
                 </Modal.Column>
               </Modal.Columns>
+              {!isAddress && <PasswordInput onEnter={_onSave}
+                passwordStates={[passwordStateHooks, password2StateHooks]}/>}
```
jacogr, line 131:
```suggestion
              {!isAddress && (
                <PasswordInput
                  onEnter={_onSave}
                  passwordStates={[passwordStateHooks, password2StateHooks]}
                />
              )}
```

Here really would prefer to no create a new object each render for passwordStates. Depending on what we need rather put the hooks inside the component itself and have a onChange that sends a single back to the parent. (When all is valid)

hanwencheng:
Here is the QR code for testing

| Address QR | Seed QR|
|---|---|
|11592217730_ pic|21592217730_ pic|

packages/page-accounts/src/Accounts/PasswordInput.tsx
```diff
@@ -0,0 +1,74 @@
...
+  const { t } = useTranslation();
+  const [isPassValid, setPassValid] = useState<boolean>(false);
+  const [{ isPass2Valid, password2 }, setPassword2] = useState({ isPass2Valid: false, password2: '' });
+
+  const _onPasswordChange = useCallback(
+    (password: string) => {
+      const isPassValid = keyring.isPassValid(password);
+
+      setPassValid(isPassValid);
+      const isValid = isPassValid && isPass2Valid;
```
jacogr, line 28:
```suggestion
      setPassValid(isPassValid);

      const isValid = isPassValid && isPass2Valid;
```

packages/page-accounts/src/Accounts/PasswordInput.tsx
```diff
@@ -0,0 +1,74 @@
...
+    },
+    [onChange, isPass2Valid]
+  );
+
+  const onPassword2Change = useCallback(
+    (password2: string) => {
+      const isPass2Valid = keyring.isPassValid(password2) && (password2 === password);
+
+      setPassword2({ isPass2Valid, password2 });
+      const isValid = isPassValid && isPass2Valid;
```
jacogr, line 40:
```suggestion
      setPassword2({ isPass2Valid, password2 });

      const isValid = isPassValid && isPass2Valid;
```

packages/page-accounts/src/Accounts/modals/Create.tsx
```diff
@@ -306,31 +301,9 @@ function Create ({ className = '', onClose, onStatusChange, seed: propsSeed, typ
...
-              value={password2}
-            />
-          </Modal.Column>
-          <Modal.Column>
-            <p>{t<string>('The password and password confirmation for this account. This is required to authenticate any transactions made and to encrypt the keypair.')}</p>
-          </Modal.Column>
-        </Modal.Columns>
+        <PasswordInput onChange={_onPasswordChange}
+          onEnter={_onCommit}
+          password={password}/>
```
jacogr, line 306:
```suggestion
        <PasswordInput
          onChange={_onPasswordChange}
          onEnter={_onCommit}
          password={password}/>
```

Not sure why the linter didn't complain here.

packages/page-accounts/src/Accounts/modals/Qr.tsx
```diff
@@ -103,6 +130,9 @@ function QrModal ({ className = '', onClose, onStatusChange }: Props): React.Rea
                   <p>{t<string>('The local name for this account. Changing this does not affect your on-line identity, so this is only used to indicate the name of the account locally.')}</p>
                 </Modal.Column>
               </Modal.Columns>
+              {!isAddress && <PasswordInput onChange={_onPasswordChange}
+                onEnter={_onSave}
+                password={password}/>}
```
jacogr, line 135:
```suggestion
              {!isAddress && (
                <PasswordInput
                  onChange={_onPasswordChange}
                  onEnter={_onSave}
                  password={password}
                />
              )}
```

jacogr:
Thank you!

PR #2969, opened 2020-06-16
25 modules translated from ES_es

wimel:
Translate to Spanish.


jacogr:
Thanks for this! 

Only the `es/translation.json` is needed since it is the combined version - en has multiples since it indicates the split in the translation UI, but actually for all languages the single combined version is loaded/used. (Hence the i18n I only outputting the single version)

wimel:
Ok, sorry I think is good if we can check the work done by modules, at least for me is easier _(that's quite a few words)_.
Do you think it's better to put it all in one `json` with name `translation.json` under `es` folder?

jacogr:
100%, but the UI is setup to only read `translation.json`. So the UI splits it by the modules, but generates a single file when using https://polkadot.js.org/apps/#/settings/i18n (which as of today also has an "Apply" button to see changes live on the UI)

Additionally then it also de-dupes keys, i.e. "Vote" may appear in a number of modules.

So I am guessing it depends on the approach, when editing the JSON directly, it certainly helps having the single JSON files, but the UI doesn't read them at all.

wimel:
> 100%, so the UI is setup to only read `translation.json`, ignoring everything else. So the UI splits it by the modules for editing, but generates a single file when using https://polkadot.js.org/apps/#/settings/i18n (which as of today also has an "Apply" button to see changes live on the UI)
> 
> Additionally then it also de-dupes keys, i.e. "Vote" may appear in a number of modules, but only one edit is actually required.
> 
> So I am guessing it depends on the approach, when editing the JSON directly, it certainly helps having the single JSON files, but the UI doesn't read them at all.

ok, one of the times in the translation I lost the connection with the node and the work was lost :sob: , from that point I started to make the process by modules and saving each one separately, thanks @jacogr

jacogr:
Grrrr. That sucks. Well, it is good feedback though, can actually do something about that. So there are def 2 gaps -

- when adding a new language without a dropdown entry, it needs to appear in the list (So "apply" would work)
- auto background save so changes are not lost

How did you combine the separate into the translation.json file? (Double-translation?)

jacogr:
Think we may be off-topic a bit - I am really happy to merge this (rather I want to ASAP), the should just remove the extra files since only the last file in the PR with the collection is actually uses. 

Love these PRs, it adds a lot of value to users :)

wimel:
Hi @jacogr again the same problem, if you lose the connection with the node don't save the work done :cry: and if you are doing some translation on a module and lose the connection, the work done is lost, I think something is worng with auto background _(I don't know how I can record this problem to try and fix it, but every x time it seems to lose the connection to the node and it hasn't saved the job properly)_.

How I've done the work, I've been translating module by module and downloading that json file, now to solve this what I am doing is "redoing" the work by hand by copying and pasting the translation, but I see that new things have been added that I have to translate, never ends  hahaha.

Again I don't know how we can solve this, but I think the key is a modular translation, not a json with everything, in this way I think that it can also be easier to correct by modules, since in Spanish sometimes the same word has been translated in several ways so that everything is better understood.
If there's anything I can do to help in this, I'm here

jacogr:
We seem to be missing each other - I cannot take all 25, since the UI i18n will only read the one single file per language. So if you want to do them separately and out of process, that is all fine, but I still cannot merge 25 files like in this PR, I need a single one.

wimel:
> We seem to be missing each other - I cannot take all 25, since the UI i18n will only read the one single file per language. So if you want to do them separately and out of process, that is all fine, but I still cannot merge 25 files like in this PR, I need a single one.

No problem, I do this "by hand"  and we can see for the next time how can we "improve" this.
Thanks!

PR #2984, opened 2020-06-19
WIP: korean translation

hskang9:


jacogr:
Going to merge this as-is alongside https://github.com/polkadot-js/apps/pull/2986 - when in the UI it means it will have greater reach (even if very much at the start) and hopefully brings in some more users to try their hand at submitting parts.

PR #3004, opened 2020-06-20
[encointer] discriminate chain types noTEE and TEE proxy 

brenzi:
with separate types and logos


jacogr:
As always, thank you!

PR #3019, opened 2020-06-21
Media adjustments with AddressSmall

jacogr:


packages/page-staking/src/Targets/Validator.tsx
```diff
@@ -99,7 +99,7 @@ function Validator ({ canSelect, filterName, info, isSelected, toggleFavorite, t
       <td className='address all'>
         <AddressSmall value={accountId} />
       </td>
-      <td className='number'>{numNominators || ''}</td>
+      <td className='number ui--media-1100'>{numNominators || ''}</td>
```
jacogr, line 102:
```suggestion
      <td className='number ui--media-1200'>{numNominators || ''}</td>
```

Align with header.

PR #3040, opened 2020-06-24
Electron build - restore zip target

krzysztof-jelski:
Extract Electron Builder config to a separate file.

Restore zip target for mac (needed for automatic updates - compare [note 2. in auto update guide](https://www.electron.build/auto-update#quick-setup-guide)).

### DIFF
--- a/electron-builder.yml
+++ b/electron-builder.yml
@@ -7,6 +7,7 @@ files:
   - ./packages/apps-electron/appleEntitlements
 appId: com.polkadotjs.polkadotjs-apps
 mac:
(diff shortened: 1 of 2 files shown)


jacogr:
This is a exceptionally confusing file to have - since `.zip` is on all platforms. Without looking at it specifically, it will be downloaded for the wrong platforms. (It has no indication that is it "Mac" or "non-user-installable" :()

krzysztof-jelski:
Added `-mac` to the mac artifact names, so the zip file will be like `Polkadot-JS-Apps-mac-0.47.0-beta.4.zip`

jacogr:
Perfect.

PR #3043, opened 2020-06-24
Log update logs to file instead of showing dialogs

vanruch:


packages/apps-electron/src/electron/autoUpdater.ts
```diff
@@ -0,0 +1,20 @@
...
+// of the Apache-2.0 license. See the LICENSE file for details.
+
+import { AppUpdater } from 'electron-updater';
+
+export async function setupAutoUpdater (): Promise<void> {
+  const { autoUpdater } = await import('electron-updater');
+
+  await setLogger(autoUpdater);
+  // eslint-disable-next-line @typescript-eslint/no-floating-promises
+  autoUpdater.checkForUpdatesAndNotify();
```
krzysztof-jelski, line 12: Adding `await` here should allow us to get rid of `eslint-disable`.

vanruch: But do we want it? I thought it was omitted on purpose so that it doesn't slow down startup

jacogr: You can just do `autoUpdater.checkForUpdatesAndNotify().catch(console.error);` - not  bad idea anyway

PR #3085, opened 2020-06-29
fix vesting

kaichaosun:


ltfschoen:
Thanks this works!

jacogr:
The module check is incorrect, but this one is the wrong way around as well. :) (Fixes for no vesting module, but doesn't appear correctly when non-zero)

PR #3093, opened 2020-06-29
Group democracy locks

### LINKED ISSUE #3088: Combine expired democracy locks in display
Currently we have the following -

image

Adjust the locks do it displays with grouping for expired and ongoing, something along these lines with the grouping using less space

```
#44 220 Locked 1x
#45 221 Locked 2x
#52 220 Locked 1x
  lock expired

#66 220 None
#67 221 Locked 1x
  ongoing referendum

#46 220 Locked6x
  expres in ...
```

jacogr:
Closes #3088

Not perfect, but better.

### DIFF
--- a/packages/react-components/src/DemocracyLocks.tsx
+++ b/packages/react-components/src/DemocracyLocks.tsx
@@ -37,7 +37,7 @@ let id = 0;
 // group by header & details
 //   - all unlockable together
 //   - all ongoing together
-//   - unlcokes are displayed individually
+//   - unlocks are displayed individually
 function groupLocks (t: <T = string> (key: string, options?: { replace?: Record<string, string> }) => T, bestNumber: BN, locks: DeriveDemocracyLock[] = []): State {
   return {
     maxBalance: bnMax(...locks.map(({ balance }) => balance)),
--- a/packages/react-components/src/DemocracyLocks.tsx
+++ b/packages/react-components/src/DemocracyLocks.tsx
@@ -5,7 +5,7 @@
 import { DeriveDemocracyLock } from '@polkadot/api-derive/types';
 
 import BN from 'bn.js';
-import React, { useState } from 'react';
+import React, { useEffect, useState } from 'react';
 import styled from 'styled-components';
 import { useApi, useCall } from '@polkadot/react-hooks';
 import { BlockToTime, FormatBalance } from '@polkadot/react-query';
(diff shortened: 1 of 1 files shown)


packages/react-components/src/DemocracyLocks.tsx
```diff
@@ -20,50 +20,101 @@ interface Props {
...
+        return sorted;
+      }, [])
+  };
+}
+
 function DemocracyLocks ({ className = '', value }: Props): React.ReactElement<Props> | null {
   const { t } = useTranslation();
   const { api } = useApi();
   const bestNumber = useCall<BN>(api.derive.chain.bestNumber, []);
   const [trigger] = useState(`${Date.now()}-democracy-locks-${++id}`);
```
Tbaut, line 88: unrelated: what is the benefit of doing this rather than using a `const`?

packages/react-components/src/DemocracyLocks.tsx
```diff
@@ -20,50 +20,101 @@ interface Props {
...
+  maxBalance: BN;
+  sorted: Entry[];
+}
+
 let id = 0;
 
+// group by header & details
+//   - all unlockable together
+//   - all ongoing together
+//   - unlcokes are displayed individually
```
Tbaut, line 40:
```suggestion
//   - unlocks are displayed individually
```

packages/react-components/src/DemocracyLocks.tsx
```diff
@@ -20,50 +20,101 @@ interface Props {
...
         icon='clock'
         tooltip={trigger}
       />
-      <FormatBalance value={max} />
+      <FormatBalance value={maxBalance} />
       <Tooltip
-        text={mapped.map(([{ balance, isFinished, referendumId, vote }, blocks], index): React.ReactNode => (
+        text={sorted.map(({ details, headers }, index): React.ReactNode => (
           <div
             className='row'
```
Tbaut, line 111: I was fighting a little in my solution to get this `row` to have an additional class with `margin-top: 0` to get the grouped element even more compactly grouped, in the case details is `undefined`. But it's a nit.

packages/react-components/src/DemocracyLocks.tsx
jacogr: The idea is to not change the state object when the structure hasn't changed - when this comment was made that check was not in yet.

packages/react-components/src/DemocracyLocks.tsx
jacogr: Details is non-optional in the entry.

Issue #2140, opened 2020-01-17
Unable to copy disabled fields on FF


jacogr:
![image](https://user-images.githubusercontent.com/1424473/72611328-14499e00-392a-11ea-814f-afcdec025791.png)

Lowhearth:
After doing some research I've found out that there is a 16 years old discussion regarding this behaviour:  https://bugzilla.mozilla.org/show_bug.cgi?id=253870

It looks like they are making disabled inputs selectable for the next Firefox version.

We might want to swap disabled input to readonly if you dont mind the text cursor showing off @jacogr

jacogr:
Welcome back! :)

Yes, there are 2 options -

- readonly (which works, we may adjust the cursor, will need to see what it looks like)
- swapping to a styled div

The first is the better approach, imho.

jacogr:
So would just wait for the next release, browser update anyway. So would just forgo doing anything at all.

jacogr:
Still an issue on 75.0 (for me). Will re-test when the next one rolls around.

Tbaut:
Still an issue on 77. I'd actually love to address, it's been bothering me for so long :D !
Alternatively, I'd even propose to add a little "copy to clipboard" button on the right of the field. To do this well I guess it would mean creating a new "CopiableField" react-component and use it for the image hash.

lmk what you think @jacogr

jacogr:
Just turn it into a static field with the correct styling. Don't want to wait another 16 years. Te copy-buttons is a slippery slope, where to add, where not to add, so not something that we really want to do.

PR #3101, opened 2020-06-30
Be able to select preimage hash in Firefox

Tbaut:
closes https://github.com/polkadot-js/apps/issues/2140

The field looks disabled. One can't edit it either, but selection is now possible on FF. 
I checked on Chrome and the behaviour is the same as on FF.

select

### DIFF
--- a/packages/page-democracy/src/Overview/PreImage.tsx
+++ b/packages/page-democracy/src/Overview/PreImage.tsx
@@ -80,7 +80,7 @@ function PreImage ({ className = '', imageHash, isImminent: propsIsImminent, onC
               onChange={setProposal}
             />
             <Input
-              className={'disabledLook'}
+              className='disabledLook'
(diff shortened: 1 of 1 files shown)


packages/page-democracy/src/Overview/PreImage.tsx
```diff
@@ -80,8 +80,8 @@ function PreImage ({ className = '', imageHash, isImminent: propsIsImminent, onC
               onChange={setProposal}
             />
             <Input
+              className={'disabledLook'}
```
jacogr, line 83:
```suggestion
              className='disabledLook'
```

Generally would prefer an actual flag instead of className pushed in - that way it can be checked. However can see how this holds up and then move to base at a later point in time. (Inputs will be replaced as they are implemented, so not worth scratching around in there too much atm)

Tbaut:
I didn't think about adding a proper flag as I thought this was a niche need.

PR #3118, opened 2020-07-03
Tests for electron app storage

vanruch:


packages/apps-electron/src/renderer/remote-electron-store.ts
```diff
@@ -3,32 +3,34 @@
 // of the Apache-2.0 license. See the LICENSE file for details.
 
 import { KeyringJson, KeyringStore } from '@polkadot/ui-keyring/types';
-import { electronMainApi } from '../api/global-exported-api';
+import { AccountStoreApi } from '../api/account-store-api';
 
 export class RemoteElectronStore implements KeyringStore {
+  constructor (private accountStore: AccountStoreApi) {}
+
```
jacogr, line 10:
```suggestion
  readonly #accountStore: AccountStoreApi;
  
  constructor (accountStore: AccountStoreApi) {
    this.#accountStore = accountStore;
  }

```

Would prefer being explicit both in assignment and making use of `#` for private fields.

(Does have an impact below, not indicated, but will just be `this.accountStore` => `this.#accountStore`)

krzysztof-jelski: Changed accordingly.

PR #3121, opened 2020-07-03
export Conviction dropdown

Tbaut:
oops, I had pointed to it directly in my test, so didn't catch this.


jacogr:
No issues. Could have also done this in your PR.

PR #3159, opened 2020-07-10
Open external links in default system browser

krzysztof-jelski:
We have at least 2 links (GitHub, Wiki) that lead to external resources. Electron by default opens them in a new Electron window, which doesn't make much sense. This PR changes this behaviour to open those links in a system browser.


krzysztof-jelski:
Maybe should be with `async - await` as in this [example](https://www.electronjs.org/docs/tutorial/security#13-disable-or-limit-creation-of-new-windows)?

Issue #2887, opened 2020-06-06
Support proxy setting via accounts (proxy module)


jacogr:
Option in accounts drop-down when proxy module is available. Open dialog to set the proxy for the account for a specific type.

jacogr:
- [x] Signer
- [x] Proxied accounts (external)
- [ ] Setup dialog

Tbaut:
I'm willing to take this on, but I'm unsure what the end result should look like. Should we add a menu to `add proxy` from the right ellipsis one, and then simply let users select the type?

Now I understand what you mean with the display headache, showing account icons for each proxy type is def too much info at once. 

One idea would be to use an `Expandable` "X accounts" and once expanded we show all the accounts with their proxy type (with `AddressMini`)
![image](https://user-images.githubusercontent.com/33178835/87157066-f009c100-c2bd-11ea-9eb2-4b91953c9f69.png)

 I just leaned about `proxy.anonymous` extrinsic, I guess this is out of the question for this issue is it?

jacogr:
The anon proxy should leave for a next round.

So the display is a huge issue -

- we have 3 columns, parent, delegation & proxy
- these are mostly empty
- they take up a lot of space with "something" in at least one

For proxy & delegation, we may actually be able to remove the columns completely and just have a badge indicator and then the management modal will show all and allow update. For these, once set it is relatively stable non-changing information, so doesn't need to be seen "all-the-time". (Additionally the Sidebar can actually show that info, not only fo own accounts, but for all)

Tbaut:
Sounds good, what kind of badge do you think about? A round icon for each badge (I guess a tooltip to explain what they are about is necessary)? Should those badges be in a dedicated column? Clicking on the delegation badge would open the managing modal (we'd use a different modal for delegation and proxy I suppose?)
Lastly, not sure how the modal should look like, we don't have anything in place like this yet do we? Should it present the information, read only, and provide buttons that would fire yet another modal on top to change things.

jacogr:
Nested modals are a no-go (even the signer modal on top of others is actually really bad...)

Modal can just show everything and allow you to edit, submit sends it to the chain. so see it as a current-settings edit. SideBar can have a static/quick view.

Badges are already in place for accounts. One global missing thing (everywhere) is a badge legend per page where they are used, will need to get in since tooltips are exceptionally expensive (hence removed from pages like staking) and without it is really nn-descriptive.

Tbaut:
[...]

PR #3202, opened 2020-07-15
Move delegation column to badge

Tbaut:
As discussed in https://github.com/polkadot-js/apps/issues/2887#issuecomment-657584543
Moves all delegation options to 1 modal (undelegate is now an additional negative button)
Slightly modified wording depending if the user is setting up a delegation or editing one.
Kept the change / undelegate menus.

delegation-2


packages/page-accounts/src/Accounts/Account.tsx
```diff
@@ -335,45 +352,20 @@ function Account ({ account: { address, meta }, className = '', delegation, filt
...
-                  <div>{delegation.conviction.toString()}</div>
-                </div>
-              }
-              value={delegation.accountDelegated}
-            />
-          )}
-        </td>
-      )}
-      <td className='number'>
+h      <td className='number'>
```
Tbaut, line 368: woops!
```suggestion
      <td className='number'>
```

packages/page-accounts/src/Accounts/modals/Delegate.tsx
```diff
@@ -31,10 +31,15 @@ function Delegate ({ amount: _amount, conviction: _conviction, delegatedAccount:
   const [delegatedAccount, setDelegatedAccount] = useState<string | null>(_delegatedAccount || null);
   const [conviction, setConviction] = useState(_conviction?.toNumber() || 1);
 
+  const isDirty = amount?.toString() !== _amount?.toString() ||
+    delegatedAccount !== _delegatedAccount ||
+    delegatingAccount !== _delegatingAccount ||
+    conviction !== _conviction?.toNumber();
```
jacogr, line 37: I find this incredibly hard to read for a couple of reasons -

- we don't do the props to `_` casting anywhere else in the codebase
- not sure (just glancing) which is which
- at later points we mix the _/non_ (e.g. in label)

(More of a nitpick.)

Small issue spotted with conviction override as well, when set to "none" it will always reset here to "0". Not introduced here, but introduced in the previous.

packages/page-accounts/src/Accounts/modals/Delegate.tsx
```diff
@@ -31,10 +31,15 @@ function Delegate ({ amount: _amount, conviction: _conviction, delegatedAccount:
...
+
   return (
     <Modal
       className='staking--Delegate'
-      header= {t<string>('Delegate democracy vote')}
+      header= {_delegatedAccount ? t<string>('democracy vote delegation') : t<string>('delegate democracy vote')}
```
jacogr, line 42:
```suggestion
      header={_delegatedAccount ? t<string>('democracy vote delegation') : t<string>('delegate democracy vote')}
```

Generally would prefer the "?:" over multiple lines as well. (Not a crisis, just alignment)

packages/page-accounts/src/Accounts/modals/Delegate.tsx
Tbaut: Agreed, that `_*` naming is not great at all. I'll rename for `previous*`, it should be better.

packages/page-accounts/src/Accounts/modals/Delegate.tsx
Tbaut: Sure, changed this.

Tbaut:
Nice catch for the x0.1 bug, it's fixed.
I changed the names, I find it already much more readable. lmk if you have a better idea in case it's still not good enough.

PR #3215, opened 2020-07-16
Add proxy edition to accounts

Tbaut:
closes https://github.com/polkadot-js/apps/issues/2887

Here is the structure I thought about:
- get 1 array for previous proxies (from the props)
- create 1 array `previousProxyDisplay` for the displayed previous proxies (it changes if the user delete previous proxies)
- create 1 array `batchStackPrevious` with the extrinsics related to the previous proxies (only deletion is possible, so it's basically a list of `proxy.removeProxy`)
- create 1 array `batchStackAdded` with the extrinsics related to the newly added proxies (it needs to be updated when the user change a new proxy account or type)
- create 1 array `addedProxies` with the displayed newly added proxy

We are working heavily with the "index" of those different arrays, the extrinsic array needs to be kept in sync with what is displayed.

Both extrinsics arrays are combined for the final transaction. If the user only perform 1 action, a batch isn't used.

todo:
- [x] remove any `any`
- [x] ~~fix the bug found in https://github.com/polkadot-js/apps/pull/3215#issuecomment-660084504~~ this is on the signer, not this PR. Solved https://github.com/polkadot-js/apps/pull/3220
- [x] make sure the accounts grid is fixed in width (it's changing with the name length right now)

proxy


packages/page-accounts/src/Accounts/modals/ProxyOverview.tsx
```diff
@@ -0,0 +1,296 @@
...
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-return
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+  return api.tx.utility.batch([...batchPrevious, ...batchAdded]);
+}
+
+function ProxyOverview ({ className, onClose, previousProxy, proxiedAccount }: Props): React.ReactElement<Props> {
+  const { t } = useTranslation();
+  const { api } = useApi();
+  // the stack of extrinsics on the previous proxies
+  const [batchStackPrevious, setBatchStackPrevious] = useState<any[]>([]);
```
Tbaut, line 44: todo: get rid of that `any[]` and the corresponding eslint silencing

packages/page-accounts/src/Accounts/modals/ProxyOverview.tsx
```diff
@@ -0,0 +1,296 @@
...
+  // actualized list of new proxies (additions, if any)
+  const [addedProxies, setAddedProxies] = useState<[AccountId, ProxyType][]>([]);
+
+  useEffect(() => {
+    if (batchStackPrevious.length || batchStackAdded.length) {
+      setExtrinsic(createExtrinsic(api, batchStackPrevious, batchStackAdded));
+    }
+  }, [api, batchStackPrevious, batchStackAdded]);
+
+  const typeOpts = [
```
Tbaut, line 62: @jacogr can I get this dynamically from the api rather than declaring them myself?

packages/page-accounts/src/Accounts/modals/ProxyOverview.tsx
```diff
@@ -0,0 +1,296 @@
...
+      grid-column: 1;
+    }
+    .buttons-column{
+      grid-column: 2;
+      padding-top: 0.3rem;
+    }
+  }
+
+  .add-proxy-button{
+    width: 100%;
```
Tbaut, line 294: I tried a new type of button, to make it clear. lmk what you think.

packages/page-accounts/src/Accounts/modals/ProxyOverview.tsx
Tbaut: done

Tbaut:
Found a bug (pretty hard to reproduce) where a new proxy isn't effectively added. I'll track it down.

Tbaut:
I finally figured out what that bug was. It seems related to the signer, together with "Staking" type proxy.
At least I know how to reproduce it, but I don't understand how it happens.

What happens is that the signer, per default, let me sign a batch transaction with a proxy account that does not have the sufficient rights to do. The Tx is marked as successful, but it must fail down the line silently.

#### How to 
- Set a proxy for "Staking" type on any account (e.g Alice set Bob as Staking proxy)
- Send a "bacth" form this first (Alice) account (e.g add 2 more proxies, or remove 1 and add 1.. it needs to be a batch)
--> the signer, per default, is using the proxy account "Bob" for signing, although this proxy is only for Staking.

Here is a demo:
- 2 proxies (Staking) are set.
- I remove both them manually (that's a batch)
- the signer per default let me sign with Bob.. who only had staking proxy --> extrinsic is successful, but it's not effective
- I do the same, but untick the proxy on the signer --> it succeed and is effective.
of a failed, then a success (ticking the checkbox to remove the proxy on the signer)
bug

jacogr:
Sorry, my humble apologies - wanted to get it in by today, but was severely side-tracked elsewhere.

PR #3245, opened 2020-07-21
Check for window top

jacogr:


jacogr:
Will know in around 12-15 minutes once CI id deployed.

PR #3348, opened 2020-08-05
Take unlocking and redeemable into account to bond more funds

### LINKED ISSUE #3232: Use unbonding amount in staking warning calculations
It is possible to handicap yourself with "Bond more funds" simply by submitting an amount that exceeds balance. It's even worse with polkadot, funds will be locked until balance transfers enabled.

Tbaut:
closes #3232 
We now take any unbonding amount as well as any amount that hasn't been redeemed yet.

I've tested it with both unbonding (also several) and redeemable amounts.

### DIFF
--- a/packages/page-staking/src/Actions/Account/BondExtra.tsx
+++ b/packages/page-staking/src/Actions/Account/BondExtra.tsx
@@ -37,7 +37,10 @@ function BondExtra ({ controllerId, onClose, stakingInfo, stashId }: Props): Rea
 
   useEffect((): void => {
     if (stakingInfo && stakingInfo.stakingLedger && stashBalance) {
-      const available = stashBalance.freeBalance.sub(stakingInfo.stakingLedger.active.unwrap());
+      const sumUnlocking: BN = stakingInfo.unlocking?.reduce((acc, { value }) => acc.add(value), BN_ZERO) || BN_ZERO;
+      const redeemable: BN = stakingInfo.redeemable || BN_ZERO;
+
+      const available = stashBalance.freeBalance.sub(stakingInfo.stakingLedger.active.unwrap()).sub(sumUnlocking).sub(redeemable);
 
       setStartBalance(
         available.gt(api.consts.balances.existentialDeposit)
--- a/packages/page-staking/src/Actions/Account/BondExtra.tsx
+++ b/packages/page-staking/src/Actions/Account/BondExtra.tsx
@@ -37,9 +37,8 @@ function BondExtra ({ controllerId, onClose, stakingInfo, stashId }: Props): Rea
 
   useEffect((): void => {
     if (stakingInfo && stakingInfo.stakingLedger && stashBalance) {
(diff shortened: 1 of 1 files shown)


jacogr:
The sub should be .add

image

So on the current master, that allows 0.039 (instead of 0.049). In your version is does less.

packages/page-staking/src/Actions/Account/BondExtra.tsx
```diff
@@ -37,7 +37,9 @@ function BondExtra ({ controllerId, onClose, stakingInfo, stashId }: Props): Rea
 
   useEffect((): void => {
     if (stakingInfo && stakingInfo.stakingLedger && stashBalance) {
-      const available = stashBalance.freeBalance.sub(stakingInfo.stakingLedger.active.unwrap());
+      const sumUnlocking = stakingInfo.unlocking?.reduce((acc, { value }) => acc.add(value), BN_ZERO) || BN_ZERO;
+      const redeemable = stakingInfo.redeemable || BN_ZERO;
+      const available = stashBalance.freeBalance.sub(stakingInfo.stakingLedger.active.unwrap()).sub(sumUnlocking).sub(redeemable);
```
jacogr, line 42:
```suggestion
      const available = stashBalance.freeBalance.sub(stakingInfo.stakingLedger.active.unwrap()).add(sumUnlocking);
```

jacogr:
Sorry, not 100% sure about my add/sub, it was late. (The redeemable I'm sure that it should not be there though, based on the screenshot above) Will double-check now against where I have unlocking.

Tbaut:
Thanks, this got me confused yesterday, but it was indeed too late for me to sanely check. I'll push my test further for the redeemable, because from what I could see, it was needed because the calculation would have been wrong as soon as the unbonding period was hit. Will post the results of my test shortly.

jacogr:
The above screenshot is a good test.

1. The account had 0.05 to start
2. It bonder 0.01
3. In unbonded 0.01, that not is redeemable
4. So what it can currently bond is 0.04 (ignoring the 0.001 fees that were taken, so 0.039 exact)

However we show here only 0.029 can be bonded, should be 0.039. (0.049 when we have redeemed)

Tbaut:
Just did a test, you can not bond any fund that has not been redeemed yet, so we're good on that.

> However we show here only 0.029 can be bonded, should be 0.039. (0.049 when we have redeemed)

Nope, with the current code, we're actually showing `0.039`, because the `sumUnlocking` becomes 0 as soon as the `redeemable` become `0.01`, this is the way the derive are designed at least (and the reason I introduce the `redeemable` into the equation).

Here is a video about the flow: https://youtu.be/rI1UySyXEOo

jacogr:
Ok, the small numbers confused me. Yes, it is correct, the extra 0.01 I wqs looking for is actually the existential amount. So it won't allow you to go under that. (So the missing piece)

Issue #3157, opened 2020-07-10
Add forums links for tips on the right side, next to "endorse" button


rrtti:
Add links for tips on the right side next to "endorse" button, in the same way we can see on any proposal. Polkassembly now included on-chain tips descriptions and discussions, allowing us to add the link to the tip in case more info is needed. We can add this link + Polkascan and Subscan. If Commonwealth includes this functionality we should add as well.

Tbaut:
Thanks Raul, it def makes sense. We should note that tips have no id, but a hash. I think Polkassembly needs to first be able to request tips by hash before we can have a link. I'll log this.

Tbaut:
Should be good now, I'll submit a PR soon.

Tbaut:
totally forgot about this one sorry, will do it in the next days.

PR #3355, opened 2020-08-06
Add tip link to Polkassembly

Tbaut:
Closes https://github.com/polkadot-js/apps/issues/3157
Looks like this:
image

### DIFF
--- a/packages/page-treasury/src/Overview/Tip.tsx
+++ b/packages/page-treasury/src/Overview/Tip.tsx
@@ -7,7 +7,7 @@ import { AccountId, Balance, BlockNumber, OpenTip, OpenTipTo225 } from '@polkado
 import BN from 'bn.js';
 import React, { useEffect, useState } from 'react';
 import styled from 'styled-components';
-import { AddressSmall, AddressMini, Expander, Icon, TxButton } from '@polkadot/react-components';
+import { AddressSmall, AddressMini, Expander, Icon, LinkExternal, TxButton } from '@polkadot/react-components';
 import { useAccounts } from '@polkadot/react-hooks';
 import { BlockToTime, FormatBalance } from '@polkadot/react-query';
 import { BN_ZERO, formatBalance, formatNumber } from '@polkadot/util';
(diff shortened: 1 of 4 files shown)


jacogr:
Exactly what I was actually complaining about yesterday (in private chats), that take is way too busy and now starts breaking lines as well. Not your issue, but basically -

- those links are way too obtuse and basically almost unusable
- all these tables are too busy
- ... not sure where to put them though

So complaining, been driving me crazy for a while, no solutions...

packages/page-treasury/src/Overview/Tip.tsx
```diff
@@ -157,6 +157,13 @@ function Tip ({ bestNumber, className = '', hash, isMember, members, tip }: Prop
           />
         )}
       </td>
+      <td className='mini ui--media-1000'>
```
jacogr, line 160: We need this is the header as well. (Should also adjust the classes, maybe 1200 or 1400 until we can sort out the busy nature)

Tbaut: ah right totally missed it.

jacogr: Don't worry, I miss that a lot since they stuff is in 2 places - so don't feel all alone.

Tbaut:
> not sure where to put them though

A couple idea from the top of my head:
I could see these live in a "..." menu, it's 1 more click but would allow to show the full name of the links. Not sure how much they are used.

In terms of user friendliness, the platform's logo could also help keep these short, yet give an idea right away where they go.

jacogr:
Agreed, the logos are actually (not a solution) but would make is much cleaner and much less obtuse. I actually use these a lot, not sure about others...

But the table issue is not confined to here, referendums are equally bad. As is accounts. And that is what makes it tricky since it needs to be a consistent solution. Just a lot of info to convey.

Tbaut:
Just realized now what those `media` classes are. From what I could see, the table doesn't shrink past 150x, so the links wouldn't even be visible, I went for 1600 then.

jacogr:
The media here is rough - it helps, but table are... problematic with the large amount of data deemed „useful“. For instance on tips the deposit field is actually useless with no real value. Can be removed.

Issue #3360, opened 2020-08-07
Council voting displays the wrong maximum amount


Tbaut:
as per https://github.com/polkadot-js/apps/issues/2618#issuecomment-670416618

PR #3363, opened 2020-08-07
Fix council voting balance

Tbaut:
closes https://github.com/polkadot-js/apps/issues/3360

Tested both democracy and council voting with a treasury deposit (reserved).

Example:
Account has 1000
bond 500 in a treasury
--> can vote 500 for democracy
--> can vote 1000 for council

### DIFF
--- a/packages/react-components/src/VoteValue.tsx
+++ b/packages/react-components/src/VoteValue.tsx
@@ -40,7 +40,7 @@ function VoteValue ({ accountId, autoFocus, isCouncil, onChange }: Props): React
     // if the set accountId changes and the new balances is for that id, set it
     (accountId !== selectedId) && allBalances && allBalances.accountId.eq(accountId) && setValue({
       selectedId: accountId,
-      value: allBalances?.lockedBalance
+      value: allBalances.lockedBalance
     });
   }, [allBalances, accountId, maxVotingBalance, selectedId]);
 
--- a/packages/react-components/src/VoteValue.tsx
+++ b/packages/react-components/src/VoteValue.tsx
@@ -5,7 +5,7 @@
 import { DeriveBalancesAll } from '@polkadot/api-derive/types';
 
 import BN from 'bn.js';
-import React, { useCallback, useEffect, useState } from 'react';
+import React, { useCallback, useEffect, useState, useMemo } from 'react';
 import { useApi, useCall } from '@polkadot/react-hooks';
 import { BalanceVoting } from '@polkadot/react-query';
 import { isBn } from '@polkadot/util';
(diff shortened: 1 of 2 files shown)


packages/react-components/src/VoteValue.tsx
```diff
@@ -30,14 +30,19 @@ function VoteValue ({ accountId, autoFocus, isCouncil, onChange }: Props): React
...
+      ? allBalances?.votingBalance.add(allBalances?.reservedBalance)
+      : allBalances?.votingBalance
+  , [allBalances?.reservedBalance, allBalances?.votingBalance, isCouncil]);
 
   useEffect((): void => {
     // if the set accountId changes and the new balances is for that id, set it
     (accountId !== selectedId) && allBalances && allBalances.accountId.eq(accountId) && setValue({
       selectedId: accountId,
-      value: allBalances.lockedBalance
+      value: maxVotingBalance
```
jacogr, line 43: Would prefer to keep it as the current locked. This just means that because of the overlaps he/she doesn't inadvertently lock more.

packages/react-components/src/VoteValue.tsx
```diff
@@ -54,7 +59,10 @@ function VoteValue ({ accountId, autoFocus, isCouncil, onChange }: Props): React
   return (
     <InputBalance
       autoFocus={autoFocus}
-      defaultValue={accountId !== selectedId ? undefined : allBalances?.lockedBalance}
+      defaultValue={accountId !== selectedId
```
jacogr, line 62: Same as above, prefer being a bit less aggressive. (Another thing is that we don't have an easy way to unlock these atm, missing features, so it is problematic)

jacogr:
For future reference, he is the locking on votes -

https://github.com/paritytech/substrate/blob/833fe6259115625f61347c8413bab29fded31210/frame/elections-phragmen/src/lib.rs#L388-L397

```rust
			// Amount to be locked up.
			let locked_balance = value.min(T::Currency::total_balance(&who));

			// lock
			T::Currency::set_lock(
				T::ModuleId::get(),
				&who,
				locked_balance,
				WithdrawReasons::except(WithdrawReason::TransactionPayment),
			);

```

packages/react-components/src/VoteValue.tsx
```diff
@@ -30,14 +30,19 @@ function VoteValue ({ accountId, autoFocus, isCouncil, onChange }: Props): React
...
+      ? allBalances?.votingBalance.add(allBalances?.reservedBalance)
+      : allBalances?.votingBalance
+  , [allBalances?.reservedBalance, allBalances?.votingBalance, isCouncil]);
 
   useEffect((): void => {
     // if the set accountId changes and the new balances is for that id, set it
     (accountId !== selectedId) && allBalances && allBalances.accountId.eq(accountId) && setValue({
       selectedId: accountId,
-      value: allBalances.lockedBalance
+      value: allBalances?.lockedBalance
```
jacogr, line 43:
```suggestion
      value: allBalances.lockedBalance
```

Checked for existence already.

Tbaut:
Ah, I understand now, took me a while. We want to show whatever used to be locked (0 or the previous amount) rather than the maximum. Totally makes sense, I've reverted.

Issue #3003, opened 2020-06-20
Add a batch msg to the last vote on council motions


rrtti:
Add a batch msg to the last vote on council motions for the last voter to close the proposal and pay for it.

Tbaut:
Sounds super valuable. Just a comment not directly related to this issue, I think the same could be applied to tips isn't it?
Anyway I'll take a look at this one.

jacogr:
Thought about tips, but not really possible.

So in tips once there is enough it enters a countdown phase, after that only it can be closed.

Tbaut:
> after that only it can be closed.

Ah right, I wasn't entirely sure about that, you're right, there's this `end` state field that gets populated when the tip status moves to "closing", and until this block is reached nothing can be done.

PR #3368, opened 2020-08-07
Batch vote and close for the last needed vote

Tbaut:
closes https://github.com/polkadot-js/apps/issues/3003
- checks if the vote will make the number of aye or nay reach the vote treshold.
- if the current vote allows the proposal to be closes it shows a warning and the vote button mentions "Vote XYZ and Close"

I was able to test with the council, but I'd require help for the techCommittee, not sure what needs to be changed in my (already modified) node to enable it. Right now I can't submit candidacy. Running 0.8.22 with `--chain kusama-dev pruning archive --alice`


packages/page-council/src/Motions/Motion.tsx
```diff
@@ -81,14 +81,15 @@ function Motion ({ className = '', isMember, members, motion: { hash, proposal,
         votes={nays}
       />
       <td className='button'>
-        {isVoteable && !isCloseable && (
```
Tbaut, line 84: I figured out this `isClasable` wasn't needed because it's mutually exclusive with `isVotable` afaiu.

packages/react-components/src/VoteActions.tsx
```diff
@@ -28,11 +32,17 @@ function VoteActions ({ accountId, aye, className = '', isDisabled, onClick, par
...
       <TxButton
         accountId={accountId}
+        extrinsic={extrinsic}
         icon='check'
         isDisabled={!accountId || isDisabled}
         label={aye
-          ? t<string>('Vote Aye')
-          : t<string>('Vote Nay')}
+          ? isClosing
+            ? t<string>('Vote Aye and Close')
```
Tbaut, line 40: Not the nicest button but I thought it's always better to explicitly tell users what this batch is supposed to do.

Tbaut:
Figured there's a flaw when the number of council members is even. It looks like the threashold is member.length/2+1.
Fixing that.

packages/page-council/src/Motions/Voting.tsx
```diff
@@ -42,6 +49,16 @@ function Voting ({ hash, idNumber, isDisabled, members, prime, proposal }: Props
 
   const isPrime = prime?.toString() === accountId;
 
+  const voteExtrinsic = api.tx.council.vote(hash, idNumber, voteValue);
+  const closeExtrinsic = api.tx.council.close.meta.args.length === 4
+    ? api.tx.council.close(hash, idNumber, proposalWeight, proposalLength)
+    : api.tx.council.close(hash, idNumber, 0, 0);
```
jacogr, line 55: This can't be right, see https://github.com/polkadot-js/apps/blob/master/packages/page-council/src/Motions/Close.tsx#L73-L77

So, 

- if args === 4 and passing, it needs the weights (since it will apply)
- if args === 4 and failing, the weights are 0 (it doesn't apply and we save fees)
- if args !== 4, don't pass the last 2 (here you pass them as 0, 0, so it breaks on older versions)

Tbaut: Ah thanks. I actually had something more correct but super hacky (wouldn't pass linting) concerning the last point
>     if args !== 4, don't pass the last 2 (here you pass them as 0, 0, so it breaks on older versions)

I was doing something like `api.tx.council.(close as any)(hash, idNumber);`.
What would you recommend to support older versions?

jacogr: Would just do the call and override with `// @ts-ignore` for the older one. Example here - https://github.com/polkadot-js/apps/pull/3358/files#diff-e9dca1c535128a81cb2067002a8d88c5R62-R66

It is messy, but don't have a better (non-any) solution. It is valid in _some_ older cases.

packages/page-council/src/Motions/Voting.tsx
```diff
@@ -56,7 +56,9 @@ function Voting ({ hash, idNumber, isDisabled, members, prime, proposal, votes }
   const voteExtrinsic = api.tx.council.vote(hash, idNumber, voteValue);
   const closeExtrinsic = api.tx.council.close.meta.args.length === 4
     ? api.tx.council.close(hash, idNumber, proposalWeight, proposalLength)
-    : api.tx.council.close(hash, idNumber, 0, 0);
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore new version
```
jacogr, line 60:
```suggestion
    // @ts-ignore older version (2 params)
```

packages/page-tech-comm/src/Proposals/Voting.tsx
```diff
@@ -51,7 +51,9 @@ function Voting ({ hash, prime, proposal, proposalId, votes }: Props): React.Rea
   const voteExtrinsic = api.tx.technicalCommittee.vote(hash, proposalId, voteValue);
   const closeExtrinsic = api.tx.council.close.meta.args.length === 4
     ? api.tx.technicalCommittee.close(hash, proposalId, proposalWeight, proposalLength)
-    : api.tx.technicalCommittee.close(hash, proposalId, 0, 0);
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore new version
```
jacogr, line 55:
```suggestion
    // @ts-ignore older version (2 params)
```

jacogr:
I think the logic seems sound. I just want to do a live test/play, should be in a day or two.

packages/page-council/src/Motions/Voting.tsx
```diff
@@ -42,6 +53,18 @@ function Voting ({ hash, idNumber, isDisabled, members, prime, proposal }: Props
...
+  const closeExtrinsic = api.tx.council.close.meta.args.length === 4
+    ? api.tx.council.close(hash, idNumber, proposalWeight, proposalLength)
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore older version (2 params)
+    : api.tx.council.close(hash, idNumber);
+
+  // vote and close if this vote ends the vote
+  const extrinsic = willPass || willFail
+    ? api.tx.utility.batch([voteExtrinsic, closeExtrinsic])
+    : voteExtrinsic;
```
jacogr, line 66: We don't want to create these each render - this is quite expensive.

packages/page-tech-comm/src/Proposals/Voting.tsx
```diff
@@ -40,6 +48,18 @@ function Voting ({ hash, prime, proposalId }: Props): React.ReactElement<Props>
...
+  const closeExtrinsic = api.tx.council.close.meta.args.length === 4
+    ? api.tx.technicalCommittee.close(hash, proposalId, proposalWeight, proposalLength)
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore older version (2 params)
+    : api.tx.technicalCommittee.close(hash, proposalId);
+
+  // vote and close if this vote ends the vote
+  const extrinsic = willPass || willFail
+    ? api.tx.utility.batch([voteExtrinsic, closeExtrinsic])
+    : voteExtrinsic;
```
jacogr, line 61: Same comment as in Council.

packages/react-components/src/VoteActions.tsx
jacogr, line 40:
```suggestion
            ? t<string>('Vote Aye/Close')
```

packages/react-components/src/VoteActions.tsx
```diff
@@ -28,11 +32,17 @@ function VoteActions ({ accountId, aye, className = '', isDisabled, onClick, par
...
+            : t<string>('Vote Aye')
+          : isClosing
+            ? t<string>('Vote Nay and Close')
```
jacogr, line 43:
```suggestion
            ? t<string>('Vote Nay/Close')
```

packages/page-council/src/Motions/Voting.tsx
Tbaut: Right, I'll use `useMemo`

Tbaut:
I've an error that I've never have in the past testing this one when doing the batch:
> DRR: Error: 1010: Invalid Transaction: Transaction would exhausts the block limits

I'll test further

Tbaut:
The "Transaction would exhausts the block limits" happened only when the motion (not external motion) submitted  was a `system.setCode`. I don't understand why, but since it was also the first time I've tested with this extrinsic, I imagine this is expected?

Other than that I've fixed the above mentionned glitch by adding a fail safe to prevent an account to needlessly vote twice.

vote-close

packages/page-council/src/Motions/Voting.tsx
```diff
@@ -88,13 +118,32 @@ function Voting ({ hash, idNumber, isDisabled, members, prime, proposal }: Props
                 <div>{t<string>('You are voting with this collective\'s prime account. The vote will be the default outcome in case of any abstentions.')}</div>
               </article>
             )}
+            {willFail && isNewVote && (
+              <article className='warning'>
```
jacogr, line 122:
```suggestion
              <article className='full warning'>
```

packages/page-council/src/Motions/Voting.tsx
```diff
@@ -88,13 +118,32 @@ function Voting ({ hash, idNumber, isDisabled, members, prime, proposal }: Props
...
+                <div>{t<string>('Your "Nay" vote will end this proposal. This will close it.')}</div>
+              </article>
+            )}
+            {willPass && isNewVote && (
+              <article className='warning'>
```
jacogr, line 127:
```suggestion
              <article className='full warning'>
```

packages/page-council/src/Motions/Voting.tsx
```diff
@@ -88,13 +118,32 @@ function Voting ({ hash, idNumber, isDisabled, members, prime, proposal }: Props
...
+                <div>{t<string>('Your "Aye" vote will end this proposal. This will close it.')}</div>
+              </article>
+            )}
+            {!isNewVote && (
+              <article className='warning'>
```
jacogr, line 132:
```suggestion
              <article className='full warning'>
```

(Just stretches it to the left edge)

packages/page-council/src/Motions/Voting.tsx
```diff
@@ -61,9 +57,15 @@ function Voting ({ hash, idNumber, isDisabled, members, prime, proposal, votes }
     : api.tx.council.close(hash, idNumber);
 
   // vote and close if this vote ends the vote
-  const extrinsic = willPass || willFail
-    ? api.tx.utility.batch([voteExtrinsic, closeExtrinsic])
-    : voteExtrinsic;
+  const extrinsic = useMemo(() =>
```
jacogr, line 65: This is great, we still have the vote stuff above doing the same - so this will execute each render anyway since those objects will change. You probably want to encapsulate the close/vote extrinsics inside this memo. (Not needed externally)

Tbaut: Thanks, totally makes sense.

packages/page-tech-comm/src/Proposals/Voting.tsx
```diff
@@ -34,12 +48,27 @@ function Voting ({ hash, prime, proposalId }: Props): React.ReactElement<Props>
...
+      ? api.tx.technicalCommittee.close(hash, proposalId, proposalWeight, proposalLength)
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore older version (2 params)
+      : api.tx.technicalCommittee.close(hash, proposalId);
+
+    return willPass || willFail
+      ? api.tx.utility.batch([voteExtrinsic, closeExtrinsic])
+      : voteExtrinsic;
+  }
+  , [api.tx.council.close.meta.args.length, api.tx.technicalCommittee, api.tx.utility, hash, proposalId, proposalLength, proposalWeight, voteValue, willFail, willPass]);
```
jacogr, line 66:
```suggestion
  },
  [api, hash, proposalId, proposalLength, proposalWeight, voteValue, willFail, willPass]);
```

packages/page-council/src/Motions/Voting.tsx
```diff
@@ -36,12 +52,27 @@ function Voting ({ hash, idNumber, isDisabled, members, prime, proposal }: Props
...
+      ? api.tx.council.close(hash, idNumber, proposalWeight, proposalLength)
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore older version (2 params)
+      : api.tx.council.close(hash, idNumber);
+
+    return willPass || willFail
+      ? api.tx.utility.batch([voteExtrinsic, closeExtrinsic])
+      : voteExtrinsic;
+  }
+  , [api.tx.council, api.tx.utility, hash, idNumber, proposalLength, proposalWeight, voteValue, willFail, willPass]);
```
jacogr, line 70:
```suggestion
  },
  [api, hash, idNumber, proposalLength, proposalWeight, voteValue, willFail, willPass]);
```

packages/page-tech-comm/src/Proposals/Voting.tsx
```diff
@@ -34,12 +48,27 @@ function Voting ({ hash, prime, proposalId }: Props): React.ReactElement<Props>
     []
   );
 
+  const isPrime = accountId === prime?.toString();
+
+  // vote and close if this vote ends the vote
+  const extrinsic = useMemo(() => {
+    const voteExtrinsic = api.tx.technicalCommittee.vote(hash, proposalId, voteValue);
+    const closeExtrinsic = api.tx.council.close.meta.args.length === 4
```
jacogr, line 56:
```suggestion
    const closeExtrinsic = api.tx.technicalCommittee.close.meta.args.length === 4
```

Tbaut:
There should be a CI pop-up before committing saying "DID YOU CHECK EDGEWARE?" :dancers: 
Joke aside, thanks, and sorry to have yet again forgotten about them.

Tbaut:
Added an if to prevent things from breaking. I don't think the warning about the last vote needs to be explicitly removed since there's no mention of a batch or anything, it just says:
> Your "XYZ" vote will end this proposal. This will close it.
which remains true.

jacogr:
They are in the process of upgrading and once substrate 2.0 is out, need to have a long-hard think about how long all these code hacks should still live. (Not 6 days, but also not 6 months)

Tbaut:
hmm I didn't realize this one hadn't been merged. I had fixed the comments before my time off and I think it was in a good shape. Will see if I can make sense of these conflicts.

jacogr:
I played with it on various proposals and it seems ok in the normal form. The reason why it is not in is that there are all kinds of edge-cases here with ordering, votes for aye/nay and potential closes and have not gone through all of them.

Tbaut:
My VSCode doesn't apply linting on save any more. Is that a new rule to make things a little snappier?

jacogr:
No, something is very weird atm with the current eslint version - I'm hoping this will fix it - https://github.com/polkadot-js/dev/pull/382 (somewhere with typescript-eslint & eslint and an update it works and then doesn't work, which is exceptionally frustrating)

Tbaut:
Any comment on this one? I don't mind closing it along with the issue if you're not ok with the feature. It's been sitting here for quite a while now.

jacogr:
Not sure as of now -

- have seen cases where you don’t want to close
- have had complaints about clicking twice 
- have had requests for fine-grained control 

So it is one of those damned-if-you-do and damned-if-you-don’t features. 

Apart from that my initial delay was just my own testing.

jacogr:
Cleanup closing. Honestly still up-in-the air on this. 

However, now with the new .28 runtime doing this is actually more expensive ... for the close operation, it is free, when embedded in a batch, you actually pay the weight for the batch call. (One vote per motion is also free)

So extra click, however have not heard too much complaints... and cannot beat free.

(Have however made a note of this one in my revisit list...)

PR #3371, opened 2020-08-07
A proof of concept for frontend test with substrate in docker 

krzysztof-jelski:
Looking for a way to start writing UI tests.

Right now the UI components are tightly coupled to the basic modules (api, crypto, keyring). So I figured it's easier to start with more integration/e2e kind of tests.

I'm open to discussion how much do you like this approach.

Also doing a PR to see if it would run on CI 🙂 .


jacogr:
Have a couple of comments -

- If this is done, I would suggest nightly as opposed to each merge/PR. The overhead associated is quite large on running this type of thing.
- We are shooting for a moving target, let me explain...

We actually had e2e tests in docker in the API and I completely removed all of it since the overhead in maintenance became a full-time job. Substrate changes rapidly so the calls, the params thereof and the types actually returned are non-static. This actually adds a massive overhead in dev as well (since old, new and in-between are supported), see for instance - https://github.com/polkadot-js/apps/blob/master/packages/page-council/src/Motions/Slashing.tsx#L131-L133

In other cases it needs graceful handling of non-existing features, such as https://github.com/polkadot-js/apps/blob/master/packages/page-accounts/src/Accounts/Account.tsx#L83 and https://github.com/polkadot-js/apps/blob/master/packages/page-explorer/src/Summary.tsx#L30-L47

So trying to maintain the tests against this backdrop is a very real challenge since there is no single version and the support is actually against a range of chains each with nuances. Substrate also needs upgrades, for instance a lot of time changes are in master which breaks the UI completely, just due to lag between code there and code here. (The release version of Substrate, e.g. rc5 lags too much to be useful and current master at any point may be ok, or may not be - it generally is, the devs are relatively good in alerting on breaking changes, but Rust updates still break calls completely at points)

TL;DR The idea is good, the maintenance of any tests that has to rely on a node is actually overwhelming and would end up being a full-time job. (Like in the API, they would just completely rot otherwise and have no use at all)

krzysztof-jelski:
Maybe I'm missing something, but I don't see a big maintenance problem here. Let me explain.

I assume the following:
1. There has to be a version of substrate that our UI is compatible with.
2. There exists a docker image of that substrate version in the `parity/substrate` docker hub.

(1.) seems like a must :-) (2.) from what you're saying can be false. In this case, we still have the following options, each adding a minimal maintenance overhead:
  * Run the tests nightly against substrate `latest` and regard the result only as a feedback as to whether we lag behind with the UI or not. So the tests would be broken for max a couple of days until we get the UI in line with substrate changes.
  * Build our own substrate image from a fixed commit and publish it to another dockerhub repo. We'd need to update the image once in a while, but that should be a matter of running a script with a commit hash as an argument.

My point is: from the point of view of the tests we should not care about API specifics, as the tests simply mimic use behaviours in the UI. So if the UI talks to the right API the right way, tests pass.

An alternative to using docker container in tests that I have considered is stubbing the API. This could be done (A) at the REST calls level (with a tool like `nock`) or (B) we could mock the `@polkadot/api/ApiBase` class.

Yet another option (C) is to add another "Service" layer between the React components and the Api (it's the [ultimate way of solving any problem](https://en.wikipedia.org/wiki/Fundamental_theorem_of_software_engineering), obviously 😉 ). For example, we could have `AccountService`, `StakingService`, etc., that would raise the level of abstraction and hide details of interacting with the Api. Then we would stub those services in tests. Stable UI tests, no dependence on Api nor substrate.

My preference here would be option (C). That would require an upfront work of extracting a service before writing a test. I'd say that overhead is acceptable. I really wouldn't like to do (A) or (B), as I believe those options tighten us more closely with substrate and would lead to the maintenance problems you described.

jacogr:
So really not against the idea, since yes, there is a version that it is compatible with, 100%. My point it it doesn't cover everything, sadly no the actual real-bug that pop up.

Let's take examples like polling or parachains, it is not available on Substrate, only on Polkadot. So we cannot test these in the Substrate image. In the case of Substrate, you have either Babe or Aura, these have different behaviours in staking, only one is switched on. so generally we actually break Babe since it is the non-default.

So the places where we break is not due to incompatibilities with Substrate, it is the places where we always assume compatibility with the base Substrate. For a very real example, not all chains have the proxies, so when we put in proxy handling it worked great in Substrate, Polkadot/Kusama, but broke on chains such as Kulupu & Edgeware. (Simple `api.query.proxy` vs `api.query.proxy?`, but it ended up that accounts were unusable on those chains)

Effectively where issues comes from in these that get missed in manual testing, it is never the "happy path" but rather the assumptions that break since the support is vastly different. So the tricky bits are -

- does the new version of this interface and the old version of this still work?
- it this is not available, does all hell break loose?
- are we handling and displaying the different responses correctly

The above is where the time in testing and the bugs logged actually come from.

krzysztof-jelski:
I guess I wanted to have a test fixture quickly, which would allow me to add any UI test, and also test-drive the changes where needed. I was not yet aiming for a fixture (or fixtures) that would prevent errors from popping out.

I think adding any kind of tests will help in achieving the latter goal in the future (eg. we have more tests to run against different networks).

Also in the end we will most likely want to have tests with different fixture sizes (with container, with mocked api, with no react).

Anyway, let's try to decide on a way to move forward with the tests. I see the following options:
1) leave this PR as-is, just configure CI to run container tests nightly
2) close this PR and create another one with tests without container, but with mocked api

jacogr:
As of now, think the container is the best idea - the API stuns will require a lot more effort (actually proper API stubbing has been on the cards for a while now, but time...)

jest/globalSetup.ts
```diff
@@ -0,0 +1,26 @@
+// Copyright 2017-2020 @polkadot/app-accounts authors & contributors
+// This software may be modified and distributed under the terms
+// of the Apache-2.0 license. See the LICENSE file for details.
+import { GenericContainer, Wait } from 'testcontainers';
```
jacogr, line 4:
```suggestion
// of the Apache-2.0 license. See the LICENSE file for details.

import { GenericContainer, Wait } from 'testcontainers';
```

jest/globalSetup.ts
```diff
@@ -0,0 +1,26 @@
...
+import { SubstrateTestsGlobal } from './substrateTestsGlobal';
+declare const global: SubstrateTestsGlobal;
```
jacogr, line 6:
```suggestion
import { SubstrateTestsGlobal } from './substrateTestsGlobal';

declare const global: SubstrateTestsGlobal;
```

jest/globalTeardown.ts
```diff
@@ -0,0 +1,13 @@
+// Copyright 2017-2020 @polkadot/app-accounts authors & contributors
+// This software may be modified and distributed under the terms
+// of the Apache-2.0 license. See the LICENSE file for details.
+import { SubstrateTestsGlobal } from './substrateTestsGlobal';
+declare const global: SubstrateTestsGlobal;
```
jacogr, line 5:
```suggestion
// of the Apache-2.0 license. See the LICENSE file for details.

import { SubstrateTestsGlobal } from './substrateTestsGlobal';

declare const global: SubstrateTestsGlobal;
```

jest/jest.setup.ts
```diff
@@ -0,0 +1,4 @@
+// Copyright 2017-2020 @polkadot/app-accounts authors & contributors
+// This software may be modified and distributed under the terms
+// of the Apache-2.0 license. See the LICENSE file for details.
+jest.setTimeout(25000);
```
jacogr, line 4:
```suggestion
// of the Apache-2.0 license. See the LICENSE file for details.

jest.setTimeout(25000);
```

jest/substrateTestsGlobal.ts
```diff
@@ -0,0 +1,9 @@
+// Copyright 2017-2020 @polkadot/app-accounts authors & contributors
+// This software may be modified and distributed under the terms
+// of the Apache-2.0 license. See the LICENSE file for details.
+import { StartedTestContainer } from 'testcontainers';
```
jacogr, line 4:
```suggestion
// of the Apache-2.0 license. See the LICENSE file for details.

import { StartedTestContainer } from 'testcontainers';
```

packages/page-accounts/src/CreateAccount.spec.tsx
```diff
@@ -0,0 +1,62 @@
+// Copyright 2017-2020 @polkadot/app-accounts authors & contributors
+// This software may be modified and distributed under the terms
+// of the Apache-2.0 license. See the LICENSE file for details.
+import AccountsApp from '@polkadot/app-accounts';
```
jacogr, line 4:
```suggestion
// of the Apache-2.0 license. See the LICENSE file for details.

import AccountsApp from '@polkadot/app-accounts';
```

packages/page-accounts/src/test-support/MemoryStore.ts
```diff
@@ -0,0 +1,34 @@
+// Copyright 2017-2020 @polkadot/app-accounts authors & contributors
+// This software may be modified and distributed under the terms
+// of the Apache-2.0 license. See the LICENSE file for details.
+import { KeyringJson, KeyringStore } from '@polkadot/ui-keyring/types';
```
jacogr, line 4:
```suggestion
// of the Apache-2.0 license. See the LICENSE file for details.

import { KeyringJson, KeyringStore } from '@polkadot/ui-keyring/types';
```

packages/page-accounts/src/test-support/MemoryStore.ts
```diff
@@ -0,0 +1,34 @@
+// Copyright 2017-2020 @polkadot/app-accounts authors & contributors
```
jacogr, line 1: This file/folder probably needs to sit in the root. (It will be the same for most)

package.json
```diff
@@ -71,7 +71,7 @@
     "lint:css": "stylelint './packages/**/src/**/*.tsx'",
     "postinstall": "polkadot-dev-yarn-only",
     "postinstall:electron": "electron-builder install-app-deps",
-    "test": "polkadot-dev-run-test packages/page-claims/src packages/apps-electron",
+    "test": "polkadot-dev-run-test packages/page-claims/src packages/apps-electron packages/page-accounts",
```
jacogr, line 74: This is the only reason why not merging as-is since this will do these tests on all runs. Would suggest moving this into a seperate test call (which we can then do with a specific GH workflow).

packages/page-accounts/src/CreateAccount.spec.tsx
```diff
@@ -19,7 +19,7 @@ const WaitForApi = ({ children }: { children: React.ReactNode }): PropsWithChild
   return api.isApiReady ? (children) : null;
 };
 
-describe('Account Create', () => {
+describe('--SLOW--: Account Create', () => {
```
jacogr, line 22: Also don’t have a better idea just yet, but not sure this is a scalable solution. However, more than good enough to get it running and see it working.

krzysztof-jelski: Some alternative might be to keep the slow/docker tests in separate folder. Another would be to have a distinct extension (eg.`.slow.spec.ts`). However, I prefer the option to specify type of test in test name. Seems most natural to me.

packages/page-accounts/src/test-support/MemoryStore.ts
krzysztof-jelski: That's right. However, not yet sure in which package to put common testing stuff. As it's used only in `page-accounts`, I'd leave it here for now.

jacogr (approved):

Issue #3275, opened 2020-07-27
Council remove vote


xlc:
Should have a button to call `remove_voter` to allow user to unlock funds and unreserve bonds.

PR #3382, opened 2020-08-10
Add a "clear all" button to council vote

Tbaut:
closes https://github.com/polkadot-js/apps/issues/3275

### DIFF
--- a/packages/page-council/src/Overview/Vote.tsx
+++ b/packages/page-council/src/Overview/Vote.tsx
@@ -109,6 +109,18 @@ function Vote ({ electionsInfo }: Props): React.ReactElement<Props> {
             </Modal.Columns>
           </Modal.Content>
           <Modal.Actions onCancel={toggleVisible}>
+            {defaultVotes.length > 0 && <TxButton
+              accountId={accountId}
+              icon='trash-alt'
+              label={t<string>('Clear all')}
+              onStart={toggleVisible}
+              params={[]}
+              tx={
+                api.tx.electionsPhragmen
+                  ? 'electionsPhragmen.removeVoter'
+                  : 'elections.removeVoter'
+              }
+            />}
             <TxButton
               accountId={accountId}
               isDisabled={!accountId || votes.length === 0 || voteValue.lten(0)}
--- a/packages/page-council/src/Overview/Vote.tsx
+++ b/packages/page-council/src/Overview/Vote.tsx
@@ -109,20 +109,19 @@ function Vote ({ electionsInfo }: Props): React.ReactElement<Props> {
             </Modal.Columns>
           </Modal.Content>
           <Modal.Actions onCancel={toggleVisible}>
-            {defaultVotes.length > 0 && (
-              <TxButton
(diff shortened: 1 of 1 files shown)


packages/page-council/src/Overview/Vote.tsx
```diff
@@ -109,6 +109,18 @@ function Vote ({ electionsInfo }: Props): React.ReactElement<Props> {
...
+              icon='trash-alt'
+              label={t<string>('Clear all')}
+              onStart={toggleVisible}
+              params={[]}
+              tx={
+                api.tx.electionsPhragmen
+                  ? 'electionsPhragmen.removeVoter'
+                  : 'elections.removeVoter'
+              }
+            />}
```
jacogr:
```suggestion
            {defaultVotes.length > 0 && (
              <TxButton
                accountId={accountId}
                icon='trash-alt'
                label={t<string>('Clear all')}
                onStart={toggleVisible}
                params={[]}
                tx={
                  api.tx.electionsPhragmen
                    ? 'electionsPhragmen.removeVoter'
                    : 'elections.removeVoter'
                }
              />
            }
```

Tbaut: :+1: fixed this as I realized it a tiny bit before!

packages/page-council/src/Overview/Vote.tsx
```diff
@@ -109,6 +109,20 @@ function Vote ({ electionsInfo }: Props): React.ReactElement<Props> {
             </Modal.Columns>
           </Modal.Content>
           <Modal.Actions onCancel={toggleVisible}>
+            {defaultVotes.length > 0 && (
+              <TxButton
+                accountId={accountId}
+                icon='trash-alt'
+                label={t<string>('Clear all')}
```
jacogr, line 116: I would actually suggest making this `isDisabled` instead of hidden.

So basically when you vote the first time you see it - it won't help, but at least some will remember and go there to clear.

Tbaut: agreed, I'll do that.

packages/page-council/src/Overview/Vote.tsx
```diff
@@ -109,6 +109,19 @@ function Vote ({ electionsInfo }: Props): React.ReactElement<Props> {
             </Modal.Columns>
           </Modal.Content>
           <Modal.Actions onCancel={toggleVisible}>
+            <TxButton
+              accountId={accountId}
+              icon='trash-alt'
+              isDisabled={!defaultVotes.length}
+              label={t<string>('Clear all')}
```
jacogr, line 116:
```suggestion
              label={t<string>('Unvote prior')}
```

Tried "clear prior", "remove prior" - what about the above? (It current be confused with clearing the selection above, hence the rename). Naming stuff is hard.

Tbaut: Because the selection can change (and prior might be confusing), what do you think about 
```suggestion
              label={t<string>('Unvote all')}
```

Issue #3073, opened 2020-06-28
Display fees alongside registration


jacogr:
Pulled from constants - overall & per field

Tbaut:
Do you mean when setting identity in accounts, calculate the total based on the following ? 
![image](https://user-images.githubusercontent.com/33178835/89788022-c93de500-db1e-11ea-9505-9cd09d01d526.png)

jacogr:
Yes. Just an overall summary of the values.

PR #3387, opened 2020-08-11
Show identity deposit info

Tbaut:
closes: #3073 

This looks horrible but I'm at the point where I've been testing so many bad positions, that I need to get more input where you think this should sit.
I've been testing next to the toggle (on the right), but it's hacky because the toggle is in absolute positioning and it's a lot of text.

Suggestions:
- show the deposit below (like now) but only when the field is selected
- or maybe we should just add a modal column that says "Each field is XYZ Unit", because repeating the same doesn't add much value (afaiu the deposit is per field, the amount of data will just change the Tx fee)
- I'd consider making the total deposit stand out a bit without the `article` (so no white background), but with a bigger font, not sure if there's anything similar in the UI.


jacogr:
I would suggest just showing the total. It is just way too busy otherwise. (And effectively that is the only surprising bit, i.e. how much do I need to pay)

Tbaut:
Back to something simple, with a helper text:
image

jacogr:
From a UI consistency poc, would just suggest an `InputBalance` with `isDisabled` and a default value, e.g. as in transfer -

image

(There is one dialog where this approach is not consistent, the signer dialog - there the retrieved value rather align with the breakdown for the extrinsic)

Tbaut:
The calculation is wrong, when testing the deposit tells me 0.04 whereas I get 0.01 reserved. Taking a look.

Tbaut:
This ends up being extremely simple. The fields that the UI proposes to add are part of the `BasicDeposit`.
The `FieldDeposit` is actually only used for `extra_fields` [here](https://github.com/paritytech/substrate/blob/master/frame/identity/src/lib.rs#L771). All the `display`, `legal` etc.. aren't extra fields as defined in `IdentityInfo` [here](https://github.com/paritytech/substrate/blob/master/frame/identity/src/lib.rs#L333)

jacogr:
Ok, so the extra fields are for the additional fields - we don't allow entry of these (but actually should soon, there are some ideas floating around on profile info, so once that takes shape, can extend in that direction)

Tbaut:
Should I actually add an issue to be able to add `extra_fields` to this modal or do we deem this as "nice to have" and keep it unlogged for now?

jacogr:
No, let's keep it outside - so the idea is to let the validators log extra info/fields (potentially), but in that case it will be "additional", but we will keep control over the names to make sense. So don't want to open up completely. So I guess it will look the same as above, but be stored as additional under known keys.

But still up in the air, atm it is not 100% workable since any change requires a re-verification, which is well, problematic.

Tbaut:
> we will keep control over the names to make sense

Understood, then we can log this when the needs is more pressing and we can set these names in ~~stone~~ block

packages/page-accounts/src/Accounts/modals/IdentityMain.tsx
```diff
@@ -27,7 +27,7 @@ interface WrapProps {
 }
 
 interface ValueState {
-  info: Record<string, unknown>;
+  info: Record<string, any>;
```
jacogr, line 30: Just a question around this one - where was it needed? (Just interested since the best-practice is to use `unknown` (and then cast as required), but alas, there is still a lot of work left, just been touching them as it comes into focus.

Tbaut: It's not needed anymore, reverting. The problem was when I did something like:
```js
Object.values(info).forEach((i: Record<string, string>) => console.log(i.bla));
```
where it was complaining on the `i`:
> Argument of type '(i: Record<string, string>) => void' is not assignable to parameter of type '(value: unknown, index: number, array: unknown[]) => void'.

jacogr:
Thank you!

PR #3388, opened 2020-08-11
Add clear identity button

Tbaut:
Allows to remove an identity when on was set.
Will probably minimally conflict with #3387


jacogr:
Thank you!

PR #3391, opened 2020-08-11
Erk.

jacogr:


jacogr:
It is not funny when you actually get a bug report. Tested it... and then made this change and didn't re-test. And it broke.

Issue #3161, opened 2020-07-10
Remove *.google font refs


jacogr:
Due to semantic-ui - but it flies in the face of decentralized and not relying on big borther.

https://github.com/polkadot-js/apps/pull/3158/files#r452846065

jacogr:
Closing, will be done automagically once SUI is removed.

PR #3399, opened 2020-08-12
Sidebar redesign

alexcherman:


packages/apps/src/Apps.tsx
```diff
@@ -199,7 +197,7 @@ export default React.memo(styled(Apps)`
 
   &.menu-open {
     .apps--SideBar-Wrapper {
-      width: 12rem;
+      max-width: 230px;
```
jacogr, line 200: Would really prefer to stick to rem since it is the way it is applied everywhere. (Same for padding, etc.) Only fallback to px when there is no other choice.

jacogr:
Way off-base questions here, just random thoughts, ignore at will.

If we are reworking the navigation, we may want to explore if a top-bar is not more useful. So we really have issues atm with horizontal space as opposed to vertical. So with a two-level topbar it may work -

- first level is grouping of functionality, i.e. Governance (Democracy, Council, Treasury, Bounties)
- second level is what is currently the page headers

The issue with the grouping is obviously that we want to avoid clicks, so when hover it should just open so the user have a less-cluterred experience, but can still get to everything with 1 click (and some mouse movement)

This PR (which I have not looked at) sparked this thought. If we are redoing, we don't need to keep the status quo - since we really want to make things as optimal as possible.

(Like I said, ignore at will, not against the sidebar otherwise we won't have one, but if there are much-better approaches, happy to explore)

packages/apps-config/src/ui/icons/accounts.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg">
```
jacogr, line 1: Would really prefer not adding icons and just stick to the fa set. (Happy to change the set, but needs to be consistent)

The issue here is that the base here is not "just for apps", a lot of developers actually fork and add their own custom modules. So doing this has an impact downstream as well.

Additionally, it is easier "picking from a set" than finding correctly licensed svgs.

packages/apps-routing/src/accounts.ts
```diff
@@ -12,7 +13,7 @@ export default function create (t: <T = string> (key: string, text: string, opti
     display: {
       needsApi: []
     },
-    icon: 'users',
+    icon: Icon as string,
```
jacogr, line 16: As explained above, this really has wide-ranging impacts. So would keep to the fa set (imperfect as it may be)

packages/apps/public/index.html
```diff
@@ -6,6 +6,7 @@
     <link rel="manifest" href="manifest.json">
     <link rel="shortcut icon" href="favicon.ico">
     <title><%= htmlWebpackPlugin.options.PAGE_TITLE %></title>
+    <link href="https://fonts.googleapis.com/css2?family=Nunito:ital,wght@0,200;0,400;0,800;1,200;1,400&display=swap" rel="stylesheet">
```
jacogr, line 9: Really would prefer not pulling additional fonts from google. See this issue as well https://github.com/polkadot-js/apps/issues/3161 (which is WIP, a big driver behind removing semantic-ui). 

If we want a consistent font, need one with the correct licenses added specifically to the repo. (Really trying to get to self-contained without relying on centralized services)

packages/apps/src/SideBar/ChainInfo.tsx
```diff
@@ -24,35 +24,61 @@ function ChainInfo ({ className = '', isToggled, onClick }: Props): React.ReactE
...
-      onClick={onClick}
-    >
-      <div className='apps--SideBar-logo-inner'>
-        <ChainImg />
-        <div className='info'>
-          <Chain className='chain' />
-          {runtimeVersion && (
-            <div className='runtimeVersion'>{t<string>('version {{version}}', { replace: { version: runtimeVersion.specVersion.toNumber() } })}</div>
+    <div className={className}>
+      <div className='apps--Sidebar-top'>{'polkadot{.js}'}</div>
```
jacogr, line 28: I would remove this label completely, see for instance https://portal.chain.centrifuge.io/#/explorer

packages/apps/src/SideBar/ChainInfo.tsx
```diff
@@ -24,35 +24,61 @@ function ChainInfo ({ className = '', isToggled, onClick }: Props): React.ReactE
...
+      <div
+        className='apps--SideBar-logo'
+        onClick={onClick}
+      >
+        <p className='apps--SideBar-label'>Network:</p>
```
jacogr, line 33: All string should be translatable. Having said that, not sure this adds any additional value having this here.

packages/apps/src/Apps.tsx
```diff
@@ -113,33 +113,43 @@ export default React.memo(styled(Apps)`
   flex-direction: row;
   height: 100vh;
 
+  * {
+    font-family: 'Nunito', sans-serif !important;
```
jacogr, line 117: This is problematic since it applies to the UI, but not to the modals. So we have a mixture of fonts, see for instance -

image

jacogr:
Closing, replaced nave with top. There is a bunch of effort still needed there, e.g. stuff like this https://github.com/polkadot-js/apps/pull/3459#issue-471405544 and https://github.com/polkadot-js/apps/pull/3422#issue-468275204 (and tons more I have not even thought about) but it moved the needle into the right direction.

Issue #3224, opened 2020-07-17
Display sub accounts (when parent) in sidebar


jacogr:

Tbaut:
I'll tackle this one next

Tbaut:
Having a deeper look, I think the best way to do this is to add the property `childs?: [AccountId]` to the `DeriveAccountRegistration` type in the `api`. I'd add this right next to the parent retrieval [here](https://github.com/polkadot-js/api/blob/master/packages/api-derive/src/accounts/info.ts#L85). Does that sound like the way you'd do that?

jacogr:
Generally I would agree. In this case I would rather not and keep the call specific to the component/container. 

The issue atm is that we are already overloading due to the number of queries and the sheer volume of data. See for instance staking, where we add an additional subsOf for all 500 (currently) on Kusama.

If this is only used here, would rather just do the `identity.subsOf` call straight in the sidebar, as it gets opened.

PR #3149, opened 2020-07-09
Fix accounts not showing with no democracy

jacogr:
Closes https://github.com/polkadot-js/apps/issues/3148 (Introduced in https://github.com/polkadot-js/apps/pull/3122)

Honestly dunno how I tested this and missed something this simple.

PR #3405, opened 2020-08-12
Show sub accounts in sidebar if any

Tbaut:
closes https://github.com/polkadot-js/apps/issues/3224

Looks like this:
image

image


packages/page-accounts/src/Sidebar/Identity.tsx
```diff
@@ -90,6 +92,20 @@ function Identity ({ address, identity }: Props): React.ReactElement<Props> | nu
                 </div>
               </div>
             )}
+            {!!subs?.length && (
+              <div className='tr subs'>
+                <div className={subs.length > 1 ? 'th top' : 'th'}>{t<string>('subs')}</div>
```
Tbaut, line 97: Can't remember how you go about plurals. If you cater for it or not. I'm happy to add a condition and have "sub" and "subs"

Tbaut: Thinking about it, it's actually much better, committing that.

packages/page-accounts/src/Sidebar/Sidebar.tsx
```diff
@@ -249,7 +253,7 @@ export default React.memo(styled(FullSidebar)`
     > * {
       margin-bottom: 0.4rem;
 
-      &:not(:first-child) {
+      &:not(:first-sub) {
```
jacogr, line 256: ? Not sure about this...

Tbaut: nice catch, sorry about that, a search&replace that went wrong.
```suggestion
      &:not(:first-child) {
```

jacogr:
So not 100% sure about the first one in your example, since we don' actually show those as linked. But I think it has grown on me, since it is almost like a identity easter-egg where we do show relationships. So it is actually cool.

Have some issues with this -

image

And there are a lot of those. So would suggest making -

- the subs expandable with `subs (<count>)`
- move them to the bottom (below fields like website)

Tbaut:
Looks like this now:
image

packages/page-accounts/src/Sidebar/Identity.tsx
```diff
@@ -22,6 +23,7 @@ function Identity ({ address, identity }: Props): React.ReactElement<Props> | nu
   const { api } = useApi();
   const { isRegistrar, registrars } = useRegistrars();
   const [isJudgementOpen, toggleIsJudgementOpen] = useToggle();
+  const subs = useCall<[BalanceOf, AccountId[]]>(api.query.identity.subsOf, [address])?.[1];
```
jacogr, line 26:
```suggestion
  const subs = useCall<[BalanceOf, AccountId[]]>(api.query.identity?.subsOf, [address])?.[1];
```

Tbaut:
> Almost forgot...

I'm pretty sure I'll forget every single time if there's no TS error on that. Do you know why there isn't any?

jacogr:
The base Substrate includes it. The issue it that not all chains do. :(

Tbaut:
Ah sure, I'll make sure to keep that in mind as a fail safe for later.

jacogr:
Always a bitch,  also step into it, but the major source of issues - verifying it works is easy, verifying it doesn't break, is difficult -

https://github.com/polkadot-js/apps/pull/3299/files
https://github.com/polkadot-js/apps/pull/3149/files

PR #3410, opened 2020-08-13
Allow saving custom endpoints

### LINKED ISSUE #2826: Multiple custom endpoints
I assume many contributors will keep switching between Kusama, Polkadot and other Substrate networks, ideally using their own nodes. In that scenario, the current UX is not very practical as there is no way to save multiple custom endpoints.

A decent alternative is to use nginx with multiple "location" pointers so one only has to type/adjust the name of the network they want to connect to:
wss://1.2.3.4/polkadot
wss://1.2.3.4/kusama
[...]

Or probably simpler but not as easy to memorize, manually adjusting the port
ws://127.0.0.1:9944 (say Polkadot)
ws://127.0.0.1:9945 (say Kusama)
[...]

A low priority issue certainly, but one nonetheless.

Tbaut:
closes #2826

edit: 
- you can switch to a custom endpoint without saving it
- saving a custom endpoint switches to it directly
- deleting a custom endpoint removes it from the list but you can save it again right away
custom-endpoints-2

----
obsolete: A first shot with some questions

- You can't delete a custom endpoint yet. I'd like to add this functionality. UX-wise, I was thinking about changing the save icon to a "trash" once we select a custom endpoint. Clicking it would simply delete it from the list. Does that sound good to you?
- There's already a functionality to add custom endpoints via env vars. Not entirely sure if it would make sense to merge those 2?
- What's your standard way to share constants between packages such as a localStorage key (see comment in code)?

Here's how it looks like.
[custom-endpoints](https://user-images.githubusercontent.com/33178835/90123194-b6b1ee80-dd5e-11ea-8286-52f9192b732d.gif)


packages/apps-config/src/settings/endpoints.ts
```diff
@@ -17,6 +17,27 @@ interface EnvWindow {
   }
 }
 
+function createOwn (t: TFunction): LinkOption[] {
+  try {
+    const CUSTOM_ENDPOINT_KEY = 'polkadot-app-custom-endpoints';
```
Tbaut, line 22: This needs to be shared between this package and apps. I was thinking about setting and exporting it in this package, and pull if in apps from `@polkadot/apps-config/settings`. How does this sound?

packages/apps/src/SideBar/constants.ts
```diff
@@ -0,0 +1,5 @@
+// Copyright 2017-2020 @polkadot/apps authors & contributors
+// This software may be modified and distributed under the terms
+// of the Apache-2.0 license. See the LICENSE file for details.
+
+export const CUSTOM_ENDPOINT_KEY = 'polkadot-app-custom-endpoints';
```
Tbaut, line 5: that's obviously wrong here, I'll move it.

jacogr:
1. Yes, agreed.
2. I would not bother with it atm, it really is not in wide-spread use at all (most installs actually customize the apps config for pin-point accuracy)
3. What you have there with the constants file looks spot-on

Tbaut:
> What you have there with the constants file looks spot-on

ah, so then I miss something. My constant file is in `apps` right now. Do you mean I should export it and pull it in the other package from `@polkadot/apps`? This looks weird to me.

jacogr:
Ahh, apps-config for the shared if it is used there. (example https://github.com/polkadot-js/apps/blob/master/packages/apps-config/src/api/constants.ts)

Yes, apps should not be depended on by anybody. (Well, elctron can)

Tbaut:
Saw your PR about useEffect + useState -> useMemo. Will do that here as well.

jacogr:
Looks good, but a small bug.

screencast 2020-08-14 11-09-12

Tbaut:
Nice catch, should be the save button indeed.

Tbaut:
It should be what we expect now:
- you can switch to a custom endpoint without saving it
- saving a custom endpoint switches to it directly
- deleting a custom endpoint removed it from the list but you can save it again right away
custom-endpoints-2

jacogr:
thank you!

Issue #3411, opened 2020-08-13
Account adding. Wrong account generation from Raw seed


StepanLavrentev:
**Detail description:**
For "Raw seed" the address is generated incorrectly when you specify a password in the derivation path

*Case:*

1. Go to accounts
2. Click add account
3. Choose mnemonic type - Raw seed
4. Fill in the required fields
5. Remember the address that was generated
6. Enter derivation path - "///password"

**Actual result:**
Address has not changed after entering ///password

<details>
  <summary>Spoiler!</summary>
  
![image](https://user-images.githubusercontent.com/40560660/90132518-ae65ad80-dd76-11ea-922e-79298db796b7.png)
![image](https://user-images.githubusercontent.com/40560660/90132537-b7567f00-dd76-11ea-8200-69f7fd6a5717.png)

</details>

Tbaut:
I could reproduce, but the bug is rather that any derivation path should be ignored with a raw seed. In other words, not only `///password` should be ignored, but any derivation path like `/soft//hard///password`, the UI should probably hide this field altogether then. I'll dig a bit into that because this is only from memory, there used to be a mention about this in the kusama wiki, but I couldn't find it.

Tbaut:
I found it, it's on substrate.dev about subkey: https://substrate.dev/docs/en/knowledgebase/integrate/subkey#password-protected-keys

> Note that the "Secret seed" is not password protected and can still recover the account. The "Secret phrase" however, is not sufficient to recover the account without the password.

The UI here mentions that a hex or a string is accepted for "raw seed". I imagine that this is for backward compatibility reasons.
![image](https://user-images.githubusercontent.com/33178835/90140370-f0dcb980-dd79-11ea-8360-a05e535a2df6.png)

@jacogr will surely know more about the why.

jacogr:
Ok, that makes sense. Should at least error or show as invalid.

So `///password` is applied when converting from mnemonic to entropy.

Tbaut:
From my tests, it looks like the UI is behaving exactly like `subkey` (this contradicts part of my first message, I edited it):
- derivation paths (hard and soft) have an incidence on both the mnemonic and the hex seed
- any `///password` is ignored for hex seed but taken into account for mnemonic (as the doc says)
```bash
subkey generate -w 12
  Secret phrase `trophy welcome benefit junk simple post still zoo athlete tape dove explain`
[...]

PR #3412, opened 2020-08-13
Error out when using hex seed and a passowrd in derivation path

Tbaut:
closes https://github.com/polkadot-js/apps/issues/3411


jacogr:
Thank you!

PR #3422, opened 2020-08-15
Nav re-imagined

jacogr:
Based on the comment in https://github.com/polkadot-js/apps/pull/3399#issuecomment-673349202

screencast 2020-08-15 17-02-05

What it does do is get rid of the ever-growing sidebar. Additionally it actually makes the content more usable, the current horizontal layouts take up a lot of space. 

Net-negative in terms of code as it stands here, which is good.

It needs a lot of TLC -

- ~~settings needs to be a clickable item (instead of menu)~~
- ~~send/transfer needs to be a clickable item (not in menu, quick access)~~
- ~~accounts needs to go first~~
- ~~styling (menu is good, however the menu popups are not perfect color-wise)~~ (better, not perfect)
- ~~no actual over-item hover here, highlights are good~~
- ~~NodeInfo does not appear (RPC, API & UI versions)~~
- because of the location of the "indicated item", there is left-to-right movement of the menu
- ~~there is code that can be dumped, notably the app styling relating to menu~~
- ~~Apps Content can probably use an extra 1rem each left/right side now~~
- ~~useCounter indicators need to be re-applied (top level may be additive?)~~
- ~~Society is a bit lost (under social)~~
- ~~Don't have a place for the wiki & GitHub links (and can actually extend with other community links)~~

Some stuff that may not be done here, but follow-up -

- with this we can split contacts from accounts again (in menu, own route)
- rpc (under developer), can move from toolbox to own app (again)
- split treasury tips into tab (not app)
- split dispatch/scheduled on democracy into tab
- status, status, status - they need to be condensed & out of the way 

Holes left -

- with this, it looks nice, but the rest of the app does not - so there is a general revamp needed
- ... not sure what the above means, but needs to "be nice"

... etc


jacogr:
Alternative topbar/navbar -

image

PR #3435, opened 2020-08-18
italicize "New" as per style guide

joshua-mir:
I'm not certain how to italicize https://github.com/polkadot-js/apps/blob/7bc5c314c4dd7d18999cae0cea56fb236513f18f/packages/react-api/src/Api.tsx#L101 though


jacogr:
That is the reason why it is not done, since it is not going to be consistent. It all works with text-only and the API itself does the bulk of the formatting returning a text string. So if done at one place, it needs to be be done everywhere. So this adds more confusion as opposed to less.

jacogr:
Closing, format removed.

PR #3447, opened 2020-08-19
Update Kulupu RPC endpoint

sorpaas:
The old domain is usually troubled by adblocker for some reasons. Moving this to `rpc.kulupu.corepaper.org` to improve situations on that.


jacogr:
I wonder if it is the `.network` extension then. The W3F nodes have exactly the same issue.

jacogr (approved):
Thank you!

PR #3451, opened 2020-08-20
add crab (darwinia canary network)

WoeOm:
website:
https://darwinia.network/

telemetry:
https://telemetry.polkadot.io/#/Crab

Crab like kusama and polkadot's relation.

Later we'll bridge the crab to darwinia. The tokens can be transfer between darwinia and crab with ratio 1:100.


packages/apps-config/src/api/spec/crab.ts
```diff
@@ -0,0 +1,230 @@
+// Copyright 2017-2020 @polkadot/apps-config authors & contributors
+// This software may be modified and distributed under the terms
+// of the Apache-2.0 license. See the LICENSE file for details.
+
+// structs need to be in order
+/* eslint-disable sort-keys */
+
+export default {
+  '__[frame.system]__': {},
```
jacogr, line 9: These are weird. Would really like to track where they come from. (All good, just seems like something that should not be decorated by the API at all)

packages/apps-config/src/settings/endpoints.ts
```diff
@@ -85,6 +85,12 @@ function createLive (t: TFunction): LinkOption[] {
       value: 'wss://kusama.polkadot.cloud.ava.do'
     },
     // alphabetical based on chain name
+    {
```
jacogr, line 88: Just move it one down, `ce` before `cr`

packages/apps-config/src/api/spec/crab.ts
jacogr: Was happy with them if needed to operate, just weird if they do get decorated by the API (that would point to some issue in that layer).

jacogr (approved):
Thank you!

jacogr:
There are a couple missing - 

image

WoeOm:
> There are a couple missing -
> 
> image

I will fill in this type as soon as possible

jacogr:
No issues, just wanted to make sure you are aware (these are still tricky... we will get to types-via-metadata and all this effort will fall away)

WoeOm:
> No issues, just wanted to make sure you are aware (these are still tricky... we will get to types-via-metadata and all this effort will fall away)

Crab has two tokens, cring and ckton. Many modifications are made on the basis of substrate. Even if apps are submitted with a type, they will still report errors on somee pages. Is there a good solution for this part?

Can we submit pr to apps repo to fix these problems for crab?

jacogr:
You can have 2 types of overrides -

- per spec
- per chain

So the types for spec are applied first, if there are chain-specific types (different or non-existent in others) you can place then under chain. (Just the deltas needed. The Edgeware chain is an example here although it is actually done incorrectly since it has the spec and chain duplicates all of them again - this extra duplication is not needed... I know, I should get to fiing that)

WoeOm:
> You can have 2 types of overrides -
> 
> * per spec
> * per chain
> 
> So the types for spec are applied first, if there are chain-specific types (different or non-existent in others) you can place then under chain. (Just the deltas needed. The Edgeware chain is an example here although it is actually done incorrectly since it has the spec and chain duplicates all of them again - this extra duplication is not needed... I know, I should get to fixing that)

For example, staking and crab support cring (equivalent to dot) and ckton. At that time, apps only had one input box to fill in the amount of cring, and there was no second box to fill in ckton.

There are some content that cannot be adapted in the api -> api-derive module, especially staking, which will cause the page to crash

jacogr:
The apps UI only deals with the Substrate-default staking module.

WoeOm:
> The apps UI only deals with the Substrate-default staking module.

Thank you~

jacogr:
On those will take a peek, there may be some low-hanging stuff that can at least make it not crash. Will check the assumptions and what is there.

PR #3497, opened 2020-08-24
rococo acala

aniiantt:
Hi, We built a acala Parachain based on Rococo, and now we hope to put it into Polkadot Apps.


jacogr:
As always, thank you!

PR #3518, opened 2020-08-25
add rococo laminar

aniiantt:
Hi, we've wired the Laminar Chain into Rococo and now want it to display on the Polkadot Apps.

Because a type is missing, only after this PR merge can it work well.

https://github.com/polkadot-js/api/pull/2519


jacogr:
Will update the apps UI API and pull this in. Thank you very much for handling both sides of the divide.

jacogr (approved):

Issue #3541, opened 2020-08-27
Multisig Creation accepts threshold of 1 but runtime requires min. 2


lucasvo:
I am able to create a multisig with 1 signer in the UI and it shows up as an account letting me believe I can actually use the account. However the runtime requires you to have a threshold of at least 2:
https://github.com/paritytech/substrate/blob/master/frame/multisig/src/lib.rs#L531

Here's my multisig that I created on https://polkadot.js.org/apps/#/accounts

![screen_2020-08-27-12-36-13](https://user-images.githubusercontent.com/31727/91442806-ee8e5b00-e861-11ea-9a31-f6fc732835ef.png)

But the transaction gets rejected: https://polkadot.subscan.io/extrinsic/1331769-3

lucasvo:
FYI, I just locked up 10DOTs because of this indefinitely.

lucasvo:
@jacogr I would argue it's a bit more than an `[enhancement]` we did loose 50$ because of this :( But thanks for fixing it quickly.

jacogr:
The UI doesn't actually do any checks in most places, so it is an enhancement from that pov. It doesn't actually duplicate constants from Rust generally. Not saying it shouldn't, like here having it certainly protects the user and there is a lot more places where it should actually do checks to prevent invalid transactions. 

In this case it is def.more important having checks than other places where you only lose tx fees since here it first locks before failing on the subsequent tx.

gavofyork:
Consider using this rather than disabling completely: https://github.com/paritytech/substrate/blob/master/frame/multisig/src/lib.rs#L233

jacogr:
Yes, we do have an issue logged for multisig of 1, specifically https://github.com/polkadot-js/apps/issues/3862

PR #3542, opened 2020-08-27
Adjust min multisig threshold to 2

jacogr:
Closes https://github.com/polkadot-js/apps/issues/3541


gavofyork:
@jacogr probably just want to use this, eventually: https://github.com/paritytech/substrate/blob/master/frame/multisig/src/lib.rs#L233

jacogr:
Yes, logged in https://github.com/polkadot-js/apps/issues/3862 (Actually have an implementation of that, but it is not 100% yet, there are some gaps where the complicity for either-or just made things slightly more difficult than it should be. So it still needs quite a bit of TLC)

PR #3606, opened 2020-09-04
fill in all empty values in all languageCaches.

motinados:
This is bug fix.

On the settings page, when I select a language, the old value remains.
I think this is because the browser language is not initialized.

In my case, if I select Korean and then Japanese, the old value remains.

(My browser language is 'ja')


jacogr (approved):
Thank you!

PR #3619, opened 2020-09-06
Fix translation key.

motinados:
This is bug fix.
Values greater than 1 will also be translated


jacogr (commented):
Can you please explain the change to me? At this point I see it will break all the translated strings that have been made for this since `{{days}} days` (as translated) !== `{{d}} days`.

However the translate object is still equivalent since (apart from the name), the value is till passed.

jacogr:
I do see this hapenning as well...

PR #3627, opened 2020-09-06
Update translation.json

fomod-in:
Improved staking and account modules for better UI understanding. Additional updates will follow also thanks to italian community contribution


jacogr:
... as always, thanks a million!

PR #3660, opened 2020-09-11
show warning when the password contains a /

### LINKED ISSUE #3647: Potential problem with using the "/" inside password part in DP
Potential problem with using the "/" symbol the password part for the derivation path.
When creating an account in the secret derivation path we can specify a hard, soft path as well as a password and after entering 3 characters "/" - everything is considered as a password but the user is not notified of this.

For example:

<details>
  <summary>On subkey app</summary>
  
</details>

<details>
  <summary>On JS app</summary>
  
</details>

**Case 1:**

1. User creating account
2. Set derivation path with password and make a mistake - type 4 characters "/"
(////password)

**Result:**
Account creation with password - "/password", but user wanted to create an account with "password"

----
**Case 2:**

1. User creating account
2. Set derivation path - "///password/soft//hard"

**Result:**
Account creation with password - "password/soft//hard"

----

Maybe should limit the entry of the "/" symbol or give a more detailed description?

Tbaut:
closes #3647
image

### DIFF
--- a/packages/page-accounts/src/modals/Create.tsx
+++ b/packages/page-accounts/src/modals/Create.tsx
@@ -35,8 +35,8 @@ type SeedType = 'bip' | 'raw' | 'dev';
 
 interface AddressState {
   address: string | null;
-  deriveError: string | null;
   derivePath: string;
+  deriveValidation? : DeriveValidationOutput
   isSeedValid: boolean;
   pairType: KeypairType;
   seed: string;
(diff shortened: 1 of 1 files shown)


jacogr (approved):
Thank you!

jacogr:
Logged https://github.com/polkadot-js/apps/issues/3662 alongside this (I'm sure it is not the only place)

Issue #3662, opened 2020-09-11
Make derivation errors translatable


jacogr:
I suppose when the error gets thrown we use an index string and then a lookup on the UI alongside the translatable text.

As seen in https://github.com/polkadot-js/apps/pull/3660/files#diff-a07a1ad3dc171e647cb05d98c01a53eeR60

Tbaut:
Ah totally, tackling that now.

jacogr:
Sure it is not the only place, I fixed some yesterday that I found on my travels (for an options dropdown). The errors are generally trickier, but if found, may as well log and attempt.

PR #3663, opened 2020-09-11
Translate error and warning messages

Tbaut:
closes https://github.com/polkadot-js/apps/issues/3662

I did not type the index keys stronger because the validation can also spit out an unknown string [here](https://github.com/polkadot-js/apps/blob/b1e07bee2cdfe28bcd71d06c67bcc58e90d8c038/packages/page-accounts/src/modals/Create.tsx#L82)

### DIFF
--- a/packages/page-accounts/src/modals/Create.tsx
+++ b/packages/page-accounts/src/modals/Create.tsx
@@ -64,17 +64,17 @@ function deriveValidate (seed: string, seedType: SeedType, derivePath: string, p
 
     // show a warning in case the password contains an unintended / character
     if (password?.includes('/')) {
-      result = { warning: 'Your password contains at least one "/" character. Disregard this warning if it is intended.' };
+      result = { warning: 'WARNING_SLASH_PASSWORD' };
     }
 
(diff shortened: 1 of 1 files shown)


packages/page-accounts/src/modals/Create.tsx
```diff
@@ -381,10 +386,10 @@ function Create ({ className = '', onClose, onStatusChange, seed: propsSeed, typ
                     value={derivePath}
                   />
                   {deriveValidation?.error && (
-                    <article className='error'>{deriveValidation.error}</article>
+                    <article className='error'>{errorIndex[deriveValidation.error]}</article>
```
jacogr, line 389: The caught error case is not handled, so `errorIndex[...] || deriveValidation.error`

Tbaut: Indeed, just realized that. Thanks. Should be fixed by https://github.com/polkadot-js/apps/pull/3663/commits/ef2d6270d30a1a31b81fc5e976b074a41fee9cb0

jacogr (approved):
Thank you!

PR #3711, opened 2020-09-18
New account creation modal

krzysztof-jelski:
Account creation modal with new designs.

I'm open to suggestions as to ways of separating new and old designs (some parts use copied components, some use just additional styling).

Probably still some styling issues exist, though I believe it's close enough to open a PR 🙂


Tbaut:
Nice!! I played with it super briefly, found a couple nits here and there:
- next step button is lacking the icon
- back button (when on the second step) seems disabled because it has the same color as the background
- shouldn't the "advanced" part be hidden by default?
- the warning about the seed phrase doesn't have the same styling as the other warnings in the interface.
- It seems that the PR is based on an older master, I could see some warning [recently implemented](https://github.com/polkadot-js/apps/pull/3660) were not present. Merging master would be needed I guess.

jacogr:
Things I really like:

- The TextArea instead of the single-line entry for the seed
- The back button to the left instead of squashed in the middle. (Have not looked at the impl. but it useful everywhere, so passing it to Button.Group for alignment makes sense)
- Clear split between setup and the final creation
- The top close, although generally would still prefer a cancel on the bottom as well

What is missing:

- The Advanced type & derivation selection needs to be on the first - all these together creates the account
- There is no warning anymore that says "don't do this in web"
- The guidance on the right is completely gone (it was actually added as a result of early mock-ups)
- We are missing the confirmation (which is not great) and backup step in the current create

What doesn't work:

- The i18n text with ALL CAPS (if this is needed, do it via CSS)
- The button styling here doesn't align with the UI and have ill-effects elsewhere, i.e. looks the same as toggle groups and just the changes here made all disabled buttons everywhere have a gray background (so they all stick out now)
- The text-area labels doesn't align with what we have everywhere else (done this way for the insane recursiveness of e.g. extrinsics or proposals where outer labels just don't make sense and yield and even-more unreadable mess)
- We have lost the per-field help labels
- We are adding more google-stuff, trying to get rid of that and the tracking it brings (see eg. https://github.com/polkadot-js/apps/issues/3161). Happier with slightly-less pretty fonts and browser defaults, than Google-centralized.

Overall, I think the functional ideas here (with a couple of additions/tweaks can work really well, possibly even taking us to 3 overall steps in this dialog), however from a styling pov there is work to be done to make things fit and pleasing.

krzysztof-jelski:
Thanks for the feedback!

> * On the first screen, the warning about the seed phrase doesn't have the same styling as the other warnings in the interface.

As we're targeting changing the design for the whole app, I'm afraid there will be some time when we'll face some UI inconsistencies. So I propose to leave it like this and gradually roll out new styles to other places.

> * It seems that the PR is based on an older master, I could see some warning [recently implemented](https://github.com/polkadot-js/apps/pull/3660) were not present. Merging master would be good I guess.

Thanks for the catch! That's definitely my mistake during merging

krzysztof-jelski:
Thanks for your feedback! Agree with most of it 🙂. We'll go through the designers with some.

> * We are missing the confirmation (which is not great) and backup step in the current create

So I'd say that in the current flow copying/printing of the seed plays the role of creating backup. That's why I consciously removed the download step.

As for confirmation step, do you really consider it useful? It just shows the address, the name, the cryptography used and derivation path -- all the info you reviewed at previous step. So the role would be, for example, to check if you haven't misspelled the derivation path?

> * The text-area labels doesn't align with what we have everywhere else (done this way for the insane recursiveness of e.g. extrinsics or proposals where outer labels just don't make sense and yield and even-more unreadable mess)

My take on this would be to make it two different components. So we'd have an input-like text area in modals and text area with description inside for extrinsics and such. What do you think?

> * We have lost the per-field help labels

Do you mean the guidance on right or something else?

> * We are adding more google-stuff, trying to get rid of that and the tracking it brings (see eg. #3161). Happier with slightly-less pretty fonts and browser defaults, than Google-centralized.

The font has an [open licence](https://www.fontsquirrel.com/license/nunito-sans), so we'll add it to the repo instead of fetching from G.

jacogr:
The download and confirmation is important -

- for confirmation is has been explicitly asked for since people would like to double and tripple-check. Sadly people make mistakes
- the download once again, has proven very useful - it is an extra protecting against losing access

The confirmation is not great as it stands, but also don't want to swing the pendulum towards "having to fill in some words", rather just allowing final steps and checks.

On what is missing -

- the help text on the right
- the help text per label

Both of these may not add an immense amount of extra information to something that people know, however the concepts all-over are exceptionally new and complex in most cases, so the guidance and consistency is important.

krzysztof-jelski:
@jacogr @Tbaut Ok guys, I believe you can have a second look now.

It's quite exhausting keeping up with changes in master, so I'd be happy we get it merged as soon as possible, and then follow up on improvements in consecutive PRs.

Also (obviously) don't expect the dark theme to work with yet.

Next step would be to apply new styles to the Derive component, which should also remove some duplication among components and styles we have now.

packages/page-accounts/src/CreateAccount.spec.tsx
```diff
@@ -5,10 +5,31 @@ import AccountsApp from '@polkadot/app-accounts';
...
 import '@polkadot/react-components/i18n';
+import { ThemeDef } from '@polkadot/react-components/types';
 import { useApi } from '@polkadot/react-hooks';
-import { fireEvent, render } from '@testing-library/react';
+import { fireEvent, render, waitForElementToBeRemoved } from '@testing-library/react';
 import React, { PropsWithChildren } from 'react';
 import { MemoryRouter } from 'react-router-dom';
+import { ThemeProvider } from 'styled-components';
+
+// copied from packages/apps/src/themes.ts, as we cannot reference apps module
```
jacogr, line 15: Single this is a test, we can just pull the import. (And it is in account already)

packages/page-accounts/src/modals/Create.tsx
```diff
@@ -1,27 +1,38 @@
 // Copyright 2017-2020 @polkadot/app-accounts authors & contributors
 // SPDX-License-Identifier: Apache-2.0
 
+import NewPasswordInput from '@polkadot/app-accounts/modals/PasswordInputNew';
```
jacogr, line 4: Please change the ordering back.

- first types
- then externals
- (in the same block) polkadot-js packages
- then locals

packages/page-accounts/src/modals/Create.tsx
```diff
@@ -1,27 +1,38 @@
...
+import CreateConfirmation from '@polkadot/app-accounts/modals/CreateConfirmation';
+import { useTranslation } from '@polkadot/app-accounts/translate';
+import { ModalProps } from '@polkadot/app-accounts/types';
```
jacogr, line 7: These are all local, no package imports, just relative.

packages/page-accounts/src/modals/Create.tsx
```diff
@@ -381,60 +424,232 @@ function Create ({ className = '', onClose, onStatusChange, seed: propsSeed, typ
...
+          label={t<string>('Next step')}
+          onClick={_nextStep}
+        />
+      </div>
+      }
+
+      {step === 3 &&
+      <div className='ui--Modal-Footer'>
+        <BackButton className='ui--Modal-back-button'
+          onClick={_previousStep}/>
```
jacogr, line 532:
```suggestion
        <BackButton
          className='ui--Modal-back-button'
          onClick={_previousStep}
        />
```

And the same styling above.

packages/page-accounts/src/modals/Create.tsx
```diff
@@ -381,60 +424,232 @@ function Create ({ className = '', onClose, onStatusChange, seed: propsSeed, typ
...
+
+export default styled(Create)`
+  & * {
+    font-family: 'Nunito Sans', sans-serif;
+  }
+
+  &.ui.modal > .header:not(.ui) {
+    font-family: 'Nunito Sans', sans-serif;
+    text-transform: initial;
+  }
```
jacogr, line 556: This is now inconsistent with the rest of the UI. Prefer not doing that.

packages/page-accounts/src/modals/InfoBox.tsx
```diff
@@ -0,0 +1,46 @@
...
+  type: 'alert' | 'error';
+  upperCase?: boolean;
+  value: string;
+}
+
+function InfoBox ({ className, icon, value }: Props): React.ReactElement<Props> {
+  return (
+    <article className={className}>
+      { icon && <Icon icon={icon}
+        size='2x' /> }
```
jacogr, line 20:
```suggestion
      {icon && (
        <Icon
          icon={icon}
          size='2x'
        />
      }
```

packages/page-accounts/src/modals/InfoBox.tsx
```diff
@@ -0,0 +1,46 @@
...
+      <div>{value}</div>
+    </article>
+  );
+}
+
+export default styled(InfoBox)`
+    display: flex;
+    align-items: center;
+    width: 100%;
```
jacogr, line 29: (nit) indents here seem to be 4 spaces (or tabs?)

packages/react-components/src/BackButton.tsx
```diff
@@ -0,0 +1,35 @@
...
+  className?: string;
+  onClick: () => void;
+}
+
+function BackButton ({ className, onClick }: Props): React.ReactElement<Props> {
+  return (
+    <button className={className}
+      onClick={onClick}>
+      <Icon icon='arrow-left' />
+    </button>
```
jacogr, line 18: Rather use the standard `<Button .. />` for consistency

packages/react-components/src/Button/Button.tsx
```diff
@@ -132,4 +133,23 @@ export default React.memo(styled(Button)(({ theme }: ThemeProps) => `
...
+      box-shadow: none;  
+    }
+
+    &:not(.hasLabel) {
+      line-height: initial;
+    }
+
+    &.hasLabel {
+      padding: 0.65rem 1.1rem 0.65rem ${1.1 - ICON_PADDING}rem;
+    }
```
jacogr, line 153: This should really not know about anything in `apps-*` - specific overrides should go in there.

packages/react-components/src/Button/types.ts
```diff
@@ -9,7 +9,7 @@ export type Button$Callback = () => void | Promise<void>;
 export interface ButtonProps {
   children?: React.ReactNode;
   className?: string;
-  icon: IconName;
+  icon?: IconName;
```
jacogr, line 12: Icons are required on all Buttons. We really don't want to do this, it was literally months of work to make sure we have icons everywhere enabling us to switch this on.

packages/react-components/src/CopyToClipboard.tsx
```diff
@@ -0,0 +1,50 @@
+// Copyright 2017-2020 @polkadot/react-components authors & contributors
+// SPDX-License-Identifier: Apache-2.0
+
+import { Icon } from '@polkadot/react-components';
+import StatusContext from '@polkadot/react-components/Status/Context';
+import { useTranslation } from '@polkadot/react-components/translate';
+import React, { useCallback, useContext } from 'react';
```
jacogr, line 7: Import ordering.

packages/react-components/src/BackButton.tsx
```diff
@@ -0,0 +1,35 @@
+// Copyright 2017-2020 @polkadot/react-components authors & contributors
+// SPDX-License-Identifier: Apache-2.0
+
+import React from 'react';
+import Icon from './Icon';
```
jacogr, line 5: Import order - this (since it is local) in a new section.

packages/page-accounts/src/modals/PasswordInputNew.tsx
```diff
@@ -0,0 +1,86 @@
+// Copyright 2017-2020 @polkadot/react-components authors & contributors
+// SPDX-License-Identifier: Apache-2.0
+
+import React, { useCallback, useEffect, useState } from 'react';
+import { PasswordNew, PasswordStrength } from '@polkadot/react-components';
+import keyring from '@polkadot/ui-keyring';
+import { useTranslation } from '../translate';
+import styled from 'styled-components';
```
jacogr, line 8: Ordering.

packages/page-accounts/src/modals/InfoBox.tsx
```diff
@@ -0,0 +1,46 @@
+// Copyright 2017-2020 @polkadot/app-accounts authors & contributors
+// SPDX-License-Identifier: Apache-2.0
+
+import React from 'react';
+import { Icon } from '@polkadot/react-components';
+import styled from 'styled-components';
```
jacogr, line 6: Ordering.

packages/react-components/src/Dropdown.tsx
```diff
@@ -108,25 +107,36 @@ function BaseDropdown<Option> ({ allowAdd = false, children, className = '', def
...
-      >
-        {dropdown}
-        {children}
-      </Labelled>
-    );
+  return (
+    <div className={className}>
+      {isButton
+        ? <SUIButton.Group>{dropdown}{children}</SUIButton.Group>
+        : <Labelled
```
jacogr, line 114:
```
: (
  <Labelled ... />
)
```

packages/react-components/src/Dropdown.tsx
```diff
@@ -108,25 +107,36 @@ function BaseDropdown<Option> ({ allowAdd = false, children, className = '', def
...
 
 const Dropdown = React.memo(styled(BaseDropdown)`
+  .ui.selection.dropdown {
+    font-size: 1rem;
+    color: #000000;
+    font-weight: 400;
+    border: 1px solid #DFDFDF;
+    border-top-left-radius: 0;
+    border-bottom-left-radius: 0;
+  }
```
jacogr, line 139: Sadly this does break and conflict with what is in `styles/semantic` - prefer keeping it in there explicitly (with the theme overrides)

And additional comment, rather never use pure black, #4e4e4e is the base color (per theme) for light.

packages/react-components/src/DropdownNew.tsx
```diff
@@ -0,0 +1,190 @@
...
+          isFull={isFull}
+          isOuter
+          label={label}
+          labelExtra={labelExtra}
+          withEllipsis={withEllipsis}
+          withLabel={withLabel}
+        >
+          {dropdown}
+        </Labelled>
+
```
jacogr, line 123:
```suggestion
```

packages/react-components/src/DropdownNew.tsx
```diff
@@ -0,0 +1,190 @@
...
+    font-size: 1.15rem;
+    line-height: 1.5rem;
+    color: #000000;
+    font-weight: 400;
+    border: 1px solid #DFDFDF;
+
+    .dropdown.icon {
+      padding: 0 1.15rem;
+    }
+  }
```
jacogr, line 143: Ad per comments in the Dropdown class.

packages/react-components/src/DropdownNew.tsx
```diff
@@ -0,0 +1,190 @@
...
+        left: -2.6rem;
+        top: -1.15rem;
+        opacity: 1;
+      }
+
+      .ui--Dropdown-name {
+        margin-left: 0;
+      }
+    }
+  }
```
jacogr, line 178: Same here, overrides in semantic.ts

packages/react-components/src/Expander.tsx
```diff
@@ -56,7 +57,25 @@ function formatMeta (meta?: Meta): React.ReactNode | null {
...
+function ExpandIcon ({ className = '', isExpanded, isVisible = true }: ExpandIconProps): React.ReactElement<ExpandIconProps> {
+  return <Icon
+    className={className}
+    color={isVisible ? undefined : 'transparent'}
+    icon={
+      isExpanded
+        ? 'caret-up'
+        : 'caret-down'
+    }
+  />;
```
jacogr, line 75:
```js
return (
  <Icon ...
);
```
For the sake of readability.

packages/react-components/src/Expander.tsx
```diff
@@ -78,6 +97,8 @@ function Expander ({ children, className = '', help, helpIcon, isOpen, isPadded,
         className='ui--Expander-summary'
         onClick={toggleExpanded}
       >
+        {iconPlacement === 'left' && <ExpandIcon className='ui--Expander-icon-left'
+          isExpanded={isExpanded} />}
```
jacogr, line 101: Wth multiple attributes, attribute each on a single line with () wrapping for readability.

packages/react-components/src/Expander.tsx
```diff
@@ -91,14 +112,8 @@ function Expander ({ children, className = '', help, helpIcon, isOpen, isPadded,
...
-        <Icon
-          color={hasContent ? undefined : 'transparent'}
-          icon={
-            isExpanded
-              ? 'caret-up'
-              : 'caret-down'
-          }
-        />
+        {iconPlacement === 'right' && <ExpandIcon className='ui--Expander-icon-right'
+          isExpanded={isExpanded} />}
```
jacogr, line 116: As previous.

packages/react-components/src/Expander.tsx
```diff
@@ -172,4 +199,15 @@ export default React.memo(styled(Expander)`
       text-overflow: ellipsis;
     }
   }
+  
+  &.accounts--Creator-advanced > .ui--Expander-summary > .ui--Expander-summary-header {
```
jacogr, line 203: No app-specific overrides in here.

packages/react-components/src/Expander.tsx
```diff
@@ -172,4 +199,15 @@ export default React.memo(styled(Expander)`
...
+    font-weight: 800;
+    font-size: 10px;
+    line-height: 14px;
+    text-transform: uppercase;
+    color: #000000;
```
jacogr, line 208: As per theme, no pure black.

packages/react-components/src/InputSection.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+import React, { ReactNode } from 'react';
+import styled from 'styled-components';
+
+interface InputSectionProps {
+  children: ReactNode,
+  className?: string,
+}
+
+const InputSection = ({ children, className }: InputSectionProps) => (
```
jacogr, line 12: Always use a function, no `const` shortcuts. (consistency & mainatainability, i.e. there is just one way things look)

packages/react-components/src/InputSection.tsx
```diff
@@ -0,0 +1,20 @@
...
+  <div className={className}>{children}</div>
+);
+
+export default React.memo(styled(InputSection)`
+  & + & {
+    margin-top: 17px;
```
jacogr, line 18: No pixels, please. We use em everywhere. (So margins should align between the inputs so it can just be use anywhere)

packages/react-components/src/Labelled.tsx
```diff
@@ -147,5 +148,46 @@ export default React.memo(styled(Labelled)`
...
     }
+
+    & > .ui--Labelled-content > .ui--Static {
+      ${(props) => props.modalNew ? 'border: 1px solid #DFDFDF;' : ''}
+      ${(props) => props.modalNew ? 'background-color: #FFFFFF;' : ''}
+      ${(props) => props.modalNew ? 'color: #8B8B8B;' : ''}
+    }
+  }
+  
+  .ui--CreateAccount-new-create-modal div.ui--Modal-Column > &.ui--Labelled,  .ui--CreateAccount-new-create-modal div:not(.ui--Modal-Column) > &.ui--Labelled {
```
jacogr, line 159: Ad per crsoo package css - this will get really mess if we have the specific CSS everywhere for each modal.

packages/react-components/src/PasswordStrength.tsx
```diff
@@ -1,52 +1,79 @@
...
-    (/\d/.test(password) ? 1 : 0) +
-    // a special character
-    (/[^A-Za-z0-9]/.test(password) ? 1 : 0)
-  );
+  strengthTester.config({
+    allowPassphrases: true,
+    maxLength: 128,
+    minLength: 8,
+    minPhraseLength: 20
+  });
```
jacogr, line 23: This should be done once, not on each check. So move it to file-level.

packages/react-components/src/PasswordStrength.tsx
```diff
@@ -1,52 +1,79 @@
...
 
   return (
-    <div className={className}>
-      <div
-        className='highlight--bg'
-        style={style}
-      />
+    <div
+      className={className}
+      style={{ display: !alwaysVisible ? (passwordStrength ? 'flex' : 'none') : 'flex' }}
```
jacogr, line 39: On inline styles, don't make it re-create on each render.

packages/react-components/src/Row.tsx
```diff
@@ -83,6 +83,7 @@ export const styles = `
     font-size: 1.25em;
     padding: 0;
     margin-bottom: 0.25rem;
+
```
jacogr, line 86:
```suggestion
```

packages/react-components/src/Row.tsx
```diff
@@ -144,6 +145,27 @@ export const styles = `
   .ui--Row-tags-input {
     margin-bottom: -1.4em;
   }
+
+  .ui--CreateAccount-new-create-modal && {
```
jacogr, line 149: As per page-specific comments.

packages/react-components/src/TextArea.tsx
```diff
@@ -0,0 +1,33 @@
+// Copyright 2017-2020 @polkadot/react-components authors & contributors
+// SPDX-License-Identifier: Apache-2.0
+
+import styled, { css } from 'styled-components';
+
+interface Props {
+  withError?: boolean;
+}
+
+const TextInput = css(({ withError }: Props) => `
```
jacogr, line 10: Def. prefer none of these components. Just standard React components instead of the style-component shorthands. 

Yes, this is the way _they_ recommend, but sadly something that I find non-maintainable, nor readable.

packages/react-components/src/index.tsx
```diff
@@ -31,8 +32,10 @@ export { default as Columar } from './Columar';
 export { default as Column } from './Column';
 export { default as ConvictionDropdown } from './ConvictionDropdown';
 export { default as CopyButton } from './CopyButton';
+export { default as CopyToClipboard } from './CopyToClipboard';
```
jacogr, line 35: Really not sure about this - prefer a single consistent component for all, so would rather re-use the CopyButton. If it needs replacing, rather do it consistently everywhere.

packages/react-components/src/util/toShortAddress.ts
```diff
@@ -6,7 +6,7 @@ import { AccountId, AccountIndex, Address } from '@polkadot/types/interfaces';
 export default function toShortAddress (_address?: AccountId | AccountIndex | Address | string | null | Uint8Array): string {
   const address = (_address || '').toString();
 
-  return (address.length > 13)
-    ? `${address.slice(0, 6)}…${address.slice(-6)}`
+  return (address.length > 31)
+    ? `${address.slice(0, 15)}…${address.slice(-15)}`
```
jacogr, line 10: Rather not. I'm sure this break all the mini layouts.

jacogr (commented):
Some comments on an initial read-through (will play in the next day to see how it all fits together)

jacogr:
Some initial functional comments -

- I really like the flow and steps here, it is 100% clear
- There is no reason no to re-use the existing modals, i.e. all that is effectively added is the gray bar at the bottom (this can be applied consistently, although I'm unsure about the utility thereof)
- The labelling of the inputs are not going to work globally. I understand there is resistance to the way it is done, but if you look at for instance extrinsics (schedule a call as an example), you will see that the nesting and not having defined areas is not going to work. (As it stands there, in e.g. lists the labels are creating huge issues)
- I don't see a reason for non-standard button styling here, it looks out-of-place. If the buttons styling needs to change (which I would argue it doesn't since it ends up being "too much" on screens, it needs to be done globally)
- I really find the warning text here way too small and exceptionally difficult to read, I still have no idea what it says since I couldn't see it
- The expander "ALL CAPS" won't work globally, since it is used extensively for breakdowns in e.g. calls/events and the casing is important
- The responsiveness on the modals is worse than what is there, the columns don't seem to have the correct behavior
- Sadly we have lost the backup JSON, this is important (we have had endless issues where people don't do this an still lost access to seeds)

Overall for me personally -

- Love the flow and guidance here
- Really don't care for the styling to global items at all, it is inconsistent and cannot see what additional value it brings

krzysztof-jelski:
Superseded by #3880, let's iterate from there.

Issue #2085, opened 2019-12-23
Add option to show amount nominating in current session


joepetrowski:
The Staking > Account Actions page shows a list of all accounts that you are nominating with their self stake. It would be nice to have an option to see how much of your stake is nominating each account in the current session instead of the self stake of each nominee.

Ideally you could choose to display "total stake", "self stake", or "my current nominating stake" as options on this page.

For example, here I am nominating Federico with 30.858:
![image](https://user-images.githubusercontent.com/25483142/71342531-5961f680-255d-11ea-8b22-71f6adb50656.png)

Staking > Actions should show 30.858 under Federico here:
![image](https://user-images.githubusercontent.com/25483142/71342597-82828700-255d-11ea-9c95-bacbf81946c8.png)

Tbaut:
Do you know if this information is available in the storage state anywhere? Couldn't find it.

jacogr:
The exposure for the validator has this info. So for each validator in the list -

- query the exposure
- the `others` field contains the specific values

PR #3714, opened 2020-09-18
Show nomination balance below nominees

Tbaut:
closes https://github.com/polkadot-js/apps/issues/2085

Not sure if there is any more straightforward way, this is very convoluted and seems costly.
I couldn't test with more than 1 active nominees yet, I'll get there once my node is done compiling.


packages/page-staking/src/Actions/Account/ListNominees.tsx
```diff
@@ -18,6 +20,39 @@ function ListNominees ({ nominating, stashId }: Props): React.ReactElement<Props
...
   const { api } = useApi();
   const { nomsActive, nomsChilled, nomsInactive, nomsOver, nomsWaiting } = useInactives(stashId, nominating);
+  const sessionInfo = useCall<DeriveSessionProgress>(api.query.staking && api.derive.session?.progress);
+  const [nomBalanceMap, setNomBalanceMap] = useState<Record<string, BN>>({});
+
+  useEffect(() => {
+    let unsub: (() => void) | undefined;
+
+    if (nomsActive && sessionInfo?.activeEra) {
+      api.derive.staking.eraExposure(sessionInfo.activeEra, (v) => {
```
jacogr, line 30: Would prefer just using a `useCall` with a transform. (Since we don't need to manage state, as it stands this also doesn't take care of unmounts, which useCall does without worries)

jacogr:
Will check on mine, have some with multiple nominations.

jacogr:
Ok, it does show multiples, but it flickers on block updates and seems to change position. (no index)

packages/page-staking/src/Actions/Account/ListNominees.tsx
jacogr: Probably not with a transfers - useCall to retrieve the data and then a useMemo to combine it. (Since otherwise the active noms will be captured in the closure, so the memo will take both the era and noms in to get a result)

But yes, quite heavy atm since it subscribes/unsubscribes all the time.

Tbaut:
Thanks, will try to implement those inputs!

packages/page-staking/src/Actions/Account/ListNominees.tsx
```diff
@@ -17,6 +19,29 @@ function ListNominees ({ nominating, stashId }: Props): React.ReactElement<Props
   const { t } = useTranslation();
   const { api } = useApi();
   const { nomsActive, nomsChilled, nomsInactive, nomsOver, nomsWaiting } = useInactives(stashId, nominating);
+  const sessionInfo = useCall<DeriveSessionProgress>(api.query.staking && api.derive.session?.progress);
+  const eraExposure = useCall<DeriveEraExposure>(api.derive.staking.eraExposure, [sessionInfo?.activeEra]);
```
jacogr, line 23: This is one of those problematic ones that may break on Edgeware, it doesn't have that functionality. (Not sure if the derive itself does the check, it should...)

Tbaut:
No more flickering afaics.
> seems to change position. (no index).

Not sure I get that, the nominator ordering should not be changed by this PR since it comes from the `nmsActive` array ([here](https://github.com/polkadot-js/apps/pull/3714/files#diff-020b090863fcb404663ca3419356a561R70)). Or did I misunderstand?

packages/page-staking/src/Actions/Account/ListNominees.tsx
Tbaut: Oh thanks, checking that (and will add it in the derive if it's not there)

jacogr:
I think I meant to say "key or ordering". Lemme see the current version, couldn't quite put my finger on what was hapenning with the previous.

packages/page-staking/src/Actions/Account/ListNominees.tsx
```diff
@@ -17,6 +19,29 @@ function ListNominees ({ nominating, stashId }: Props): React.ReactElement<Props
   const { t } = useTranslation();
   const { api } = useApi();
   const { nomsActive, nomsChilled, nomsInactive, nomsOver, nomsWaiting } = useInactives(stashId, nominating);
+  const sessionInfo = useCall<DeriveSessionProgress>(api.query.staking && api.derive.session?.progress);
+  const eraExposure = useCall<DeriveEraExposure>(api.derive.staking?.eraExposure, [sessionInfo?.activeEra]);
```
jacogr, line 23: Cater for older chains.

```suggestion
  const eraExposure = useCall<DeriveEraExposure>(api.query.staking.erasStakers && api.derive.staking.eraExposure, [sessionInfo?.activeEra]);
```

packages/page-staking/src/Actions/Account/ListNominees.tsx
```diff
@@ -44,9 +69,10 @@ function ListNominees ({ nominating, stashId }: Props): React.ReactElement<Props
         >
           {nomsActive.map((nomineeId, index): React.ReactNode => (
             <AddressMini
+              balance={nomBalanceMap[nomineeId]}
               key={index}
               value={nomineeId}
-              withBalance={false}
+              withBalance
```
jacogr, line 75:
```suggestion
              withBalance={!!eraExposure}
```

packages/page-staking/src/Actions/Account/ListNominees.tsx
Tbaut: Thanks, I was digging through the api and was having trouble finding what it should be replaced by. Pretty hard to follow all these changes, derive are extremely valuable for that.

packages/page-staking/src/Actions/Account/ListNominees.tsx
```diff
@@ -17,6 +19,29 @@ function ListNominees ({ nominating, stashId }: Props): React.ReactElement<Props
...
+  const eraExposure = useCall<DeriveEraExposure>(api.derive.staking?.eraExposure, [sessionInfo?.activeEra]);
+  const nomBalanceMap: Record<string, BN> = useMemo(() => {
+    const res: Record<string, BN> = {};
+
+    if (nomsActive && eraExposure) {
+      // for every active nominee
+      nomsActive.forEach((nom) => {
+        // cycle through its nominator to find our current stash
+        eraExposure.validators?.[nom].others.some((o) => {
+          if (o.who.toString() === stashId) {
```
jacogr, line 32:
```suggestion
          if (o.who.eq(stashId)) {
```

jacogr (commented):
Some small comments to cater for eg. Edgeware

packages/page-staking/src/Actions/Account/ListNominees.tsx
Tbaut: TIL again, didn't think that would work with BN and string!

jacogr: who is AccountId, all the polkadot-js Codec stuff has .eq on them. (And in this case can compare hex, Uint8array, other AccountId and ss58)

jacogr (approved):
Thank you!

Tbaut:
Thank YOU :hand:

PR #3718, opened 2020-09-18
Update Kulupu treasury threshold

sorpaas:
In the current runtime it's changed to 0.5. It's planned to be changed back (of course, given the community agrees) to 0.8 when we get `MoreThanMajorityThenPrimeDefaultVote`.


jacogr:
Thank you!

Issue #3729, opened 2020-09-21
Warn when importing via JSON and genesisHash does not match


jacogr:
Either warn or swap to the genesis explicitly (or both)

This avoids users importing an the account disappearing (like in https://github.com/polkadot-js/apps/issues/3722, there has been a couple of reports)

Tbaut:
I'll have a look at this one unless you're working on it already.

jacogr:
No, feel free - I actually forgot to log it explicitly yesterday and just thought about it again when doing https://github.com/polkadot-js/apps/pull/3728 (Really need to stop forgetting to log stuff, big or small)

PR #3732, opened 2020-09-21
Show warning and genesis override when importing JSON with different genesis hash

Tbaut:
closes https://github.com/polkadot-js/apps/issues/3729

Suggestions welcome for the warning wording. I have to read it twice to understand it.

### DIFF
--- a/packages/page-accounts/src/modals/Import.tsx
+++ b/packages/page-accounts/src/modals/Import.tsx
@@ -174,7 +174,7 @@ function Import ({ className = '', onClose, onStatusChange }: Props): React.Reac
         </Modal.Columns>
         { differentGenesis &&
             <article className='warning'>
-              <p>{t<string>('The network from which this account was originaly generated is different than the network you are currently connected to.')}</p>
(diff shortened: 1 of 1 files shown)


packages/page-accounts/src/modals/Import.tsx
```diff
@@ -164,6 +172,11 @@ function Import ({ className = '', onClose, onStatusChange }: Props): React.Reac
             <p>{t<string>('The password previously used to encrypt this account.')}</p>
           </Modal.Column>
         </Modal.Columns>
+        { differentGenesis &&
+            <article className='warning'>
+              <p>{t<string>('The network from which this account was originaly generated is different than the network you are currently connected to.')}</p>
```
jacogr, line 177:
```suggestion
              <p>{t<string>('The network from which this account was originally generated is different than the network you are currently connected to. Once imported ensure you toggle the "allow on any network" option for the account to keep it visible on the current network.')}</p>
```

jacogr (approved):
Thank you!

PR #3739, opened 2020-09-22
Fix overflow in democracy proposal (polkadot)

Tbaut:
I had a look at the account creation, Council motions, payout.. and couldn't notice any break.
Was like this
image

Is now:
image

### DIFF
--- a/packages/react-components/src/Expander.tsx
+++ b/packages/react-components/src/Expander.tsx
@@ -145,6 +145,9 @@ export default React.memo(styled(Expander)`
       text-overflow: ellipsis;
       vertical-align: middle;
       white-space: nowrap;
+      span {
+        white-space: normal;
+      }
     }
 
     .ui--Icon {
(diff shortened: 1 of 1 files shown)


jacogr:
These are tricky. Actually for the Expander the summary should have ellipsis. However this works.

packages/react-components/src/Expander.tsx
```diff
@@ -145,6 +145,9 @@ export default React.memo(styled(Expander)`
       text-overflow: ellipsis;
       vertical-align: middle;
       white-space: nowrap;
+      span {
```
jacogr, line 148:
```suggestion

      span {
```

jacogr (approved):
lgtm.

jacogr:
We should actually in the expander itself remove everything in-between `#<weight> ... #</weight>` (it doesn't make any sense in this context) - https://github.com/polkadot-js/apps/blob/master/packages/react-components/src/Expander.tsx#L42-L56

PR #3741, opened 2020-09-22
Filter #<weight>...</weight> from metadata

jacogr:
As per https://github.com/polkadot-js/apps/pull/3739#issuecomment-696671595

Funnily enough, it actually has (mostly) been scrubbed from the code anyway.


Tbaut (approved):
working well

PR #3753, opened 2020-09-24
Fixes staking pages and console  errors

drewstone:


jacogr (approved):
Thank you! Really excited to see Edgeware upgraded, it has been a long road.

package.json
```diff
@@ -114,5 +114,8 @@
     "webpack-cli": "^3.3.12",
     "webpack-merge": "^5.1.4",
     "webpack-plugin-serve": "^1.0.1"
+  },
+  "dependencies": {
+    "@edgeware/node-types": "^2.4.5"
```
jacogr, line 119: This is actually already available in https://github.com/polkadot-js/apps/blob/master/packages/apps-config/package.json#L16 (local to apps-config so should not need another add, probably may need a bump)

packages/react-hooks/src/useStashIds.ts
```diff
@@ -7,7 +7,9 @@ import useApi from './useApi';
 import useCall from './useCall';
 
 const transformStashes = {
-  transform: (stashes: AccountId[]) => stashes.map((accountId) => accountId.toString())
+  transform: (stashes: AccountId[]) => stashes
+  	.filter((stashes: AccountId) => !!stashes)
+  	.map((accountId) => accountId.toString())
```
jacogr, line 12:
```suggestion
  transform: (stashes: AccountId[]) => stashes
    .filter((stashes: AccountId) => !!stashes)
    .map((accountId) => accountId.toString())
```

lint fix.

package.json
drewstone: Fixed!

packages/apps-config/src/api/chain/beresheet.ts
```diff
@@ -4,19 +4,15 @@
...
-import * as edgewareDefinitions from '@edgeware/node-types/dist/interfaces/definitions';
-
-const edgTypes = Object
-  .values(edgewareDefinitions)
-  .reduce((res, { types }) => ({ ...res, ...types }), {});
+import { Beresheet } from '@edgeware/node-types';
 
 export default {
-  ...edgTypes,
+  ...Beresheet,
```
jacogr, line 10:
```suggestion
  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
  // @ts-ignore
  ...Beresheet,
```

packages/apps-config/src/api/spec/edgeware.ts
```diff
@@ -4,23 +4,15 @@
...
-import * as edgewareDefinitions from '@edgeware/node-types/dist/interfaces/definitions';
-
-const edgTypes = Object
-  .values(edgewareDefinitions)
-  .reduce((res, { types }) => ({ ...res, ...types }), {});
+import { Mainnet } from '@edgeware/node-types';
 
 export default {
-  ...edgTypes,
+  ...Mainnet,
```
jacogr, line 10:
```suggestion
  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
  // @ts-ignore
  ...Mainnet,
```

packages/apps-config/src/api/chain/beresheet.ts
```diff
@@ -4,14 +4,12 @@
...
-const edgTypes = Object
-  .values(edgewareDefinitions)
-  .reduce((res, { types }) => ({ ...res, ...types }), {});
+import { Beresheet } from '@edgeware/node-types';
 
 export default {
-  ...edgTypes,
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
+  ...Beresheet,
```
jacogr, line 12:
```suggestion
  ...Beresheet.types,
```

packages/apps-config/src/api/spec/edgeware.ts
```diff
@@ -4,23 +4,17 @@
...
-const edgTypes = Object
-  .values(edgewareDefinitions)
-  .reduce((res, { types }) => ({ ...res, ...types }), {});
+import { Mainnet } from '@edgeware/node-types';
 
 export default {
-  ...edgTypes,
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
+  ...Mainnet,
```
jacogr, line 12:
```suggestion
  ...Mainnet.types,
```

packages/react-hooks/src/useStashIds.ts
```diff
@@ -7,7 +7,9 @@ import useApi from './useApi';
 import useCall from './useCall';
 
 const transformStashes = {
-  transform: (stashes: AccountId[]) => stashes.map((accountId) => accountId.toString())
+  transform: (stashes: AccountId[]) => stashes
+    .filter((stashes: AccountId) => !!stashes)
+    .map((accountId) => accountId.toString())
```
jacogr, line 12: This def seems to be an issue on Edgeware, but not on Bereesheet nor other networks.  will leave this in for now, but will need to investigate what is going wrong since something is still mismatching.

jacogr: When I look at the Miannet types o GH, they still seem old, i.e. here - https://github.com/hicommonwealth/edgeware-node-types/blob/master/src/spec/mainnet.ts#L8-L24

So believe this is actually as a result of the old types used. So would suggest switching (for now) to Bereesheet types for Miannet as well, which seems to have the latest. (And then this is unneeded)

jacogr: That also doesn't quite do the trick... (In my tests, there is some mismatch somewhere)

jacogr: Ok, there is def something weird in the keys, some of them are null (this is from `api.quesry.staking.validators.keys()` as used by the derive here) -

```
[
  "ndCyA9VQAkDpw1uDedPQm9tEsJcFdnfrHNnxom3CMW9uvop",
  "ndAMaNishd8z6YcR843jdqmUUcAgCLiFhNovXDsvmgkwdba",
  "nDqE5fV5YBhCFQCt19pS3k2uDyu5ecRhE75sKxrocHnAjEp",
  "mwpdcgb8KQYwi5pCg7K174rpSZtk8QEW95VnJQaMmeaK31H",
  "hZyPPSc8iDEdmLxHoUp9cpcZavZxkVnKGWqpYeCSt5vakHe",
  "nx7Be1NK8b2uV2bSHSxDPMJqNGAr6sc2i982shToqVHBVH7",
  "kYgxuZ7Ad7vHeyN85t6PLicyKxmbE72P46iQncJJxXTKhDU",
  "hbvLTD4NR2fKziSLeH6j6sFwycHyA8ch9VRtTxiD54JkwFC",
  "mB6GJCxuDB4WzRf6yBa83eUXn7SaMTg67i54cKrKZZ6f5KY",
  null,
...
```

jacogr: Ok, investigating, doing this -

```
const keys = await api.query.staking.validators.keys();
const stashes = await api.derive.staking.stashes();
const args = keys.map((key, index) => [key.toHex(), key.args[0], stashes[index]])

console.log(JSON.stringify(args, null, 2))
```

Which yields (trimmed) -

```
...
[
    "0x5f3e4907f716ac89b6347d15ececedca88dcde934c658227ee1dfafcd6e169030f608263d3785577a20c71fe648acea8ea13e7ce0d507cfb00c1788f1eb402b756ba7dbd4dba783a",
    "mB6GJCxuDB4WzRf6yBa83eUXn7SaMTg67i54cKrKZZ6f5KY",
    "mB6GJCxuDB4WzRf6yBa83eUXn7SaMTg67i54cKrKZZ6f5KY"
  ],
  [
    "0x5f3e4907f716ac89b6347d15ececedca88dcde934c658227ee1dfafcd6e1690311320ae78b67bc3cc132f878791c063a4085531c801f4f2f2e216ebefa7ab69d",
    null,
    null
  ],
  [
    "0x5f3e4907f716ac89b6347d15ececedca88dcde934c658227ee1dfafcd6e16903171b1b5ab8e7b8f3e4b531c4f877411b4c46bc0e45e5ce94629a17b111ba715ced098b6c0116f562",
    "ngVZ5aqPr5ja2xoTYN2zVXCnPdKiXoQMbp6Ay4DAao8zJav",
    "ngVZ5aqPr5ja2xoTYN2zVXCnPdKiXoQMbp6Ay4DAao8zJav"
  ],
...
```

The ones where `null` comes through seems to be shorter than the rest, i.e. they don't seem fully encoded. Which is weird.

jacogr: Ok, see the same when I query the entries.So I'm hoping this is a once-off. Anyway, digging through to see wtf can be done.

image

Issue #3765, opened 2020-09-25
Loaded JSON files accept any password to create account; cannot send funds from this account


laboon:
DESCRIPTION:
User can load JSON file to account with any password.  No matter what they type in, an account will be generated.

However, users will be unable to sign transactions from that account.

TO REPRODUCE:
1. Create a new account on Polkadot-JS
2. Give password 'password'
3. Forget account
4. Load account from JSON file
5. Enter password '123456'

EXPECTED BEHAVIOR:
Error

OBSERVED BEHAVIOR:
Account is loaded into polkadot-JS

----

TO REPRODUCE:
1. Try to send funds from that account
2. Enter correct password

EXPECTED BEHAVIOR:
Funds sent

OBSERVED BEHAVIOR:
Appears as though password is rejected

laboon:
After loading this JSON file with the incorrect password (`123456`), I cannot send from it even using the correct password (`password`).  The UI just freezes momentarily.

If I try to send and enter the incorrect password `123456` it is correctly seen as an invalid password.

Essentially, if you mistype the password when loading the JSON file, you will not be able to send from that account.

laboon:
Even after loading the JSON file with the correct password, I am unable to send any DOT from that account.

Tbaut:
Thanks for the details Bill, taking a look.

Tbaut:
Seeing this PR a little late, that's where things went wrong: https://github.com/polkadot-js/apps/pull/3758/files

PR #3766, opened 2020-09-25
Fix account JSON import

Tbaut:
closes https://github.com/polkadot-js/apps/issues/3765

A first shot that works. Wondering if you're still willing to use `addKeyPair` @jacogr and verify the password in a different manner, or should I refactor the import around this solution?


jacogr:
The correct way is either -

- ui-keyring saveAccount needs to decode with the password before saving (have not actually looked at to why it saves with incorrect - that is the root cause)

... or 

- decodePkcs8 upon adding, when failing, show error

Issue #3786, opened 2020-09-28
Apps shows zero balances for all accounts of Subsocial Network


siman:
Even if an account has a positive balance, apps still show zero.

As an example, here is a list of accounts with a positive balance (should be <= 2 tokens). Apps show zero for them.

```
3oYZERLKmdzNdnmzEA8wiP3Zsbj2vfuL32CgdUR2yqUbah8T
3tAgsRwfPpzf8Zik3ZerHpxxg7joA3Nnm3ritFXXZKpFBex9
3sQBzkHPTD7E8rQ2V6mMAmoQgf3LdC66J3qquJH5f4nUSyML
3qH7oSCrGxPFTY889z4fEJcTKDJuMTaQNAQtq86mTXrjTKMC
3r8bJMcMrufzH8fzi2Lk3s84oFD9QiHnqsqexu9g4kWJkMMx
3sDL3YVNEh9eoCPrMa6J88XDsa5RWVSbghSQ82Kf5YGt6sxS
5E4KMLjFReA4rQuyiFfiGyKvkqPk3DhpSxxrUt6Vvyj8qoL6
3rg8w7wUUgUfuSAn2zrEBestVokZAUUL8TFhPCuafhG54Wsu
3tFfN6S7ncBBCsEQv39F2JQY8swBM1G2d6WPKWPJdY4a121J
3thKzNVpErTN1D2iUeZccoQNTuWJAF7n3dnt9WcbtNHjhHKu
```

Where to check:
https://polkadot.js.org/apps/?rpc=wss://rpc.subsocial.network

jacogr:
You need to update the types for RefCount in your types, it ill be `u8` for non-2.0-final version. (In this case the apps UI itself has no overrides since it all gets pulled in from the external packages, for others it is done)

siman:
Not sure where should it be updated. 
1. Our Rust runtime? 
2. Our TypeScript type definitions?

siman:
Ok, I see your updated comment. Seems like it should be an easy one.

siman:
fixed 👍

PR #3789, opened 2020-09-29
Update @subsocial/types

olehmell:
Add ```RefCount: 'u8'``` in types as fix this [issue](https://github.com/polkadot-js/apps/issues/3786)


jacogr:
Thank you! I just replaced this with #3790 where `yarn` was also run to update the lockfile.

PR #3810, opened 2020-10-02
Introduce a new navigation menu's design

JustCaptcha:
image
image
image
image


krzysztof-jelski:
I really like it! Comments after playing with it a bit (haven't looked at the code yet):

- Font size in 2nd row is too big. In the design it's the same as in first row (~14px)
- Let's add hover to not-selected items in 2nd row. Could be same dash as currently in selected item, but in the color of hover from the previous design
- When I make the screen narrow, a scrollbar shows up for the 2nd row item. Not sure if that's improvement over the previous version. I mean, now at least all items are available :-) Though if we can get them to show up without a scrollbar that would be even better.

krzysztof-jelski (changes requested):
Left a comment.

jacogr:
Some functional comments -

- The sub-menu grouping doesn't make sense in this context. In this example, the governance items has no bearing on Treasury, yet it appears below.

image

Any submenu items should relate to the parent.

- Additionally with the above we now have a moving target on the main menubar with items appearing and disappearing.

- The 2-row layout combining version and blocknumber needs to be split. We have 3 rows available and actually it makes either hard to see at a glance with the single row

image

- On smaller screens the setting snow disappears, these are critical.

image

- From the same screen show above, we have a serious inconsistency where some "single menu items" setting snow appear right and others (see tech-comm above) appear in their position. I have mentioned this before as a concern - everything is dynamic. Move the settings back to the left, it is one of the app navigation items.

- Once again with the above in-min, the actual drop-icon like we have in the submenu needs to appear with the menu items so it is clear they are not single clickable items. (So in the misplaced submenu it is correct)

- The font actually looks nice on the full interface, but there is something wrong with it on the menu items, it actually looks out of place in some weird way. Not sure if it is sizing or something else, but something is "off" (maybe the caret fixes it)

- The submenu items are off-center and it detracts. It needs to be vertically aligned

image

- Not a fan of the small menus here, it looks like the topmenu items and the actual menu has different colors or a shadow is applied. 

- The padding on highlighted items are way too small in the topmenu

Overall, I'm not sure if it is currently better - it is different, has some interesting aspects atm, but overall it is UX-wise a step back from consistency and friendlyness.

packages/apps-routing/src/index.ts
```diff
@@ -17,12 +17,12 @@ import js from './js';
 import parachains from './parachains';
 import poll from './poll';
 import rpc from './rpc';
-import settings from './settings';
 import signing from './signing';
 import society from './society';
 import staking from './staking';
 import storage from './storage';
 import sudo from './sudo';
+import settings from './settings';
```
jacogr, line 25: Should be alphabetical, please move back up.

packages/apps/src/Menu/ChainInfo.tsx
```diff
@@ -33,13 +33,15 @@ function ChainInfo ({ className }: Props): React.ReactElement<Props> {
...
-          {runtimeVersion && (
-            <div className='runtimeVersion'>{t<string>('version {{version}}', { replace: { version: runtimeVersion.specVersion.toNumber() } })}</div>
-          )}
-          <BestNumber
-            className='bestNumber'
-            label='#'
-          />
+          <div className='info-text-wrapper'>
+            {runtimeVersion && (
+              <div className='runtimeVersion'>{t<string>('version {{version}}', { replace: { version: runtimeVersion.specVersion.toNumber() } })}</div>
```
jacogr, line 38: As commented, please revert.

packages/apps/src/Menu/Grouping.tsx
```diff
@@ -6,90 +6,228 @@ import { Group } from './types';
...
+        );
+      }
+    }
+  }
+
+  switch (variant) {
+    case 'active-tab': {
+      return (
+        <li className={className}>
+          {activeRoute &&
```
jacogr, line 52: Use the 

```
{activeRoute && ( 
...
)}
```

form since it is easier to see the start/end

packages/apps/src/Menu/Grouping.tsx
```diff
@@ -6,90 +6,228 @@ import { Group } from './types';
...
+            <ul className='groupMenu tab highlight--bg-light'>
+              {routes.map((route): React.ReactNode => {
+                const itemClassName = (activeRoute && activeRoute.text === route.text) ? 'active highlight--color' : '';
+
+                return <Item
+                  className={itemClassName}
+                  key={route.name}
+                  route={route}
+                />;
+              })}
```
jacogr, line 73:
```suggestion
              {routes.map((route): React.ReactNode => (
                <Item
                  className={(activeRoute && activeRoute.text === route.text) ? 'active highlight--color' : ''}
                  key={route.name}
                  route={route}
                />;
              ))}
```

Additionally that also allows us to get rid of the `return <Item ` (which should be wrapped in ()'s for readability)

packages/apps/src/Menu/Grouping.tsx
```diff
@@ -6,90 +6,228 @@ import { Group } from './types';
...
+              />
+            ))}
+          </ul>
+        </li>
+      );
+    }
+  }
 }
 
+const MenuActiveItem = styled.div`
```
jacogr, line 100: Please don't use this form in the apps codebase. Which this is the recommended form from styled-components, it is hell on maintenance. (Where it is used elsewhere, e.g. the extension, these is an active effort to remove all - some are a nightmare to see what is local, what is external, etc.)

packages/apps/src/Menu/Grouping.tsx
```diff
@@ -6,90 +6,228 @@ import { Group } from './types';
...
+    .dropdown {
+      transform: rotate(180deg);
+    }
+
+    .groupHdr::before {
+      content: '';
+      position: absolute;
+      bottom: 0;
+      left: 0;
+      background: #fff;
```
jacogr, line 214: This should come from the theme - it is different for dark and light

packages/apps/src/Menu/Grouping.tsx
```diff
@@ -6,90 +6,228 @@ import { Group } from './types';
...
+      height: 3.92rem;
+      transform: translateY(.5rem);
+      border-top-left-radius: .15rem;
+      border-top-right-radius: .15rem;
+      z-index: -1;
     }
 
     .groupHdr {
       box-shadow: 0 -${SHA_OFF} ${SHA_OFF} -${SHA_OFF} ${SHA_COL}, ${SHA_OFF} 0 ${SHA_OFF} -${SHA_OFF} ${SHA_COL}, -${SHA_OFF} 0 ${SHA_OFF} -${SHA_OFF} ${SHA_COL};
+      color: #1A1B20;
```
jacogr, line 225: This should come from the theme.

packages/apps/src/Menu/Item.tsx
```diff
@@ -18,10 +18,13 @@ const DUMMY_COUNTER = () => 0;
 
 function Item ({ className = '', isToplevel, route: { Modal, href, icon, name, text, useCounter = DUMMY_COUNTER } }: Props): React.ReactElement<Props> {
   const [isModalVisible, toggleModal] = useToggle();
+  const [colorState, setColorState] = useState('gray');
   const count = useCounter();
 
   return (
-    <li className={`${className}${count ? ' withCounter' : ''} ${isToplevel ? 'topLevel  highlight--color-contrast' : 'highlight--hover-color'}`}>
+    <li className={`${className}${count ? ' withCounter' : ''} ${isToplevel ? 'topLevel  highlight--color-contrast' : 'highlight--hover-color'}`}
+      onMouseEnter={() => setColorState('counter')}
```
jacogr, line 26: Please no inline functions. Where required, use `useCallback` - we don't want to recreate these on each render.

packages/apps/src/Menu/index.tsx
```diff
@@ -44,6 +42,8 @@ function logDisabled (route: string, message: string): void {
 }
 
 function checkVisible (name: string, { api, isApiConnected, isApiReady }: ApiProps, hasAccounts: boolean, hasSudo: boolean, { isHidden, needsAccounts, needsApi, needsSudo }: Route['display']): boolean {
+  if (name === 'settings') return false;
```
jacogr, line 45: This is very problematic since we can never move settings to a menu item. (Like is planned, since we actually want to split meta alongside)

As shows in the actual functional views as well, this is also problematic.

packages/apps/src/Menu/index.tsx
```diff
@@ -87,25 +87,29 @@ function extractGroups (routing: Routes, groupNames: Record<string, string>, api
...
 
   const externalRef = useRef(createExternals(t));
 
   const groupRef = useRef({
     accounts: t('Accounts'),
     developer: t('Developer'),
     governance: t('Governance'),
-    network: t('Network'),
-    settings: t('Settings')
+    network: t('Network')
```
jacogr, line 106: Revert change. (Keep as right)

packages/apps/src/Menu/index.tsx
```diff
@@ -87,25 +87,29 @@ function extractGroups (routing: Routes, groupNames: Record<string, string>, api
...
   });
 
   const routeRef = useRef(createRoutes(t));
 
+  const settingsRoute = extractSettingsRoute(routeRef.current);
```
jacogr, line 111: This is done on each render - rather use a ref.

packages/apps/src/Menu/index.tsx
```diff
@@ -162,9 +161,9 @@ function Menu ({ className = '' }: Props): React.ReactElement<Props> {
 export default React.memo(styled(Menu)(({ theme }: ThemeProps) => `
   align-items: center;
   display: flex;
-  justify-content: space-between;
   padding: 0;
   z-index: 220;
+  font-family: 'Nunito Sans',sans-serif;
```
jacogr, line 166: Please remove. We just use the global fond, it doesn't make sense to add these all-over the app wherever required.

packages/apps/src/Menu/index.tsx
```diff
@@ -177,39 +176,49 @@ export default React.memo(styled(Menu)(({ theme }: ThemeProps) => `
       filter: grayscale(1);
     }
 
+    .menuActive::before {
+      background: #f5f3f1;
```
jacogr, line 180: This is only for light, it should be theme-specific.

packages/react-components/src/Badge.tsx
```diff
@@ -95,15 +95,9 @@ export default React.memo(styled(Badge)`
...
-  &.counterInvertColor {
-    margin: 0 0.5rem;
-    vertical-align: middle;
-  }
-
   &.grayColor {
-    background: #eeedec !important;
-    color: #aaa9a8;
+    background: #4D4D4D !important;
+    color: #fff;
```
jacogr, line 100: Rather no full-white fonts.

packages/react-components/src/Tabs/Tab.tsx
```diff
@@ -55,20 +55,39 @@ function Tab ({ basePath, className = '', count, hasParams, index, isExact, isRo
...
+  padding: 0 1.5rem;
+  height: 100%;
+  display: flex;
+  align-items: center;
+  color: #8B8B8B !important;
 
   &.tabLinkActive {
-    border-bottom-color: #e6e6e6;
+    font-weight: bold;
+    color: #000 !important;
```
jacogr, line 72: No full-black colors. The them does have a color for these.

packages/react-components/src/Tabs/TabsSectionDelimiter.tsx
```diff
@@ -0,0 +1,31 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+import React from 'react';
+import styled from 'styled-components';
+
+interface TabsSectionDelimiterProps {
+  className?: string;
+}
+
+export const TabsSectionDelimiter = ({ className }: TabsSectionDelimiterProps): React.ReactElement => (
```
jacogr, line 11: Always export with `function` and then a memo. No shothands.

packages/react-components/src/Tabs/TabsSectionDelimiter.tsx
```diff
@@ -0,0 +1,31 @@
...
+      viewBox='0 0 17 65'
+      width='17'
+      xmlns='http://www.w3.org/2000/svg'>
+      <path d='M1 1L16 32.5L1 64'
+        stroke='#DFDFDF'/>
+    </svg>
+  </Wrapper>
+);
+
+const Wrapper = styled.div`
```
jacogr, line 24: As explained above, please don't use this form, it is a nightmare for anything that is not write-once.

packages/react-components/src/Tabs/index.tsx
```diff
@@ -38,38 +56,98 @@ function Tabs ({ basePath, className = '', hidden, isSequence, items }: Props):
...
+    () => !!sudoKey && allAccounts.some((address) => sudoKey.eq(address)),
+    [allAccounts, sudoKey]
+  );
+
+  const groupRef = useRef({
+    accounts: t('Accounts'),
+    developer: t('Developer'),
+    governance: t('Governance'),
+    network: t('Network')
+  });
```
jacogr, line 69: Ok, all these changes needs to be reverted. 

- We really don't want to maintain this in two places
- We now have the components fully tied to apps (the medium-term is to move them out)
- Functionally, this approach is confusing as mentioned

packages/react-components/src/Tabs/index.tsx
```diff
@@ -38,38 +56,98 @@ function Tabs ({ basePath, className = '', hidden, isSequence, items }: Props):
...
   background: ${theme.bgTabs};
   border-bottom: 1px solid ${theme.borderTabs};
   display: flex;
   margin: 0 -1.5rem;
-  padding: 1.5rem 1.5rem 0;
+  padding: 0 1.5rem 0;
   text-align: left;
+  height: 4.35rem;
   z-index: 1;
+  font-family: 'Nunito Sans',sans-serif;
```
jacogr, line 124: No per-component fonts, as above. It is either global, or not applied.

packages/react-components/src/Tabs/index.tsx
```diff
@@ -38,38 +56,98 @@ function Tabs ({ basePath, className = '', hidden, isSequence, items }: Props):
...
 
   &::-webkit-scrollbar {
     display: none;
     width: 0px;
   }
 `));
+
+const TabList = styled.ul`
```
jacogr, line 132: Remove this form.

jacogr (commented):
Some initial comments from an initial look-through.

packages/react-components/src/Tabs/index.tsx
```diff
@@ -107,6 +108,7 @@ function Tabs ({ basePath, className = '', hidden, isSequence, items }: Props):
           </li>
         ))}
       </ul>
+      <HelpOverlay/>
```
jacogr: Would be happy dropping this here. 

It is a good idea, but actually so outdated info-wise that it is probably dangerous. 

(There is one exception - the developer tab in settings. But there can find another solution if dropped)

jacogr:
Closing, replaced.

PR #3813, opened 2020-10-02
italian translation typo fix

roccomuso:


jacogr:
@fomod-in I would just like a check (It seems 100% sane to me, but I don't speak Italian :)

fomod-in:
> @fomod-in I would just like a check since you did 99.9% of this (It seems 100% sane to me, but I don't speak Italian :)

it's ok, let's merge

jacogr (approved):

PR #3823, opened 2020-10-06
Update Bahasa Indonesia Translation

nzjourney:
Update Bahasa Indonesia Translation.


jacogr (approved):
As always, thank you!

PR #3832, opened 2020-10-09
Style identity bar for dark theme

Tbaut:
Now looks like this (it was originally unreadable in white on white): 
image
image


jacogr (approved):
Thank you. Can't believe I didn't click through on that in dark...

PR #3172, opened 2020-07-13
Pass a set of types to the API

jacogr:

PR #3857, opened 2020-10-12
Acala upgrade testnet

aniiantt:
Acala recently upgraded its testing network. But we also wanted to be compatible with the old test network, so we added the typesBundle configuration. Since there were no typesBundle examples before, I'm not sure this is the right way to write it.

It will work after Polkadotjs solved the issue of the priority of type overrides. https://github.com/polkadot-js/api/pull/2716


jacogr:
Right approach. There is an example of typesBundle here - https://github.com/polkadot-js/apps/pull/3172 (still intend to complete it for all as soon as there is a gap)

jacogr:
The only reason why I have not merged, is the discussion in the API repo which this one relies upon. (Honestly don't want to update PRs hanging around forever since it gets in the way of doing work)

If you give the ok, will merge.

aniiantt:
> The only reason why I have not merged, is the discussion in the API repo which this one relies upon. (Honestly don't want to update PRs hanging around forever since it gets in the way of doing work)
> 
> If you give the ok, will merge.

The fact that typesSpec and typesBundle exist together can cause confusion. So I want to drop the typesSpec and just use the typesBundle. All types of Acala are specified in typesBundle.

jacogr (approved):
Thank you!

PR #3859, opened 2020-10-13
Fix contract metadata backround in dark mode

Tbaut:
Now looks:

image


jacogr (approved):
Sure I've done this before. Thank you.

PR #3880, opened 2020-10-15
Updated flow for creating accounts

krzysztof-jelski:


packages/page-accounts/src/modals/CreateConfirmation.tsx
```diff
@@ -10,10 +10,10 @@ import { isHex } from '@polkadot/util';
 import { useTranslation } from '../translate';
 
 interface Props {
-  address: string;
+  address?: string;
```
jacogr, line 13: Why won't the address and name be available? If it is not, it means that we have done something wrong and should not be in here. So it really _should_ always require it.

packages/react-api/src/Api.tsx
```diff
@@ -139,7 +147,7 @@ async function loadOnReady (api: ApiPromise, store: KeyringStore | undefined, ty
   TokenUnit.setAbbr(tokenSymbol);
 
   // finally load the keyring
-  keyring.loadAll({
+  isKeyringLoaded() || keyring.loadAll({
```
jacogr, line 150: This should always be the case. If not, we have a bug.

packages/react-components/src/AddressRow.tsx
```diff
@@ -1,20 +1,21 @@
...
 import React from 'react';
 import styled from 'styled-components';
-import { useAccountInfo } from '@polkadot/react-hooks';
-import BaseIdentityIcon from '@polkadot/react-identicon';
-
-import { toShortAddress } from './util';
 import IdentityIcon from './IdentityIcon';
 import Row, { RowProps } from './Row';
 
+import { toShortAddress } from './util';
```
jacogr, line 13: I believe a newline snuck in before this line.

packages/react-components/src/Checkbox.tsx
```diff
@@ -23,18 +23,21 @@ function Checkbox ({ className = '', isDisabled, label, onChange, value }: Props
   );
 
   return (
-    <div className={`ui--Checkbox${isDisabled ? ' isDisabled' : ''} ${className}`}>
+    <div className={`ui--Checkbox${isDisabled ? ' isDisabled' : ''} ${className}`}
+      onClick={_onClick}>
```
jacogr, line 27: Small request -

```suggestion
    <div
      className={`ui--Checkbox${isDisabled ? ' isDisabled' : ''} ${className}`}
      onClick={_onClick}
    >
```

The linting rules should pick this up, so that is weird.

packages/react-components/src/Checkbox.tsx
```diff
@@ -48,6 +51,12 @@ export default React.memo(styled(Checkbox)`
     display: inline-block;
     margin: 0 0.5rem;
     opacity: 1;
+    cursor: pointer;
+    -webkit-touch-callout: none; /* iOS Safari */
+    -webkit-user-select: none; /* Safari */
+    -moz-user-select: none; /* Old versions of Firefox */
+    -ms-user-select: none; /* Internet Explorer/Edge */
```
jacogr, line 58: Not against this, but since we are targetting newer browsers, it really should not be required.

packages/react-components/src/CopyButton.tsx
```diff
@@ -33,8 +34,13 @@ function CopyButton ({ children, className, icon = 'copy', isAddress = false, va
         message: t<string>('address copied'),
         status: 'queued'
       });
+      isMnemonic && queueAction && queueAction({
+        action: t<string>('clipboard'),
+        message: t<string>('mnemonic copied'),
+        status: 'queued'
+      });
```
jacogr, line 41: This can just be combined with the one above, just do the `?:` alongside the message.

packages/react-components/src/PasswordStrength.tsx
```diff
@@ -1,52 +1,78 @@
 // Copyright 2017-2020 @polkadot/react-components authors & contributors
 // SPDX-License-Identifier: Apache-2.0
 
+import { useTranslation } from '@polkadot/react-components/translate';
```
jacogr, line 4: `import { useTranslation } from  './translate'` - this is local to the component (so also needs to move down in position)

packages/react-components/src/PasswordStrength.tsx
```diff
@@ -1,52 +1,78 @@
...
-    (/\d/.test(password) ? 1 : 0) +
-    // a special character
-    (/[^A-Za-z0-9]/.test(password) ? 1 : 0)
-  );
+  strengthTester.config({
+    allowPassphrases: true,
+    maxLength: 128,
+    minLength: 8,
+    minPhraseLength: 20
+  });
```
jacogr, line 22: This should be done in the file as a global, no need to set it each time - once set, it is used.

packages/react-components/src/PasswordStrength.tsx
```diff
@@ -1,52 +1,78 @@
...
+  background: #ECECEC;
+}
+& .ui--Strength-bar-highlighted {
+  position: absolute;
+  top: -0.07rem;
+  height: 0.6rem;
+  width: 100%;
+  border-radius: 0.15rem;
+  background: linear-gradient(90.43deg, #FF8B00 0%, #FFBB50 112.75%);
+}
```
jacogr, line 77: Indent all.

packages/react-components/src/PasswordStrength.tsx
```diff
@@ -1,52 +1,78 @@
...
-  }
+align-items: center;
+margin-top: 0.5rem;
+margin-left: 2rem;
+font-weight: 800;
+font-size: 10px;
+line-height: 14px;
+text-transform: uppercase;
+color: #8B8B8B;
+& .ui--Strength-bar {
```
jacogr, line 61: Is the & here really needed? (also the next section below)

packages/react-components/src/PasswordStrength.tsx
```diff
@@ -1,52 +1,78 @@
...
-    height: 0.5rem;
-  }
+align-items: center;
+margin-top: 0.5rem;
+margin-left: 2rem;
+font-weight: 800;
+font-size: 10px;
+line-height: 14px;
+text-transform: uppercase;
+color: #8B8B8B;
```
jacogr, line 60: This probably will create some issues with the themes?

packages/react-components/src/PasswordStrength.tsx
```diff
@@ -1,52 +1,78 @@
...
-
-  > div {
-    height: 0.5rem;
-  }
+align-items: center;
+margin-top: 0.5rem;
+margin-left: 2rem;
+font-weight: 800;
+font-size: 10px;
+line-height: 14px;
```
jacogr, line 58: em please, so we have consistency in sizes. (We don't have this size anywhere else)

packages/react-components/src/PasswordStrength.tsx
```diff
@@ -1,52 +1,78 @@
...
-  margin-top: 0.25rem;
-  overflow: hidden;
-
-  > div {
-    height: 0.5rem;
-  }
+align-items: center;
+margin-top: 0.5rem;
+margin-left: 2rem;
+font-weight: 800;
```
jacogr, line 56: 700 max I believe (not that we use bold styles anywhere in the UI atm?)

packages/react-components/src/TextArea.tsx
```diff
@@ -0,0 +1,31 @@
+// Copyright 2017-2020 @polkadot/react-components authors & contributors
+// SPDX-License-Identifier: Apache-2.0
+
+import styled, { css } from 'styled-components';
+
+interface Props {
+  withError?: boolean;
+}
+
+const TextInput = css(({ withError }: Props) => `
```
jacogr, line 10: Really don't like these components al all. They are a nightmare to maintain if we want to add and enhance and will require a full re-do. Create an actual component.

packages/react-components/src/TextArea.tsx
```diff
@@ -0,0 +1,31 @@
...
+  background: #FAFAFA;
+  border-color: ${withError ? '#E42F2F' : '#DDE1EB'};
+  border-radius: 4px;
+  border: 1px solid #DDE1EB;
+  box-sizing: border-box;
+  color: ${withError ? '#E42F2F' : '#242529'};
```
jacogr, line 16: This needs to adjust to align with -

- other inputs
- themes

packages/react-components/src/TextAreaWithDropdown.tsx
```diff
@@ -0,0 +1,43 @@
...
+export default function TextAreaWithDropdown ({ children, id, isError, isFocused, isReadOnly, onChange, rowsCount, value }: Props): React.ReactElement<Props> {
+  const _onChange = useCallback(
+    ({ target: { value } }: React.ChangeEvent<HTMLTextAreaElement>): void => {
+      onChange && onChange(value);
+    },
+    [onChange]
+  );
+
+  return (
+    <>
```
jacogr, line 27: `<div>` so we can expand, add Labelled and classes for other component adjustments.

packages/react-components/src/TextAreaWithLabel.tsx
```diff
@@ -0,0 +1,106 @@
...
+const isCut = (key: string, isPreKeyDown: boolean): boolean =>
+  isPreKeyDown && key === KEYS.X;
+
+const isPaste = (key: string, isPreKeyDown: boolean): boolean =>
+  isPreKeyDown && key === KEYS.V;
+
+const isSelectAll = (key: string, isPreKeyDown: boolean): boolean =>
+  isPreKeyDown && key === KEYS.A;
+
+function TextAreaWithLabel ({ children, className, help, isReadOnly, label, onChange, seed }: Props): React.ReactElement<Props> {
```
jacogr, line 61: This should just be TextArea - we have created 3 components here for this now, I have no idea which to use. For all inputs we always pass `withLabel`, if false the Labelled component won't render a label.

jacogr (changes requested):
Needs adjustments and some cleanups.

packages/page-accounts/src/modals/CreateConfirmation.tsx
krzysztof-jelski: So the thing is I'm keeping the `AddressRow` on top of the Modal on each of the three steps, confirmation included. This makes more sense to me, as I have one place to configure its props and looks. but `CreateConfirmation` is currently used in `Derive` as well. So I wanted to keep it showing the `AddressRow` there, but hide it here -- so effectively optional `address` is just a way to hide the `AddressRow`. Maybe it makes more sense to change `Derive` in this PR as well in similar fashion.

packages/react-api/src/Api.tsx
krzysztof-jelski: That's a dirty hack to allow running more than one test. The keyring is one hell of a global variable, and there's no way to reset the damn thing. Also, as `Api` is initialised in a component, there's no easy way to have it's instance created once before all tests and reused. So that's the smallest non-breaking change that I could think of that was needed to be made to production code that allows to have more than 1 test relying on Api.

packages/react-components/src/CopyButton.tsx
krzysztof-jelski: Do you mean sth like:
```
      (isAddress || isMnemonic) && queueAction && queueAction({
        account: value,
        action: t<string>('clipboard'),
        message: t<string>(`${isAddress ? 'address' : 'mnemonic'} copied`),
        status: 'queued'
      });
```
?

packages/react-components/src/TextArea.tsx
```diff
@@ -0,0 +1,144 @@
...
+/>;
+
+const TextAreaInput = styled(TextInput)(({ theme }: ThemeProps) => `
+  background: ${theme.bgInput};
+  border-radius: 4px;
+  border: 1px solid #DDE1EB;
+  box-sizing: border-box;
+  color: ${theme.color};
+  display: block;
+  font-size: 1.15rem;
```
jacogr, line 38:
```suggestion
```
Would suggest not having an override, it actually looks inconsistent with the rest of the UI.

packages/react-components/src/TextArea.tsx
```diff
@@ -0,0 +1,144 @@
...
+}
+
+const TextInput = ({ withError, ...textAreaProps }: TextInputProps) => <textarea
+  {...textAreaProps}
+  className={[textAreaProps.className, withError ? 'ui-textArea-withError' : ''].join(' ')}
+/>;
+
+const TextAreaInput = styled(TextInput)(({ theme }: ThemeProps) => `
+  background: ${theme.bgInput};
+  border-radius: 4px;
```
jacogr, line 33:
```suggestion
  border-radius: 0.25rem;
```

Just aligns with what we have everywhere.

packages/react-components/src/TextArea.tsx
```diff
@@ -0,0 +1,144 @@
...
+function TextArea ({ children, className, help, isError, isReadOnly, label, onChange, seed, withLabel }: Props): React.ReactElement<Props> {
+  return (
+    <Labelled
+      className={className}
+      help={help}
+      label={label}
+      withLabel={withLabel}
+    >
+      <div className='TextAreaWithDropdown'>
+        <TextAreaWithDropdown
```
jacogr, line 102: Really not sure why we have these amount of components - we can just flatten into a single? I struggle to follow this with the 3 layers as it stands.

packages/react-components/src/CopyButton.tsx
jacogr: I missed this question originally, sorry about that. Yes.

packages/react-components/src/PasswordStrength.tsx
```diff
@@ -1,52 +1,80 @@
...
-
-  > div {
-    height: 0.5rem;
+  font-size: 1rem;
+  text-transform: uppercase;
+  color: ${theme.colorLabel};
+  .ui--Strength-bar {
+    position: relative;
+    height: 0.6rem;
+    width: 14rem;
```
jacogr, line 66: This should take up the full width, same size as the inputs.

jacogr (commented):
Couple of small comments from playing through.

packages/page-accounts/src/modals/Create.tsx
```diff
@@ -262,147 +278,172 @@ function Create ({ className = '', onClose, onStatusChange, seed: propsSeed, typ
...
+                  isMnemonic
+                  value={seed}
+                />
+              </TextArea>
+            </Modal.Column>
+            <Modal.Column>
+              <p>{t<string>('The secret seed value for this account. Ensure that you keep this in a safe place, with access to the seed you can re-create the account.')}</p>
+            </Modal.Column>
+          </Modal.Columns>
+          <ExternalWarning />
```
jacogr, line 329: I would suggest moving this last, it looks a bit out of place especially when advanced is used. (Since it pertains to the overall account, not only the seed)

Would, alongside the move to the last component, also have it available on both steps.

packages/page-accounts/src/modals/Create.tsx
```diff
@@ -262,147 +278,172 @@ function Create ({ className = '', onClose, onStatusChange, seed: propsSeed, typ
...
-        )
-        : (
-          <Modal.Content>
-            <Modal.Columns>
-              <Modal.Column>
-                <AddressRow
-                  defaultName={name}
-                  noDefaultNameOpacity
-                  value={isSeedValid ? address : ''}
+      <Modal.Content>
```
jacogr, line 281: It would be good to have something like "step 1 or 2" in the header.

packages/page-accounts/src/modals/Create.tsx
```diff
@@ -262,147 +278,172 @@ function Create ({ className = '', onClose, onStatusChange, seed: propsSeed, typ
...
-                  <p>{t<string>('The derivation path allows you to create different accounts from the same base mnemonic.')}</p>
-                </Modal.Column>
-              </Modal.Columns>
-            </Expander>
-          </Modal.Content>
-        )
-      }
+          </Expander>
+          <Checkbox
+            label={<>{t<string>('I have saved my mnemonic seed safely')}</>}
```
jacogr, line 383: This checkbox doesn't "pop" - not your fault, since it applies label, but it is not quite obvious enough.

packages/page-accounts/src/modals/Create.tsx
krzysztof-jelski: Thought about it, just wasn't sure how well it will play with some of the translations,`ar` in particular. First, it's RTL. Second, some of the translations with numbers do not include number in translation (eg. `"1 day": "يوم واحد"`). So if I did sth like ${t('header') (1/2)}, would it still make sense in arabic?

packages/page-accounts/src/modals/Create.tsx
krzysztof-jelski: You mean doesn't catch user's attention visually? Or do you mean the styling when selected?

packages/react-components/src/TextArea.tsx
krzysztof-jelski: OK, done. I'm pretty sure it will make it harder to reuse TextArea in other places, but for now you're right, it simplifies the code greatly.

jacogr: We really should just use the `TextArea` directly from react-components. I'm sure there are some gaps, but will see them when using.

packages/page-accounts/src/modals/Create.tsx
jacogr: Arabic is currently a problem child. :( It is actually wrong in some places (visibly so). So my view on the header is changing the string to -

`t<string>('add {{step}}/2', { replace: { step } })`

(Will just need to make a sweep to add these to the existing translation files, but shouldn't break - happy to NOT do it is Arabic, I struggle with my tools to even fix the obviously wrong lines)

packages/page-accounts/src/modals/Create.tsx
jacogr: It is not obvious that it needs to be checked at all. I even had to look around before figuring out - the actions should be more in the face, that component certainly doesn't do it. Possibly -

- centering it (it is not consistent, but would allow for more visual draw possibly?)
- overriding the color on that one to be "normal"

So both are inconsistent, however unless we come up with something else, cannot see how to just draw the user's attention to actually making an action.

jacogr (approved):
Brilliant! Thank you very much. Tis is a huge improvement.

I'm still a bit concerned with the checkbox, the position is great, but the checkbox we have is just horrible in terms of visibility here. Will need to figure out something, but it really should not hold up this PR.

PR #3881, opened 2020-10-15
Add content-container

JustCaptcha:
image
image
image
image
image


packages/react-components/src/styles/components.ts
```diff
@@ -118,4 +118,10 @@ export default css`
     z-index: 1002;
     max-width: 300px;
   }
+
+  .content-container {
+    width: 100%;
+    max-width: 1088px;
```
jacogr, line 124: Too small. This breaks democracy and things like tips. For optimal viewing pleasure 1792px. So this should only apply on exceptionally large screens.

Additionally, all the apps _should_ be wrapped in `<main>...</main>` so it can possibly apply at that level.

So overall -

- probably don't need to apply it anywhere explicitly, just add to main (unless it doesn't work) with center
- keep it to Jaco's screen default of 1792px (1800px screen), anything less than that and we start dropping columns

jacogr: image

JustCaptcha: If add .content-container class to main, then tabs will be broken.
image

packages/page-contracts/src/index.tsx
```diff
@@ -74,69 +84,62 @@ function ContractsApp ({ basePath, className = '' }: Props): React.ReactElement<
   );
 
   return (
-    <main className={`contracts--App ${className}`}>
-      <HelpOverlay md={introMd as string} />
-      <header>
-        <Tabs
-          basePath={basePath}
-          items={itemsRef.current}
+    <React.Fragment>
```
jacogr, line 87: Always prefer a simple `<>` when using Fragment. 

Having said that, for all these you would generally want a `<div className={className}>` everywhere since some do have styles and consistency needs to be applied.

krzysztof-jelski: Can you provide some use cases or examples of `className` usage on app level? Couldn't find it.

As we want to render `main` in the `apps/src/Content` now, I thought we could possibly just add `<main className={`App--${name}`}>` there.

krzysztof-jelski:
Superseded by #4170

Issue #3833, opened 2020-10-09
Address book, Event calendar in "apps-routing" don't show up in i18n translation


Tbaut:
I had a look but couldn't make sense out of it.
Those files ([here](https://github.com/polkadot-js/apps/blob/master/packages/apps-routing/src/addresses.ts)) in `apps-routing` all look the same. The scanner seems to see the ns `apps-routing`, yet new strings aren't picked up when running `yarn build:i18n`.

Tbaut:
That's solved :rocket:

jacogr:
Re-opening, since it has to be manually added... something is _very_ weird and can't figure it out.

Tbaut:
Ah damn, I've also spent a lot of time digging this before opening the issue :(

PR #3887, opened 2020-10-16
Fix i18n generation

jacogr:
Closes https://github.com/polkadot-js/apps/issues/3833

Seems it was the `t<string>(...)` vs `t(..)` in normal ts files. Checked by removing contents and then running `yarn build:i18n`.

... and we have some additional keys in here now


i18next-scanner.config.js
```diff
@@ -10,7 +10,7 @@ const findPackages = require('./scripts/findPackages');
 function transform (file, enc, done) {
   const { ext } = path.parse(file.path);
 
-  if (ext === '.tsx') {
+  if (['.ts', '.tsx'].includes(ext)) {
```
jacogr, line 13: This is the actual fix.

PR #3903, opened 2020-10-20
Update translation.json

maestro779:
revisions to the existing sentences, making the translation completion rate close to 99%


jacogr (approved):
Thanks an absolute million for this. This is brilliant.

PR #3911, opened 2020-10-22
Added hanonycash Network logos & endpoints

Hanonycash:
Added endpoints, logos and spec of hanonycash Network

http://hanonycash.com


packages/apps-config/src/settings/endpoints.ts
```diff
@@ -122,6 +122,13 @@ function createLiveNetworks (t: TFunction): LinkOption[] {
       textBy: t('rpc.hosted.by', 'hosted by {{host}}', { ns: 'apps-config', replace: { host: 'Commonwealth Labs' } }),
       value: 'wss://mainnet1.edgewa.re'
     },
+    {
+      dnslink: 'hanonycash',
```
jacogr, line 126:
```suggestion
```

For now removing it until I've managed to cleanup the deployment of these and then it should be applied consistently for all. (it is an option for the dotapps.io domain which is hosted on IPFS)

PR #3920, opened 2020-10-22
Adjust contract Abi formatters for Balance

jacogr:


packages/react-components/src/util/getContractAbi.ts
```diff
@@ -19,6 +19,8 @@ export default function getContractAbi (address: string | null): Abi | null {
   try {
     const data = meta.contract && JSON.parse(meta.contract.abi) as AnyJson;
 
+    console.error(api.registry.getChainProperties()?.tokenDecimals.toHuman(), api.registry.getChainProperties()?.ss58Format.toHuman());
+
```
jacogr, line 23:
```suggestion
```

PR #3931, opened 2020-10-25
corrected internal link

NukeManDan:
seems ui/general is depricated for ui/colors.ts


jacogr (approved):
Thank you! Believe I renamed it to ensure consistency and ease-of-use. However missed a rather critical link.

PR #3938, opened 2020-10-28
Fix an important translation error in Chinese

starit:
here `sudo` has been translated to `undo`, `撤销` means `cancel` or `undo` in Chinese and nothing to do with `sudo`...


jacogr (approved):
Thank you!

Issue #3927, opened 2020-10-24
Cleanup style theme expansion


jacogr:
Currently we use this form `background: ${({ theme }) => theme.bgPage}` is all base component styles, e.g. here https://github.com/polkadot-js/apps/blob/master/packages/react-components/src/styles/index.ts#L216

Clean it up to have the closure at the top, e.g. `createGlobalStyle<Props & ThemeProps>((props) => `...`)`

This applies to all files in the linked area. (the included ones may be need a tweak, to be called with the props), I reverted it elsewhere (for now) since it steps on open PRs.

PR #3943, opened 2020-10-28
Cleanup style expansion

Tbaut:
closes https://github.com/polkadot-js/apps/issues/3927


jacogr (commented):
lgtm. Just holding off a little bit since it will conflict with the fonts PR.

PR #3946, opened 2020-10-29
Fix refcount for Edgeware to fix balance displays

drewstone:
Most balances look wack! I almost had a heart-attack lol


jacogr:
Nice. Happy that you are up to date. Will merge, but since it is the Substrate default could just remove. Will do so in a follow-up.

drewstone:
Cool, yea we'll have a new types update soon and remove it too

PR #3958, opened 2020-11-02
Electron app housekeeping

krzysztof-jelski:
Couple of small changes to the Electron app.
* Upgrade Electron deps versions
* Add additional hash for the index.html top script (I guess the whitespaces are different when packaged)
* Externalise packaging configuration

### DIFF
--- a/packages/apps-electron/src/electron/contentSecurityPolicy.ts
+++ b/packages/apps-electron/src/electron/contentSecurityPolicy.ts
@@ -11,12 +11,13 @@ export const setupContentSecurityPolicy = (_: string): void => {
         ...details.responseHeaders,
         // unsafe-eval is needed for the WASM content - same as the extension
         // script hash here is for the window.top script (not technically needed)
-        'Content-Security-Policy': ["default-src 'self' 'unsafe-eval' 'sha256-02/ejyoV/iwRdJ4NAsxjzF6WVUtLMPM6Nv96EbAm6u8=';" +
+        'Content-Security-Policy': ["default-src 'self';" +
         " style-src-elem 'self' https://fonts.googleapis.com/css 'unsafe-inline';" +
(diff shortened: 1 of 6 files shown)


packages/apps-electron/src/electron/contentSecurityPolicy.ts
```diff
@@ -9,14 +9,15 @@ export const setupContentSecurityPolicy = (_: string): void => {
...
+        'Content-Security-Policy': ["default-src 'self';" +
         " style-src-elem 'self' https://fonts.googleapis.com/css 'unsafe-inline';" +
         " font-src data: 'self' https://fonts.gstatic.com;" +
         " style-src 'unsafe-inline';" +
         " connect-src 'self' wss:;" +
-        " img-src 'self' data:"]
+        " img-src 'self' data:;" +
+        // unsafe-eval is needed for the WASM content - same as the extension
+        // script hashes here are for the window.top script (not technically needed)
+        " script-src 'self' 'unsafe-eval' 'sha256-02/ejyoV/iwRdJ4NAsxjzF6WVUtLMPM6Nv96EbAm6u8=' 'sha256-wW/WsLudCDaPo/ibpeK0KslHqYpCzcAKNFxFBXwCHJg='"]
```
jacogr, line 20: Where does this magic hash come from? I have been wondering. And now that there is a change, wondering what cause that change.

krzysztof-jelski: So what I did, I ran the packaged app through https://github.com/bytedance/debugtron and inspected the renderer console. It complained about the script not being executed due to the policy. But when I changed the hash, it complained when running through `start:electron`. So what is happening here is probably whitespaces are different in the builds. It's a bit of guessing, because I really can't tell the difference. Anyhow, adding 2 hashes make it not complain.

jacogr (approved):

Issue #3519, opened 2020-08-25
Support Ethereum PairType


crystalin:
CRYPTOS needs to include
```
{
  info: 'ethereum',
  text: 'Ethereum (BTC/ETH compatible)',
  value: 'ethereum'
}
```
(and ECDSA needs to remove the `(BTC/ETH compatible)`

crystalin:
It also needs to be supported there:
https://github.com/polkadot-js/apps/blob/562e11dc06c656da92ec957cf9de063ebb507cb6/packages/react-components/src/util/toAddress.ts#L16

This is failing when using the ethereum type.

jacogr:
Closing thanks to the tireless effort by @joelamouche

PR #3959, opened 2020-11-02
WIP: Add FE modifications to support Moonbeam

joelamouche:
Part of #3519


packages/react-components/src/util/toAddress.ts
```diff
@@ -12,9 +13,13 @@ export default function toAddress (value?: string | Uint8Array | null, allowIndi
         ? hexToU8a(value)
         : keyring.decodeAddress(value);
 
-      assert(allowIndices || u8a.length === 32, 'AccountIndex values not allowed');
+      assert(allowIndices || u8a.length === 32 || u8a.length === 20, 'AccountIndex values not allowed');
```
jacogr, line 16: This is one of the new changes after my original PR.

packages/react-api/src/Api.tsx
```diff
@@ -155,13 +155,15 @@ async function loadOnReady (api: ApiPromise, store: KeyringStore | undefined, ty
 
   setDeriveCache(api.genesisHash.toHex(), deriveMapCache);
 
+  const isEthereum:boolean = (api.runtimeVersion.specName.eq('node-moonbeam') || api.runtimeVersion.specName.eq('moonbase-alphanet'));
```
jacogr, line 158: This is not quite scalable. It was already horrible with Käufer the one check, but think it will get completely out of control with multiples. 

This is ok as a „get it going“ (for this PR and now) but probably need to move something like this into api-config, where all the chain/spec config sits.

jacogr:
So this is tricky, not even sure what is all affected. So I guess my original approach was to run a local node, crest an account and get that bit working. (Which is create and then the subsequent display - here already not sure about the funded dev accounts for the chain, ie. Alice, Bob, etc. for normal Substrate dev chains)

Next up would be using, selecting from the Dropdown menu and see what breaks in terms of transfer. 

I wish I could say I had a checklist for this one, but the reality is that is has been built over almost 3 years with one type of id in mind. So there may well holes I cannot even think about off-hand.

packages/react-api/src/Api.tsx
joelamouche: Ok sounds good. Should I create an issue for this in your repo? I can assign it to myself if you want

packages/react-components/src/util/toAddress.ts
joelamouche: Where can I find those changes? Should I just rebase to master?

packages/react-api/src/Api.tsx
jacogr: It is fine, I can also do so. (Should be a quick one, just don’t want to lose track)

packages/react-components/src/util/toAddress.ts
jacogr: Yes. Would just rebase on master. This may be the only === 32 Check here, but a search for 32 should throw out the rest. 

Previous to this we never actually checked the sizes, however doing so does create some issues in chains where the config may be different. (Hence being explicit)

joelamouche:
@jacogr So after rebasing I see

`systemName === 'node-moonbeam'
                    ? uiSettings.availableCryptosEth
                    : uiSettings.availableCryptos`

Shouldn't we use the isEthereum flow like in your PR?

jacogr:
Agreed. That is much cleaner.

jacogr:
It may be easier to just apply the required changes (they are not that many), fixing conflicts is never fun.

joelamouche:
closed in favor of https://github.com/polkadot-js/apps/pull/3961/files

PR #3961, opened 2020-11-03
Support ethereum addresses for Moonbeam

joelamouche:
Finish https://github.com/polkadot-js/apps/pull/3959 and #3519 

Use isEthereum from the api to determine wether to use Ethereum addresses with 20 length instead of 32

### Accounts
- [x] Create Eth address account
  - [x] Using the 'add account' button
  - [x] Using the vanity generator
- [x] See Eth Address in the account list
  - [x] Working identicon with paste
  - [x] See Balance
- [x] Transfer from one account to another
  - [x] With MetaMask
  - [x] With Interface (account page)
  - [x] With Interface (address page)
  - [ ] With Interface (extrinsic page)

### Addresses
- [x] Add an Eth address
- [x] Display Address correctly
- [x] Reload addresses correctly

### Chain State
-  [x] Display AccountInfo for an Eth address (Moon-119)

### Verify
-  [ ] postponed to #4006 

Should also tackle https://github.com/polkadot-js/apps/issues/3986, https://github.com/polkadot-js/apps/issues/4000


joelamouche:
Minor issues:

- Clicking on Ethereum accounts should copy address

- icons don’t display for the ethereum accounts in the chainstate page (but also on account pages)

The two latter issues are due to the utilization of BaseIdentityIcon from @polkadot/react-identicon, which only works with 256 addreses.

Solution:

- create our own identicon component with https://github.com/MyCryptoHQ/ethereum-blockies-base64

- get permission from polkadot to modify their react-identicon lib

What do you think @jacogr ?

jacogr:
For the icons, I believe it probably makes sense to use blockies, since it is the Eth (well, discounting Metamask) default.

packages/page-addresses/src/modals/Create.tsx
```diff
@@ -45,7 +45,7 @@ function Create ({ onClose, onStatusChange }: Props): React.ReactElement<Props>
 
         address = keyring.encodeAddress(publicKey);
         isAddressValid = keyring.isAvailable(address);
-        isPublicKey = publicKey.length === 32;
+        isPublicKey = isEthereum? publicKey.length === 20:publicKey.length === 32;
```
crystalin, line 48: I believe the Ethereum public key is 32 bytes (64 in reality but only 32 are useful/needed). It then gets converted to 20 bytes to retrieve the address (AccountId)

packages/react-api/src/Api.tsx
```diff
@@ -162,12 +162,15 @@ async function loadOnReady (api: ApiPromise, injectedPromise: Promise<InjectedEx
 
   setDeriveCache(api.genesisHash.toHex(), deriveMapCache);
 
+  const isEthereum:boolean = (api.runtimeVersion.specName.eq('node-moonbeam') || api.runtimeVersion.specName.eq('moonbase-alphanet'));
```
crystalin, line 165: Currently, there is also the standalone version that is using `moonbeam-standalone`

packages/page-addresses/src/modals/Create.tsx
jacogr: For Moonbeam (ethereum type) I believe we actually have the full 65 bytes uncompressed.

jacogr:
So on the icons, if you pass `theme='ethereum', it will render blockies now. We probably want an `isEthereum` check here - https://github.com/polkadot-js/apps/blob/master/packages/react-components/src/IdentityIcon/index.tsx#L26 (will figure out the selection at some point, probably just hide it - having it working with an "ignore settings when Eth" is more important so it at least shows)

packages/react-components/src/IdentityIcon/index.tsx
```diff
@@ -44,18 +44,17 @@ function IdentityIcon ({ className = '', prefix, size = 24, theme, value }: Prop
...
-  );
+      return (
+        <BaseIdentityIcon
+          Custom={Custom}
+          className={className}
+          onCopy={_onCopy}
+          prefix={prefix}
+          size={size}
+          theme={isEthereum? 'ethereum' : thisTheme as 'substrate'}
+          value={value.length==42? value+"000000":value}
```
jacogr, line 55: Does it still break here? Should just "work"...

joelamouche: It does work  without this logic. Good catch!

jacogr: ... if not, there would be a bug in the IdentityIcon piece. (It was tested, however only in displaying normal longer addresses... so worried there may be an issue which I didn't catch)

joelamouche: The ethereum addresses still show as blank  identicon, but the sr25519 addresses do display

jacogr: Grrrr.... Ok, need to track that one, it should "just work". Ok, not sure how I managed to break it. https://github.com/polkadot-js/ui/blob/master/packages/react-identicon/src/Identicon.tsx#L72-L78

joelamouche: I'm getting a lot of 'DRR: Error: Unable to retrieve accountId' error; not sure if it's related.
Also I don't know if this helps but this is the code at /apps/packages/react-components/src/util/toAddress.ts

`export default function toAddress (value?: string | Uint8Array | null, allowIndices = false): string | undefined {
  if (value) {
    try {
      const u8a = isHex(value)
        ? hexToU8a(value)
        : keyring.decodeAddress(value);

      assert(allowIndices || u8a.length === 32 || u8a.length === 20, 'AccountIndex values not allowed');

      if (u8a.length === 20) {
        return ethereumEncode(u8a);
      } else {
        return keyring.encodeAddress(u8a);
      }
    } catch (error) {
      // noop, undefined return indicates invalid/transient
    }
  }

  return undefined;
}`

jacogr: Ahhh... ok... that is a completely separate issue, which I did see initially when I played, but slipped my mind. It comes from here - https://github.com/polkadot-js/apps/blob/master/packages/react-hooks/src/useAccountInfo.ts

Basically on the input it tries to retrieve any relevant info for that address. So on the `AddressMini` component it uses the address returned from it to actually display the icon. so if the above fails (which is non-Eth aware), the actual address comes back as empty.

jacogr: So not sure which one, but some component is playing havoc :)

joelamouche: I think the problem is coming from this line (https://github.com/polkadot-js/apps/blob/master/packages/react-components/src/AccountName.tsx#L134) indeed. The info is returned only for non-ethereum addresses

Isn't this where the accountId is derived:
https://github.com/polkadot-js/api/blob/master/packages/api-derive/src/accounts/accountId.ts
?
Because it looks like this could be the cause of the problem

jacogr: Yes, that is def. an issue. The check should be swapped around in that case to rather check for index (<= 8), AccountId can have an override.

joelamouche: So how do you want to proceed to solve this issue? This is in the @polkadot/api repo right?

jacogr: Will update the apps deps as soon as api CI is done.

joelamouche: thank you

joelamouche:
https://github.com/polkadot-js/apps/issues/3967

joelamouche:
@crystalin Are we suppose to see 100 on the balance in the account page or zero?

jacogr:
Wondering about those - not sure which addresses Moonbeam uses for test accounts. If they are not the standard `//Alice` there will be adjustments. Additionally, for the default keyring those are sr25519, so not sure at all they are compatible keyring-wise. 

Well, it may be that the dev accounts are useless in that case, so we should not show them, i.e. in `isEthereum` mode pass `!isDevelopment` to the keyring. The apps UI can do magic and add specific accounts as needed for those cases

packages/react-components/src/IdentityIcon/index.tsx
joelamouche: @jacogr Updating the dependencies did solve the identicon problem.
However, now that I added an ethereum address to my address book I have a problem when loading the app.
At `/apps/packages/react-api/src/Api.tsx` at the  `// finally load the keyring` step, I have `Error: Expected a valid key to convert, with length 1, 2, 4, 8, 32, 33` even though I modified the code this way: ` isKeyringLoaded() || keyring.loadAll({
    genesisHash: api.genesisHash,
    isDevelopment,
    ss58Format,
    store,
    type: isEthereum? 'ethereum': 'ed25519'
  }, injectedAccounts);`

Also, I noticed that this file is using '@polkadot/ui-keyring' but this package is not in the dependancies of react-api

joelamouche:
@jacogr Sounds good I will pass in !isDevelopment to the key ring. Is there anyway we could generate test eth addresses?
Also, I was curious: in the lint command, why not run --fix?

jacogr:
So after passing in `!isDevelopment`, we can look at what moonbeam uses and then generate accounts the same way. Assuming the derivation is the same the `//Alice` with the `ethereum` type should work. (There may be issues, it has not been compared against a node)

`--fix` sadly does create more problem it solves. On simple things it works, but then it creates other formatting issues. It actually may be better now with the eslint-typescript usage (here has been huge improvements there), however have had endless issues with this in the past. (Granted this was via a separate tslint and early versions of typescript-eslint as it got started)

crystalin:
Thanks for your work.
@jacogr Currently, the StageNet that @joelamouche is referencing (not our AlphaNet), is using dedicated account, not Alice or similar ones. If uses Alice account, will PolkadotJs use the Ethereum Account for Alice private key or because of the keyring it will always use the sr25519 ?

@joelamouche You can check the accounts using metamask or web3 sdk, those should match (and you also need to change the `refCount: "u8"` in the types

jacogr:
So the `//Alice` derivation is only added on `--dev` (i.e. when `isDevelopment` is true). However, those are sr25519 accounts. So basically for pre-funded dev accounts (as in genesis), we would need a manual add.

packages/react-components/src/IdentityIcon/index.tsx
joelamouche: @jacogr reopened

joelamouche:
@jacogr 

>So the //Alice derivation is only added on --dev (i.e. when isDevelopment is true).

Can I modify that script to load addresses from MetaMask?

Also, I noticed that the "add account" button in the accounts page only allows to add sr25519 addresses. Should I modify this modal?

jacogr:
Not sure about the Metamask question.

For `isEthereum`, yes, would default to Ethereum.

joelamouche:
@jacogr thanks for the answers

did you get the opportunity to check out the keyring issue? https://github.com/polkadot-js/apps/pull/3961#discussion_r517348815

jacogr:
Keyring with dev always adds the default Substrate dev keys. If we wish to override, it sits in apps. (Really cannot do chain-specific stuff at that level(

joelamouche:
@jacogr I'm not sure we are refering to the same thing. Im talking about this: https://github.com/polkadot-js/apps/pull/3961#discussion_r517348815

The issue here is that the keyring loads addresses from the store and if I saved an ethereum address in the store it throws an error

jacogr:
Ahh. No Idea just looking. Needs a change somewhere.

packages/react-components/src/IdentityIcon/index.tsx
joelamouche: created an issue https://github.com/polkadot-js/apps/issues/3983

packages/apps-config/src/api/spec/moonbeam.ts
```diff
@@ -1,17 +1,77 @@
 // Copyright 2017-2020 @polkadot/apps-config authors & contributors
-// SPDX-License-Identifier: Apache-2.0
+// SPDX-License-Identifier": Apache-2.0
```
JoshOrndorff: Must be an accident, right?

jacogr: Lint has got it - https://github.com/polkadot-js/apps/pull/3961/checks?check_run_id=1365083861#step:3:182

joelamouche: good catch :)

joelamouche:
@crystalin It is normal that there is a 3 digit difference between the tokens on our testnet and the regular balance? Gerald show with 100,000 tokens

jacogr:
Quite possible. If the chain spec does not define the "tokenDecimals", the UI will apply a default automatically. Do a RPC call to `system_properties` on both chains.

joelamouche:
@jacogr How do I specify tokenDecimals? Is it in the same file as the type declaration?

jacogr:
No, it gets retrieved via the `system_properties` RPC call - it is fixed in the chain spec under the properties section.

joelamouche:
https://github.com/PureStake/moonbeam/pull/99

PR to fix decimals

jacogr:
So I was looking at the keyring... and I have no idea if and how it would work. Granted, it has been a long day here... 

I've seen that transfers and adding accounts work, so does this work -

- create an account (call it "N")
- transfer funds into "N"
- transfer funds out of "N"

Then, reload the page and try to transfer funds out of "N" again.

joelamouche:
> so does this work 

yes this works. 

Btw we are trying to get this merged by tomorrow so I will be postponing the 'verify' fix to another PR

packages/react-params/src/initValue.ts
```diff
@@ -89,7 +89,7 @@ export default function getInitValue (def: TypeDef): unknown {
       return createType(registry, 'H512');
 
     case 'H160':
-        return createType(registry, 'H160');
+      return createType(registry, 'H160');
```
jacogr, line 92:
```suggestion
      return registry.createType('H160');
```

(I have the others changed in a different PR, for some reason the `createType(registry, ...` form is still in some places in apps)

joelamouche: ty

joelamouche:
@jacogr @crystalin @JoshOrndorff  ready for review

packages/apps-config/src/api/rpc/moonbeam.ts
```diff
@@ -0,0 +1,66 @@
+// Copyright 2017-2020 @polkadot/apps-config authors & contributors
+// SPDX-License-Identifier: Apache-2.0
+
+import { DefinitionRpc } from '@polkadot/types/types';
+
+export default function ():Record<string, Record<string, DefinitionRpc>> {
```
jacogr, line 6:
```suggestion
export default function (): Record<string, Record<string, DefinitionRpc>> {
```

packages/apps-config/src/api/rpc/moonbeam.ts
```diff
@@ -0,0 +1,66 @@
...
+  const dummyDescription = {
```
jacogr, line 7: If these contain nothing, i.e. is not usable from the RPC page, we should leave it out completely. The console warning is harmless and doesn't affect operation. However, with inaccurate information here, it means it is available from the RPC page and calls won't work, so it is the worse option.

packages/apps-config/src/api/spec/moonbeam.ts
```diff
@@ -0,0 +1,77 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+// structs need to be in order
+/* eslint-disable sort-keys */
+
+export default {
+  AccountId: 'EthereumAccountId',
+  Address: 'AccountId',
+  Balance: 'u128',
+  RefCount: 'u8',
```
jacogr, line 11: We actually may want to use typesBundle, i.e. versioned types. The reason is then we can do the type overrides (such as this one which will most probably change), based on a version range. (In general, since we have support for it now, it is a better option anyway, since the changes can be made even before the chain gets upgraded and will kick in on upgrade)

packages/page-accounts/src/Vanity/index.tsx
```diff
@@ -8,7 +8,7 @@ import { GeneratorMatches, GeneratorMatch, GeneratorResult } from '@polkadot/van
 import React, { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import styled from 'styled-components';
 import { Button, Dropdown, Input, Table } from '@polkadot/react-components';
-import { useIsMountedRef } from '@polkadot/react-hooks';
+import { useIsMountedRef, useApi } from '@polkadot/react-hooks';
```
jacogr, line 11:
```suggestion
import { useApi, useIsMountedRef } from '@polkadot/react-hooks';
```

packages/react-api/src/Api.tsx
```diff
@@ -121,14 +121,14 @@ async function retrieve (api: ApiPromise, injectedPromise: Promise<InjectedExten
...
   registry.register(types);
-
   const { injectedAccounts, properties, systemChain, systemChainType, systemName, systemVersion } = await retrieve(api, injectedPromise);
   const ss58Format = uiSettings.prefix === -1
     ? properties.ss58Format.unwrapOr(DEFAULT_SS58).toNumber()
     : uiSettings.prefix;
   const tokenSymbol = properties.tokenSymbol.unwrapOr(undefined)?.toString();
   const tokenDecimals = properties.tokenDecimals.unwrapOr(DEFAULT_DECIMALS).toNumber();
   const isDevelopment = systemChainType.isDevelopment || systemChainType.isLocal || isTestChain(systemChain);
+  const isEthereum:boolean = (api.runtimeVersion.specName.eq('node-moonbeam') || api.runtimeVersion.specName.eq('moonbase-alphanet') || api.runtimeVersion.specName.eq('moonbeam-standalone'));
```
jacogr, line 131:
```suggestion
  const isEthereum: boolean = (api.runtimeVersion.specName.eq('node-moonbeam') || api.runtimeVersion.specName.eq('moonbase-alphanet') || api.runtimeVersion.specName.eq('moonbeam-standalone'));
```

We should really have this list somewhere in `apps-config`, there will be more.

packages/react-api/src/types.ts
```diff
@@ -20,6 +20,7 @@ export interface ApiState {
   hasInjectedAccounts: boolean;
   isApiReady: boolean;
   isDevelopment: boolean;
+  isEthereum:boolean;
```
jacogr, line 23:
```suggestion
  isEthereum: boolean;
```

packages/react-params/src/Param/findComponent.ts
```diff
@@ -54,6 +55,7 @@ const componentDef: TypeToComponent[] = [
   { c: Raw, t: ['Raw', 'Keys'] },
   { c: Enum, t: ['Enum'] },
   { c: Hash256, t: ['BlockHash', 'CodeHash', 'Hash', 'H256', 'SeedOf'] },
+  { c: Hash160, t: ['BlockHash', 'CodeHash', 'Hash', 'H160', 'SeedOf'] },
```
jacogr, line 58:
```suggestion
  { c: Hash160, t: ['H160'] },
```

jacogr (commented):
Some small initial comments.

packages/apps-config/src/api/rpc/moonbeam.ts
joelamouche: @jacogr I added the real rpc declarations. let me know what you think

packages/apps-config/src/api/spec/moonbeam.ts
jacogr: Really weird. I don't want to delay this one, want to get it one if everybody is happy. (The bundles can go after)

jacogr (approved):
lgtm.

packages/apps-config/src/api/rpc/moonbeam.ts
jacogr: All good.

jacogr:
This is a really unbelievable effort, thanks you very much. Really excited to see it in.

If everybody is happy, just give me the green light and will get it in - today, tomorrow, whenever you are ready to go. (+- 15mins from merge to live)

crystalin:
Awesome work folks !!
You can merge this one whenever you want.

joelamouche:
@jacogr Thank you for your help :)

PR #3994, opened 2020-11-08
update for Acala Mandala TC5

xlc:


packages/apps-config/src/api/bundle/spec/index.ts
```diff
@@ -6,5 +6,6 @@ import laminar from './laminar';
 
 export default {
   acala,
-  laminar
+  laminar,
+  mandala: acala
```
xlc, line 10: we changed the spec name from `acala` to `mandala`, keep both for compatibility with old testnet

jacogr (approved):
Thank you!

Issue #4005, opened 2020-11-10
Fix Verify flow for ethereum signatures


joelamouche:
*This problem is related to moonbeam compatibility on polkadot-js*

- [x] The verify function doesn't work because the crypto-util package needed an update on the `signatureVerify` function that would only accept a 36bytes address instead of ethereum compatible 20bytes address
- [ ] The signature function comes from the keyring, which is only compatible with ss58 
 address according to this https://www.npmjs.com/package/@polkadot/ui-keyring . Moreover, the keyring doesn't expose the private key, so we can't use the `secp256k1Sign` function used in the util-crypto tests

joelamouche:
@jacogr see my comment on signature: I think it needs an update on the keyring package, which is private

jacogr:
I don't understand, it basically would men that the exposed sign is incorrect for Ethereum - https://github.com/polkadot-js/common/blob/master/packages/keyring/src/pair/index.ts#L163

joelamouche:
Ok thanks for pointing to me the right part of the code to look at. It looks like the pair type (ethereum) needs to be passed at some point and Im trying to figure out if its during creation or when signing

jacogr:
The pair does know it is an Ethereum type there. (Once loaded on the keyring it will be indicated as ETH)

The issue with those if that I worked a bit in the dark, so very possible both the generation and verifying in that object is wrong.

joelamouche:
This was resolved by #4006 and https://github.com/polkadot-js/common/pull/767

PR #4006, opened 2020-11-10
Feature/add verify

joelamouche:
solves https://github.com/polkadot-js/apps/issues/4005

Thanks to https://github.com/polkadot-js/common/pull/767 ethereum signature verifications work.

I added the right verification in the verify component (ecdsa). This also works for non-ethereum ecdsa signatures


jacogr:
Sorry for the absolutely long delay, not on purpose - I just merged the update with the common libs so the verify now (as fixed by you), should be operational... hopefully :)

joelamouche:
No problem, thank you!

joelamouche:
@jacogr ready for review

packages/page-signing/src/Verify.tsx
```diff
@@ -56,13 +59,26 @@ function Verify ({ className = '' }: Props): React.ReactElement {
         if (isValidSr) {
           cryptoType = 'sr25519';
         } else {
-          isValid = false;
+          let verification;
```
jacogr, line 62: Should probably just use `signatureVerify` here and get rid of the specifics for sr/ed. We get the valid & type for free.

(Should actually have been done before, as it stands it is a duplication of what we have in util-crypto)

packages/react-params/src/initValue.ts
```diff
@@ -30,9 +30,7 @@ export default function getInitValue (registry: Registry, def: TypeDef): unknown
       : {};
   }
 
-  const type = [TypeDefInfo.Compact, TypeDefInfo.Option].includes(def.info)
-    ? (def.sub as TypeDef).type
-    : def.type;
+  const type = [TypeDefInfo.Compact, TypeDefInfo.Option].includes(def.info) ? (def.sub as TypeDef).type : def.type;
```
jacogr, line 33:
```suggestion
  const type = [TypeDefInfo.Compact, TypeDefInfo.Option].includes(def.info)
    ? (def.sub as TypeDef).type
    : def.type;
```

Assuming this snuck in via prettier?

packages/react-api/src/types.ts
```diff
@@ -54,7 +54,9 @@ export interface CallState {
 
 export type CallProps = ApiProps & CallState;
 
-export type BaseProps<T> = BareProps & CallProps & ChangeProps & {
+export type BaseProps<T> = BareProps &
+CallProps &
+ChangeProps & {
```
jacogr, line 59:
```suggestion
export type BaseProps<T> = BareProps & CallProps & ChangeProps & {
```

Actually weird, thought we manager to get rid of all the type with `&` and replaced with `interface`. (Probably some old reason behind this one)

packages/react-params/src/initValue.ts
```diff
@@ -140,7 +138,9 @@ export default function getInitValue (registry: Registry, def: TypeDef): unknown
       if (!warnList.includes(type)) {
         warnList.push(type);
         error && console.error(`params: initValue: ${error}`);
-        console.info(`params: initValue: No default value for type ${type} from ${JSON.stringify(def)}, using defaults`);
+        console.info(
+          `params: initValue: No default value for type ${type} from ${JSON.stringify(def)}, using defaults`
+        );
```
jacogr, line 143:
```suggestion
        console.info(`params: initValue: No default value for type ${type} from ${JSON.stringify(def)}, using defaults`);
```

Prettier? Very weird one.

jacogr (commented):
Generally looks good. Think we should just simplify the verification with `signatureVerify` instead of re-doing that here. (Was duplication anyway before this, now it is just more)

packages/react-params/src/initValue.ts
joelamouche: yes indeed. let me reverse that

packages/page-signing/src/Verify.tsx
```diff
@@ -31,38 +33,18 @@ function Verify ({ className = '' }: Props): React.ReactElement {
...
-        }
+      if (verification.crypto !== 'none') {
+        isValid = verification.isValid;
+        cryptoType = verification.crypto;
       }
     }
 
     setValidity({ cryptoType, isValid });
-  }, [data, isValidPk, isValidSignature, publicKey, signature]);
+  }, [isEthereum, data, isValidPk, isValidSignature, publicKey, signature]);
```
jacogr, line 47:
```suggestion
  }, [data, isValidPk, isValidSignature, publicKey, signature]);
```

Think this is not used now?

jacogr (approved):

jacogr:
As the custom by now, thanks a million for this.

PR #4011, opened 2020-11-11
Move to types bundle

joelamouche:
https://github.com/polkadot-js/apps/issues/4008

From Jaco: "We actually may want to use typesBundle, i.e. versioned types. The reason is then we can do the type overrides (such as this one which will most probably change), based on a version range. (In general, since we have support for it now, it is a better option anyway, since the changes can be made even before the chain gets upgraded and will kick in on upgrade)"


joelamouche:
error `DRR: Error: createType(AccountId):: Invalid AccountId provided, expected 32 bytes, found 20`

packages/apps-config/src/api/bundle/spec/moonbeam.ts
```diff
@@ -0,0 +1,79 @@
+// Copyright 2017-2020 @polkadot/apps-config authors & contributors
+// SPDX-License-Identifier: Apache-2.0
+
+import { OverrideBundleDefinition } from '@polkadot/types/types';
+
+// structs need to be in order
+/* eslint-disable sort-keys */
+
+export default {
```
jacogr, line 9: The format is slightly different here, and probably the issue.

Here is an example,

```export default [
  {
    // lowest version it applies to, highest version it applies to
    minmax: [0, undefined],
    // as per normal
    types: { ... }
  }
]
```

This allos you to manage multiples, see eg. https://github.com/polkadot-js/api/blob/master/packages/types-known/src/spec/kusama.ts where it is done for Kusama

joelamouche: thanks :)

However, doesn't that mean that the laminar and acala files use the wrong type (OverrideBundleDefinition)?

joelamouche: it looks like using OverrideVersionedType[] like in the example you provided doesn't conform to the type for Api.create of react-api

jacogr: It may be slightly different, honestly would need to look at what it does under the hood. Too many type options. 

(I have been meaning to move all types across for all chains, Acala added the required support into the UI specifically since they had upgrades lines up)

joelamouche: @jacogr have you had the occasion to look? Should I try again?

jacogr: Will take a peek and refresh.

jacogr: It should just work as defined here. 

You do `OverrideVersionedType[]` which aligns with types-known, as first used.

When passed to the API, it expects `typesBundle: OverrideBundleType` which is of this form -

```js
typesBundle: {
  spec: {
    moonbeam: OverrideVersionedType[]
  }
}
```

The api-react should take the stuff defined and just put it in that format.

jacogr: Ok, so what you should export is -

```js
export default {
  types: OverrideVersionedType[]
};
```

jacogr: (Like I said once the types is in, we can attempt to move the RPCs as well - not 100% sure the API uses it atm, the intent is there - if it doesn't can fix)

joelamouche: Thanks for the help. Should be ready now!

>Like I said once the types is in, we can attempt to move the RPCs as well 

Not too hot on adding features that aren't needed explicitly but I can add that as well if you feel this is important

jacogr: No, I'm sure that stuff is going to break and it is more for my benefit ... so I'll ply with that and see what breaks horribly. However, will get you to review once ready. (Not promising in the next couple of days, but having seen that again makes me want to get it going)

jacogr (approved):
Code looks good. 

I am assuming it works for you? Don't want to merge and break :)

Issue #4025, opened 2020-11-13
Make address addition more ethereum compatible


joelamouche:
When importing an address from MetaMask using a privatekey, we noticed some things that could be improved:
- Add ‘private key’ instead of 'rawSeed' for ethereum addresses (that wasnt clear to me)
- Make it clear that it needs to start with 0x
- Make sure checkbox is there even when the window is too small (was a problem for oen of our testers)
- when copying address to clipboard, the message shouldnt say "memnonic"

joelamouche:
what do y ou think @jacogr ? Should I go ahead and implement this?

jacogr:
Yes, it makes sense.

joelamouche:
Do you know how to make sure the tick box doesnt disappear?

jacogr:
I honestly don’t understand that one.

joelamouche:
yes me neither. what are you using for responsiveness? I guess this is not urgent

jacogr:
The breakpoints are really simple -

- on tables, there is a media class that hides columns
- on modals the only area that should get hidden is the columns/descriptions on the right

Have really not paid much attention to making sure it works across all, the above simple approach makes it passable. One day, when bored...

jacogr:
Ok, replicated... it sits in the second column on the modal, hence being hidden. Will make a fix for that one.

jacogr:
Closed in #4032

PR #4032, opened 2020-11-16
Feature/better eth address addition

joelamouche:
Addresses https://github.com/polkadot-js/apps/issues/4025

- Add ‘private key’ instead of 'rawSeed' for ethereum addresses (that wasnt clear to me)
- Make it clear that it needs to start with 0x
- when copying address to clipboard, the message shouldnt say "memnonic"

- modified the CopyButton to adapt to private key

### DIFF
--- a/packages/page-accounts/src/modals/Create.tsx
+++ b/packages/page-accounts/src/modals/Create.tsx
@@ -272,7 +272,6 @@ function Create ({ className = '', onClose, onStatusChange, seed: propsSeed, typ
     },
     [api, derivePath, isDevelopment, isValid, name, onClose, onStatusChange, pairType, password, seed, t]
   );
-
   return (
     <Modal
       className={className}
@@ -322,7 +321,7 @@ function Create ({ className = '', onClose, onStatusChange, seed: propsSeed, typ
                 />
                 < CopyButton
                   className='copyMoved'
-                  isMnemonic
+                  type={seedType === 'bip'?"mnemonic":seedType === 'raw'?isEthereum?'private key':'seed':'development seed'}
                   value={seed}
                 />
               </TextArea>
--- a/packages/page-accounts/src/modals/Create.tsx
+++ b/packages/page-accounts/src/modals/Create.tsx
@@ -204,8 +204,8 @@ function Create ({ className = '', onClose, onStatusChange, seed: propsSeed, typ
       : []
   ).concat(
     { text: t<string>('Mnemonic'), value: 'bip' },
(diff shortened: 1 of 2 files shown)


packages/page-accounts/src/modals/Create.tsx
```diff
@@ -316,9 +320,9 @@ function Create ({ className = '', onClose, onStatusChange, seed: propsSeed, typ
                   onChange={_selectSeedType}
                   options={seedOpt}
                 />
-                <CopyButton
+                < CopyButton
                   className='copyMoved'
-                  isMnemonic
+                  type={seedType === 'bip' ? 'mnemonic' : seedType === 'raw' ? isEthereum ? 'private key' : 'seed' : 'development seed'}
```
jacogr, line 325:
```suggestion
                  type={seedType === 'bip' ? t<string>('mnemonic') : seedType === 'raw' ? isEthereum ? t<string>('private key') : 'seed' : t<string>('raw seed')}
```

Just made it translatable. (and s/development/raw/)

packages/react-components/src/CopyButton.tsx
```diff
@@ -15,27 +15,27 @@ interface Props {
...
 
   const _onCopy = useCallback(
     (): void => {
-      (isAddress || isMnemonic) && queueAction && queueAction({
-        account: isAddress ? value : undefined,
+      (type !== 'other') && queueAction && queueAction({
+        account: type !== 'mnemonic' ? value : undefined,
         action: t<string>('clipboard'),
-        message: t<string>(`${isAddress ? 'address' : 'mnemonic'} copied`),
+        message: t<string>(`${type} copied`),
```
jacogr, line 34:
```suggestion
        message: t<string>('{{type}} copied`, { replace: { type: type || t<string>('other') } }),
```

packages/react-components/src/CopyButton.tsx
```diff
@@ -15,27 +15,27 @@ interface Props {
...
-  isAddress?: boolean;
+  type?: string;
   isMnemonic?: boolean;
   value: string;
 }
 
 const NOOP = () => undefined;
 
-function CopyButton ({ children, className, icon = 'copy', isAddress = false, isMnemonic = false, value }: Props): React.ReactElement<Props> {
+function CopyButton ({ children, className, icon = 'copy', type = 'other', value }: Props): React.ReactElement<Props> {
```
jacogr, line 25:
```suggestion
function CopyButton ({ children, className, icon = 'copy', type, value }: Props): React.ReactElement<Props> {
```

(Just removed it here so the string is translatable in full)

jacogr (approved):
Thank you!

PR #4037, opened 2020-11-17
Eslint imports

ekowalsk:
I did not configured alphabetical order, because it would cause changes in over 500 files. For the same reasons, I didn't touch the new line breaks between groups


packages/apps/src/Menu/index.tsx
```diff
@@ -1,13 +1,13 @@
 // Copyright 2017-2020 @polkadot/apps authors & contributors
 // SPDX-License-Identifier: Apache-2.0
 
+import { TFunction } from 'i18next';
```
jacogr, line 4: Can we configure (and use) a `import type { TFunction } ...` or is it not that aware of syntax? (I guess in this case it works since i18next is generally only touched in the `translate.ts` files, but technically not quite correct.)

packages/apps/src/Menu/types.tsx
```diff
@@ -1,8 +1,8 @@
 // Copyright 2017-2020 @polkadot/apps authors & contributors
 // SPDX-License-Identifier: Apache-2.0
 
-import { IconName } from '@fortawesome/fontawesome-svg-core';
 import { Routes } from '@polkadot/apps-routing/types';
+import { IconName } from '@fortawesome/fontawesome-svg-core';
```
jacogr, line 5: This is also a `type { IconName }` import.

packages/apps/src/Menu/index.tsx
ekowalsk: If I understand correctly, you would like to apply the `import type` syntax to configure linter, so that it separates that type-only imports into another group. If so, then I'm afraid it's not possible yet to configure it that way
image

packages/apps/src/Menu/types.tsx
ekowalsk: I've applied separate rule for that import, now it works as it was supposed to

packages/page-accounts/src/modals/Delegate.tsx
```diff
@@ -3,11 +3,11 @@
 
 import { AmountValidateState } from '../Accounts/types';
 
+import { Conviction } from '@polkadot/types/interfaces';
```
jacogr, line 6: These are all types, so I guess it is a special case again for `@polkadot/types/interfaces*` (They are wildly inconsistent in all the repos, some places they are treated correctly as types, other places not)

jacogr: This one is weird though, since I don' know why it jumped to the top of this section and not the previous section rather.

packages/page-accounts/src/modals/ProxyOverview.tsx
```diff
@@ -1,18 +1,18 @@
...
 // SPDX-License-Identifier: Apache-2.0
 
-import { ApiPromise } from '@polkadot/api';
 import { SubmittableExtrinsic } from '@polkadot/api/types';
 import { AccountId, ProxyDefinition, ProxyType } from '@polkadot/types/interfaces';
-
 import BN from 'bn.js';
 import React, { useCallback, useEffect, useMemo, useState } from 'react';
+import styled from 'styled-components';
+import { ApiPromise } from '@polkadot/api';
```
jacogr, line 9: This one is perfect as an adjustment, it caught something that was incorrect. (Although I guess `ApiPromise` was only used as a `type`)

packages/page-accounts/src/util.tsx
```diff
@@ -3,10 +3,10 @@
 
 import { SortedAccount } from './types';
 
+import { KeyringAddress } from '@polkadot/ui-keyring/types';
```
jacogr, line 6: This should be in the types section above.

packages/page-claims/src/Claim.tsx
```diff
@@ -1,13 +1,13 @@
 // Copyright 2017-2020 @polkadot/app-claims authors & contributors
 // SPDX-License-Identifier: Apache-2.0
 
-import { Option } from '@polkadot/types';
 import { BalanceOf, EthereumAddress, StatementKind } from '@polkadot/types/interfaces';
-
+import { TxCallback } from '@polkadot/react-components/Status/types';
 import React, { useEffect, useState } from 'react';
 import styled from 'styled-components';
+import { Option } from '@polkadot/types';
```
jacogr, line 8: This is correct (discounting the types just above that should move one up, but this one is indicated 100% as a non-type import from something called "types")

packages/page-claims/src/Warning.tsx
```diff
@@ -1,10 +1,9 @@
 // Copyright 2017-2020 @polkadot/app-claims authors & contributors
 // SPDX-License-Identifier: Apache-2.0
 
-import { AddressMini, Card } from '@polkadot/react-components';
-
 import React from 'react';
 import styled from 'styled-components';
+import { AddressMini, Card } from '@polkadot/react-components';
```
jacogr, line 6: +1

packages/page-contracts/src/shared/MessageSignature.tsx
```diff
@@ -3,11 +3,11 @@
 
 import { AbiMessage } from '@polkadot/api-contract/types';
 
-import React from 'react';
 import { ThemeProps } from '@polkadot/react-components/types';
+import { encodeTypeDef } from '@polkadot/types/create';
```
jacogr, line 7: This is not correct, it is a non-type import.

packages/page-democracy/src/Overview/TreasuryCell.tsx
```diff
@@ -4,11 +4,11 @@
 import { ProposalIndex, TreasuryProposal } from '@polkadot/types/interfaces';
 import { TypeDef } from '@polkadot/types/types';
 
+import { getTypeDef } from '@polkadot/types/create';
```
jacogr, line 7: As the above, non-type import, should not be moved.

jacogr (commented):
There are some interesting places here where it does better than the human :) Which is great. I think the types-part needs some massaging still.

packages/page-accounts/src/modals/Delegate.tsx
ekowalsk: It actually jumped into the types section - but as I said, I did not touched the group separation configuration, because it would cause hundreds of files to change. The breaks between the groups are indifferent to linter configuration and it does not change them. However, I will manually adjust them in a minute in the places you've pointed

packages/react-components/src/Icon.tsx
```diff
@@ -1,8 +1,8 @@
 // Copyright 2017-2020 @polkadot/react-components authors & contributors
 // SPDX-License-Identifier: Apache-2.0
 
-import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { IconName, library } from '@fortawesome/fontawesome-svg-core';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
```
jacogr, line 5: There is an issue here - this import is not a type. Only IconName is (library on the preceding is also not)

packages/react-components/src/InputRpc/options/method.tsx
```diff
@@ -3,10 +3,10 @@
 
 import { DefinitionRpcExt } from '@polkadot/types/types';
 import { DropdownOption, DropdownOptions } from '../../util/types';
+import jsonrpc from '@polkadot/types/interfaces/jsonrpc';
```
jacogr, line 6: The ordering here is funny, this should move 2 up.

jacogr:
I would not expend much effort on this at this point, as horrible and inconsistent (human-error) it is in places. Not sure the trying to force type imports here is a good thing. Think it should rather be done properly when the underlying libs has `import type` support. 

(It would be useful when touching stuff atm to start using that form though... and hope the stars align at some point soon)

jacogr:
This may also be an option after #4052 - https://github.com/lydell/eslint-plugin-simple-import-sort#readme (it is supposed to support type imports)

jacogr (approved):
Thank you!

There is one small gap here, which I've neglected to mention before. But not going to delay this one since it is quite tiring having to rebase the whole time while others go in...

Anyway, we probably want to pass the autofix flag to eslint, so it gets fixed locally as well as on CI. (Since it is not passed anywhere it is a manual process atm)

PR #4038, opened 2020-11-17
Align account name with nunito sans

MiZiet:
Hey, this should solve the problem with account name alignment after applying Nunito Sans font.


jacogr:
There is something seriously weird with this font. This is on the council page, 

image

- W3F top-aligned (with unicode as sub)
- JACO & JACK top-aligned, nothing special
- GAV & JUTTA bottom aligned (both contain unicode)

This is the current, where it seems center aligned in all cases -

image

MiZiet:
image

jacogr:
Is this really worth it? I'm at the point of giving up since it is seemingly like whacking moles for no benefit since this font behaves funny...

This is in the dropdowns -

image

This is the current -

image

(The new one is only nicer since the sub is also middle aligned based on the last push)

jacogr:
I love the groundwork that went int this overall, while not perfect (theme-wise), we managed to cleanup the fonts and align it properly with no surprises as to use.

As for this final push, I've not looked at the last iteration of actually pulling in the Nunito font since I think it is much more pain that it is worth worth little to no actual benefit. Having taken a quick peek around, I'll be more partial to something like simple such as https://newcss.net/, e.g.

```js
fontMono: 'Consolas,monaco,"Ubuntu Mono","Liberation Mono","Courier New",Courier,monospace', // 'monospace',
  fontSans: '"Inter",-apple-system,BlinkMacSystemFont,"Segoe UI",Roboto,Oxygen,Ubuntu,Cantarell,"Open Sans","Helvetica Neue",sans-serif,"Apple Color Emoji","Segoe UI Emoji","Segoe UI Symbol"',
```

If we really want to make a change. From quick looking it suffers none of the drawbacks that we have been playing whack-a-mole with, works out of the box, tries to use what is on the users system and has no impacts with extra downloads overall.

Thoughts?

packages/react-components/src/AccountName.tsx
```diff
@@ -193,17 +193,17 @@ export default React.memo(styled(AccountName)`
...
     width: 100%;
 
     .name {
       font-weight: 400 !important;
       filter: grayscale(100%);
-      line-height: 1;
       opacity: 0.6;
       overflow: hidden;
       text-overflow: ellipsis;
+      line-height: 1.358rem;
```
jacogr, line 206: 22px as per https://github.com/polkadot-js/apps/blob/master/packages/react-components/src/Badge.tsx#L56

jacogr (approved):
Thank you!

PR #4062, opened 2020-11-24
Factor out types from apps-config to new known-types package

emostov:
This PR is an attempt to factor out API types from the apps-config package, reverting apps-config to private and instead publishing a new known-types package. The rationale behind moving the types out of apps-config is that apps-config exposes exports that are only valuable to the apps project and do not need to be public (links, colors, etc). Hopefully this would lead to less worry for polkadot-js maintainers about breaking things for users when updating the apps-config setup.

If you follow through with this refactor, I imagine in the future it might make sense to move known-types to the polkadot-js/common repo (and maybe merge with @polkadot/networks package), but I thought it would be simplest to first just create the package here and move it later once you have confidence in its stability.

I am not tied to any of the changes in this PR and happy to help with related PRs if this one does not make sense. 

cc @dvdplm


jacogr:
With only  brief look, this has some serious issues which breaks the world.

- `private` does not mean anything from a script context, you need to explicitly add the `.npm-skip`
- there are already a used `types-known`, https://github.com/polkadot-js/api/tree/master/packages/types-known
- this is 100% unlikely to be split as indicated (especially not into common which is top-level to all, dependencies go down, not up and cross)

Don't forget about how people actually use this, go through the PRs where this has been adjusted - any extra pian to that process is not going to work.

jacogr:
Cleanup closing, not being worked on.

PR #4068, opened 2020-11-24
add "save event to calendar" functionality on calendar page

wirednkod:
A small calendar icon is added to each event in order to allow user to download ics file and add it to calendar Screenshot 2020-11-25 at 12 12 02 AM

### DIFF
--- a/packages/page-calendar/src/DayItem.tsx
+++ b/packages/page-calendar/src/DayItem.tsx
@@ -3,7 +3,7 @@
 
 import type { EntryInfo } from './types';
 
-import React, { useMemo, useState } from 'react';
+import React, { useMemo, useState, useCallback } from 'react';
 import styled from 'styled-components';
 import { formatNumber, isString } from '@polkadot/util';
 import { dateCalendarFormat } from './util';
@@ -25,6 +25,39 @@ function DayItem ({ className, item: { date, info, type } }: Props): React.React
   const { t } = useTranslation();
 
   const [description, setDescription] = useState<string>('');
+
+  const _exportCal = useCallback(
+    () => {
+      const startDate = dateCalendarFormat(date);
+      // For now just add 1 hour for each event
+      const endDate = dateCalendarFormat(new Date(new Date(date).setHours(new Date(date).getHours() + 1)));
+      const calData =
+        'BEGIN:VCALENDAR\n' +
+        'CALSCALE:GREGORIAN\n' +
+        'METHOD:PUBLISH\n' +
+        'PRODID:-//Test Cal//EN\n' +
+        'VERSION:2.0\n' +
+        'BEGIN:VEVENT\n' +
+        'UID:test-1\n' +
+        'DTSTART;VALUE=DATE:' + startDate + '\n' +
+        'DTEND;VALUE=DATE:' + endDate + '\n' +
+        'SUMMARY:' + description + '\n' +
+        'DESCRIPTION:' + description + '\n' +
+        'END:VEVENT\n' +
+        'END:VCALENDAR';
+      const fileNameIcs = encodeURI(description) + '.ics';
+      const data = new File([calData], fileNameIcs, { type: 'text/plain' });
+      const anchor = window.document.createElement('a');
+
+      anchor.href = window.URL.createObjectURL(data);
+      anchor.download = fileNameIcs;
+      document.body.appendChild(anchor);
+      anchor.click();
+      document.body.removeChild(anchor);
+      window.URL.revokeObjectURL(anchor.href);
+    },
+    [description, date]
+  );
 
   const desc = useMemo(
     (): React.ReactNode => {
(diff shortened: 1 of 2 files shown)


packages/page-calendar/src/DayItem.tsx
```diff
@@ -46,62 +50,116 @@ function DayItem ({ className, item: { date, info, type } }: Props): React.React
                     : ['treasurySpend'].includes(type)
                       ? <div className='itemLink'><a href='#/treasury'>{t<string>('via Treasury')}</a></div>
                       : undefined;
+      let s = '';
 
       switch (type) {
         case 'councilElection':
-          return <><div className='itemDesc'>{t<string>('Election of new council candidates')}</div>{typeLink}</>;
-
+          s = 'Election of new council candidates';
```
jacogr, line 57: So on these we need to bring back the `t<string>('Election of new council candidates')` wrapper - that allows translation. Without it (or being dynamic), the i18n functionality is lost. The output can still be uses (i.e. the result from the `t(...)`) in the iCal, it is just a string.

TL;DR For all of these, just remove the fragment/div wrappers, keep the `t(...)`

packages/page-calendar/src/DayItem.tsx
```diff
@@ -46,62 +50,116 @@ function DayItem ({ className, item: { date, info, type } }: Props): React.React
...
+          s = 'Acceptance of new members and bids';
+          break;
 
         default:
           return assertUnreachable(type);
       }
+
+      setDescription(id ? s.replaceAll('{{id}}', id) : s);
+
+      return id ? <><div className='itemDesc'>{t<string>(s, { replace: { id } })}</div>{typeLink}</> : (<><div className='itemDesc'>{s}</div>{typeLink}</>);
```
jacogr, line 113: As above, this is fine, but we have lost the much-needed i18n

packages/page-calendar/src/DayItem.tsx
```diff
@@ -46,62 +50,116 @@ function DayItem ({ className, item: { date, info, type } }: Props): React.React
...
     },
     [info, t, type]
   );
 
+  function exportToCal (fileName: string, date: Date): void {
```
jacogr, line 118: Extract this function to file-level - as it stands it gets re-created each render.

packages/page-calendar/src/DayItem.tsx
```diff
@@ -46,62 +50,116 @@ function DayItem ({ className, item: { date, info, type } }: Props): React.React
...
+
+    anchor.href = window.URL.createObjectURL(data);
+    anchor.download = fileNameIcs;
+    document.body.appendChild(anchor);
+    anchor.click();
+    document.body.removeChild(anchor);
+    window.URL.revokeObjectURL(anchor.href);
+  }
+
+  function calendarIcon (date: Date): React.ReactNode {
```
jacogr, line 148: Also extract to file-level.

packages/page-calendar/src/DayItem.tsx
```diff
@@ -46,62 +50,116 @@ function DayItem ({ className, item: { date, info, type } }: Props): React.React
...
+    if (date) {
+      return (<Button className='exportCal'
+        icon='calendar-plus'
+        onClick={() => exportToCal(description, date)} />);
```
jacogr, line 152: This we would want to do as a `useCallback`

jacogr (changes requested):
Thanks an absolute million for this - some small comments, but generally the functionality and look seems spot-on.

packages/page-calendar/src/DayItem.tsx
```diff
@@ -46,62 +50,116 @@ function DayItem ({ className, item: { date, info, type } }: Props): React.React
...
+    } else {
+      return null;
+    }
+  }
+
   return (
     <div className={className}>
       <div className='itemTime'>{date.toLocaleTimeString().split(':').slice(0, 2).join(':')}</div>
       {desc}
+      {calendarIcon(date)}
```
jacogr, line 162: Sorry, this comment got lost (for some obscure reason). The above can just simply be (without extracting the function to file-level) -

```suggestion
      {date && (
        <Button
          className='exportCal'
          icon='calendar-plus'
          onClick={_exportCal}
      )}
```

(It assumes we have a `useCallback` with the name `_exportCal`)

wirednkod:
> Thanks an absolute million for this - some small comments, but generally the functionality and look seems spot-on.

Its a pleasure :) Thanks for the feedback

jacogr (approved):
Absolutely brilliant! I'll be making extensive use of this, as are others.

Swader:
@wirednkod this is great, can you let us know a DOT address? I'd like to tip you for this work.

wirednkod:
@Swader Thank you for the offer but there is really no need for a tip, mate. It was done for pleasure - rather than payment/tip. Was interested on having a look on the code for an interview with Parity and stumbled upon this one - seemed really easy and fast to create - so why not  ¯\_(ツ)_/¯

wirednkod:
@jacogr I have found various 3rd party npms that are making the export easier to download for Google/Apple/Outlook etc, etc. I could expand this feature to that - if you are ok with it :) Do you prefer custom code or 3rd parties are fine as well?

Swader:
> @Swader Thank you for the offer but there is really no need for a tip, mate. It was done for pleasure - rather than payment/tip. Was interested on having a look on the code for an interview with Parity and stumbled upon this one - seemed really easy and fast to create - so why not ¯_(ツ)_/¯

Admirable, but it's really commonplace: https://ipfs.io/ipns/polkadot.dotapps.io/#/treasury/tips - we really do like to support our supporters, especially since it's funded by the chain itself. If you change your mind, the offer stands, all I need is an address :)

jacogr:
Really depends on the value the npms actually add - if it is left-pad like, then I would rather bring it in. 

Each dep ends up with having to be managed for the next 20 years and each bring in their level of cruft and so it just adds to the pile. Some weeks can't keep up with security advisories, so generally unless it is really novel, try to steer clear from too many external deps, there are too many already bringing in their own pile... 

Trying to clean up here as well, but current efforts in that regard actually sits in other repos, so will get here for "spring cleaning" :)

wirednkod:
> Each dep ends up with having to be managed for the next 20 years and each bring in their level of cruft and so it just adds to the pile. Some weeks can't keep up with security advisories, so generally unless it is really novel work, try to steer clear from too many external deps, there are too many already bringing in their own pile...

Makes sense :) I'll have a look to wrap it in custom code in order to avoid dependencies. Cheers

wirednkod:
> > @Swader Thank you for the offer but there is really no need for a tip, mate. It was done for pleasure - rather than payment/tip. Was interested on having a look on the code for an interview with Parity and stumbled upon this one - seemed really easy and fast to create - so why not ¯_(ツ)_/¯
> 
> Admirable, but it's really commonplace: https://ipfs.io/ipns/polkadot.dotapps.io/#/treasury/tips - we really do like to support our supporters, especially since it's funded by the chain itself. If you change your mind, the offer stands, all I need is an address :)

Well.. if its a commonplace then here is an address:
11e2DNjBooKKoiXnt9Mw9ePi3PSDF9bUvh1ntYctAfUUoYD

Cheers

PR #4114, opened 2020-11-30
added plasm dusty type definitions

hoonsubin:
This PR updates the Dusty Plasm Network type definitions and separates the types from the main network to the test network (Dusty). Notably, the Dusty network implements the following types:

- dapps staking
- ovm
- plasm rewards
- plasma

and it also removes the `RefCount` type which is specific to the main network of Plasm.


jacogr (approved):
Thank you!

PR #4137, opened 2020-12-02
Add simple bounty list

krzysztof-jelski:
Initial bounty list.

Design is somewhere between the new figma wireframes and current styles from other apps.

There's a problem with auto-updating bounties (eg. page refresh is needed when bounty state changes), help appreciated.

Display of PendingPayout-specific information is not there yet.


packages/page-bounties/src/Bounties.tsx
```diff
@@ -0,0 +1,63 @@
...
+  transform: (entries: BountyDescriptionEntry[]): BountyDescription[] => {
+    return entries.map(([key, description]) => ({ description: description.value.toHuman() as string, id: key.args[0] }));
+  }
+};
+
+function Bounties (): React.ReactElement {
+  const { t } = useTranslation();
+  const { api } = useApi();
+
+  const bounties = useCall<BountyEntry[]>(api.query.treasury.bounties.entries);
```
jacogr, line 29: This is the refresh issue - entries doesn't have subscriptions, so we never know when they update. (There is a Substrate issue logged on this)

krzysztof-jelski: Darn, thanks. Alternatives as of now?

jacogr: This is where a derive would come in nicely -

- subscribe to the number of bounties
- when changed, update `.keys` (not entries)
- subscribe to each of the keys for updates

packages/page-bounties/src/types.ts
```diff
@@ -0,0 +1,17 @@
+// Copyright 2017-2020 @polkadot/app-bounties authors & contributors
+// SPDX-License-Identifier: Apache-2.0
+import type { AccountId, BlockNumber } from '@polkadot/types/interfaces/runtime';
```
jacogr, line 3:
```suggestion
// SPDX-License-Identifier: Apache-2.0

import type { AccountId, BlockNumber } from '@polkadot/types/interfaces/runtime';
```

I guess the header rule needs to check for a newline.

jacogr (approved):
Happy to merge as-is. 

The expansion of the row is not 100% (to look at at some point) -

- the actual columns jump when expanding
- with multiples the actual rows will change colors, i.e. the expansion doesn't take the color of the parent

But that is just extra effort, as here, can get it in.

packages/page-bounties/src/types.ts
jacogr: Yes, well, it doesn't support empty-line-following

PR #4138, opened 2020-12-02
Update Phala testnet (PoC2 => PoC3)

LusWar:


LusWar:
Fix type and update logo

PR #4141, opened 2020-12-02
Show unapplied slashes for the active era

andresilva:
Not sure if there's any specific reason why we wouldn't want this. I was looking into some slash event that just happened on Kusama and it wasn't showing on the UI for this reason. I haven't tested this PR though.


jacogr:
No, it was a bug, I assumed it was end-of-era - however have just merged the fix alongside my last PR. (And it is exactly the same as yours)

jacogr:
Sadly closing as duplicate since I just merged it a short while ago. I really do appreciate the effort.

PR #4142, opened 2020-12-03
Add polkadot and kusama endpoints hosted by PatractLabs

toxotguo:


jacogr (approved):
Thank you!

A big +1 from me for having this in Asia, we need more infrastructure in that region.

PR #4168, opened 2020-12-07
Ek extract type only imports into separate group

ekowalsk:
Used beta release of `eslint plugin simple import sort ` with new option to extract all type-only imports as a separate group


jacogr (approved):
Seems sane :)

PR #4170, opened 2020-12-08
Add max-width to content

MiZiet:
These changes will set a max-width to the app content.


jacogr:
Not quire convinced, yes.

Existing - 
image

New -
image

The losing of the padding here is non-optimal.

krzysztof-jelski:
Idea -- move the `1750px` and `875px` constants into the theme. What do you think?

jacogr (approved):
Seems sane :)

Do agree with @krzysztof-jelski that these would be well-suited to extract. (Including using 1750/2, slightly less magic - somebody is going to forget to adjust at some point)

jacogr:
Brilliant, thank you.

PR #4171, opened 2020-12-08
Eq balance

pr0fedt:
Added custom derives in apps-config packages;
Added derives.balances.all for Equilibrium chain to support custom balances;


jacogr (commented):
I'm struggling with the PR, since it is all unneeded whitespace changes which makes it inconsistent with the project. 

Could you please revert the whitespace, would love to look at the functionality and actual changes.

pr0fedt:
> I'm struggling with the PR, since it is all unneeded whitespace changes which makes it inconsistent with the project.
> 
> Could you please revert the whitespace, would love to look at the functionality and actual changes.

Done

jacogr:
Much appreciated, sorry for being a pain - however, it did show a gap in the linting rules, so that has also been addressed. So net-positive :)

packages/react-api/src/Api.tsx
```diff
@@ -189,30 +191,44 @@ function Api ({ children, store, url }: Props): React.ReactElement<Props> | null
...
+      api.on('error', (error: Error) => setApiError(error.message));
+      api.on('ready', callback);
+    };
+
+    create(undefined, () => {
+      const specName = Object.keys(derivesSpec).find((specName) =>
+        api.runtimeVersion.specName.eq(specName)
+      ) as keyof typeof derivesSpec | undefined;
+
+      create(specName, () => {
```
jacogr, line 218: The problem with this is that the API init is a very heavy operation, so basically (as you'e indicated), we are hacking around the API. The correct approach would be to adjust `typesBundle` on the API to also take optional derives and let the API decorate.

Overall not unhappy with allowing the API to decorate derives based on info it has retrieved (this would be useful in a number of situations), just rather do it correctly.

Does it make sense?

pr0fedt: Thank you

jacogr:
The proper attaching of the RPCs (from typesBundle) went into the api today, https://github.com/polkadot-js/api/pull/2927 - so that is what I suggest doing with the derives as well. I'll link and issue to this, I think that after I went through the RPCs, it should be relatively straight-forward (although obviously not in the same places)

pr0fedt:
Closed in favour of https://github.com/polkadot-js/apps/pull/4182

PR #4174, opened 2020-12-08
Move moonbeam/sora RPCs to bundle-only

jacogr:
Requires https://github.com/polkadot-js/apps/pull/4173


packages/react-components/src/InputRpc/useRpcs.ts
```diff
@@ -0,0 +1,21 @@
+// Copyright 2017-2020 @polkadot/react-components authors & contributors
+// SPDX-License-Identifier: Apache-2.0
+
+// TODO: We have a lot shared between this and InputExtrinsic & InputStorage
+
+import type { DefinitionRpcExt } from '@polkadot/types/types';
```
jacogr, line 6:
```suggestion
import type { DefinitionRpcExt } from '@polkadot/types/types';
```

PR #4182, opened 2020-12-09
Eq add balance fixed

pr0fedt:


jacogr:
Brilliant! Does it do the trick for you?

pr0fedt:
Yes, only waiting for our backand devs to return correct tokenDecimals

jacogr (approved):
Thanks a lot of doing ALL of this - the API and then you interfaces.

PR #4183, opened 2020-12-09
Fix spacing (as found) after width limit

jacogr:
Introduced in https://github.com/polkadot-js/apps/pull/4170


jacogr:
Replaced by https://github.com/polkadot-js/apps/pull/4184

PR #4185, opened 2020-12-10
Adding Centrifuge Runtime Upgrade 4 types

mikiquantum:


jacogr (approved):
Thank you!

A suggestion to better manage upgrades would be to move the types to `apps-config/api/bundle` (instead of `apps-config/api/spec` - in that case it is versioned, so you can actually already apply anything pre-upgrade.

PR #4188, opened 2020-12-10
Improve bounties display

MiZiet:
This should solve the problem with table columns changing their widths when expanding Bounty component.


jacogr (approved):
Thank you.

PR #4193, opened 2020-12-11
Use bounty derive

MiZiet:
These changes will simplify the bounty component by using a bounty derive

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2000,7 +2000,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@polkadot/api@npm:^2.10.2-7":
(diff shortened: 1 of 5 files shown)


yarn.lock
```diff
@@ -2494,6 +2494,20 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@polkadot/metadata@npm:2.10.2-12":
```
jacogr, line 2497: This is weird, it is like the resolution were not updated, so there are 2 occurences here now?

jacogr: You shouldn't need to update it at all, the current version does have the derive?

jacogr (approved):
Thank you.

PR #4195, opened 2020-12-11
Move to Ledger usage with hook

jacogr:


Tbaut (approved):
tested:
- connect ledger S
- show address on device
- make keep alive transaction

PR #4197, opened 2020-12-12
Display error.message || error (when string is thrown)

jacogr:
Enzyme signing doesn't return an error object, ensure that when an error string is received, it is still displayed.


packages/react-signer/src/TxSigned.tsx
```diff
@@ -79,9 +79,9 @@ async function signAndSend (queueSetTxStatus: QueueTxMessageSetStatus, currentIt
     }));
   } catch (error) {
     console.error('signAndSend: error:', error);
-    queueSetTxStatus(currentItem.id, 'error', {}, error);
+    queueSetTxStatus(currentItem.id, 'error', undefined, error);
```
jacogr, line 82:
```suggestion
    queueSetTxStatus(currentItem.id, 'error', {}, error);
```

PR #4199, opened 2020-12-12
Update translation.json

fomod-in:
Major update UI translation


jacogr (approved):
As always, thank you very much. It is greatly appreciated.

PR #4202, opened 2020-12-14
add MathChain-Galois 

hging:


jacogr:
You should be able to auto-fix the linting issues by just running `yarn lint` locally. (It will apply `--fix`)

jacogr (approved):
Thank you very much.

The `MultiAddress` usage is interesting, since it is quite recent in Substrate and this is the first chain that I know about using it. (It does make a lot of sense since it is very flexible and provides a number of options)

packages/apps-config/src/api/spec/galois.ts
```diff
@@ -0,0 +1,11 @@
+// Copyright 2017-2020 @polkadot/apps-config authors & contributors
+// SPDX-License-Identifier: Apache-2.0
+
+// structs need to be in order
+/* eslint-disable sort-keys */
+
+export default {
+  Address: 'MultiAddress',
```
jacogr, line 8: There seems to be a couple of additional types missing, specifically -

```
REGISTRY: Unable to resolve type ExitReason, it will fail on construction
METADATA: Unknown types found, no types for ExitReason, Receipt, Transaction, TransactionStatus
```

PR #4221, opened 2020-12-15
add zero testnet

2075:
hi there,

this is to add ~~again~~ the zero alphaville testnet, ~~somehow it got wiped away~~ complimenting the other two PR in substrate and telemetry a while ago.

- we have added webpack to the dev dependencies, as otherwise the docker wont build.

greetings!


package.json
```diff
@@ -93,6 +93,8 @@
     "styled-components": "^5.2.1",
     "stylelint": "^13.8.0",
     "stylelint-config-recommended": "^3.0.0",
-    "stylelint-config-styled-components": "^0.1.1"
+    "stylelint-config-styled-components": "^0.1.1",
+    "webpack": "^5.10.2",
```
jacogr, line 97: Thank you, think this is a bit of fallout from just having swapped to WP5, would have run into this next Monday.

yarn.lock
```diff
@@ -19336,6 +19338,43 @@ fsevents@~2.1.2:
   languageName: node
   linkType: hard
 
+"webpack@npm:^5.10.2":
```
jacogr, line 19341: Can we just align this with the version in `packages/apps` (5.10.1, just so there is one version the lockfile, instead of 2 as installed atm)

jacogr (commented):
Some types are missing, I get the following connecting -

```
2020-12-15 18:40:29        METADATA: Unknown types found, no types for Campaign, EventMessage
```

package.json
2075: no worries, glad add two cents

2075:
> Some types are missing, I get the following connecting -
> 
> ```
> 2020-12-15 18:40:29        METADATA: Unknown types found, no types for Campaign, EventMessage
> ```

will this be resolved by settings>metadata>update or do you contain them here as well?

jacogr:
No. Those are per-chain types. The metadata only communicates the names of the types. Those need to be supplied, see eg. the apps-config/api/spec for types supplied by other chains that match their config. 

Without the correct types mirroring the Rust config (all those non standard or changes made) you will have incorrect decoding.

2075:
allright, i have added the current spec.

packages/apps-config/src/api/spec/zero.ts
```diff
@@ -0,0 +1,21 @@
+// Copyright 2017-2020 @polkadot/apps-config authors & contributors
+// SPDX-License-Identifier: Apache-2.0
+
+// structs need to be in order
+/* eslint-disable sort-keys */
+
+export default {
+  Address: 'AccountId',
```
jacogr, line 8: You would want `LookuoSource: 'AccountId'` as well (otherwise transfers won't work)

packages/apps-config/src/api/spec/zero.ts
```diff
@@ -0,0 +1,21 @@
+// Copyright 2017-2020 @polkadot/apps-config authors & contributors
+// SPDX-License-Identifier: Apache-2.0
+
+// structs need to be in order
+/* eslint-disable sort-keys */
+
+export default {
```
jacogr, line 7: Cool.

jacogr:
Apart from the `LookupSource` change above, it looks good. 

As an optional suggestion to make your life easier going forward. I would suggest using `typesBundle` instead of types directly -

- it allows you to inject types based on versions (i.e. when you upgrade, it comes in really handy since you can even do the type changes beforehand)
- it is very simmilar

- So to use that, just move the types file into `bundle` (as opposed to `spec`)
- wrap it in a version tag, so instead of `export default { ...all types... }`, you would have `export default { types: [{ minmax: [0, undefined], types: { ..all types... } }] }`

So the `minmax` indicated the version to apply it to, a recent example (local here) is monbeam, https://github.com/polkadot-js/apps/blob/master/packages/apps-config/src/api/bundle/spec/moonbeam.ts

packages/apps-config/src/api/spec/zero.ts
2075: done

packages/apps-config/src/api/spec/zero.ts
2075: ty

2075:
I have added `LookupSource` now.

Thanks for the hint with typesBundle, will check that out and follow up with the next minor release, ok?

jacogr:
100% on the bundle - it will just make your life easier when updates do come in. Just needs a resolve on the conflict as well.

2075:
ok, there you go.

jacogr (approved):
Thank you!

2075:
Kawabunga!
Likewise :)

Issue #4124, opened 2020-11-30
Help tool tips on inputs


jacogr:
It doesn’t work on the contract reply modal. Not sure if it is 

- ~~a local issue there~~
- a modal issue (general)
- ~~an all inputs affected error (general)~~

Logged as a reminder to take a look and fix

PR #4222, opened 2020-12-16
fix tooltip z-index

ii-ii-ii:
Hi,

I found that all modal tooltip is now not displayed.

I debugged it and found that the z-index defined by ui-Tooltip was overwritten.  Now the z-index is 999, which is less than the modal's z-index 1000, so tooltip is not displayed. So I manually set the z-index to 1002 again to solve the problem.


jacogr:
Thank you! Closes https://github.com/polkadot-js/apps/issues/4124

jacogr (approved):

PR #4231, opened 2020-12-17
Add create bounty modal

MiZiet:
Add new modal to propose bounties from app. Validation - to be added soon.


packages/apps/public/locales/en/app-bounties.json
```diff
@@ -1,11 +1,22 @@
 {
+  "Add Bounty": "Add Bounty",
+  "Add bounty": "Add bounty",
```
krzysztof-jelski, line 3: the one with capital 'B' should be enough

packages/page-bounties/src/BountyCreate.tsx
```diff
@@ -0,0 +1,91 @@
...
+  const bountyDepositPerByte = api.consts.treasury.dataDepositPerByte;
+
+  const [accountId, setAccountId] = useState<string | null>(null);
+  const [title, setTitle] = useState('');
+  const [bond, setBond] = useState(bountyDepositBase.toBn());
+  const [amount, setAmount] = useState<BN | undefined>(BN_ZERO);
+  const [isValid] = useState(true);
+  const [isOpen, toggleIsOpen] = useToggle();
+
+  const onTitleChange = (value: string) => {
```
jacogr, line 27: useCallback rather.

jacogr: ... which would probably meant you don't want to pass the consts `bountyDepositBase`, `bountyDepositPerByte` in, but rather the API. The former will change on each render, the later won't.

jacogr:
Happy as-is. If you want it in (with some work left), happy to get it in as-is since it does move things forward.

krzysztof-jelski:
Great, let's merge it in, then. We'll add validation in another PR.

jacogr (approved):
Excellent, nice move forward.

PR #4233, opened 2020-12-18
fix media query width

ii-ii-ii:
Because the media query width set in the table body is 1400 and the media query width set in the table head is 1500, there will be problems with the display when the width is between 1400 and 1500

https://github.com/polkadot-js/apps/blob/master/packages/page-staking/src/Targets/Validator.tsx#L116

image


jacogr (approved):
Thank you!

PR #4236, opened 2020-12-18
fix typo society pallet new bid page

ltfschoen:

PR #4240, opened 2020-12-19
society typo (all translation strings)

jacogr:


jacogr:
From @ltfschoen in #4236 - just expanded here.

ltfschoen:
i didn't do a very good job, i should have done a global find/replace for `canidate` and replaced with `candidate`

PR #4253, opened 2020-12-20
update encointer types, add encointer testnets to dropdown

brenzi:
* follow our updates to post-2.0.0 `MultiAddress`
* adjust logos
* add two testnets to dropdown: Gesell and Cantillon


jacogr:
For the lint fixes - if you run „yarn lint“ locally it will auto-fix what it can.

packages/apps-config/src/api/spec/encointer-node-notee.ts
```diff
@@ -6,9 +6,8 @@
 /* eslint-disable camelcase */
 
 export default {
-  Address: 'AccountId',
-  LookupSource: 'AccountId',
-  RefCount: 'u8',
+  Address: 'MultiAddress',
+  LookupSource: 'MultiAddress',
```
jacogr, line 10: Nice to see this in the wild :)

jacogr (approved):
Thank you.

Just as a suggestion for the future, you may want to look at versioned types (see config/api/bundle vs config/api/spec) - it allows you to apply types based on the spec version. Which also means it can be done pre-deployment, i.e. will only take effect when the spec version changes.

(It is the same way we manage the Kusama/Polkadot types - the format is slightly different, just being wrapped with a min/max, the moonbeam spec there is a good example)

All optional, but have found it makes people's lives easier.

PR #4262, opened 2020-12-22
Update API with frontier RPC/types support

jacogr:


packages/apps-config/src/api/bundle/spec/moonbeam.ts
```diff
@@ -23,58 +20,6 @@ export default {
           nonce: 'U256',
           balance: 'u128'
         },
-        TransactionCondition: {
-          _enum: {
-            block: 'u64',
-            time: 'u64'
-          }
-        },
-        Transaction: {
```
joelamouche, line 32: So I take it all occurrences of Transaction have been replaced with EthTransaction?

jacogr: Yes. On the metadata there is an alias (so occurrences in the metadata is transparently catered for)

PR #4264, opened 2020-12-23
Add PolkaBTC to block explorer

savudani8:
Integrates PolkaBTC Alpha (https://alpha.polkabtc.io/) with the block explorer


jacogr (approved):
Thank you!

PR #4276, opened 2020-12-28
Restore contracts banners size. Closes #4275

### LINKED ISSUE #4275: Article padding broken
Assuming is after the width adjustments in https://github.com/polkadot-js/apps/pull/4170

image

(Assuming this is the same in accounts and everywhere where these are displayed?)

MiZiet:
This changes should restore the size and paddings of contract's banner.

image


jacogr:
I think the class naming makes sense. It is also not broken in staking -> payouts, so we should probably just align where the warning is placed, instead of adding a class specifically for it. For instance, this is in the payouts -

image

So I think it may be a placement issue rather. Prefer consistency and since we are not applying the class everywhere, it means that page is different to the rest.

MiZiet:
image
So I extracted Contracts from ContractsApp (to have a consistency with other apps), I hope that's what you meant.

jacogr:
I was hoping it would be less effort, but this is great. Thank you.

jacogr:
Thank you, seems spot-on (like I said, more work that I would have liked for a quick-fix, but hopefully didn't take too much effort).

Just want to play though, it looks code-wise good, but just want to take a peek.

jacogr (approved):
Thank you!

PR #4281, opened 2020-12-29
Update Kulupu Address and LookupSource definitions

sorpaas:
The runtime just today was updated to include the new MultiAddress support https://github.com/paritytech/substrate/pull/7380 and hence we need to update the type definitions.


jacogr (approved):
Thank you.

It probably makes sense to move this to versioned types, i.e. based on the specVersion. Examples with min/max - https://github.com/polkadot-js/apps/blob/master/packages/apps-config/src/api/bundle/spec/moonbeam.ts

That means it can be done anytime before the upgrade and just takes effect,

PR #4284, opened 2020-12-29
UI tests for bounties

krzysztof-jelski:
Sorry if it's too massive, willing to split into smaller ones if desired.

Included:
- slow test (with substrate)
- fast test (with extracted context and mocked api)
- extract test support

Looking for feedback especially to the fast tests. I hate that we still need to mock the api, but quite a few low-level components depend on it. I think extracting more api-related stuff to Contexts is an option.


jacogr:
Cannot say that I'm terribly excited about the contexts here to work around testing. Contexts are typically global state, here we even tied something to a specific account in the case of Balances.

The Bounty one is slight more appealing, but it suffers from departing and not allowing the component to use the API directly. (Which is an issue since e.g. for the ongoing refactoring due to Substrate changes we generally search for `api.<tx>.<something>.<method>` in this case it will be missed completely. 

So I can see why, but thing the ongoing pain is worse than the small benefit.

krzysztof-jelski:
Thanks for the feedback.

I'm pretty sure it's exactly the opposite with the second issue you mention. Eg. with Contexts the api usage is much more localised and easier to find and change.

Contexts such as BountyContext can be limited to one app, which I think is pretty ok (not that much of a global state). As to accounts and balances, I have to think/play with it some more.

So what do you say, perhaps I split this PR in 2: one with extracting the test support and with the substrate test, and another one with the fast tests?

jacogr:
The reason why I'm saying it is not easy is that you are assuming no changes in names, modules, params. All of these change and for all of these there are support for multiple incarnations.

For instance, look at this example, the params changed between chains -

https://github.com/polkadot-js/apps/blob/master/packages/page-council/src/Motions/Close.tsx#L71-L78

Here the actual call names are different as well based on detection -

https://github.com/polkadot-js/apps/blob/master/packages/page-staking/src/Actions/NewValidator.tsx#L117

None of this is isolated, there are are multiple examples of one of both of these together, alongside the type of movement you are familiar with in module name changes. (Or in some cases modules have different names on different chains, e.g. elections vs electionsPhragmen)

So it seems localized, but we are removing the api from the components, not allowing detection and hoping that only one of the above does occur (like it already has).

krzysztof-jelski:
Extracted the fast test PoC to #4302, leaving only the substrate test here.

jacogr (approved):

PR #4290, opened 2020-12-29
Use derive call in Bounty's useCounter

MiZiet:
To correctly display the number of active bounties we should switch to `DeriveBounties` because `bountyCount` returns a number of all bounties ever created.

### DIFF
--- a/packages/page-bounties/src/useCounter.ts
+++ b/packages/page-bounties/src/useCounter.ts
@@ -8,7 +8,7 @@ import { useApi, useCall } from '@polkadot/react-hooks';
 
 export default function useCounter (): number {
   const { api, isApiReady } = useApi();
-  const bounties = useCall<DeriveBounties>(isApiReady && (api.derive.bounties || api.derive.treasury)?.bounties);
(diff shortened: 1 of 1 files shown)


packages/page-bounties/src/useCounter.ts
```diff
@@ -1,18 +1,17 @@
...
-
 import { useMemo } from 'react';
 
+import { DeriveBounties } from '@polkadot/api-derive/types';
 import { useApi, useCall } from '@polkadot/react-hooks';
 
 export default function useCounter (): number {
   const { api, isApiReady } = useApi();
-  const bounties = useCall<BountyIndex>(isApiReady && (api.query.bounties || api.query.treasury)?.bountyCount);
+  const bounties = useCall<DeriveBounties>(isApiReady && (api.derive.bounties || api.derive.treasury)?.bounties);
```
jacogr, line 11:
```suggestion
  const bounties = useCall<DeriveBounties>(isApiReady && api.derive.bounties?.bounties);
```

jacogr (approved):
Thank you!

PR #4296, opened 2020-12-29
use moonbeam types  npm package

joelamouche:
It's better to have a single source of moonbeam-specific types for both the apps and our tests.
Let me know if you'd prefer if the types were added on another level.


jacogr (approved):
Brilliant!

PR #4302, opened 2020-12-30
PoC for a fast UI bounty test.

krzysztof-jelski:


krzysztof-jelski:
Tried a different approach, with just mocking hooks. Should be simpler and more straightforward

packages/page-bounties/src/Bounties.tsx
```diff
@@ -1,23 +1,18 @@
 // Copyright 2017-2021 @polkadot/app-bounties authors & contributors
 // SPDX-License-Identifier: Apache-2.0
 
-import type { DeriveBounties } from '@polkadot/api-derive/types';
-import type { BlockNumber } from '@polkadot/types/interfaces';
-
 import React, { useRef } from 'react';
 
+import { useBounties } from '@polkadot/app-bounties/hooks';
```
jacogr, line 6: Just `./useBounties`? (it is local)

packages/page-bounties/src/hooks/useBounties.tsx
```diff
@@ -0,0 +1,42 @@
...
+export function useBounties (): BountyApi {
+  const { api } = useApi();
+  const bounties = useCall<DeriveBounties>((api.derive.bounties || api.derive.treasury).bounties);
+  const bestNumber = useCall<BlockNumber>(api.derive.chain.bestNumber);
+  const constsBase = api.consts.bounties || api.consts.treasury;
+  const bountyDepositBase = (constsBase.bountyDepositBase as BalanceOf).toBn();
+  const bountyValueMinimum = (constsBase.bountyValueMinimum as BalanceOf).toBn();
+  const maximumReasonLength = constsBase.maximumReasonLength.toNumber();
+  const dataDepositPerByte = (constsBase.dataDepositPerByte as BalanceOf).toBn();
+  const proposeBounty = (api.tx.bounties || api.tx.treasury).proposeBounty;
```
jacogr, line 31: Would not do this one, but the rest are correct.

packages/page-bounties/src/hooks/useBounties.tsx
```diff
@@ -0,0 +1,42 @@
...
+  bountyDepositBase: BN,
+  bountyValueMinimum: BN,
+  dataDepositPerByte: BN,
+  maximumReasonLength: number,
+  proposeBounty: ((...args: any[]) => SubmittableExtrinsic<'promise'>);
+};
+
+export function useBounties (): BountyApi {
+  const { api } = useApi();
+  const bounties = useCall<DeriveBounties>((api.derive.bounties || api.derive.treasury).bounties);
```
jacogr, line 24:
```suggestion
  const bounties = useCall<DeriveBounties>(api.derive.bounties.bounties);
```

Since we known the API we are working with here.

jacogr (commented):
Certainly much cleaner, looks pretty good..

jacogr (commented):
... ignore, GH hiccup....

jacogr (commented):
It think it is probably fine as a first round, but really shouldn't live too long in the exact form., The as-implemented hook-based approach has issues that should not spread. The `useBalance` hook is not really required, the functions (without filling them in) on the hook.

So the first case is explanatory, the second case is along the lines of -

Probably don't want to pass just "approve" through as a Submittable (it is already available on the api), however if the hook fully constructed it, it has real use. So basically the hook should not be a holder for the API, but actually apply"something".

So for instance for a close hook, this could be acceptable `const closeTx = useCloseHook(id)` where the closeHook constructs the extrinsic fully with params. (Which mean that any logic around that sits in the hook, not the component itself, it only passes data)

So happy to "get it in" (so the tests are there which has value), but certainly in the current form would not like to see it extend much in use.

jacogr (approved):

krzysztof-jelski:
Updated to latest master. Would happily see it merged.

My thoughts on your points below.

> It think it is probably fine hear and now as a first round, but really shouldn't live too long in the exact form.,

Definitely. I see the test suite and approach to making a codebase more testable as something that evolves as we write more and more tests, touching new use/test-cases. So I see this as a starting point as well.

> So basically the hook should not be a holder for the API, but actually apply"something".
> 
> So for instance for a close hook, this could be acceptable `const closeTx = useCloseHook(id)` where the closeHook constructs the extrinsic fully with params.

I see some value in having the hooks even if the are only a holder for the API.
 * testability itself, eg. the ability to inject values into the component under test in a clean way. I'm used to valuing that a lot, even over code cleanliness. Though usually those two go hand in hand.
 * a very much needed layer of abstraction. Reading the components code, I regard interacting with the API directly too low-level. Think of all those `.meta.args` checks. Component really looks better if it just expresses what state it needs and gets it from a hook/service or whatever intermediary out design provides. And I'm pretty sure that if we start extracting this layer, we're gonna find a lot of opportunities to reduce duplication (eg. the `.meta.args` checks on proposal creation) or extracting the domain logic outside from the view layer (eg. the derive-style transformations for proxies).

That being said, wherever possible the hook should do any transforms, param construction etc. And particularly I like the idea of `const closeTx = useCloseHook(id)`, gonna play around with that.

jacogr:
Feel free to merge as-is. Not too fond of PRs hanging around forever - especially in cases like these where they do add real value. (Even if not 100%)

PR #4304, opened 2020-12-30
Allow anyone to close a successful council proposal and minor text ch…

lovesh:
…ange.

As per the council pallet, [anyone can close](https://github.com/paritytech/substrate/blob/master/frame/collective/src/lib.rs#L612) a successful council proposal.

Signed-off-by: lovesh <lovesh.bond@gmail.com>

### DIFF
--- a/packages/page-council/src/Motions/Close.tsx
+++ b/packages/page-council/src/Motions/Close.tsx
@@ -14,11 +14,10 @@ interface Props {
   hasFailed: boolean;
   hash: Hash;
   idNumber: ProposalIndex;
-  members: string[];
   proposal: Proposal;
 }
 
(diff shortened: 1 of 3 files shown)


jacogr:
Thank you, the close part is unlikely to change - the reason for this is that the close operation will swap to the members-only `close_operational` (Substrate PR 7470) as soon as that is ready. (Obviously will fallback to the fully open for older chains).

lovesh:
Hi Jaco.

> close operation will swap to the members-only close_operational

Do you mean `close` will be replaced with `close_operational` in the apps UI?

jacogr:
Yes. The collective close will go through the members-only `close_operational`. With a fallback for older, so the logic will be more of less the following -

```js
tx={api.tx.council.closeOperational || api.tx.council.close}
```

(Same in tech comm)

lovesh:
Thanks.

PR #4311, opened 2021-01-03
Not truncating address in Sidebar

eddiemachado:
I removed the ellipsis on the address so you can see the whole thing. Also added a bit of padding.

I'd love to contribute more UX / UI improvements :D


jacogr:
Thank you! PRs are always welcome - there are a lot of improvements all around. 

Will take a look tomorrow morning, my brief code-only-look-through comment seems that it does what it says on the tin, no concerns with it as it stands.

jacogr (approved):
Seems good, thanks a lot for this.

Improvements would be to split almost-in-half, i.e. if the address width is around `24ch` it seems more visually pleasing, i.e. no single chars on a line. (The addresses are funny since it can actually vary in length, so you won't get a full 50% split in all cases)

PR #4317, opened 2021-01-04
Add new menu styling

MiZiet:
Hey, here's a new approach to menu [design](https://www.figma.com/file/5NAICV06SHNbbIoYhrUS3u/PolkaDot?node-id=748%3A1189) implementation.

image


krzysztof-jelski:
Supercedes #3810

jacogr (commented):
Will need to play through first to see what it does. Not entirely convinced by the screenshots with the tabs on top of each other (i.e. active white blurs into menu white), but really need to click before seeing what it does in-depth.

packages/apps/src/Menu/ChainInfo.tsx
```diff
@@ -91,16 +91,24 @@ export default React.memo(styled(ChainInfo)`
...
-        font-size: 0.9rem;
-        line-height: 1.2;
+        max-width: 9rem;
+        white-space: nowrap;
+        overflow: hidden;
+        text-overflow: ellipsis;
+        font-weight: 600;
+        font-size: 1.15rem;
+        line-height: 1.5rem;
+        color: #FFFFFF;
```
jacogr, line 104: This has exceptionally weird side-effects. (as the the chain name). The issue is that the font color adapts to the background, on dark it is white on light it is dark gray.

So what happens here now is that as we swap chains, the color changes from being aligned to not being aligned.

(See for instance Amber vs Cantilon testnets)

packages/react-components/src/styles/index.ts
```diff
@@ -32,7 +32,7 @@ function getContrast (uiHighlight: string | undefined): string {
 
   return brightness > BRIGHTNESS
     ? 'rgba(45, 43, 41, 0.875)'
-    : 'rgba(255, 253, 251, 0.875)';
+    : 'rgb(255, 255, 255)';
```
jacogr, line 35: We don't use pure-white anywhere. It creates too much contrast with the very dark backgrounds. (Hence the fade)

packages/apps/src/Menu/ChainInfo.tsx
jacogr: Example (the white here is very wrong for the light background) -

image

UI inconsistent with the above just after swapping, above different, this the same -

image

jacogr (commented):
Some comments.

- the menus left-right squashed on each other is distracting when going from e.g. networks to the next
- the top-left changes are inconsistent with the UI feel (apart from the comments made below on it being inconsistent and even unreadable on light highlight colors and too jarring due to the white on dark)
- the tabs look very squashed now top-bottom, probably needs more padding

Overall not sure this is a really needed change, but if it aligns willing to try. Some things are cool, i.e. the mini tab hover colors. Not better, just different.

krzysztof-jelski:
Hi @jacogr, working on improvements. Just not sure what you mean here:
> the top-left changes are inconsistent with the UI feel
Can you elaborate?

jacogr:
@krzysztof-jelski I believe that refers to the bold applied there. (Looks out of place since it is not headers or titles)

ekowalsk:
Hi Jaco, could you give it another look after the last changes?

jacogr:
Thank you. I will play-through late this afternoon, there are really some nice concepts in here.

jacogr:
Some comments -

- the node info on the right is missing (this is important info and appears in reports the whole time)
- the tabs not blending into each other is much nicer
- something is weird with the unselected font, it doesn't match the UI, seems bolder? (have not checked the code, but it looks weird, off and inconsistent). Same with the menu fonts now as well, not sure what is weird here, it actually lmost looks like a different font.
- tabs now seem to have the same issue as above?
- not 100% sure about the text vertical alignment in the tabs, something is slightly off, not sure if it is too narrow now of something else
- the left icon is not vertically centered
- we now cut off the chain name in a very short fashion
- the left spacing of the text is not appealing at all, there is a gap between the name and the info and it looks/feels weird to me
- the actual chain info adjustments now make in unreadable, I cannot see the chain best block at a glance
- the badges on the menu dropped the color, it now has a dark color - it needs to fit with the chain theme, as it was
- believe the menu right `>` should be in the theme color to align with the rest (it is too light, the color here will add without making it too much)

Those are the issues that detract I spotted. Two things are nice if all the rest gets dropped -

- the menubar on the left with `>` (colored)
- the top menus acting like menus

jacogr:
image

- pushed a highlight for the actual divider stroke (class addition on the stroke)
- the badge color is not aligning with the theme
- when on a non-menu option it is very confusing to have a menu hover background (see settings above - only when active navigation-wise, the white works, not before). Same with Wiki/Github.
- the right search bar seems non-vertically-centered

Please keep this layout below, "Rococo" above does not seem to align with anything anywhere in terms of fonts - the fact that is it larger gives exactly the emphasis needed (the layout here with the different parts above looks very messy imho, since it is not consistent)

image

I do understand we want to start adding "bold" everywhere, but it is not used at all. The sizes dictate extra emphasis.

jacogr:
Comments on the current version -

- It still shows non-menu items as menu items, e.g. 
image
- For the above Settings is also confusing, since there is no dropdown associated with it
- The fonts not looks the same as the rest, that is a great improvement
- The shortened line actually looks weird in some cases, see the Metadata option here, you expect it to align with the text, e.g. 
image
- Dark here looks much better than light, i.e. the menu doesn't blend with the tabs, e.g
image
vs
image
- The above is quite weird, it is not introduced here but it seems more noticeable here, especially if the tabs become long, here is an example of it just almost blending in light
image
- this happens on smaller screens
image

jacogr:
Current version comments:

- the explorer shows this -

image

- much better

image

- however this is off, when you go to settings and then to the menu on the left of it, it is inconsistent. Basically the eye expects consistency with a menu dropdown and doesn't get it at all. It should be aligned with the above format where there is no menu, i.e. it is clickable.

image

- This is mis-aligned, I understand it is not an easy problem, but this happens in a number of places, i.e. everywhere where there is a counter -

image

- not convinced about this, it also looks weirder now in dark mode. Rather the whole menu (including header) should be a different share, like it is in the dark mode, instead f the highlight. The highlight brings some other problems where if the first item is highlighted, it seems like the spacing is off -

image

- This really looks mis-aligned. I'm not sure it is, but believe the line makes it seem that way, or something else - it really doesn't appear centered at all -

image

natiwa:
@jacogr Hey please take a look at this options. 
Let me know are these a good direction? 
Kusama@2x
Light menu v1@2x
Light menu v2@2x

jacogr:
Didn't think on this option, but at least for the light the first one actually looks very appealing. The top-bar actually works really well, so to have the same dark styling for the menu works great. It solves a lot of issues -

- never blends with the second-bar
- is consistent with the top

jacogr:
Just a small comment to the above. We probably need to (instead of going 15% darker, or which never works), go lighter for the menu. i.e. option 2 is topColor + x%, I would probably suggest topColor - x%

The reasoning being that Kusama is #000 - so it cannot go darker, but anything lighter would apply. (The fully light as per the second light theme option I'm not that fond of, it is quite tricky get get it right across all palettes, so instead of very light, slight light should, hopefully, do the trick)

Alternatively, if darker works the best for everything else, I'm sure just a shadow would also solve "the Kusama problem".

natiwa:
@jacogr there is one issue behind going lighter, the main color does not always look good with a white overlay. 
Here is an example with 15% white opacity on pink. 
Light menu v1a@2x

natiwa:
But I have to admit that for yellow background brighter overlay looks way better :) 
Light menu yellow B@2x
Light menu yellow D@2x

jacogr:
The pink is always an issue... Could do something like we do with the text color detection - on dark it goes white, on light it gos dark.

So in this case for the pink (white text), make the background darker, on the yellow (black text), make the background lighter.

For Kusama, well, let's just hope that the border shadows are enough. (Since that one is a problem child)

jacogr:
PS: The split above between the "hover part" and the "menu items" is actually quite interesting. So each item (menu or not) will have a consistent "hover part" and then where a menu is available it will show independent. 

Need to see it is action, but seems to be really nice from a consistency pov.

jacogr (approved):
Brilliant! this is excellent and really an improvement on what is there.

Issue #4273, opened 2020-12-28
Add "send to council" from bounties


jacogr:
It would do a council proposal with `approveBounty(bountyId)` - at this point council can vote.

threshold is the same as for treasury proposals.

krzysztof-jelski:
How about allowing council members to approve/reject and vote without leaving the Bounties page? Compare https://www.figma.com/file/TnSJFgORMMwJAnKRxHHh7W/Bounty?node-id=368%3A67
What do you think about it?

jacogr:
The issue is that the council members go to the council page for their actions. 

The alternative is that they need to go to bounties and treasury and council to vote. And that mean you lose focus and drop stuff. 

It is the same reason the treasury voting is only under council now - the previous split voting just didn’t work for active council members. 

Once created it shows up under council anyway since it is a council approve/reject, so no need to duplicate.

natiwa:
hey, 

I think that from the user perspective it will be great to have one consistent flow for similar actions. Now, we can vote "Nay" or "Yay" on the modal. Having the same experience for one step earlier (approving or rejecting the bounty/proposal) will make navigating the application intuitive for the user. 

Now it might sound like a lot of changes have to be done, but it may be worth implementing.

krzysztof-jelski:
@jacogr by the way, what was "the previous split voting" you mentioned?

natiwa:
In my opinion, it’s more convenient for the user to have voting visible in these two places. Duplicating actions it’s not a bad thing, it’s making an application more approachable for the user.

jacogr:
So the voting was duplicated in Treasury & Council. (It will always appear in council since it is a council proposal). This was exceptionally confusing -

- council didn't just have  place to vote
- worse, the same votes showed in 2 places and allowed actions

So the slight issue atm is that the council proposal breakdown is a bit generic, so in the case of treasury it should actually just show the value/beneficiary. (And allow a link to the treasury proposal extenally). So there will be some UI duplication.

jacogr:
@natiwa No. As a council member and actual user of this functionality, I respectfully disagree. It is exceptionally confusing. It is not something that should make any comeback.

krzysztof-jelski:
OK, I think that sounds good. Can we just specify exactly what status do we want to show in the Bounties list? Is what we have in designs accurate, or do we need more/less info? 
![image](https://user-images.githubusercontent.com/430616/103287162-18a78380-49e2-11eb-8776-b5d9bc21834e.png)

[...]

PR #4320, opened 2021-01-05
Initiate voting from bounty list - accept / reject bounty

ekowalsk:
This PR adds initiate voting functionality from the bounty list, which enables accepting or rejecting bounty for council members. It uses bounty proposals implementation from api, that wasn't merged yet.


jacogr:
I'm not quite on-board with having the voting directly here, as explained earlier. It is an absolute nightmare trying to align the voting functionality between the apps. This is the reason it was removed from treasury.

As suggested, a summary view (non-action) would be ok, but anything voting related should stick to council.

krzysztof-jelski:
There's no voting in this PR (or planned). This is just equivalent to "To council" in treasury proposals. Though following the designs we decided to name things differently, thus "Initiate voting" on the button.

packages/page-bounties/src/BountyInitiateVoting.tsx
```diff
@@ -0,0 +1,113 @@
...
+  const approveBountyProposal: SubmittableExtrinsic<'promise'> = (api.tx.bounties || api.tx.treasury).approveBounty(index);
+  const closeBountyProposal: SubmittableExtrinsic<'promise'> = (api.tx.bounties || api.tx.treasury).closeBounty(index);
+
+  const isVotingInitiated = useMemo(() => proposals?.filter(({ proposal }) => BOUNTY_METHODS.includes(proposal.method)).length !== 0, [proposals]);
+
+  return isMember && !isVotingInitiated
+    ? (
+      <>
+        <Button
+          icon='plus'
```
jacogr, line 46: Would at least align the icon with treasury, quite skittish of having the same type of functionality different everywhere.

packages/page-bounties/src/BountyInitiateVoting.tsx
```diff
@@ -0,0 +1,113 @@
...
+            <Modal.Actions onCancel={toggleOpen}>
+              <TxButton
+                accountId={accountId}
+                icon='check'
+                isDisabled={false}
+                label={t<string>('Approve')}
+                onStart={toggleOpen}
+                params={api.tx.council.propose.meta.args.length === 3
+                  ? [threshold, approveBountyProposal, approveBountyProposal.length]
+                  : [threshold, approveBountyProposal]
```
jacogr, line 89: In this case you can stick to the first only - bounties are only on current chains so doesn't need to align with older.

packages/page-bounties/src/BountyInitiateVoting.tsx
```diff
@@ -0,0 +1,113 @@
...
+                tx={api.tx.council.propose}
+              />
+              <TxButton
+                accountId={accountId}
+                icon='ban'
+                isDisabled={false}
+                label={t<string>('Reject')}
+                onStart={toggleOpen}
+                params={api.tx.council.propose.meta.args.length === 3
+                  ? [threshold, closeBountyProposal, closeBountyProposal.length]
```
jacogr, line 100: As above, only this.

krzysztof-jelski:
Linting fails because this PR relies on newer API (including proposal in derived bounties). Should be ok after that lands.

jacogr:
Just merge master, see https://github.com/polkadot-js/apps/pull/4321

packages/page-bounties/src/BountyInitiateVoting.tsx
```diff
@@ -0,0 +1,107 @@
...
+  const [threshold, setThreshold] = useState<BN>();
+
+  useEffect((): void => {
+    members && setThreshold(
+      new BN(Math.ceil(members.length * getTreasuryProposalThreshold(api)))
+    );
+  }, [api, members]);
+
+  const approveBountyProposal: SubmittableExtrinsic<'promise'> = (api.tx.bounties || api.tx.treasury).approveBounty(index);
+  const closeBountyProposal: SubmittableExtrinsic<'promise'> = (api.tx.bounties || api.tx.treasury).closeBounty(index);
```
jacogr, line 38: Don't believe you want to create this each render? (They are quite heavy, a new class instance). Rather create the full proposal - probably in the effect above. then just pass it as `extrinsic={approve/close}` instead of `tx/params`

jacogr:
Closes #4273

packages/page-bounties/src/BountyInitiateVoting.tsx
ekowalsk: Would the useRef do the work?

jacogr (approved):
Thank you, seems sane.

PR #4334, opened 2021-01-07
add jupiter testnet

zzcwoshizz:
hi there,

this PR is to add jupiter testnet.

greetings!


packages/apps-config/src/api/spec/jupiter.ts
```diff
@@ -0,0 +1,9 @@
+// Copyright 2017-2021 @polkadot/apps-config authors & contributors
+// SPDX-License-Identifier: Apache-2.0
+
+// structs need to be in order
+/* eslint-disable sort-keys */
+
+export default {
+  LookupSource: 'MultiAddress'
```
jacogr, line 8: Shouldn't we also have `Address: 'MultiAddress`?

atenjin: Hi @jacogr 
We define for `type Address` is same to `node-template` define. Thus there are two points related to this:
1. `node-template` have not changed the define of 
    ```json
    types: {
      Address: 'AccountId',
      LookupSource: 'AccountId'
    }
    ```
    We think this should be changed after the pr https://github.com/paritytech/substrate/pull/7380 in substrate, 

    using:
    ```json
    {
      "LookupSource": "MultiAddress",
      "Address": "LookupSource"
    }
    ```
    We may create a pr to fix this if you will, or you may have some  your own considerations.
2. In our chain, using `polkadot.js` would use default config, thus, in `node` config, the default already has:
    ```json
    "Address": "LookupSource"
    ```
    thus we just override
    ```json
    LookupSource: 'MultiAddress'
    ```
     is ok. But you are right, override both `LookupSource` and `Address` may be better, we will change later, thanks!

zzcwoshizz:
Hi, @jacogr .

I have added the ```Address: 'MultiAddress'```types.

jacogr (approved):
Thank you

PR #4336, opened 2021-01-07
Add bounties summary

MiZiet:
Hey, here's the summary sections for app-bounties. 
image


jacogr (approved):
Looks good, thank you.

The treasury burn should not be applicable here, so I would suggest it is removed. (It has no impact on the bounties at all)

PR #4339, opened 2021-01-07
font-weight: 400 -> theme.fontWeightNormal

jacogr:
~~Doesn't quite catch all, didn't want to go through and add all the themeProps where missing~~


jacogr:
We probably want to move the theme props into css vars.

PR #4340, opened 2021-01-07
Allow Ledger support for Dock

lovesh:
Signed-off-by: lovesh <lovesh.bond@gmail.com>

### DIFF
--- a/packages/apps-config/src/api/constants.ts
+++ b/packages/apps-config/src/api/constants.ts
@@ -20,3 +20,5 @@ export const POLKADOT_GENESIS = getGenesis('polkadot');
 export const POLKADOT_DENOM_BLOCK = new BN(1248328);
 
 export const KULUPU_GENESIS = getGenesis('kulupu');
+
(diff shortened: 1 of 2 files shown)


packages/apps-config/src/api/constants.ts
```diff
@@ -20,3 +20,5 @@ export const POLKADOT_GENESIS = getGenesis('polkadot');
 export const POLKADOT_DENOM_BLOCK = new BN(1248328);
 
 export const KULUPU_GENESIS = getGenesis('kulupu');
+
+export const DOCK_GENESIS = '0xf73467c6544aa68df2ee546b135f955c46b90fa627e9b5d7935f41061bb8a5a9';
```
jacogr, line 24: It is available, see https://github.com/polkadot-js/common/blob/master/packages/networks/src/index.ts#L229 

We don't want o maintain the chain-specific stuff here, i.e. just get it from the same route as above. (The networks package determines what is available for ss58/genesis here and in the extension)

TL;DR `getGenesis('dock-mainnet')`

lovesh: Thanks for reminding me. Fixed.

jacogr (approved):
Thank you!

Issue #4349, opened 2021-01-09
Desktop App Fails to Connect to Local Node Due to Content Security Policy


zicklag:
Hey there!

I just tried to connect the desktop app Appimage to a local polkadot node started by running `polkadot --dev`, but the app throws a console error because it is apparently configured with a content security rule that requires all websocket connections to be `wss`.

![image](https://user-images.githubusercontent.com/25393315/104079193-5c229e80-51e7-11eb-8130-86a048dd446b.png)
![image](https://user-images.githubusercontent.com/25393315/104079199-65137000-51e7-11eb-976f-e95b43df32cd.png)

jacogr:
Thanks for the report. https://github.com/polkadot-js/apps/blob/master/packages/apps-electron/src/electron/contentSecurityPolicy.ts#L16 wold need to be extended to `ws`.

PR #4350, opened 2021-01-09
Update CSP for ws

jacogr:
Closes https://github.com/polkadot-js/apps/issues/4349

cc @krzysztof-jelski - would like you eyes on this as well


krzysztof-jelski:
Looks good. BTW we don't need those Google Fonts URLs anymore, do we?

krzysztof-jelski (approved):

jacogr:
Yes, you are 100% correct - although the couple of semantic-ui components left (& css) still does Google fonts (which we override anyway), but just worried that may break something in there.

PR #4360, opened 2021-01-11
Added Plasm Network Rococo parachain endpoint

akru:
plasm-parachain-pc2

**Changes**

* Added parachain collator endpoint
* Added chain color association
* Added node name association


jacogr (commented):
Thank you.

Will merge as soon as the registration is open and it has been added.

PR #4361, opened 2021-01-11
Remove colon from win path

### LINKED ISSUE #4348: Polkadot/Substrate Portal forgot accounts in Windows 10 app
I'm use Parity Polkadot v0.8.26 api v3.3.1 apps v0.74.1 on Windows 10 x64 Rus
I new user of Polkadot. I'm download and install app for Windows there days ago. After I add two new accounts in app by Accounts\Accounts menu:
image

Now if close app and open it one more, app don't remember my accounts. I need push "+Add account", enter my seed phrase and password and restore it from Internet everytime. Is it normal situation or bag?

krzysztof-jelski:
Fixes #4348 

Perhaps should be rather handled in [file keystore](https://github.com/polkadot-js/ui/blob/master/packages/ui-keyring/src/stores/File.ts ), but this workaround is good for now.

### DIFF
--- a/packages/apps-electron/src/main/account-store.ts
+++ b/packages/apps-electron/src/main/account-store.ts
@@ -13,12 +13,20 @@ import { registerIpcHandler } from './register-ipc-handler';
 
 const ACCOUNTS_SUBFOLDER = 'polkadot-accounts';
 
+function safeWriteKey (key: string) {
+  return process.platform === 'win32' ? key.replaceAll(':', '-') : key;
+}
+
+function safeReadKey (key: string) {
+  return process.platform === 'win32' ? key.replaceAll('-', ':') : key;
+}
+
 export const accountStoreIpcHandler = (fileStore: FileStore): IpcMainHandler => ({
   'account-store-all': () => {
     let result: { key: string, value: KeyringJson }[] = [];
 
     const collect = (key: string, value: KeyringJson) => {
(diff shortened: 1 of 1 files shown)


jacogr (commented):
We should probably just apply it to all without platform detection. (Basically suggesting always replacing)

Agreed that this is a "good enough" step.

packages/apps-electron/src/main/account-store.ts
```diff
@@ -13,12 +13,20 @@ import { registerIpcHandler } from './register-ipc-handler';
 
 const ACCOUNTS_SUBFOLDER = 'polkadot-accounts';
 
+function safeWriteKey (key: string) {
+  return process.platform === 'win32' ? key.replaceAll(':', '-') : key;
```
jacogr, line 17:
```suggestion
  return key.replaceAll(':', '-');
```

packages/apps-electron/src/main/account-store.ts
```diff
@@ -13,12 +13,20 @@ import { registerIpcHandler } from './register-ipc-handler';
...
+}
+
+function safeReadKey (key: string) {
+  return process.platform === 'win32' ? key.replaceAll('-', ':') : key;
```
jacogr, line 21:
```suggestion
  return key.replaceAll('-', ':');
```

jacogr (approved):

PR #4363, opened 2021-01-11
add onfinality rococo endpoint

ianhe8x:
I have to abandon the usage of `linked`, otherwise can not add another provider for rococo.

### DIFF
--- a/packages/apps-config/src/endpoints/testing.ts
+++ b/packages/apps-config/src/endpoints/testing.ts
@@ -15,7 +15,7 @@ import { createRococo } from './testingRococo';
 export function createTesting (t: TFunction): LinkOption[] {
   return [
     // polkadot test relays
-    createRococo(t),
+    ...createRococo(t),
     // alphabetical based on chain name
     {
       info: 'centrifuge',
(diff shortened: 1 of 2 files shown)


jacogr:
Thank you. 

Yes, that seems like a glaring oversight on my part. Was actually expecting just another Rococo entry (without the linked entries) to just "work" since there is still filtering. 

It could make sense to turn `textBy` and `value` into Array. Which is what it tries to do anyway flatenning these. It is a bit messy atm, there is way too much stuff to carry across when adding, hence trying to make it (slightly) simpler with the linked.

jacogr:
I'm playing a bit, see if I can cleanup.

jacogr:
This should solve the issue - https://github.com/polkadot-js/apps/pull/4365

However... I've also found another workaround for this which I'll get in first. (Since there is already one parachain waiting.) So basically with the work-around can get in OnFinality ASAP and then the above one can wait a bit until thing quiet down. (Which is not going to be today)

jacogr:
Immediate workaround in #4366 - will get that in and close this one.

Thank you!

ianhe8x:
Thank you Jaco for supporting us. The ticket was created in midnight before I went to bed and in the morning it all has been merged.

jacogr:
No issues, this was def. an oversight on my part - so the outcome is all good, I believe it is actually much, much cleaner now with less duplication - https://github.com/polkadot-js/apps/blob/master/packages/apps-config/src/endpoints/testingRococo.ts#L21-L24

The config stuff is nowhere 100%, this brings it one step closer to being "the right fit for the problem".

PR #4396, opened 2021-01-13
Render extended statuses when voting on bounty

ekowalsk:


packages/page-bounties/src/VotingDescription.tsx
```diff
@@ -0,0 +1,41 @@
...
+      )}
+    </>
+  );
+}
+
+export default React.memo(styled(VotingDescription)`
+    font-size: 0.7rem;
+    color: #8B8B8B;
+    margin-top: 10px;
+`);
```
jacogr, line 41: OCD, the indent here looks weird.... 4 spaces? tab?

packages/page-bounties/src/VotingDescription.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { getVotingDescription } from './helpers/extendedStatuses';
+import { useTranslation } from './translate';
+
+interface Props {
+  className?: string;
+  proposals: DeriveCollectiveProposal[];
+  status: BountyStatus;
+}
+
+function VotingDescription ({ className = '', proposals, status }: Props): JSX.Element {
```
jacogr, line 19:
```suggestion
function VotingDescription ({ className = '', proposals, status }: Props): React.ReactElement<Props> {
```

Just aligns with what we do everywhere

packages/page-bounties/src/helpers/extendedStatuses.ts
```diff
@@ -0,0 +1,59 @@
...
+  CuratorProposed: ['unassignCurator', 'closeBounty'],
+  Funded: ['proposeCurator', 'closeBounty'],
+  PendingPayout: ['unassignCurator'],
+  Proposed: ['approveBounty', 'closeBounty']
+};
+
+function votingDescription (proposalName: string | undefined, t: TFunction): string | null {
+  switch (proposalName) {
+    case ('approveBounty'):
+      return t('Approval under voting');
```
jacogr, line 23: These are problematic - becuase it is outside of a `useTranslation`, it doesn't get included in the `en/app-bounties.json`, which in turn means it won't appear for translation at all.

packages/page-bounties/src/helpers/extendedStatuses.ts
```diff
@@ -0,0 +1,59 @@
...
+    case ('proposeCurator'):
+      return t('Curator under voting');
+    case ('unassignCurator'):
+      return t('Unassign curator under voting');
+    default:
+      return null;
+  }
+}
+
+function bestValidMethod (bountyProposals: DeriveCollectiveProposal[], status: BountyStatus) {
```
jacogr, line 35: Function should always have a type specifier for the return value.

krzysztof-jelski: Strange that linter didn't pick it up 😞

jacogr (approved):
Thank you! Have not played, but it looks completely sane.

PR #4408, opened 2021-01-14
fix Galois cannot construct AccountServiceEnum

hging:
fix cannot construct AccountServiceEnum


packages/apps-config/src/api/chain/galois.ts
```diff
@@ -0,0 +1,42 @@
+// Copyright 2017-2021 @polkadot/apps-config authors & contributors
+// SPDX-License-Identifier: Apache-2.0
+
+// structs need to be in order
+/* eslint-disable sort-keys */
+
+export default {
+  Address: 'MultiAddress',
```
jacogr, line 8: Not 100% sure why there is a spec & chain entry - this means the types would be included twice. Only the spec should be needed. (chain is only used in the case where a chain has the same spec, but is vastly different, be it for testing purposes or otherwise. In that case the specVersioned is actually preferred)

jacogr (commented):
Thank you! One comment around duplication.

packages/apps-config/src/api/chain/galois.ts
hging: because our chain runtime name is wrong. but it can't change.
so we need to use chain.
so i will remove the spec files

jacogr (approved):
Thank you!

PR #4416, opened 2021-01-14
Add create bounty script

MiZiet:
Hi, here's a little script that creates a list of bounties on a local substrate. Merging this will make developing and testing bounties faster and easier for us :)


jacogr (approved):
Cool :)

PR #4419, opened 2021-01-15
mandala pc2

ntduan:
acala rococo

I encountered an error while debugging. I'm not sure if it's because acala isn't connected to rococo yet.

image

### DIFF
--- a/packages/apps-config/src/endpoints/testingRococo.ts
+++ b/packages/apps-config/src/endpoints/testingRococo.ts
@@ -52,6 +52,14 @@ export function createRococo (t: TFunction): EndpointOption {
       },
       // add any additional parachains here, alphabetical
       {
+        info: 'rococoAcala',
+        paraId: 5000,
+        text: t('rpc.rococo.acala', 'Mandala PC2', { ns: 'apps-config' }),
+        providers: {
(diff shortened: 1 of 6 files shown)


jacogr:
Will look at that - that is not your issue not cause by you, I think we are missing an exclusion there, will take a look.

packages/apps-config/src/endpoints/testingRococo.ts
```diff
@@ -51,6 +51,14 @@ export function createRococo (t: TFunction): EndpointOption {
         }
       },
       // add any additional parachains here, alphabetical
+      {
+        info: 'rococoAcala',
+        paraId: 5000,
```
jacogr, line 56: Not sure this is correct, 666?

jacogr (commented):
Thank you! One small comment.

packages/apps-config/src/endpoints/testingRococo.ts
jacogr, line 56:
```suggestion
        paraId: 666,
```

jacogr (approved):
Going to merge as-is without a deploy. Will push a fix fo the Id.

PR #4420, opened 2021-01-15
Added robonomics rococo parachain endpoint

akru:
**Description**

* Fix chain types
* Added rococo parachain endpoint


jacogr (approved):
Thank you! This looks great - I'll merge alongside the chain being approved. (Probably actually sooner since it may be a bad idea to let these just queue up)

PR #4427, opened 2021-01-15
Fix dock logos being blurry/low res

SamHellawell:
Description

- Updated Dock mainnet and testnet logos to be of higher resolution


jacogr (approved):
Thank you.

PR #4435, opened 2021-01-17
[Hardware Wallet Support Grant Milestone 2] ECDSA Custom Signature Pallet UI implementation

hoonsubin:
## Feature Summary

> Related to [this grant milestone](https://github.com/w3f/Grant-Milestone-Delivery/pull/83)

This pull request adds the Custom Signature page under the Account tab.
The Custom Signature page allows the user to submit a transaction that was signed with an Ethereum private key and use it as a native ECDSA account in Substrate. Plus, you are able to use MetaMask for signature requests, which allows users to use their Ethereum address inside Ledger and Trezor to create a transaction.

### Ledger

https://user-images.githubusercontent.com/40356749/104814290-7137a700-5851-11eb-8cef-00b9495da0d6.mp4

### Trezor

https://user-images.githubusercontent.com/40356749/127802497-e7aba206-6c28-403a-89f3-f46d4689469b.mov

IMG_0796

Screen Shot 2021-01-15 at 1 06 30 AM

## Notable Changes

- add `assert` browser polyfill that is used by the `ethereumjs-util` package
- add basic zero-dependency MetaMask support to the Custom Signature page (only works on that page)
- add `custom-signature` page that will be available for chains that implement the `custom-signatures` pallet (https://github.com/PlasmNetwork/Plasm/tree/dusty/frame/custom-signatures) as `EthCall` (https://github.com/PlasmNetwork/Plasm/blob/feature/vectorized-custom-signatures/bin/node/runtime/src/lib.rs#L649)
- add Ethereum account to SS58 ECDSA account converter

## Usage

Currently, this page is made with MetaMask in mind. Most browser-injected ethereum wallets that use the `window.ethereum` object and accept the `personal_sign` method should be compatible, but please understand that there may be unexpected errors.

You can learn more about the usage from this doc: https://docs.plasmnet.io/build/metamask-signatures


jacogr (commented):
It is great that it works, well done. Having said that, there are a couple of issues I see -

1. It is not actually integrated with the signer, so we have a completely seperate process
2. Accounts are completely seperate from the account accounts page

To do it correctly, when the call is available, it needs to allow addition on the Accounts page.

Then when sent, the Signer will detect the account and perform wrapping as required.

An additional concern is the way the extension is integrated, see for instance https://github.com/polkadot-js/extension/pull/566 which the uses the existing extension interface.

hoonsubin:
thank you for your feedback @jacogr.
My original plan was to isolate the MetaMask instance used within the Custom Signature page to prevent any conflicts with other Ethereum extension implementation done by other teams (plus this is a `custom-signatures` pallet specific feature), but having full integration with the existing account component instead of a separate hook would definitely be consistent and clean for everyone. I'll see what I can do to improve on that!
Thanks again.

hoonsubin:
whoops. I shouldn't have rebased it. Sorry, this was an amateur mistake on my side. I'll close this PR and open a clean one later!

PR #4439, opened 2021-01-18
Add subscan rococo extra link

freehere107:


packages/apps-config/src/links/subscan.ts
```diff
@@ -23,7 +23,8 @@ export default {
     Plasm: 'plasm',
     Polkadot: 'polkadot',
     Stafi: 'stafi',
-    Westend: 'westend'
+    Westend: 'westend',
+    Rococo: 'rococo'
```
jacogr, line 27: Thank you. Looks good, just needs to be moved alphabetical here, before Stafi (have not checked the CI failures, but I do believe that is why it is complaining)

jacogr (approved):
Thank you!

PR #4441, opened 2021-01-18
chore: add kilt endpoint to rococo

wischli:
As KILT is the next candidate to join the Rococo-v1 network as a parachain, we would be happy if our endpoint could be added as well.


jacogr (approved):
Thank you!

Orginally intended to let these pile up and merge as available, but earlier is better since it creates no conflicts. (Plus it shows in the proposals as well, so it shows you are doubly-prepared :))

It is really nice having the endpoints, even if you are just using the parachains view - since it can get stuff like the best block directly from the chain.

Issue #3788, opened 2020-09-28
(FR) add any address / remember account after unplugging Ledger


drandreaskrueger:
Great function, that "Add Ledger" in '[accounts](https://polkadot.js.org/apps/#/accounts)'. Very easy to use, and comfortable. Just one wish:
### Feature Request: remember account address

When I unplug the Ledger Device, my browser always forgets the related account, not even the address is remembered.

What about: Offering an "**Add read-only account**" to the buttons, then I could simply add the address of _any account that also contains my coins_ (with perhaps in column 'type' as 'read-only' or 'address only'?).  Or when the Ledger is plugged in, and an account had been added via "Add Ledger ... Save", then one more menu (three vertical dots) option right below "Forget this account" - but with a new function "**Remember this account**" (and then in column 'type' saying 'ledger, unplugged' or simply still 'ledger')

Then e.g. my balance and other info is always there, regardless whether I have plugged in the device or not.

Thanks for considering this small enhancement. Best regards.

(P.S.: If the Ledger has already timed out, and is in "vires in numeris / PIN code" mode, and I then try to "Add Ledger ... Save", I get an "Unknown Status Code: 26628" error message. Probably you want to catch that.)

Swader:
When you add a Ledger account to your apps UI, you will have it automatically added to your Address Book (under contacts). You can see info about the account there, including balance.

drandreaskrueger:
You mean here: https://polkadot.js.org/apps/#/addresses ?  

Yes automatically would be good. I have now added my ledger addresses there, manually.

Swader:
It appeared for me automatically, but it may have happened because I renamed my account (clicking on the edit icon in the account's sidebar).

anjaw3f:
Is there a reason why the accounts page does not remember the Ledger account? It remembers all the other accounts that are created with external tools. This is confusing a lot of users.

jacogr:
It is not usable. So remembering it there means it needs to be usable- which means that everything else needs to have the Ledger attachment logic. 

So basically to do the above we need to introduce an account inactive state and consistently handle it in the UI. 

(This one alongside the other open Ledger PRs are quite delayed since atm I don’t have a Ledger anymore - horrible accident involving kids, so none of these can be tested and features can’t be done blind. So while I wait for Ledger no 3, no movement on any features)

chevdor:
Not a solution but it may help people running into the issue. It also will not reach the usability threshold from @jacogr but it works as woraround atm.

[...]

PR #4444, opened 2021-01-18
Check for Ledger available status

jacogr:
closes https://github.com/polkadot-js/apps/issues/3788#issuecomment-762499748

Needs https://github.com/polkadot-js/ui/pull/431 (aka Ledger account is always visible under accounts)

@Tbaut Would appreciate a check... still flying "slightly" blind...


packages/react-signer/src/TxSigned.tsx
```diff
@@ -271,26 +286,29 @@ function TxSigned ({ className, currentItem, requestAddress }: Props): React.Rea
     (): void => {
       setBusy(true);
 
-      setTimeout((): void => {
+      setTimeout(():void => {
```
jacogr, line 289:
```suggestion
      setTimeout((): void => {
```

packages/react-signer/src/TxSigned.tsx
```diff
@@ -274,23 +289,26 @@ function TxSigned ({ className, currentItem, requestAddress }: Props): React.Rea
...
-          errorHandler(error as Error);
-        }
+        _unlock()
+          .then((isUnlocked): void => {
+            if (isUnlocked) {
+              isSubmit
+                ? currentItem.payload
+                  ? _onSendPayload(queueSetTxStatus, currentItem, senderInfo)
+                  : _onSend(queueSetTxStatus, currentItem, senderInfo).catch(errorHandler)
+                : _onSign(queueSetTxStatus, currentItem, senderInfo).catch(errorHandler);
```
jacogr, line 304: The catch here should be the one below, however I really didn't want to mess around too much with previously working.

Tbaut (approved):
- Add ledger
- refresh, is still there
- unplug it
- send a tx from it, the browser complains it finds no device 
- plug in with the browser popup still on, the device shows up
- click send again, it shows up on the ledger.

I also tried the same as above, but when plugin it again, I just unlocked it, and did not enter the app, then the signer complains, I entered the app on the ledger and clicked sign again, and it's sent to the ledger as expected.

:100:

jacogr:
@Tbaut Thanks a lot for playing my QA team - this should hopefully sort it itself out by tomorrow (if all goes well). Extra eyes are always great, but it is horrible the untested stuff crosses to your side.

PR #4447, opened 2021-01-18
Fix #4419 - using bounties without treasury

ekowalsk:
Fixes #4419

### DIFF
--- a/packages/react-hooks/src/useTreasury.ts
+++ b/packages/react-hooks/src/useTreasury.ts
@@ -27,7 +27,7 @@ export function useTreasury (): Treasury {
   const [burn, setBurn] = useState<BN>();
 
   const treasuryBalance = useCall<DeriveBalancesAccount>(api.derive.balances.account, [TREASURY_ACCOUNT]);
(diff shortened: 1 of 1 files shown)


packages/react-hooks/src/useTreasury.ts
```diff
@@ -27,9 +27,13 @@ export function useTreasury (): Treasury {
   const [burn, setBurn] = useState<BN>();
 
   const treasuryBalance = useCall<DeriveBalancesAccount>(api.derive.balances.account, [TREASURY_ACCOUNT]);
-  const spendPeriod = api.consts.treasury.spendPeriod;
+  const spendPeriod = api.consts.treasury ? api.consts.treasury.spendPeriod : api.createType('BlockNumber', 0);
```
jacogr, line 30:
```suggestion
  const spendPeriod = api.consts.treasury ? api.consts.treasury.spendPeriod : BN_ZERO;
```

Just the BN_ZERO import from `@polkadot/util` (since we don't rely on it being an actual BlockNumber, just a BN, all fine). Generally try to steer clear of `createType` unless _really_ needed.

jacogr (commented):
One small comment, otherwise looks good.

jacogr (approved):

PR #4455, opened 2021-01-20
Bump moonbeam-types-bundle version to support upcoming v5

joelamouche:


joelamouche:
🤔 never seen that error before... Can't figure out how to rerun the CI. Also I tried running `npm run lint` and it linted 2 files I had't touched

jacogr:
You just need to add the `yarn.lock` file.

PR #4458, opened 2021-01-20
Updated wasm & util-crypto (incl. chain types bumps)

jacogr:
Replaces #4455


packages/page-accounts/src/modals/Transfer.tsx
```diff
@@ -68,7 +68,7 @@ function Transfer ({ className = '', onClose, recipientId: propRecipientId, send
   const noReference = accountInfo
     ? isRefcount(accountInfo)
       ? accountInfo.refcount.isZero()
-      : (accountInfo.consumers.isZero() && accountInfo.providers.isZero())
+      : accountInfo.consumers.isZero()
```
jacogr, line 71: The providers check fails, on a new account it is 1.

PR #4460, opened 2021-01-20
Adjust components placement in bounties list

MiZiet:
Hi, a few changes that will bring the app-bounties closer to the design.
image


jacogr:
Excellent, great to see more and more movement. There are a couple of issues, just looking at the screenshots, take with a grain of salt -

1. we don't want to row de-linianated, but rather colored on even/odd (although, this may be worth a try here, but not something that can be applied to any busy lists list the visual noises becomes overwhelming, aka we have been there) - so not a show stopper, can keep, but one unless proven on larger lists, it probably will have to change at some point

2. The title should be column 1, it is the identifier for the actual bounty. status is secondary to "what it is for"

3. labels are all caps here, that is a bit inconsistent

4. Couple to the above we jump to Aye to AYE in a couple of places, which is inconsistent even here

5. Not convinced on the bordered down button, but probably something that can justified here since t have "some effect" - just mentioning

Will take a peek at the code in the morning.

packages/page-bounties/src/Bounties.tsx
```diff
@@ -51,8 +56,56 @@ function Bounties (): React.ReactElement {
...
+export default React.memo(styled(Bounties)`
+  .bounties-table-wrapper table {
+    background: none;
+
+    &::before {
+      background: none;
+    }
+
+    tr {
+
```
jacogr, line 72:
```suggestion
```

packages/page-bounties/src/Bounties.tsx
```diff
@@ -51,8 +56,56 @@ function Bounties (): React.ReactElement {
...
+    opacity: 1;
+  }
+
+  .ui--Table th:first-child h1 {
+    padding: 0;
+    font-weight: bold;
+    font-size: 0.7rem;
+    line-height: 0.85rem;
+    text-transform: uppercase;
+    color: #4D4D4D;
```
jacogr, line 104: Not sure this is intentional? It is different from all other headers and will mean we need to circle back when making any updates?

packages/page-bounties/src/Bounties.tsx
```diff
@@ -51,8 +56,56 @@ function Bounties (): React.ReactElement {
...
+    font-size: 1rem;
+    line-height: 1.7rem;
+    text-transform: initial;
+    filter: initial;
+    opacity: 1;
+  }
+
+  .ui--Table th:first-child h1 {
+    padding: 0;
+    font-weight: bold;
```
jacogr, line 100: This should use the theme boldness?

packages/page-bounties/src/Bounties.tsx
```diff
@@ -51,8 +56,56 @@ function Bounties (): React.ReactElement {
...
+        }
+      }
+    }
+  }
+  
+  .ui--IdentityIcon {
+    margin-right: 0.42rem;
+  }
+
+  .via-identity .name {
```
jacogr, line 90: Not sure this is correct - it has a consistent styling in the UI?

packages/page-bounties/src/Bounty.tsx
```diff
@@ -55,144 +58,254 @@ function Bounty ({ bestNumber, bounty, className = '', description, index, propo
...
+          {blocksUntilPayout
+            ? <DueBlocks
+              dueBlocks={blocksUntilPayout}
+              until={'payout'}
+            />
+            : ''}
+          {blocksUntilUpdate
+            ? <DueBlocks
+              dueBlocks={blocksUntilUpdate}
+              until={'update'}/>
```
jacogr, line 88:
```suggestion
          {blocksUntilUpdate
            ? <DueBlocks
              dueBlocks={blocksUntilUpdate}
              until={'update'}
            />
```

Consistency with the block just above and elsewhere.

packages/page-bounties/src/Bounty.tsx
```diff
@@ -55,144 +58,254 @@ function Bounty ({ bestNumber, bounty, className = '', description, index, propo
...
+          <div className='fast-actions-row'>
+            <LinkExternal
+              data={index}
+              isLogo
+              type='bounty'
             />
+            <div className='table-column-icon'
+              onClick={handleOnIconClick}>
+              <Icon
+                icon={
```
jacogr, line 120: Since attribute can go just next to `Icon`

packages/page-bounties/src/Bounty.tsx
```diff
@@ -55,144 +58,254 @@ function Bounty ({ bestNumber, bounty, className = '', description, index, propo
...
 export default React.memo(styled(Bounty)`
+  .description-column {
+    max-width: 200px;
+
+    div {
+      white-space: nowrap;
+      overflow: hidden;
+      text-overflow: ellipsis;
+    }
+  }
```
jacogr, line 203:
```suggestion
  }

```

packages/page-bounties/src/Bounty.tsx
```diff
@@ -55,144 +58,254 @@ function Bounty ({ bestNumber, bounty, className = '', description, index, propo
...
   & .links {
     display: inline-flex;
   }
+  & .fast-actions {
```
jacogr, line 207:
```suggestion

  & .fast-actions {
```

packages/page-bounties/src/Bounty.tsx
```diff
@@ -55,144 +58,254 @@ function Bounty ({ bestNumber, bounty, className = '', description, index, propo
...
+    display: flex;
+    align-items: center;
+  }
+
+  .td-row {
+    display: flex;
+    justify-content: space-between;
+    align-items: center;
+  }
+  .bounty-action-row {
```
jacogr, line 285:
```suggestion

  .bounty-action-row {
```

packages/page-bounties/src/Bounty.tsx
```diff
@@ -55,144 +58,254 @@ function Bounty ({ bestNumber, bounty, className = '', description, index, propo
...
+    display: flex;
+    justify-content: flex-end;
+    align-items: center;
+    margin-left: auto;
+
+    & > * + * {
+      margin-left: 0.6rem;
+    }
+  }
+  .block-to-time {
```
jacogr, line 295:
```suggestion

  .block-to-time {
```

packages/page-bounties/src/Description.tsx
```diff
@@ -0,0 +1,28 @@
...
+      {description}
+    </div>
+  );
+}
+
+export default React.memo(styled(Description)`
+  margin-top: 0.28rem;
+  font-size: 0.7rem;
+  line-height: 0.85rem;
+  color: #8B8B8B;
```
jacogr, line 27: Not sure the hardcoded color is intentional? (e.g. what does it relate to elsewhere)

packages/page-bounties/src/Voting/VotersColumn.tsx
```diff
@@ -0,0 +1,79 @@
...
+  className?: string;
+  option: 'ayes' | 'nays';
+  proposals: DeriveCollectiveProposal[];
+  status: BountyStatus;
+}
+
+const styleOptions = {
+  ayes: {
+    icon: 'check',
+    label: 'AYE:'
```
jacogr, line 24: We need this translatable.

packages/page-bounties/src/Voting/VotersColumn.tsx
```diff
@@ -0,0 +1,79 @@
...
+  },
+  nays: {
+    icon: 'times',
+    label: 'NAY:'
+  }
+};
+
+function VotersColumn ({ className, option, proposals, status }: Props): JSX.Element {
```
jacogr, line 32:
```suggestion
function VotersColumn ({ className, option, proposals, status }: Props): React.ReactElement<Props> {
```

packages/page-bounties/src/Voting/VotersColumn.tsx
```diff
@@ -0,0 +1,79 @@
...
+  const votes = useMemo(() => option === 'ayes' ? proposal?.votes?.ayes : proposal?.votes?.nays, [proposal, option]);
+
+  const voters = useMemo(() => votes?.map((accountId) =>
+    <div
+      className='voter'
+      data-testid={`voters_${option}_${accountId.toString()}`}
+      key={accountId.toString()}
+    >
+      <AddressSmall value={accountId}/>
+    </div>), [option, votes]);
```
jacogr, line 43:
```suggestion
    </div>),
    [option, votes]
  );
```

packages/page-bounties/src/Voting/VotersColumn.tsx
```diff
@@ -0,0 +1,79 @@
...
+export default React.memo(styled(VotersColumn)`
+  .vote-numbers {
+    display: flex;
+    align-items: center;
+    margin-bottom: 0.85rem;
+  }
+
+  .vote-numbers-icon svg {
+    max-width: 10px;
+    color: #9E9E9E;
```
jacogr, line 68: Colors?

packages/page-bounties/src/Voting/VotingSummary.tsx
```diff
@@ -0,0 +1,57 @@
...
+      )}
+    </>
+  );
+}
+
+export default React.memo(styled(VotingSummary)`
+  .voting-summary-text {
+    font-size: 0.85rem;
+    line-height: 0.5rem;
+    color: #1A1B20;
```
jacogr, line 50: Colors? (This probably won't work at all on dark)

packages/page-bounties/src/index.tsx
```diff
@@ -38,4 +40,77 @@ function BountiesApp ({ basePath }: Props): React.ReactElement<Props> {
...
 
-export default React.memo(BountiesApp);
+export default React.memo(styled(BountiesApp)`
+  && thead th {
+    padding: 0.75rem 1rem 0.57rem;
+    font-weight: bold;
+    font-size: 0.7rem;
+    line-height: 0.85rem;
+    text-transform: uppercase;
+    color: #4D4D4D;
```
jacogr, line 50: ?

packages/page-bounties/src/index.tsx
```diff
@@ -38,4 +40,77 @@ function BountiesApp ({ basePath }: Props): React.ReactElement<Props> {
...
+    border: none;
+    background: none;
+  }
+
+  tbody td {
+    padding: 1.4rem 1rem;
+    vertical-align: baseline;
+    font-size: 1rem;
+    line-height: 1.7rem;
+    color: #1A1B20;
```
jacogr, line 64: As above.

packages/page-bounties/src/index.tsx
```diff
@@ -38,4 +40,77 @@ function BountiesApp ({ basePath }: Props): React.ReactElement<Props> {
...
+    &:nth-child(odd) {
+      background: #fff;
+    }
+
+    &:nth-child(odd) td {
+      vertical-align: middle;
+    }
+
+    &:nth-child(even) {
+      background: #fafafa;
```
jacogr, line 86: This is really light-only.

jacogr (commented):
I don't think this move in the right direction. I'm not on-board with but willing to give it a try before ripping it out.

An example is the specific table styling here (as mentioned) for instance there are certainly ways to still allow even/odd row coloring while inserting a row to look like the parent - may mean passing explicit isOdd/isEven classes through, but not impossible.

TL;DR Not going to stand completely in the way of this, but really itching to give it a make-over.

krzysztof-jelski:
Thanks for the comments!

Replies below.

> 1. we don't want to row de-linianated, but rather colored on even/odd (although, this may be worth a try here, but not something that can be applied to any busy lists list the visual noises becomes overwhelming, aka we have been there) - so not a show stopper, can keep, but one unless proven on larger lists, it probably will have to change at some point

Let's keep it like this for a while. We'll see how that plays in the Bounties and can decide to keep or change later.

> 2. The title should be column 1, it is the identifier for the actual bounty. status is secondary to "what it is for"

OK, we'll do this change.

> 3. labels are all caps here, that is a bit inconsistent

I understand you mean column headers and labels in the expanded row.

The design idea behind is to have it consistent with the styling in Extension. I know it's different than in other tables. Again, I'd propose we stick with the new design for a while, and see whether we go with updating other tables in this manner, or the other way around.

> 4. Couple to the above we jump to Aye to AYE in a couple of places, which is inconsistent even here

Here the "Aye/Nay" is used in what we call "Info column" (place for general info about the bounty, dependant on the status). The "AYE/NAY" in the expanded view are more like other labels. So those are different component roles, hence different capitalisation.

If you're not ok with, we could change the ones in Info to "AYE/NAY".

> 5. Not convinced on the bordered down button, but probably something that can justified here since t have "some effect" - just mentioning

I'm really afraid that users won't know that the rows can be expanded. The border draws the attention a bit more, so I'd keep it. I even had this idea of making the whole role clickable, similar to https://yearn.finance/vaults.

jacogr:
Te apps UI and extension are very different animals - the styling does not overlap between the two :)

So on the capitilization, the issue is exactly the extension -

- in the extension labels are all capital
- in the apps UI it is never used, everything is lowercase

On the bordering -

- in the extension explicit borders are added
- in the apps UI it tries to be a (mostly) border-less layout (assuming that it is done so the explicit borders are not needed)

So I really would not confuse the two at all.

jacogr (approved):
It can go as-is since it adds functionality which is great for moving things ahead. (As discussed, not keen on a number of the visual elements at all since for instance borders, explicit buttons, headers outside of the tables have all been tried for quite some time and discarded)

I will create a PR alongside when I have a gap to re-align with the Apps UI, really interested to see the collapse/expand work with consistent row highlights since it is certainly a great solution in a number of places.

packages/apps/src/themes.ts
```diff
@@ -26,10 +28,14 @@ export const darkTheme: ThemeDef = {
   bgToggle: '#58595f',
   borderTable: 'transparent',
   borderTabs: 'transparent',
+  bountyBGTableOddRow: '#111218',
+  bountyBorderTable: '#2F313C',
+  bountyColor: '#eeeeee',
```
jacogr, line 33: Note sure about these here at all - since they are only bounty-specific and won't be used anywhere so doesn't have applicability in the globals.

I suppose the right pattern in bounties would therefore be `${theme == 'dark' ? '#111218' : '#fff'}`

This localizes it to bounties where it departs from the standard. (If there is every more use it can be extracted into the global)

jacogr: Looks all good with the recent changes.

PR #4464, opened 2021-01-21
fix Chinese translation of 'comm.'

pfcoder:
string 'comm.' is used in https://github.com/polkadot-js/apps/blob/0360df1ccf47886c1247d0fa6d67578b230bbcb6/packages/page-staking/src/Targets/index.tsx#L274

here it is abbr of 'commission' which in Chinese should be '佣金'


jacogr (approved):
Thank you.

PR #4468, opened 2021-01-22
Added Phala PC1 Rococo V1 endpoint

tolak:
Hi, we'd like to add our rococo v1 testnet entry to the endpoint list, pls approve.


jacogr (approved):
Thank you.

Issue #4078, opened 2020-11-26
Displaying account balances on a custom chain


tomusdrw:
I presume account balance on the balances tab is coming from `frame_system` storage.

For some cases, when the chain supports multiple tokens there are two instances of `balances` pallet and neither should be favoured, like here:
https://gitlab.com/interlay/btc-parachain/-/blob/dev/parachain/runtime/src/lib.rs#L267

Is there a way to configure the UI to display balances from different `AccountStorage`?

To test, you can connect to: wss://alpha.polkabtc.io/api/parachain
<details><summary>and use these custom types</summary>
<pre>
{
 "H256Le": "H256",
    "DOT": "u128",
    "PolkaBTC": "Balance",
    "BTCBalance": "u128",
    "StatusCode": {
      "_enum": ["Running", "Error", "Shutdown"]
    },
    "Address": "AccountId",
    "LookupSource": "AccountId",
    "Status": "StatusCode",
    "ErrorCode": {
      "_enum": [
        "None",
        "NoDataBTCRelay",
        "InvalidBTCRelay",
        "OracleOffline",
        "Liquidation"
      ]
    },
    "VaultStatus": {
      "_enum": ["Active", "Liquidated", "CommittedTheft"]
    },
    "RawBlockHeader": {
      "0": "[u8; 80]"
    },
    "RichBlockHeader": {
      "block_hash": "H256Le",
      "block_header": "BlockHeader",
      "block_height": "u32",
      "chain_ref": "u32"
    },
    "BlockHeader": {
      "merkle_root": "H256Le",
      "target": "U256",
      "timestamp": "u32",
      "version": "i32",
      "hash_prev_block": "H256Le",
      "nonce": "u32"
    },
    "BlockChain": {
      "chain_id": "u32",
      "start_height": "u32",
      "max_height": "u32",
      "no_data": "BTreeSet<u32>",
      "invalid": "BTreeSet<u32>"
    },
    "Wallet": {
      "addresses": "BTreeSet<H160>",
      "address": "H160"
    },
    "Vault": {
      "id": "AccountId",
      "to_be_issued_tokens": "PolkaBTC",
      "issued_tokens": "PolkaBTC",
      "to_be_redeemed_tokens": "PolkaBTC",
      "wallet": "Wallet",
      "banned_until": "Option<BlockNumber>",
      "status": "VaultStatus"
    },
    "ActiveStakedRelayer": {
      "stake": "DOT"
    },
    "StatusUpdate": {
[...]

PR #4477, opened 2021-01-22
Extend PolkaBtc definitions (balances modules)

jacogr:
Needs https://github.com/polkadot-js/api/pull/3094

This add the new `interfaces` endpoint to the PolkaBtc bundle. This allows to add overrides for the modules, at this point is is only used/implemented inside balances derives.

Part of https://github.com/polkadot-js/apps/issues/4078


packages/apps-config/src/api/specVersioned/polkabtc.ts
```diff
@@ -5,4 +5,10 @@ import type { OverrideBundleDefinition } from '@polkadot/types/types';
 
 import definitions from '@interlay/polkabtc-types';
 
-export default definitions as OverrideBundleDefinition;
+// This ia a temp work-around to add the interfaces key to the actual types
```
jacogr, line 8:
```suggestion
// This is a temp work-around to add the interfaces key to the actual types
```

jacogr:
It only displays the primary atm, but a start - it needs the other PRs and pulling them through before it will display both. (And needs some UI TLC as well - actually they are returned already, so at least display-wise the UI work for that is independent)

image

PR #4484, opened 2021-01-24
"isolate opacity change to format balance error text"

jhonalino:
Screen Shot 2021-01-24 at 9 23 08 AM
Screen Shot 2021-01-24 at 9 24 39 AM

I noticed the opacity changed made it harder to see the emphasis on the whole numbers. this update isolates the change to only the error text


jacogr (approved):
Thank you.

PR #4488, opened 2021-01-25
Equilibrium networks account tab hotfix

pr0fedt:


jacogr (approved):
Sorry about this. It was added since the base derives now support retrieving from any balance location as well as having the balances module nam changable. See https://github.com/polkadot-js/apps/issues/4078#issuecomment-765704835

Since you are overriding the vesting module, we can actually make that configurable next in the same fashion in the API which would remove the need for this. Will play around a bit and see if I can get it to work (only much later today)

PS: Just needs a linting fix.

PR #4499, opened 2021-01-26
Add slash curator action

MiZiet:
Hi, this changes will add an action to unassign a curator by council, user (in the "inactive curator" scenario) or the curator himself. 

image


jacogr (approved):
Thank you.

PR #4500, opened 2021-01-26
Hide bounty unhappy path actions

ekowalsk:
Hides unhappy path actions like close bounty or reject curator under "show more" button


jacogr (commented):
Lgtm, but CI is not happy.

ekowalsk:
After the merge of slash curator actions I've added them as an extra action under the "show more" button and most recently I enabled multiple slash curator actions if the user has several roles, for example curator and council member at the same time

jacogr (approved):

PR #4510, opened 2021-01-27
Small bounty UI adjustments

jacogr:
image

- title always goes first - it is what users look at
- align the value formatting with the UI, the small font here is exceptionally hard to read
- make the time left the prominent info (block countdown is less than useful, it is technical, time has meaning to users and was previously secondary and hard to read in small font)
- add i18n as applicable (strings were missed and/or assumed English ordering)
- ~~ensure that the expanded row matches the parent background formatting~~ (Reverted isOdd/isEven for now since the overrides here are all non-standard - will level them until we align these containers with the rest of the UI for consistency)


krzysztof-jelski:
Thanks, looks good. Just regarding the title/status: we intentionally left it like this, as status is the first visual cue for filtering bounties (as in: you're likely to search for a particular bounty by its status).

jacogr:
Then statusses can be -

- grouped
- put in different containers (which is great since it allows you specific info _only_ to that in a consistent fashion)

krzysztof-jelski:
OK, just note we add some additional info under the status (eg. "Curator under voting" or "Claimable").

Issue #4370, opened 2021-01-11
Chain State tab only visible when accounts are present


JoshOrndorff:
It is a common use case that I open Apps to inspect the chain state with no intentions to submit transactions. In this case, clicking through the "add account" workflow costs several seconds.

jacogr:
Indeed. This was added after your first issue where having no accounts renders wrong for items where an account selector is needed. As of now, it is unlikely to change since any rework to the account selector is not short-term on the table. At the same time doing the type inspection on all items for filtering the list will yield queries around "this is not showing" and be expensive to boot.

So the options atm are -

- having it half broken where accounts are required
- not having it account queries are in a large number of items

Currently it sits in the second bucket.

PR #4514, opened 2021-01-27
Allow storage without accounts

jacogr:
Closes https://github.com/polkadot-js/apps/issues/4370

### DIFF
--- a/packages/react-components/src/InputAddress/index.tsx
+++ b/packages/react-components/src/InputAddress/index.tsx
@@ -14,6 +14,7 @@ import { createOptionItem } from '@polkadot/ui-keyring/options/item';
 import { isNull, isUndefined } from '@polkadot/util';
 
 import Dropdown from '../Dropdown';
+import Static from '../Static';
 import { getAddressName } from '../util';
 import addressToAddress from '../util/toAddress';
 import createHeader from './createHeader';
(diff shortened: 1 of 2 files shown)


packages/react-components/src/InputAddress/index.tsx
```diff
@@ -138,7 +139,17 @@ class InputAddress extends React.PureComponent<Props, State> {
...
-      return null;
+      // This is nasty, but since this things is non-functional, there is not much
+      // we can do (well, wrap it, however that approach is deprecated here)
+      return (
+        <Static
+          className={className}
+          help={help}
+          label={label}
+        >
+          No accounts are available for selection.
```
JoshOrndorff, line 150: Any chance we could get a text field to paste in an account we copied from somewhere else (like a block explorer, node log, text message, etc)?

I know I'm asking a lot here, and I'm already happy with how this is. Just thought I'd share the idea.

jacogr: No. The issue is that it can be multiple -

- just accounts
- just addresses
- both of them

So there needs to be logic to check the type and then swap to the simple input component. Maybe in the future when the component is reworked, but atm piling more stuff onto something (I believe) creaky is not an area I'm willing to wander into.

PR #4527, opened 2021-01-29
Display bounty ID

MiZiet:
Hi, since in council motions bounties are identified by ID, we should display ID in app-bounties.
image


jacogr (approved):
Thank you. 

It should actually be prominent (like ids are generally displayed), since it also links to external sources and it is the one link that everybody has. But better having it than not.

PR #4542, opened 2021-01-30
Add Zenlink Rococo parachain endpoint

Hayden0323:


packages/apps-config/src/api/spec/zenlink.ts
```diff
@@ -0,0 +1,28 @@
+// Copyright 2017-2021 @polkadot/apps-config authors & contributors
+// SPDX-License-Identifier: Apache-2.0
+
+export default {
+  AssetId: {
```
jacogr, line 5: I would just suggest adding the `Address` and `LookupSource` types explicitly. The reason is that the defaults are changing on the API layer, see https://github.com/polkadot-js/api/releases/tag/v3.5.1

(One of these types https://polkadot.js.org/docs/api/FAQ#i-cannot-send-transactions-sending-yields-address-decoding-failures)

packages/apps-config/src/api/spec/zenlink.ts
```diff
@@ -0,0 +1,28 @@
+// Copyright 2017-2021 @polkadot/apps-config authors & contributors
+// SPDX-License-Identifier: Apache-2.0
+
```
jacogr, line 3:
```suggestion

/* eslint-disable sort-keys */

```

We need to add this line for the type files. I think your last2 linting cleanups cleanup broke the ordering because of this - it does make sense elsewhere, but for types since it is a binary format in the SCALE-encoded blob, the ordering as originally specified is critical.

jacogr (commented):
Thank you. Some small comments.

packages/apps-config/src/api/spec/zenlink.ts
Hayden0323: Make sense. Thank you!

jacogr (approved):
Thank you!

PR #4544, opened 2021-01-30
Added ChainX logos & endpoints

qinghuan-chain:


packages/apps-config/src/api/spec/chainx.ts
```diff
@@ -0,0 +1,446 @@
+/* eslint-disable header/header */
```
jacogr, line 1:
```suggestion
// Copyright 2017-2021 @polkadot/apps-config authors & contributors
// SPDX-License-Identifier: Apache-2.0

```

packages/apps-config/src/api/spec/chainx.ts
```diff
@@ -0,0 +1,446 @@
...
+/* eslint-disable sort-keys */
+// [object Object]
+// SPDX-License-Identifier: Apache-2.0
```
jacogr, line 4:
```suggestion

```

packages/apps-config/src/api/spec/chainx.ts
```diff
@@ -0,0 +1,446 @@
...
+export default {
+  Chain: {
```
jacogr, line 6: I would just suggest adding the `Address` and `LookupSource` types explicitly. The reason is that the defaults are changing on the API layer, see https://github.com/polkadot-js/api/releases/tag/v3.5.1

(One of these types https://polkadot.js.org/docs/api/FAQ#i-cannot-send-transactions-sending-yields-address-decoding-failures)

jacogr (commented):
Thank you very much. Some small comments, but looks good otherwise.

packages/apps-config/src/api/spec/chainx.ts
qinghuan-chain: Ok. Thanks for your advice! it has been repaired.

jacogr: No issues, just trying to make sure you don't have pain later :)

jacogr (approved):
Thank you!

PR #4564, opened 2021-02-01
Update bounty summary

MiZiet:
Hey, this changes will remove `burn` from summary in app-bounties, since burn doesn't affect on bounties.
image


jacogr (approved):
Bit props for removing more lines than added - those are always the best PRs :)

PR #4586, opened 2021-02-03
Remove type prop from address inputs in app-bounties

MiZiet:
It will fix the problem with selecting accounts in propose curator and award bounty actions.


jacogr (approved):
Thank you.

PR #4594, opened 2021-02-04
edits to text in the implementation - bounties UI

rrtti:
Some changes in the text on display. To this, let me add:

- "Bounty approval under voting" - "Bounty rejection under voting" info should be located to the right, under "Voting results"
- The button to assign a curator to a funded bounty should read "Propose curator"
- "Curator under voting" info should be located to the right, under "Voting results"


jacogr (commented):
Thank you. this should actually be edited directly in the JS code, not in the translation file. (The translation files are generated from the JS source itself)

It is fine doing it here, but a dev would need to just transpose, otherwise it will get lost on merge.

krzysztof-jelski:
> Thank you. this should actually be edited directly in the JS code, not in the translation file. (The translation files are generated from the JS source itself)

Yup, will handle it.

krzysztof-jelski:
Resolved via #4615

PR #4597, opened 2021-02-04
Add SubDAO parachain info to Rococo.

chenwei767:
Add [SubDAO](https://github.com/SubDAO-Network) parachain info to Rococo.


packages/apps-config/src/ui/colors.ts
```diff
@@ -115,7 +117,8 @@ export const chainColors: Record<string, any> = [
   ['Galois', chainGalois],
   ['ZΞRO Alphaville', chainZero],
   ['Zenlink PC1', chainZenlink],
-  ['Uniarts', chainUniarts]
+  ['Uniarts', chainUniarts],
+  ['SubDAO PC1', chainSubDAO]
```
jacogr, line 121: Would have preferred this and the next one alphabetical, but will fix up afterwards - seems I was not awake on some and the order has gone slightly messy above.

jacogr (approved):
Thank you.

packages/apps-config/src/ui/colors.ts
chenwei767: I fix it right now to end up the mess.

https://github.com/polkadot-js/apps/pull/4602

PR #4606, opened 2021-02-05
Refactor the Architecture towards reusability and scalability, fix the Monorepo structure

achimcc:
I think its great that canvas-ui is using a Monorepo structure and follows an architecture pattern. But the local Yarn workspaces are currently broken, that is the current architecture as described by the current local sub-repo dependencies:

canvas-ui-dep-graph-original

Now, after adding the local dependencies to each package, this is the current apps architecture:

canvas-ui-dependencies

Actually, there are some black arrows in this diagram indicating cyclic references. I figured out that one main reason for this is that we have two stores in the app, one for persisting app internals like the session id, the other one is storing the Smart Contract related data, and rename it from `store` to e.g. `contractStore`.

I suggest to isolate the second store in its own repo first. Afterwards, we could make the underlying layers of the architecture as agnostic as possible from the store. This would allow for a better re-usablity, testability and scalability of the components and methods from the lower layers.


jacogr:
Closing, this is not canvas-ui.

PR #4657, opened 2021-02-11
Add Litentry Parachain info to Rococo

chenzongxiong:


packages/apps-config/src/endpoints/testingRococo.ts
```diff
@@ -211,6 +211,14 @@ export function createRococo (t: TFunction): EndpointOption {
         providers: {
           TrustBase: 'wss://rococo.trustednodes.net'
         }
+      },
+      {
```
jacogr, line 215: Can you just move this alphabetical? (I know, I let a couple sneak in...)

jacogr (commented):
One small comment, looks great otherwise. Thank you.

packages/apps-config/src/endpoints/testingRococo.ts
jacogr: I need to move the rest anyway, will include it there.

jacogr (approved):

packages/apps-config/src/endpoints/testingRococo.ts
chenzongxiong: I re-ordered all info correctly

PR #4693, opened 2021-02-18
Add unique network types, logo, and colors

gregzaitsev:


packages/apps-config/src/api/spec/unique.ts
```diff
@@ -0,0 +1,116 @@
...
+
+import type { OverrideBundleDefinition } from '@polkadot/types/types';
+
+// structs need to be in order
+/* eslint-disable sort-keys */
+
+const definitions: OverrideBundleDefinition = {
+  types: [
+    {
+        "AccessMode": {
```
jacogr, line 12: This should auto-fix if you run `yarn lint` locally.

gregzaitsev: Thanks, Jaco! Running it now...

packages/apps-config/src/api/spec/unique.ts
```diff
@@ -0,0 +1,116 @@
+// Copyright 2017-2021 @polkadot/apps-config authors & contributors
+// SPDX-License-Identifier: Apache-2.0
+
+import type { OverrideBundleDefinition } from '@polkadot/types/types';
+
+// structs need to be in order
+/* eslint-disable sort-keys */
+
+const definitions: OverrideBundleDefinition = {
+  types: [
```
jacogr, line 10: The bundle definitions look slightly different since they cater for multiple versions, i.e. you can actually add already before being upgraded. (Also allows you to add additional RPCs, derives, etc...)

The format is as follow (which is why the build fails) -

```js
const definitions: OverrideBundleDefinition = {
  types: [
    {
      // on all versions
      minmax: [0, undefined],
      types: {
        Address: 'AccountId',
        LookupSource: 'AccountId',
        // all other types goes here (the above examples)
      }
    }
  ]
};

export default definitions;
```

So basically what you have it right for the `types` key, it just needs one more level of nesting.

jacogr (approved):
Thank you!

PR #4707, opened 2021-02-22
introduce alternative chain colors

### LINKED ISSUE #4706: buttons and account menu triggers doesn't render correctly while using linear-gradient color in the chainColor
linear-gradient creates an image and can't be used in css images(e.g. background-color/color). The menu triggers are transparent and buttons always grey if linear-gradient colors are used.
For example, the account menu buttons doesn't render in clover
image

The radio button always looks grey
image

pangwa:
This pull request fixes #4706.

Apps allows customizing the chain theme color by using chainColor
configuration.
Some chains uses css linear-gradient image as the chainColor
which works not well in some cases. E.g. svg colors and css background
colors  does not support images at all.

To fix this, we introduce alternative chain colors configuration.
The alternative chain color will be used in cases that chainColor not
work. It will fallback to use chainColor if there is no alternative
color configured for a chain.

### DIFF
--- a/packages/react-components/src/styles/index.ts
+++ b/packages/react-components/src/styles/index.ts
@@ -203,7 +203,7 @@ export default createGlobalStyle<Props & ThemeProps>(({ theme, uiAltHighlight, u
     .ui--Toggle.isChecked {
       &:not(.isRadio) {
         .ui--Toggle-Slider {
-          background-color: ${getHighlight(uiAltHighlight)} !important;
+          background: ${getHighlight(uiHighlight)} !important;
 
           &:before {
             border-color: ${getHighlight(uiHighlight)} !important;
--- a/packages/react-components/src/styles/index.ts
+++ b/packages/react-components/src/styles/index.ts
@@ -14,6 +14,7 @@ import cssTheme from './theme';
 
 interface Props {
   uiHighlight?: string;
+  uiAltHighlight?: string;
 }
 
 const BRIGHTNESS = 128 + 32;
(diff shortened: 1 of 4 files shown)


packages/react-components/src/styles/index.ts
```diff
@@ -202,7 +203,7 @@ export default createGlobalStyle<Props & ThemeProps>(({ theme, uiHighlight }: Pr
     .ui--Toggle.isChecked {
       &:not(.isRadio) {
         .ui--Toggle-Slider {
-          background-color: ${getHighlight(uiHighlight)} !important;
+          background-color: ${getHighlight(uiAltHighlight)} !important;
```
jacogr, line 206: This can simply be `background: ${getHighlight(uiHighlight)} !important;` with non of the alt colors.

packages/react-components/src/styles/index.ts
```diff
@@ -174,7 +175,7 @@ export default createGlobalStyle<Props & ThemeProps>(({ theme, uiHighlight }: Pr
 
       .ui--Icon {
         background: transparent;
-        color: ${getHighlight(uiHighlight)};
+        color: ${getHighlight(uiAltHighlight)};
```
jacogr, line 178: Here it can change to -

```js
color: inherit;
color: ${getHighlight(uiHighlight)};
```

So you never lose the button.

jacogr (commented):
Not convinced, at it stands, the colors are already way too complex.

For the slider there is an alternative solution as well as for the in-table buttons (this one maybe not as colorful).

packages/react-components/src/styles/index.ts
pangwa: cool, will fix it!

packages/react-components/src/styles/index.ts
pangwa: > Not convinced, at it stands, the colors are already way too complex.
> 
> For the slider there is an alternative solution as well as for the in-table buttons (this one maybe not as colorful).

understood, the in-table buttons renders as black, it looks acceptable 0,0

PR #4715, opened 2021-02-23
minor fix for button/sliders colors

### LINKED ISSUE #4706: buttons and account menu triggers doesn't render correctly while using linear-gradient color in the chainColor

superpw:
This pull request fixes #4706 according toe the discussion on #4707. 

in table buttons will render if the theme color contains linear gradient
and sliders looks grey.

This commit make a minor fix for them.

### DIFF
--- a/packages/react-components/src/styles/index.ts
+++ b/packages/react-components/src/styles/index.ts
@@ -174,6 +174,7 @@ export default createGlobalStyle<Props & ThemeProps>(({ theme, uiHighlight }: Pr
 
       .ui--Icon {
         background: transparent;
+        color: inherit;
         color: ${getHighlight(uiHighlight)};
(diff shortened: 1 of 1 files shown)


jacogr (approved):
Thank you. Strangely clover is not the only chain that uses gradients - so weird nobobody complained. (I don't get to play with all the chains as much as I would like... too many, so have not seen it myself) 

I do agree with you in the original PR that the button svg here would look nice with a color. Not sure how to get there is a non-complex way.

PR #4724, opened 2021-02-24
Create add-accounts.json

jimmy-tudeski:
Creation of add-accounts.json file for further work on Polish translation for polkadot.js.apps


jacogr:
Thanks an absolute million for this!

One small request - only the `translation.json` file is actually loaded. so the different `app-*.json` files are great to do it part-by-part. So in this case, since we only have one would just renamed it to `translation.json`

Any additional keys (as more app-* are done), can then just be added in the same file.

Alternatively if this feels like a better approach to not lose track, we can just commit the `app-*` files and just combine them right at the end.

jacogr (approved):
Will combine at the end. (There will probably be some duplication with that approach)

jimmy-tudeski:
Sorry, i was not able to get on it Yesterday. Will start working on it 
today.

Have a Good one!

W dniu 2021-02-26 09:58, Jaco Greeff napisał(a):

Issue #4722, opened 2021-02-24
Adapt interface to new eth style derivation


joelamouche:
- make `m/44'/60'/0'/0/0` the default "secret derivation path" when deriving from mnemonic
- verify that derivation path is m/number(') etc

joelamouche:
I'll do it

joelamouche:
@jacogr Do you mind republishing https://www.npmjs.com/package/@polkadot/ui-keyring ? It doesn't seem like it was published again and this is the package used in the app.
Thanks in advance

jacogr:
I am in the process of going wasm -> common -> api -> ui -> apps. It takes a while, at step 2 atm :)

joelamouche:
Great thanks 🙏

joelamouche:
I also want to set “slab nerve salon plastic filter inherit valve ozone crash thumb quality whale” as the default mnemonic for ethereum addresses, which is derived from seed "ethereum".
I propose this becomes also the default set of dev addresses loaded by the api in dev mode

jacogr:
The seed is not a Substrate-standard, so unlikely to go in.

jacogr:
Final step at https://github.com/polkadot-js/apps/pull/4723 (UI doesn't need to be bumped, the apps UI uses resolutions, so will use the latest util/util-crypto/keyring)

joelamouche:
Ok so trying to understand your answer "The seed is not a Substrate-standard, so unlikely to go in." and realised there are still some things I dont understand:
- createTestKeyring in common: is that where the dev addresses are declared? So as is, Gerald will be loaded?
- In this function, the seed is the sr25519 seed?
- are you saying that changing the default mnemonic is fine but I will have to declare the dev addresses as private/public key pair?
- what does it have to do with substrate?

jacogr:
Gerald won’t be loaded. The test keyring only used sr25519. (I did comment on this)

True seed is standard in Substrate for dev accounts. In subkey when you just input „//Alice“ (no mnemonic) it uses the standard Substrate mnemonic for dev seeds. This is what the keyring emulates.

joelamouche:
@jacogr 

So I'm trying to debug the app and I'm getting  "Invalid publicKey provided" when I create the key
which is coming from 
```

  saveAccount(pair, password) {
    this.addTimestamp(pair);
    const json = pair.toJson(password);
    this.keyring.addFromJson(json);
    this.accounts.add(this._store, pair.address, json, pair.type);
[...]

PR #4732, opened 2021-02-25
Derive path for eth derivation

joelamouche:
related to https://github.com/polkadot-js/apps/issues/4722


joelamouche:
@jacogr ready for review

packages/page-accounts/src/modals/Create.tsx
```diff
@@ -61,10 +61,17 @@ interface DeriveValidationOutput {
...
+  const firstLetter = path[0];
+
+  return firstLetter === 'm' || firstLetter === 'M' || firstLetter === "m'" || firstLetter === "M'";
+}
+
 function getSuri (seed: string, derivePath: string, pairType: PairType): string {
   return pairType === 'ed25519-ledger'
     ? u8aToHex(hdLedger(seed, derivePath).secretKey.slice(0, 32))
-    : `${seed}${derivePath}`;
+    : isHDDerivePath(derivePath) ? `${seed}/${derivePath}` : `${seed}${derivePath}`;
```
jacogr, line 74: This only needs to apply to the specific ethereum-like type. Not to all. So in the case where Ethereum is detected (assuming the path is validated with the `hd` add it via `#{seed}/${path}` (All the rest left as-is)

packages/page-accounts/src/modals/Create.tsx
```diff
@@ -61,10 +61,17 @@ interface DeriveValidationOutput {
 const DEFAULT_PAIR_TYPE = 'sr25519';
 const STEPS_COUNT = 3;
 
+function isHDDerivePath (path: string) {
```
jacogr, line 64: For validation, use `hdValidatePath` from `util-crypto`. (It is slightly more strict, no uppercase "M" and it needs to not be naked, i.e. `m/` is not valid, `m/0` would be. Also checks for valid numeric derivations and sizes)

packages/page-accounts/src/modals/Create.tsx
```diff
@@ -225,24 +249,30 @@ function Create ({ className = '', onClose, onStatusChange, seed: propsSeed, typ
...
   );
 
   const _onChangePairType = useCallback(
-    (newPairType: PairType) => setAddress(
-      updateAddress(seed, '', seedType, newPairType)
-    ),
-    [seed, seedType]
+    (newPairType: PairType) =>
+      setAddress(
+        updateAddress(seed, derivePath, seedType, newPairType)
```
jacogr, line 272: This is going to cause some issues. We explicitly cleared the path on all derivations previously. So this aligns with what is done in the initial state, where the path is set to the original version.

packages/page-accounts/src/modals/Create.tsx
```diff
@@ -225,24 +249,30 @@ function Create ({ className = '', onClose, onStatusChange, seed: propsSeed, typ
...
 
   const _onChangePath = useCallback(
-    (newDerivePath: string) => setAddress(
-      updateAddress(seed, newDerivePath, seedType, pairType)
-    ),
+    (newDerivePath: string) =>
+      setAddress(
+        updateAddress(seed, newDerivePath, seedType, pairType)
+      )
+    ,
```
jacogr, line 256: Would revert this and the change below, nothing functional added.

packages/page-accounts/src/modals/Create.tsx
```diff
@@ -197,7 +217,11 @@ function Create ({ className = '', onClose, onStatusChange, seed: propsSeed, typ
   const { t } = useTranslation();
   const { api, isDevelopment, isEthereum } = useApi();
   const { isLedgerEnabled } = useLedger();
-  const [{ address, derivePath, deriveValidation, isSeedValid, pairType, seed, seedType }, setAddress] = useState<AddressState>(() => generateSeed(propsSeed, '', propsSeed ? 'raw' : 'bip', isEthereum ? 'ethereum' : propsType));
+  const [{ address, derivePath, deriveValidation, isSeedValid, pairType, seed, seedType }, setAddress] = useState<AddressState>(() => generateSeed(propsSeed,
+    isEthereum
+      ? "m/44'/60'/0'/0/0"
+      : '',
+    propsSeed ? 'raw' : 'bip', isEthereum ? 'ethereum' : propsType));
```
jacogr, line 224:
```suggestion
  const [{ address, derivePath, deriveValidation, isSeedValid, pairType, seed, seedType }, setAddress] = useState<AddressState>(() => generateSeed(
    propsSeed,
    isEthereum ? "m/44'/60'/0'/0/0" : '',
    propsSeed ? 'raw' : 'bip', isEthereum ? 'ethereum' : propsType
  ));
```

packages/page-accounts/src/modals/Create.tsx
```diff
@@ -387,13 +417,15 @@ function Create ({ className = '', onClose, onStatusChange, seed: propsSeed, typ
...
                       placeholder={
-                        seedType === 'raw'
-                          ? pairType === 'sr25519'
-                            ? t<string>('//hard/soft')
-                            : t<string>('//hard')
-                          : pairType === 'sr25519'
-                            ? t<string>('//hard/soft///password')
-                            : t<string>('//hard///password')
+                        pairType === 'ethereum'
+                          ? "m/44'/60'/0'/0/0"
```
jacogr, line 421: This should probably be a constant since it is duplicated in a number of places.

```js
const ETH_DEFAULT_PATH = "m/44'/60'/0'/0/0";
```

packages/page-accounts/src/modals/Create.tsx
```diff
@@ -134,20 +141,33 @@ function generateSeed (_seed: string | undefined | null, derivePath: string, see
...
       address = addressFromSeed(seed, derivePath, pairType);
     } catch (error) {
       console.error(error);
 
       isSeedValid = false;
+      deriveValidation = { error: (error as Error).message ? (error as Error).message : (error as Error).toString() };
+    }
+  } else {
+    deriveValidation = deriveValidate(seed, seedType, derivePath, pairType);
+    isSeedValid = seedType === 'raw'
```
jacogr, line 159: Not sure at all why the complete movement here.

For all types, 

- the path needs validation - For ethereum, you would use the `hdValidatePath`, for the rest as-is.
- You need to validate the seeds in all cases for validity, not ignore in the case of Ethereum
- then the creation happens

So would put the original flow back and only have the specific path validation parts different here.

packages/page-accounts/src/modals/Create.tsx
joelamouche: The problem is that this function is called when clicking on the "advanced" dropdown and we need the derivation to be set to a valid derivationPath or the derivation won't work

jacogr: That is fine - but my point is that it is an ethereum-specific issue - so it should have ethereum-specific logic.

 I'm not against changing it globally, but then we raise the testing surface for this completely, i.e. it means all the types and incarnations needs to be completely re-tested since this changes the existing in-place behavior.
 
 So my comments here are all around not changing existing code flow or behavior, but rather to add the the option for ethereum.

packages/page-accounts/src/modals/Create.tsx
joelamouche: >You need to validate the seeds in all cases for validity, not ignore in the case of Ethereum

I don't understand that part. deriveValidate used keyExtractSuri. what verification is keyExtractSuri doing?

packages/page-accounts/src/modals/Create.tsx
joelamouche: Module '"../../../../node_modules/@polkadot/util-crypto"' has no exported member 'hdValidatePath'.ts(2305)

packages/page-accounts/src/modals/Create.tsx
joelamouche: sounds good

packages/page-accounts/src/modals/Create.tsx
joelamouche: oh should I just use ledgerValidatePath?

jacogr: Merge master.

packages/page-accounts/src/modals/Create.tsx
jacogr: So that means deriveValidate should just, in the case of an `ethereum` pairType, use the `hdValidatePath` - sadly only true/false, but that is good enough as a start.

joelamouche:
@jacogr updated and ready for review

packages/page-accounts/src/modals/Create.tsx
```diff
@@ -64,12 +66,12 @@ const STEPS_COUNT = 3;
 function getSuri (seed: string, derivePath: string, pairType: PairType): string {
   return pairType === 'ed25519-ledger'
     ? u8aToHex(hdLedger(seed, derivePath).secretKey.slice(0, 32))
-    : `${seed}${derivePath}`;
+    : pairType === 'ethereum' ? `${seed}/${derivePath}` : `${seed}${derivePath}`;
```
jacogr, line 69:
```suggestion
    : pairType === 'ethereum'
      ? `${seed}/${derivePath}`
      : `${seed}${derivePath}`;
```

Just since we are using indents above, probably looks clearer. (At least to my eyes)

packages/page-accounts/src/modals/Create.tsx
```diff
@@ -134,19 +140,21 @@ function generateSeed (_seed: string | undefined | null, derivePath: string, see
 }
 
 function updateAddress (seed: string, derivePath: string, seedType: SeedType, pairType: PairType): AddressState {
-  const deriveValidation = deriveValidate(seed, seedType, derivePath, pairType);
+  let address: string | null = null;
+  let isSeedValid = true;
+  let deriveValidation: DeriveValidationOutput = {};
```
jacogr, line 145:
```suggestion
  let address: string | null = null;
```

packages/page-accounts/src/modals/Create.tsx
```diff
@@ -134,19 +140,21 @@ function generateSeed (_seed: string | undefined | null, derivePath: string, see
...
 
-  let isSeedValid = seedType === 'raw'
+  deriveValidation = deriveValidate(seed, seedType, derivePath, pairType);
+  isSeedValid = seedType === 'raw'
```
jacogr, line 148:
```suggestion
  const deriveValidation = deriveValidate(seed, seedType, derivePath, pairType);
  const isSeedValid = seedType === 'raw'
```

packages/page-accounts/src/modals/Create.tsx
```diff
@@ -240,9 +252,9 @@ function Create ({ className = '', onClose, onStatusChange, seed: propsSeed, typ
 
   const _onChangePairType = useCallback(
     (newPairType: PairType) => setAddress(
-      updateAddress(seed, '', seedType, newPairType)
+      updateAddress(seed, isEthereum ? derivePath : '', seedType, newPairType)
```
jacogr, line 255:
```suggestion
      updateAddress(seed, isEthereum ? ETH_DEFAULT_PATH : '', seedType, newPairType)
```

Since we are swapping from something else to Ethereum, the path cannot be the previous path, but needs to be ETH-specific.

packages/page-accounts/src/modals/Create.tsx
jacogr: Not applied, hence the current build errors.

joelamouche:
@jacogr updated

jacogr (approved):
Seems sane, with no impact on existing. Hope it works exactly as expected.

Issue #4735, opened 2021-02-26
Add message that the extension protects you from potential phishing scams


anjaw3f:
Could we add a message on polkadot.js that the extension helps protect users from scams, even if they don't manager their accounts with it? 

Right now a message is displayed that sort of says "Hey you've got the extension installed, but you're not actually using it" if they have no accounts set up in it. We could use this space to add "Well done, you're gonna get an alert every time you're accessing a potential phishing site. This is a great way to protect yourself from scams!" 

The extension is very useful in scam protection and we could promote it more this way. Maybe we could even prompt users to install it, if they haven't already?

jacogr:
It does prompt when not installed - so it would make sense to have the message in both places.

PR #4741, opened 2021-02-28
Phishing protected message for extensions

jacogr:
Closes https://github.com/polkadot-js/apps/issues/4735

### DIFF
--- a/packages/page-accounts/src/Accounts/BannerExtension.tsx
+++ b/packages/page-accounts/src/Accounts/BannerExtension.tsx
@@ -26,7 +26,7 @@ function BannerExtension (): React.ReactElement | null {
   const { t } = useTranslation();
   const { hasInjectedAccounts } = useApi();
   const upgradableCount = useExtensionCounter();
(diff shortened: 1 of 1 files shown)


packages/page-accounts/src/Accounts/BannerExtension.tsx
```diff
@@ -26,6 +26,7 @@ function BannerExtension (): React.ReactElement | null {
   const { t } = useTranslation();
   const { hasInjectedAccounts } = useApi();
   const upgradableCount = useExtensionCounter();
+  const phishing = useRef<string>(t<string>('Since some extensions, such as the polkadot{.js} extension protects you against all community-reported phishing sites, there are valid reasons to use them, even if you are not storing accounts in it.'));
```
jacogr, line 29:
```suggestion
  const phishing = useRef<string>(t<string>('Since some extensions, such as the polkadot{.js} extension, protects you against all community reported phishing sites, there are valid reasons to use them for additional protection, even if you are not storing accounts in it.'));
```

PR #4757, opened 2021-03-01
Fix eth privkey account addition UI

joelamouche:
- disable derivation path for private key
- modify warning texts


packages/page-accounts/src/modals/Create.tsx
```diff
@@ -394,7 +394,8 @@ function Create ({ className = '', onClose, onStatusChange, seed: propsSeed, typ
                 <Modal.Columns>
                   <Modal.Column>
                     <Input
-                      help={t<string>('You can set a custom derivation path for this account using the following syntax "/<soft-key>//<hard-key>". The "/<soft-key>" and "//<hard-key>" may be repeated and mixed`. An optional "///<password>" can be used with a mnemonic seed, and may only be specified once.')}
+                      help={t<string>(pairType === 'ethereum' ? 'You can set a custom derivation path for this account using the following syntax "m/<purpose>/<coin_type>/<account>/<change>/<address_index>' : 'You can set a custom derivation path for this account using the following syntax "/<soft-key>//<hard-key>". The "/<soft-key>" and "//<hard-key>" may be repeated and mixed`. An optional "///<password>" can be used with a mnemonic seed, and may only be specified once.')}
```
jacogr, line 397: We cannot put the conditional inside the `t('...`)` so needs to take the following form -

```
<Input
  help={
    pairType === 'ethereum'
      ? t<string>('...')
      : t<string>('....')
  }
  ...
```

packages/page-accounts/src/modals/Create.tsx
```diff
@@ -394,7 +394,8 @@ function Create ({ className = '', onClose, onStatusChange, seed: propsSeed, typ
...
+                      isDisabled={pairType === 'ethereum' && seedType === 'raw'}
```
jacogr, line 398: In this case I would actually hide it - it is confusing being disabled since it has a value. So it would look like it gets applied.

```
{(pairType !== 'ethereum' || seedType !== 'raw') && (
  <Input
    help={....}
     ...
   />
)}

File: packages/page-accounts/src/modals/Create.tsx [hunk 3]
```diff
@@ -420,7 +421,7 @@ function Create ({ className = '', onClose, onStatusChange, seed: propsSeed, typ
                     )}
                   </Modal.Column>
                   <Modal.Column>
-                    <p>{t<string>('The derivation path allows you to create different accounts from the same base mnemonic.')}</p>
+                    <p>{t<string>(pairType === 'ethereum' && seedType === 'raw' ? 'The derivation path is only relevant when deriving keys from a mnemonic.' : 'The derivation path allows you to create different accounts from the same base mnemonic.')}</p>
```
jacogr, line 424: Same comment as above with the conditional outside of the `t` (Otherwise the string ends up non-translatable)

joelamouche:
updated

jacogr (approved):

PR #4764, opened 2021-03-02
Fix unique types import

usetech-llc:
Adding the import for unique types in `apps-config/src/api/spec/index.ts`


jacogr (approved):
Thank you.

PR #4771, opened 2021-03-05
Update ED and reaping info in app

NukeManDan:
https://wiki.polkadot.network/docs/en/learn-accounts#existential-deposit-and-reaping


jacogr (commented):
Thank a lot for this. 

I should actually just cut all these - they are all very outdated and forgotten about, not maintained. (The only place where it is more-or-less relevant is in the types info under developer settings)

jacogr (approved):

NukeManDan:
Yeah, maybe wherever possible a link to the wiki would be better? So it mostly will remain updated, except when breaking changes to the wiki structure are made

PR #4812, opened 2021-03-10
Update KILT types for upcoming release

wischli:
This PR updates the KILT types to be compliant with our upcoming (~1h from now) release 0.24.0 running on Substrate 3.0.0. It also adds the _KILT_ prefix to the `Mashnet` chain name.

According to my tests I suppose the `Address` and `LookupSource` keys are not required to be set manually anymore? If this is not the case, we would like to re-add them before this PR is merged. The same applies to the RefCount, for which we just follow Substrate.


jacogr:
So if Address = MultiAddress, don't need to set it.

jacogr (approved):

Issue #4736, opened 2021-02-26
[UX] Add a more friendly alternative for the Address_index of the derivation path


albertov19:
When users use a mnemonic seed with ECDSA the default derivation path is `m/44'/60'/0'/0/0`. The user needs to manually increment the address_index to get the subsequent address from the same mnemonic seed.

It would a good idea to implement either a dropdown menu that shows all the addresses (either by groups of 5) or just use the next one (if it finds that the previous index has been used).

This is similar to the behavior when using MetaMask, from a mnemonic seed you just click on a button that says "Add Account" and automatically creates an address from the next address_index of the derivation path.

In such a way users will be able to just add multiple accounts from the same mnemonic seed without needing to know/understand how derivation paths work.

jacogr:
I am assuming you are referring to the Frontier-like chains here which are using Ethereum-compatible keys, e.g. Moonbeam? The standard Substrate Ecdsa doesn't use the bip39 derivation paths.

Willing to accept PRs for it, e.g. in the Ledger derivation this is exactly what is done - it "hides" the full path until right at the end.

albertov19:
Yes you are totally right, I was referring to Ethereum-compatible keys. I'm with the Moonbeam team :) Thanks Jaco!

joelamouche:
>In such a way users will be able to just add multiple accounts from the same mnemonic seed without needing to know/understand how derivation paths work.

But just to be clear: the mnemonic is randomly generated everytime, so one would have to re-enter the mnemonic to derive multiple accounts from the same mnemonic

jacogr:
I would suggest connecting to Polkadot/Kusama and try the Ledger paths - they are not perfect, but may give some ideas. (At least wopuld align it with MM-like selection, so would lean towards the defaults they have)

albertov19:
@joelamouche Yes but let's say I have my mnemonic from MetaMask for example, or I have it written down or whatever and I want to import multiple accounts from it.

joelamouche:
@jacogr how do I try the ledger path? Where is it on the app?

jacogr:
Connect to one of the Ledger supported chains - Kusama, Polkadot. Dock and add an account. Then select the type as Ledger-compatible.

joelamouche:
>Then select the crypto type as Ledger-compatible.

How do I do that? I wasn't presented with this option when I added an address

jacogr:
[...]

PR #4863, opened 2021-03-16
Add derivation eth options

joelamouche:
solves https://github.com/polkadot-js/apps/issues/4736


joelamouche:
@albertov19 DO you mind taking a look at this and testing it?

joelamouche:
@jacogr ready for review

joelamouche:
kust need to change smth

joelamouche:
@jacogr now were ready :)

jacogr:
Will take a look today, thank you.

packages/page-accounts/src/modals/Create.tsx
```diff
@@ -354,65 +320,71 @@ function Create ({ className = '', onClose, onStatusChange, seed: propsSeed, typ
...
-                    : isLedgerEnabled
-                      ? settings.availableCryptosLedger
-                      : settings.availableCryptos
-                }
-                tabIndex={-1}
-              />
-            </Modal.Columns>
+            {pairType === 'ethereum'
+              ? null
+              : <Modal.Columns hint={t<string>('If you are moving accounts between applications, ensure that you use the correct type.')}>
```
jacogr, line 325: Just

```js
{ pairType !== 'ethereum && (
  <Modal.Columns hint={t<string>('If you are moving accounts between applications, ensure that you use the correct type.')}>
...
)}
```

packages/page-accounts/src/modals/Create.tsx
```diff
@@ -354,65 +320,71 @@ function Create ({ className = '', onClose, onStatusChange, seed: propsSeed, typ
...
-                          ? ETH_DEFAULT_PATH
-                          : seedType === 'raw'
+              : pairType === 'ethereum'
+                ? <CreateEthDerivationPath
+                  addressFromSeed={addressFromSeed}
+                  derivePath={derivePath}
+                  deriveValidation={deriveValidation}
+                  onChange={_onChangePath}
+                  seed={seed}
+                  seedType={seedType} />
```
jacogr, line 355:
```suggestion
                ? (
                  <CreateEthDerivationPath
                    addressFromSeed={addressFromSeed}
                    derivePath={derivePath}
                    deriveValidation={deriveValidation}
                    onChange={_onChangePath}
                    seed={seed}
                    seedType={seedType}
                  />
                )
```

packages/page-accounts/src/modals/Create.tsx
```diff
@@ -354,65 +320,71 @@ function Create ({ className = '', onClose, onStatusChange, seed: propsSeed, typ
...
+                : (
+                  <Modal.Columns hint={
+                    t<string>('The derivation path allows you to create different accounts from the same base mnemonic.')
+                  }>
+                    {(
```
jacogr, line 360: Not sure why this Input is wrapped in a `{(` here

packages/page-accounts/src/modals/CreateEthDerivationPath.tsx
```diff
@@ -0,0 +1,147 @@
...
+  onChange,
+  seed,
+  seedType }: Props): React.ReactElement<Props> {
+  const { t } = useTranslation();
+  const [addIndex, setAddIndex] = useState(0);
+  const [customIndex, setCustomIndex] = useState(0);
+  const addressList = new Array(5).fill(0).map((_, i) => ({
+    key: i,
+    text: t('Address index {{index}} - {{address}}', {
+      replace: { address: addressFromSeed(seed, `m/44'/60'/0'/0/${i}`, 'ethereum'), index: i }
```
jacogr, line 37: Would not suggest doing this up-front. This is a very costly operation.

Additionally, it gets done on each render, not only when the seed changes.

packages/page-accounts/src/modals/CreateEthDerivationPath.tsx
```diff
@@ -0,0 +1,147 @@
...
+    setUseCustomPath(!useCustomPath);
+  };
+
+  const _toggleCustomIndex = () => {
+    setUseCustomIndex(!useCustomIndex);
+  };
+
+  const onChangeCustomIndex = (e: BN) => {
+    setCustomIndex(Number(e));
+  };
```
jacogr, line 63: These should all be with `useCallback` (otherwise it changes each render, meaning even children gets re-rendered)

packages/page-accounts/src/modals/CreateEthDerivationPath.tsx
```diff
@@ -0,0 +1,147 @@
...
+              />
+            </div>
+            {useCustomIndex
+              ? (
+                <InputNumber
+                  help={t<string>('You can set a custom derivation index for this account')}
+                  isDecimal={false}
+                  label={t<string>('Custom index')}
+                  onChange={onChangeCustomIndex}
+                  value={new BN(customIndex)}
```
jacogr, line 99: Just `customIndex`

packages/page-accounts/src/modals/CreateEthDerivationPath.tsx
```diff
@@ -0,0 +1,147 @@
...
+  const _toggleCustomPath = () => {
+    setUseCustomPath(!useCustomPath);
+  };
+
+  const _toggleCustomIndex = () => {
+    setUseCustomIndex(!useCustomIndex);
+  };
+
+  const onChangeCustomIndex = (e: BN) => {
+    setCustomIndex(Number(e));
```
jacogr, line 62: We don't want to convert here. Rather convert as-needed where used. (This issue is that one each change we now create a new BN value, which then calls into this, when then sets the state... and the cycle continues)

jacogr (commented):
Some initial comments.

packages/page-accounts/src/modals/CreateEthDerivationPath.tsx
joelamouche: the functions bellow, right?

joelamouche: ha no ok the functions above

joelamouche:
@jacogr ready for another review

packages/page-accounts/src/modals/CreateEthDerivationPath.tsx
```diff
@@ -0,0 +1,159 @@
...
+  const [useCustomPath, setUseCustomPath] = useState(false);
+  const [useCustomIndex, setUseCustomIndex] = useState(false);
+
+  const errorIndex = useRef<Record<string, string>>({
+    INVALID_DERIVATION_PATH: t<string>('This is an invalid derivation path.'),
+    PASSWORD_IGNORED: t<string>('Password are ignored for hex seed'),
+    SOFT_NOT_ALLOWED: t<string>('Soft derivation paths are not allowed on ed25519'),
+    WARNING_SLASH_PASSWORD: t<string>(
+      'Your password contains at least one "/" character. Disregard this warning if it is intended.'
+    )
```
jacogr, line 48:
```suggestion
    WARNING_SLASH_PASSWORD: t<string>('Your password contains at least one "/" character. Disregard this warning if it is intended.')
```

packages/page-accounts/src/modals/CreateEthDerivationPath.tsx
```diff
@@ -0,0 +1,159 @@
...
+  const { t } = useTranslation();
+  const [addIndex, setAddIndex] = useState(0);
+  const [customIndex, setCustomIndex] = useState(new BN(0));
+  const [addressList, setAddressList] = useState<{ key: number; text: ReactNode; value: number; }[]>([{
+    key: 0,
+    text: t('Address index 0'),
+    value: 0
+  }]);
+  const [useCustomPath, setUseCustomPath] = useState(false);
+  const [useCustomIndex, setUseCustomIndex] = useState(false);
```
jacogr, line 40: There is also a `useToggle` hook from `@polkadot/react-hooks` that may be of interest here since it seems the callbacks, well, just toggle. As an example -

```js
const [isCustomPath, toggleCustomPath] = useToggle();
```

Then insteda of the callbacks defined below, just pass through `toggleCustomPath`

packages/page-accounts/src/modals/CreateEthDerivationPath.tsx
```diff
@@ -0,0 +1,159 @@
...
+    setUseCustomPath(!useCustomPath);
+  }, [setUseCustomPath, useCustomPath]);
+
+  const _toggleCustomIndex = useCallback(() => {
+    setUseCustomIndex(!useCustomIndex);
+  }, [useCustomIndex, setUseCustomIndex]);
+
+  const _onChangeCustomIndex = useCallback((e: BN) => {
+    setCustomIndex(e);
+  }, [setCustomIndex]);
```
jacogr, line 61: I suppose this could be called directly, e.g. where used just pass through `setCustomIndex` since this does not apply any additional logic?

packages/page-accounts/src/modals/CreateEthDerivationPath.tsx
```diff
@@ -0,0 +1,159 @@
...
+
+  useEffect((): void => {
+    if (!useCustomIndex) {
+      console.log(1);
```
jacogr, line 65: Stray log? (As well as 2 lines below... I always seem to let one slip through in my code and then find it a day later...)

packages/page-accounts/src/modals/CreateEthDerivationPath.tsx
```diff
@@ -0,0 +1,159 @@
...
+    const list = new Array(5).fill(0).map((_, i) => ({
+      key: i,
+      text: t('Address index {{index}} - {{address}}', {
+        replace: { address: addressFromSeed(seed, `m/44'/60'/0'/0/${i}`, 'ethereum'), index: i }
+      }),
+      value: i
+    }));
+
+    setAddressList(list);
+  }, [seed, setAddressList, addressFromSeed, t]);
```
jacogr, line 83: Since we just effect and set, this is probably better suited with a `useMemo`? i.e.

```js
const addressList = useMemo(() => {
...
}, [seed, addressFromSeed, t]);
```

Having said that, the `addressFromSeed` is quite expensive, so not sure we want to calculate that many. Would suggest rather on index change to do 1. (With the derivation there is a noticeable delay that will happen here in the main thread)

packages/page-accounts/src/modals/CreateEthDerivationPath.tsx
joelamouche: thanks

packages/page-accounts/src/modals/CreateEthDerivationPath.tsx
joelamouche: >Would suggest rather on index change to do 1.

Not sure what you mean

jacogr: So we are creating 5 entries - the entries are not the issue, deriving from the seed and making an address is horrible.

So the option is to not derive 5 addresses - only setup the indexes, create the address only for the current selected index - as it changes.

joelamouche: The ability to see multiple addresses in the dropdown is a feature that our Developer Relations Engineer requested in order to mimic the 'standard' ux for hd derivation as seen in MetaMask or ganache. On my computer it loads fast enough. If you think this feature is too ressource intensive I can circle back to my DevRel colleague to see if we can reach a compromise

jacogr: I'll merge with it in, but I am going to rip it out :) It really is not a great idea unless it is done in a worker.

joelamouche: I don't understand what you are saying. I don't want you to merge something that you will take out. If you don't want this feature in your app, say it and I will remove it.

jacogr: I've mentioned a couple of time now that it is an exceptionally bad idea, that should not be deployed :)

joelamouche: Ok I understand. Since I'm not used to webworkers and this is not high priority I'm just going to remove the address generation in the dropdown

joelamouche:
updated

jacogr:
Thank you. will get to it later today - just want to play though on the non-ETH-address side. (Always have to be slightly careful there since, as you've seen, it is not pretty, so just want to make sure nothing unintended slipped in)

joelamouche:
@jacogr fixed and ready to merge

joelamouche:
@jacogr Please take a look at this PR when you get the opportunity

jacogr:
My appologies, it slipped off the radar. Added it to the weekend's list.

jacogr (approved):
Thank you. Tested through, does exactly what it says.

Issue #4905, opened 2021-03-23
Parachain Onboarding Issue Tracker


shawntabrizi:
Opening this issue to track various optimizations or issues around the Parachain Onboarding systems.

Crowdloan:

- [ ] When creating a crowdloan, a user shouldn't be given an empty field to type a ParaId, but actually see a drop down of the ParaIds they own.
- [ ] When a crowdloan has no contributions, it an be canceled, so showing a button for that might be nice. (in case some error in the creation of it)

jonasW3F:
I don't know if it belongs here but what are those numbers referring to:

![Screenshot from 2021-03-25 12-26-47](https://user-images.githubusercontent.com/67730419/112466004-a4247c80-8d65-11eb-81ea-9def96c1aeb6.png)

And is the "value" on the auction page already the bid * leasing periods or only the bid? There might be enough space to show both to make it clearer.

jonasW3F:
I don't know if it is feasible from a technical standpoint, but would it be possible to add another table on the auction page, which shows the actual winners of the corresponding leasing period packages and the winning bid as if the auction ends now?

shawntabrizi:
@jonasW3F those numbers are their Parachain ID

shawntabrizi:
<img width="770" alt="image" src="https://user-images.githubusercontent.com/1860335/112473580-d5557a80-8d6e-11eb-9425-964e6b82c1fc.png">

shawntabrizi:
Something funny with winning crowdloan:

<img width="1757" alt="image" src="https://user-images.githubusercontent.com/1860335/112730013-5fe4d800-8f2f-11eb-921f-6440278ad983.png">

Shouldn't be dissolveable.

jacogr:
Yes. Aware of that one. Couldn’t fix this morning, was traveling today.

shawntabrizi:
There is still an issue with the UI starting a crowdloan for the active auction on the current least period:

<img width="1063" alt="image" src="https://user-images.githubusercontent.com/1860335/112752030-39777900-8fd1-11eb-9136-7c8b9758e3fd.png">

shawntabrizi:
<img width="1055" alt="image" src="https://user-images.githubusercontent.com/1860335/117203705-56b41a00-adbd-11eb-822a-58f9d3b3e511.png">

Now we 8 lease periods, the `first - last` in the top left corner of this image is wrong

PR #4906, opened 2021-03-23
Para updates

jacogr:
Current items from https://github.com/polkadot-js/apps/issues/4905

- show manager Ids (upcoming only, as it stands on parachains list it needs a rework, too much already - splitting the chain info e.g. messaging is on the way)
- dropdown for managed ids (& filtering of address only by those where there are managers)
- allow crowdloan dissolve with no contributions (dissolve itself at the end & withdraw is not there as of yet)
- None for no leases
- disabling of buttons when there can be no actions taken (e.g. no manager)
- misc. layout adjustments

Needs https://github.com/polkadot-js/api/pull/3348


packages/page-parachains/src/Crowdloan/Fund.tsx
```diff
@@ -83,6 +94,16 @@ function Fund ({ bestNumber, className, isOngoing, value: { info: { cap, deposit
               raised={raised}
             />
           )}
+          {canDissolve && (
+            <TxButton
+              accountId={depositor}
+              icon='times'
+              isDisabled={!isDepositor}
+              label={t<string>('Cancel')}
```
jacogr, line 102:
```suggestion
              label={t<string>('Dissolve')}
```

PR #4920, opened 2021-03-24
Add Litentry Testnode entry on PolkadotJS

chenzongxiong:
We want to display our testnet entrypoint on the PolkadotJS app


jacogr (commented):
Thank you, couple of comments.

- There are some missing types for this chain, specifically - `Unknown types found, no types for DataSource, QueryKey`
- It doesn't seem to be running or rather stays at block 0?

chenzongxiong:
> * ``
> 
> * It doesn't seem to

It stayed at block 0. We will fix that soon.

chenzongxiong:
> Thank you, couple of comments.
> 
>     * There are some missing types for this chain, specifically - `Unknown types found, no types for DataSource, QueryKey`
> 
>     * It doesn't seem to be running or rather stays at block 0?

It stayed at block 0. We will fix that soon

jacogr:
Not ignoring it. It really just needs the missing types before merging - which is important anyway, as they get added it can align.

chenzongxiong:
>     * running

The testnet is **running**

chenzongxiong:
> Thank you, couple of comments.
> 
>     * There are some missing types for this chain, specifically - `Unknown types found, no types for DataSource, QueryKey`
> 
>     * It doesn't seem to be running or rather stays at block 0?

The testnet is **running**

chenzongxiong:
> Not ignoring it. It really just needs the missing types before merging - which is important anyway, as they get added it can align.

Can you give me some hints, I don't where to fix them. 
Do you have any samples ?

jacogr:
Here are the types as defined by the Litentry team - not sure if they are current - https://github.com/polkadot-js/apps/blob/master/packages/apps-config/src/api/spec/litentry.ts

jacogr:
Will merge this, since the above really are only warnings, but it should be kept updated.

chenzongxiong:
> Here are the types as defined by the Litentry team - not sure if they are current - https://github.com/polkadot-js/apps/blob/master/packages/apps-config/src/api/spec/litentry.ts

I see. I will remove it since I also copied it from other teams'. 
At this moment, we didn't use it actually. I will update it soon. 

Thanks

jacogr (approved):

PR #4970, opened 2021-03-29
[CI] Fix ipfsUpload Crust Pinner

zikunfan:
Missed the `new` of class


jacogr (approved):
Thank you. 

Will do a small bump a bit later to deploy.

PR #4977, opened 2021-03-30
Add logo, name, url for PolkaFoundry

thanhtung6824:
This PR aims to add incoming name, logo, and URL for PolkaFoundry. Please let me know if I made any mistakes. Thank you


packages/apps-config/src/ui/logos/index.ts
```diff
@@ -248,6 +252,7 @@ export const namedLogos: Record<string, unknown> = {
   polkabtc: nodePolkaBTC,
   polkadex: nodePolkadex,
   polkadot: nodePolkadot,
+  polkafoundry: nodePolkaFoundry,
```
jacogr, line 255: You would just need an entry for `rococoPolkaFoundry: nodePolkaFoundry` as well (the `info` in the RPC endpoint matches on this name)

thanhtung6824: @jacogr TY. I just added it.

jacogr (commented):
Just run yarn lint locally - have not checked CI, but it has a small complaint ;)

thanhtung6824:
@jacogr Thank you. I fixed it

jacogr (approved):
Thank you.

thanhtung6824:
@jacogr 
image
The color is not affected. Maybe I need to change the name to match the info in the RPC endpoint? In this case is "Polkafoundry PC1" instead of "Polkafoundry"

jacogr:
The colors, as mentioned in the code is based on there the `system.chain` RPC result or the actual software name via the `system.name` RPC.

Best is always to check it locally.

PR #5017, opened 2021-04-02
Adds sunrock parachain info

crystalin:
This is a test network, the provider is ws (not wss) on purpose


packages/apps-config/src/endpoints/testingRococo.ts
```diff
@@ -294,6 +294,14 @@ export function createRococo (t: TFunction): LinkOption[] {
         providers: {
           'PHOENIX Protocol': 'wss://phoenix-ws.coinid.pro'
         }
+      },
+      {
+        info: 'rococoSunrock',
+        paraId: 499,
+        text: t('rpc.rococo.sunrock', 'Sunrock', { ns: 'apps-config' }),
+        providers: {
+          Sunrock: 'ws://sunrock.kaki.dev:9944'
```
jacogr, line 303: The issue with `ws` is that is not usable on the https:// site - you cannot make connections to non-secure origins from from HTTPS unless it is localhost.

crystalin: @jacogr can I simply remove the provider ? 
This is an ephemeral network (but Shawn asked me to still add info in the UI) and I don't want to add SSL support

crystalin: I reset it to wss, and will take care of the ssl part later today

jacogr (approved):

PR #5024, opened 2021-04-04
mybank.network

armatrix:
MyBank is a decentralized financial platform based on Polkadot, aiming to establish a blockchain network that realizes asset appreciation and promotes asset flow. MyBank is not only aimed at the existing DeFi user group. MyBank is an inclusive finance platform based on the mutual trust established by social networks in the real world, which allows most people, including poor groups, to enjoy the convenience brought by finance and blockchain.


jacogr (approved):
Thank you.

PR #5025, opened 2021-04-05
Include Steam parachain for Rococo

johnwhitton:
I have tested this and the user interface looks good.
I have also added the type definitions [here](https://github.com/eavenetwork/apps/blob/master/packages/apps-config/src/api/spec/eave.ts), however I was unsure how to ensure they were imported.
Please advise and I will update this pull request or create a new one as you wish.


jacogr:
The link from types to spec needs to be added here - https://github.com/polkadot-js/apps/blob/master/packages/apps-config/src/api/spec/index.ts

Just run yarn —lint locally, it will fix the issues (seems like quotes only) locally

johnwhitton:
@jacogr Thanks for the help.
Updated and linted and it looks like it works locally.
Let me know
🙏

johnwhitton:
Side issue `dawn` is from an old build, not sure why it is still being pulled in, it should be `steam`.
When I figure this out will push again with `steam`.

packages/apps-config/src/api/spec/index.ts
```diff
@@ -96,6 +97,8 @@ const spec: Record<string, OverrideBundleDefinition> = {
   'dock-testnet': dock,
   'dotmog-node': dotmog,
   dusty4: dusty,
+  eave: eave,
+  'dawn':eave,
```
jacogr, line 101:
```suggestion
  eave: eave,
  dawn: eave,
```

It may complain about it being alphabetical here - do probably should just move dawn.

jacogr (commented):
It looks sane (apart from the one small comment) - happy to merge if it works for you.

johnwhitton:
Thanks @jacogr good to go. 
Once I figure out how to modify the chainspec from `dawn` to `eave` I'll publish a separate PR.
Let's get this one in.
🙏

jacogr (approved):

jacogr:
Thank you.

PR #5042, opened 2021-04-08
Update test uart types

tianxiemaochiyu:


packages/apps-config/src/api/spec/uart.ts
```diff
@@ -21,17 +21,20 @@ const definitions: OverrideBundleDefinition = {
...
+        ProposalId: 'u64',
+        TokenBalance: 'Balance',
         LotteryKind: {
-          _enum: {
-            Routine: 'Null',
-            TreasuryFunded: 'ReasonIndex'
-          }
+          _enum: [
+            'Routine',
+            'TreasuryFunded(ReasonIndex)'
```
jacogr, line 30: The original enum here was correct, `TreasuryFunded(ReasonIndex)` is not a valid type.

packages/apps-config/src/api/spec/uart.ts
```diff
@@ -21,17 +21,20 @@ const definitions: OverrideBundleDefinition = {
...
+          ]
         },
         LottoResult: {
-          _enum: {
-            Routine: '(AccountId, Balance)',
-            TreasuryFunded: 'Balance'
-          }
+          _enum: [
+            'Routine(AccountId, Balance)',
+            'TreasuryFunded(Balance)'
```
jacogr, line 36: As above, the original types was correct, this makes it invalid (or rather a c-like enum which won't be able to decode the values)

packages/apps-config/src/api/spec/uart.ts
```diff
@@ -159,8 +201,9 @@ const definitions: OverrideBundleDefinition = {
         },
         ReFungibleItemType: {
           Collection: 'u64',
-          Owner: 'Vec<Ownership>',
-          Data: 'Vec<u8>'
+          Owner: 'Vec<Ownership<AccountId>>',
```
jacogr, line 204:
```suggestion
          Owner: 'Vec<Ownership>',
```

Unlike on the Rust side, the types parser is not clever enough to be passed a generic parameter.

packages/apps-config/src/api/spec/uart.ts
tianxiemaochiyu: ok

tianxiemaochiyu:
@jacogr Thanks for the help.Updated.

packages/apps-config/src/api/spec/uart.ts
jacogr, line 30: The enums are still incorrect, the correct form is indeed as it was in the original.

```js
Routine: 'Null', // this one is has not extra data
TreasuryFunded: 'ReasonIndex' // This one contains data
```

As mentioned originally, this affects the `LotteryKind` (original correct, new version is not), `LotteryResult` (same where original is correct, new version is incorrect as will match as a c-like enum)

tianxiemaochiyu:
@jacogr Sorry,updated.

jacogr (approved):

PR #5065, opened 2021-04-14
update parachain id

yooml:


packages/apps-config/src/endpoints/testingRococo.ts
```diff
@@ -73,7 +73,7 @@ export function createRococo (t: TFunction): LinkOption[] {
       },
       {
         info: 'rococoBifrost',
-        paraId: 107,
+        paraId: 10,
```
jacogr, line 76: Any id < 1000 is now reserved for system parachains. Not sure if this is correct.

jacogr: Confirmed with Shawn, it needs to be > 1000

yooml: now update to 1024

jacogr (approved):
Thank you.

PR #5077, opened 2021-04-15
add prism parachain

chrissoso:
Add prism parachain network to the list of parachains


packages/apps-config/src/endpoints/testing.ts
```diff
@@ -205,6 +205,13 @@ export function createTesting (t: TFunction): LinkOption[] {
         Interlay: 'wss://beta.polkabtc.io/api/parachain'
       }
     },
+    {
+      info: 'prism',
+      text: t('rpc.prism', 'Prism', { ns: 'apps-config' }),
```
jacogr, line 210: Is this a parachain (as per description) or a test chain? Either is fine, however if (as mentioned) it is a parachain, it needs to move to the testingRococo.ts file and have a paraId.

chrissoso: Currently only a test chain

PR #5116, opened 2021-04-22
Kulupu: add dual ref count type definition for runtime version 20

sorpaas:


sorpaas:
(Retracting this for now. It seems to be a false alarm.)

PR #5119, opened 2021-04-23
update some details

Zombieliu:
haha,i see the list not update the web3games so i find some details . Thanks


jacogr:
The details here is already in master, see https://github.com/polkadot-js/apps/blob/master/packages/apps-config/src/api/spec/sgc.ts (A merge of master will show these changes are the same)

Zombieliu:
> The details here is already in master, see https://github.com/polkadot-js/apps/blob/master/packages/apps-config/src/api/spec/sgc.ts (A merge of master will show these changes are the same)

Thank you , i already changed it .  you can close this

PR #5132, opened 2021-04-23
KILT types definition repo

Dudleyneedham:
Adding KILT types definition repo.


packages/apps-config/package.json
```diff
@@ -14,6 +14,7 @@
     "@edgeware/node-types": "^3.3.4",
     "@equilab/definitions": "1.0.3",
     "@interlay/polkabtc-types": "^0.6.2",
+    "@kiltprotocol/type-definitions": "latest",
```
jacogr, line 17: This is great. Could you rather point to the version - the reason being is that I do update the dependencies (weekly, sometimes more) and `yarn upgrade-interactive` I can see what got bumped. It is unlikely that a `rm -rf yarn.lock` is called.

jacogr (commented):
One small comment for my sanity. Love these, makes it much easier from both sides.

packages/apps-config/package.json
Dudleyneedham: Sure! No problem. I was wondering which would be preferred. 👍  I have updated it now.

Dudleyneedham:
I agree. It is something we have been meaning to do for a while.

PR #5142, opened 2021-04-25
Adding support for dotscanner.com

Deekor:
DotScanner doesnt support the hash lookup of a block (only number) - once thats supported I will open another PR to add block to the paths list in `dotscanner.ts`


jacogr (approved):
Thank you.

PR #5177, opened 2021-04-30
Check if accounts are loaded before hiding a tab in staking

krzysztof-jelski:
Trying to solve a frustration: opening a "tab-level" url directly redirects to root of the section (eg. `#/staking/actions`).

It happens because accounts are not loaded at first, and the lack of accounts marks the "actions" tab as hidden. This in turn triggers a `window.location.hash` change to `/staking` root.

Provided solution checks whether accounts are loaded and only then hides (or not) the tab.


jacogr (approved):
Thank you

PR #5207, opened 2021-05-05
update BitCountry latest customType and upgrade paraId to 1008

justinphamnz:
Hi team,

We have updated the Bit.Country Parachain node to latest rococo-v1, so new custom types and updated paraId have been included in this PR. 

Thank you.


jacogr (approved):
Thank you.

PR #5213, opened 2021-05-05
Add tokens alias for Zeitgeist

lsaether:
Adding an alias for the Tokens pallet so that the clash of `AccoundData` types between that pallet and Balances' AccountData doesn't cause Apps to complain. Something of a stop-gap for now until we start generating the full bundle in our own package.


jacogr (approved):
Thank you.

PR #5216, opened 2021-05-05
adapted webpack and added packages

joelamouche:
draft to prepare for extension

### DIFF
--- a/package.json
+++ b/package.json
@@ -99,5 +99,13 @@
     "stylelint-config-styled-components": "^0.1.1",
     "webpack": "^5.36.1",
     "webpack-cli": "^4.6.0"
+  },
+  "dependencies": {
+    "@metamask/detect-provider": "^1.2.0",
+    "assert": "^2.0.0",
+    "https-browserify": "^1.0.0",
+    "os-browserify": "^0.3.0",
+    "stream-http": "^3.2.0",
+    "web3": "^1.3.5"
   }
 }
--- a/packages/apps/webpack.base.cjs
+++ b/packages/apps/webpack.base.cjs
@@ -169,6 +169,10 @@ function createWebpack (context, mode = 'production') {
       },
       extensions: ['.js', '.jsx', '.mjs', '.ts', '.tsx'],
       fallback: {
+        http: require.resolve('stream-http'),
+        https: require.resolve('https-browserify'),
+        os: require.resolve('os-browserify/browser'),
+        assert: require.resolve('assert'),
         crypto: require.resolve('crypto-browserify'),
         path: require.resolve('path-browserify'),
         stream: require.resolve('stream-browserify')
--- a/yarn.lock
+++ b/yarn.lock
@@ -5,10 +5,10 @@ __metadata:
   version: 4
   cacheKey: 7
 
-"7zip-bin@npm:~5.0.3":
-  version: 5.0.3
-  resolution: "7zip-bin@npm:5.0.3"
-  checksum: d78384cae472fdd5c2b490d320697196ae385b55b341006258e30e056e94b3de7ca5105fd04561c92baac61a5de225c1176c11347e50fadce116272661483977
+"7zip-bin@npm:~5.1.1":
+  version: 5.1.1
+  resolution: "7zip-bin@npm:5.1.1"
+  checksum: bb53a13a256dd8797f2cb8946c1ab6eacef25b8a874926f7fa73fa61a17e3577152b3767921f0e28128d48a1f99640750d5c7137112b721729f4300055938381
   languageName: node
   linkType: hard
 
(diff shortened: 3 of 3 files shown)


packages/react-api/src/Api.tsx
```diff
@@ -82,13 +84,14 @@ async function getInjectedAccounts (injectedPromise: Promise<InjectedExtension[]
 
     const accounts = await web3Accounts();
 
-    return accounts.map(({ address, meta }, whenCreated): InjectedAccountExt => ({
+    return accounts.map(({ address, meta, type }, whenCreated): InjectedAccountExt => ({
```
jacogr, line 87: On line 85 above it should explicitly ask for the `ethereum` types, when the flag is passed. (Comments in extension on this - I'm guessing this is based on the `isEthereum` flag)

jacogr: EDIT: Since we do filter on the type in the keyring anyway, it is most probably not needed, since we do have the same effect. (Probably just more descriptive always being explicit, i.e. no need to even pass these through unless needed)

joelamouche: So what do you think I should do? Leave as is?

joelamouche:
waiting on https://github.com/polkadot-js/extension/pull/566 to be merged and then remove dependencies as per https://github.com/polkadot-js/apps/pull/5216#discussion_r642820160

packages/react-api/src/Api.tsx
jacogr: Ahhhh.... ok this links through to this comment as well https://github.com/polkadot-js/extension/pull/566#discussion_r689374988

In the keyring itself, with the injected load, we don't actually check the type at all, e.g. https://github.com/polkadot-js/ui/blob/master/packages/ui-keyring/src/Keyring.ts#L249-L260 - we just inject what is there. The best place to add a solution is either -

- in the extension, i.e. add an extra flag (may not be the prettiest, as described in the linked comment it is kinda ugly)
- on the keyring itself, i.e. filtering by type there (just need to follow there again, you did do some type filtering already there, maybe I'm missing something in the link)

Would not add a solution inside the apps UI since there are a lot of extension-dapp users, and don't want them to have to make all the same changes.

Anyway, we can test this once all ok.

joelamouche: I added the filter in the extension dapp (in web3accounts), so I will pass down the type flag here

package.json
```diff
@@ -101,15 +101,8 @@
...
   },
-  "dependencies": {
-    "@metamask/detect-provider": "^1.2.0",
-    "assert": "^2.0.0",
-    "https-browserify": "^1.0.0",
-    "os-browserify": "^0.3.0",
-    "stream-http": "^3.2.0",
-    "web3": "^1.3.5"
-  }
+  "packageManager": "yarn@3.0.1"
```
joelamouche: Now that I removed this, I get the errors again, even though the dependency was updated @jacogr 
```
➜  apps git:(jlm-adapt-app-to-metamask-extension) ✗ yarn start
$ polkadot-dev-clean-build 
[webpack-cli] Failed to load '/Users/antoineestienne/GitHubRepo/apps/packages/apps/webpack.serve.cjs' config
[webpack-cli] Error: Cannot find module 'stream-http'
```

joelamouche:
@jacogr updated. Works locally with `yarn start`
I don't think I understand how dependencies work... (see comment above)

jacogr:
I added all the deps in another PR, including the required webpack overrides. (Had to do this when updating the `extension*` packages to latest)

I did have to make a small tweak to not create havoc out there for users of the `extension-dapp` package - I split the metamask compat into a separate package (code as-is), basically to make it optional for others. See https://github.com/polkadot-js/extension/pull/810

So obviously in apps we will always include it, but it just allows other users to correctly choose instead of having it forced on them (which also means updating their builds in the same way as done here)

The extension PR does mean a slight change to use, i.e. the function needs to be passed in as part of the web3Enable, e.g. https://github.com/polkadot-js/extension/pull/810/files#diff-60639b27a793d918f655ef618cd1778eb219812f1200d48079560408ac4ae3f0R66

Something like -

```js
import initMetaMask from '@polkadot/extension-compat-metamask';
...
web3Enable('polkadot-js/apps', [initMetaMask]).then(...);
```

joelamouche:
Had to add @types/node because of https://github.com/facebook/jest/issues/11640

joelamouche:
@jacogr updated

jacogr:
Is this still relevant? (ie. functionality missing?)

joelamouche:
@jacogr Yes this is still relevant: the missing feature is the injection of MetaMask account sinto the app for ethereum parachains. I just tested it and it still works. HOWEVER, it uses the bytecode signing from MetaMask that triggers the relevant warning

joelamouche:
@jacogr I don't understand why Im getting `export 'default' (imported as 'initMetaMask') was not found in '@polkadot/extension-compat-metamask' (possible exports: packageInfo)`

ap211unitech:
Closing due to further activity

PR #5225, opened 2021-05-07
update and clean galital standalone chain specs and type

masterdubs:


packages/apps-config/src/endpoints/testing.ts
```diff
@@ -139,8 +139,7 @@ export function createTesting (t: TFunction): LinkOption[] {
       }
     },
     {
-      info: 'rococoGalital',
-      paraId: 1230,
+      info: 'Galital',
```
jacogr, line 142: We need a paraId to do the mapping.

masterdubs: Thanks, fixed in next commit.

jacogr: Still seems to be missing :) Just re-add the `paraId: 1230` line (assuming that is still the selected one)

masterdubs: I can add it but it's an independent one, not a parachain version. Still need to add it ?

jacogr: ^&*%$ My error, sorry. Dunno how that slipped in there initially, so the removal of it is 100% correct.

Ok, need some coffee it seems, not even half awake :)

jacogr (approved):
Thank you.

PR #5232, opened 2021-05-07
Update galital type

masterdubs:
Update type for AccountInfo as it was making the block detail view crashing.


jacogr (approved):
Thank you.

PR #5267, opened 2021-05-12
update moonbeam type dependency

joelamouche:
Update moonbeam-types -bundle to add upcoming `AuthorId: "AccountId",` change in v36


jacogr:
Thank you. This was all bumped to latest in https://github.com/polkadot-js/apps/pull/5268

Not sure how the bump was done here, but quite sceptical without local testing of just bumping unrelated stuff to their latest patches, eg. babe - have really had issues in the past with blind patches. 

But apart from that, do really appreciate this one - I generally only bump types, along with other stuff around once a week, so this helps a lot.

PR #5282, opened 2021-05-14
Fix translations loading for ever

fadomire:
Currently when going to #/settings/i18n we have the spinner loading forever.

It is caused by 404 on locales/pl/translation.json missing.

bug introduced by this PR:
https://github.com/polkadot-js/apps/pull/4724

To avoid this in future and instead of creating a locales/pl/translation.json (that can be done in another PR) ii catch any error on retrieveJson and return an empty {}

### DIFF
--- a/packages/page-settings/src/I18n/index.tsx
+++ b/packages/page-settings/src/I18n/index.tsx
@@ -40,12 +40,13 @@ async function retrieveJson (url: string): Promise<any> {
     return cache.get(url);
   }
 
-  const response = await fetch(`locales/${url}`);
-  const json = await response.json() as unknown;
+  const json = await fetch(`locales/${url}`)
+    .then((response) => response.json())
+    .catch((e) => console.error(e)) as unknown;
 
   cache.set(url, json);
 
-  return json;
+  return json || {};
 }
 
 async function retrieveEnglish (): Promise<StringsMod> {


packages/page-settings/src/I18n/index.tsx
```diff
@@ -65,12 +65,12 @@ async function retrieveAll (): Promise<Defaults> {
...
   const english = await retrieveEnglish();
   const translations = missing.length
-    ? await Promise.all(missing.map((lng) => retrieveJson(`${lng}/translation.json`)))
+    ? await Promise.allSettled(missing.map((lng) => retrieveJson(`${lng}/translation.json`)))
     : [];
 
   // setup the language cache
   missing.forEach((lng, index): void => {
-    languageCache[lng] = translations[index] as Record<string, string>;
+    languageCache[lng] = translations[index] as unknown as Record<string, string>;
```
jacogr, line 73: This is not quite correct. So `allSettled` returns either one of these two objects -

- `{status: "fulfilled", value: <translation value> }`
- `{status: "rejected",  reason: Error: <an error> }`

So I'm guessing it probably should be something like -

```js
const { value } = translations[index];

if (value) {
  languageCache[value] = value as Record<string, string>;
}

fadomire: ok thanks i'll check that

fadomire: i edited the PR and made it simpler
no changes in typing anymore

jacogr (approved):
Thank you very much.

PR #5291, opened 2021-05-15
feat: add KILT paraId

wischli:
Adds KILT Parachain ID to the live parachains tab. Please note that the linked websocket address is not live yet.


jacogr (approved):
Thank you.

PR #5320, opened 2021-05-18
Extract paraId info for selector

jacogr:
While cool, slower with a performance impact on the selector only.


jacogr:
Won't be merged.

PR #5330, opened 2021-05-20
Update Crust Shadow logo and color

zikunfan:


packages/apps-config/src/endpoints/productionRelayKusama.ts
```diff
@@ -48,19 +48,19 @@ export function createKusama (t: TFunction): EndpointOption {
         }
       },
       {
-        info: 'crab',
-        paraId: 2006,
-        text: t('rpc.kusama.crab', 'Darwinia Crab', { ns: 'apps-config' }),
+        info: 'shadow',
```
zikunfan, line 51: > Added alphabetical based on chain name
I added with Chain name started with `C`, is that ok with different info and chain name?

jacogr (approved):

packages/apps-config/src/endpoints/productionRelayKusama.ts
jacogr: All fine. The "info" just maps through to knowing which logo to use.

PR #5334, opened 2021-05-20
Add Thales parachain

bit-god:
Add Thales parachain


jacogr:
A bit confused - since this is not a parachain as per the description. They need to go under Rococo with the correct paraId?

packages/apps-config/package.json
```diff
@@ -23,6 +23,7 @@
     "@sora-substrate/type-definitions": "^1.3.4",
     "@subsocial/types": "^0.5.2",
     "@zeitgeistpm/type-defs": "^0.1.51",
+    "thales-types-bundle": "0.8.2",
```
jacogr, line 26:
```suggestion
    "thales-types-bundle": "^0.8.2",
```

Just commit the `yarn.lock` as well. (I'm assuming that is why CI is failing, i.e. it doesn't auto-install, rather just keeps to the deps as added)

bit-god:
because it install a new dependency package, but it not allowed:

```
At least 1 approving review is required by reviewers with write access. 
```

and:
```
yarn install --immutable
```

https://yarnpkg.com/cli/install

```
--immutable | Abort with an error exit code if the lockfile was to be modified

```

jacogr:
100% - that is by design. CI should not randomly add packages that have not been through a testing and review process.

bit-god:
how to resolve it? should i commit yarn.lock?

jacogr:
Yes, please.

PR #5340, opened 2021-05-21
add kylin network to apps explorer

samelamin:
Adding Kylin parachain to Rococo. I have used parachain id 2004 but let me know if I should use a different ID


samelamin:
@jacogr could you please review when you get a chance

jacogr:
No need to ping, all the repos do have alerts and I see all PRs, support requests, bug reports and anything else across polkadot-js. (I may not get to them immediately...)

jacogr:
> I have used parachain id 2004 but let me know if I should use a different ID

As long as it is registered and your are the owner on-chain of the id, all ok.

samelamin:
> No need to ping, all the repos do have alerts and I see all PRs, support requests, bug reports and anything else across polkadot-js. (I may not get to them immediately...)

I apologise mate!  

Regarding the registration, with rococo being reset its currently set to 2000? I reserved 2000 but I still have not registered the parachain because I am not clear if I should call `setKeys` on my validator (which is still syncing) or on the public validator which I assumed has already synced

jacogr:
Ok, if you registered 2000, then use that one, not 2004 :) 2000 is special being "first" :)

packages/apps-config/src/endpoints/testing.ts
```diff
@@ -192,6 +192,13 @@ export function createTesting (t: TFunction): LinkOption[] {
         'KILT Protocol': 'wss://full-nodes.kilt.io:9944/'
       }
     },
+    {
+      info: 'kylin',
+      text: t('rpc.test.kylin', 'Kylin Testnet', { ns: 'apps-config' }),
+      providers: {
+        kylin: 'wss://rpc.kylin-node.co.uk'
+      }
+    },
```
jacogr, line 201: If this is a parachain, it should only be on the next list - this part is for stand-alone chains. (Since the RPCs match here, they are the same)

```suggestion
```

packages/apps-config/src/endpoints/testingRelayRococo.ts
```diff
@@ -241,6 +241,13 @@ export function createRococo (t: TFunction): EndpointOption {
         providers: {
           'Konomi Network': 'wss://rococo.konomi.tech'
         }
+      }, {
```
jacogr, line 244:
```suggestion
      },
      {
```

Have not checked, but quite possible why CI is complaining.

samelamin:
> Ok, if you registered 2000, then use that one, not 2004 :) 2000 is special being "first" :)

Ok no worries I will register for 2001, but the json already has 2001 being taken up by `rococoCrust`

packages/apps-config/src/endpoints/testing.ts
samelamin: Can you clarify what you mean by next list?

jacogr: This one :)

image

packages/apps-config/src/endpoints/testingRelayRococo.ts
```diff
@@ -242,6 +242,14 @@ export function createRococo (t: TFunction): EndpointOption {
           'Konomi Network': 'wss://rococo.konomi.tech'
         }
       },
+      {
+        info: 'rococoKylin',
+        paraId: 2004,
```
jacogr, line 247:
```suggestion
        paraId: 2000,
```

As registered by you.

samelamin: Thats the thing, I technically still havent registered because I am still unclear on where to run the `setKeys` function, my own validator thats still syncing or any public validator on the rococo network?

I would assume once its syncs completely it should be the same

packages/apps-config/src/endpoints/testing.ts
samelamin: Got it, so I actually added it. Thanks mate!

jacogr:
The existing chains all need to change their Ids, hence them being disabled. Rococo was reset. So, the steps are -

- click "+ ParaId", register
- use that one in your genesis & PRs here

So assuming you are this account, you are all ok -

image

samelamin:
> The existing chains all need to change their Ids, hence them being disabled. Rococo was reset. So, the steps are -
> 
> * click "+ ParaId", register
> * use that one in your genesis & PRs here
> 
> So assuming you are this accoun and you have completed the steps abovet, you are all ok -
> 
> image

Got it, what about the validator `setKeys` where I submit the validatorId. Should I be doing it on my own validator (while its syncing) or a public one?

jacogr:
You don't need to worry about validators or setKeys - Parity is running all of them (900+ and counting) on Rococo, parachains don't need to supply their own.

samelamin:
> You don't need to worry about validators or setKeys - Parity is running all of them (900+ and counting) on Rococo, parachains don't need to supply their own.

Got it, I have to say a massive massive THANK YOU! for both your patience and time to go through this PR and help me

packages/apps-config/src/endpoints/testingRelayRococo.ts
```diff
@@ -242,6 +242,14 @@ export function createRococo (t: TFunction): EndpointOption {
           'Konomi Network': 'wss://rococo.konomi.tech'
         }
       },
+      {
+        info: 'rococoKylin',
+        paraId: 2000,
```
jacogr, line 247: One last question - is this indeed yours? ;)

packages/apps-config/src/api/spec/index.ts
```diff
@@ -139,6 +140,7 @@ const spec: Record<string, OverrideBundleDefinition> = {
   'jupiter-rococo': jupiterRococo,
   khala: phalaParachain,
   'kilt-parachain': kilt,
+  kylin,
```
jacogr, line 143: Just move it one down, see https://github.com/polkadot-js/apps/runs/2642112178?check_suite_focus=true#step:3:486

packages/apps-config/src/endpoints/testingRelayRococo.ts
samelamin: Def, I own this account and I just registered the parathread to this account
`5Gn1igfpf4hP7iG1Gsm1AbwPBCpR8BmHK4b6i2VrGHQS1kAJ`


image

packages/apps-config/src/api/spec/index.ts
samelamin: Done

packages/apps-config/src/endpoints/testingRelayRococo.ts
jacogr: No issues, always like "2000" - will always be first :)

samelamin:
How do I rerun the failed build?

jacogr (approved):
Thank you.

samelamin:
@jacogr thank you so so so much for your help in this, can I dm you on element with a few quick questions?

jacogr:
I'm really not active on Element - I do try to check in once a day more-or-less, but GH keeps me "quite" busy, so I focus all energy here.

samelamin:
Ah ok I see. Fair enough mate I understand.

Can you help me understand why my node is still not registering on rococo.
I got the below error when i registered

```

1registrar.HeadDataTooLarge

```

Also the logs for my collator show me that its still syncing. Do I not
require to register a validator anymore?

jacogr:
No idea on the above - the best is to ask the in the Element Rococo channel where the pallet devs hang out (or even better, look at the Rust code to see in which circumstances that error is emitted, that will pinpoint exactly, aka https://github.com/paritytech/polkadot/blob/master/runtime/common/src/paras_registrar.rs#L521)

samelamin:
OK no worries, I will. 

Thank you so much for your help

PR #5354, opened 2021-05-25
Add thales network type bundle

bit-god:
Add thales network type bundle


bit-god:
@jacogr can you do me a favor, merge my pull request

jacogr:
1. it has conflicts with master
2. it fails CI checks

So it is not merge-able at all.

PR #5363, opened 2021-05-25
Moonbeam: display account id from author id

joelamouche:
For moonbeam, the author provided in the last header is the node address and is used to retrieve the actual author account id from an onchain mapping
- if no author field, retrieve address from events and get account id from api
- add HeaderExtendedWithMapping type to include iptional authorFromMapping string field


joelamouche:
@jacogr ready for review

yarn.lock
```diff
@@ -1980,14 +1980,14 @@ __metadata:
   linkType: hard
 
 "@ledgerhq/hw-transport-webusb@npm:^5.53.0":
-  version: 5.53.0
-  resolution: "@ledgerhq/hw-transport-webusb@npm:5.53.0"
+  version: 5.53.1
```
jacogr, line 1983: This one needs to be bumped in common, so not the right place.

jacogr (commented):
Would keep it to the moonbeam-types only - all the rest are as part of the rolling bundles updates.

packages/page-explorer/src/BlockHeader.tsx
```diff
@@ -27,9 +31,11 @@ function BlockHeader ({ value }: Props): React.ReactElement<Props> | null {
       </td>
       <td className='all hash overflow'>{hashHex}</td>
       <td className='address'>
-        {value.author && (
-          <AddressMini value={value.author} />
-        )}
+        {value.authorFromMapping
```
jacogr, line 34: +1 Makes sense.

packages/react-query/src/BlockAuthors.tsx
```diff
@@ -48,10 +53,19 @@ function BlockAuthorsBase ({ children }: Props): React.ReactElement<Props> {
       }).catch(console.error);
 
       // subscribe to new headers
-      api.derive.chain.subscribeNewHeads((lastHeader): void => {
+      api.derive.chain.subscribeNewHeads(async (lastHeader: HeaderExtendedWithMapping): Promise<void> => {
         if (lastHeader?.number) {
           const blockNumber = lastHeader.number.unwrap();
-          const thisBlockAuthor = lastHeader.author?.toString();
+          let thisBlockAuthor = ''; //= lastHeader.author?.toString();
```
jacogr, line 59:
```suggestion
          let thisBlockAuthor = '';
```

packages/react-query/src/BlockAuthors.tsx
```diff
@@ -48,10 +53,19 @@ function BlockAuthorsBase ({ children }: Props): React.ReactElement<Props> {
...
+
+          if (lastHeader.author) {
+            thisBlockAuthor = lastHeader.author.toString();
+          } else if (lastHeader.digest.logs && lastHeader.digest.logs[0] && lastHeader.digest.logs[0].asConsensus[1]) {
```
jacogr, line 63: `lastHeader.digest.logs[0].asConsensus[1]` this may fail, there is no `isConsensus` check.

Additionally would do a check for `isFunction(api.query.authorMapping?.mapping)` first (probably not in the subscribe itself, can set a flag outside) - only use this branch in those cases.

jacogr (commented):
Couple of small comments.

joelamouche:
@jacogr updated

packages/react-query/src/BlockAuthors.tsx
```diff
@@ -38,20 +43,30 @@ function BlockAuthorsBase ({ children }: Props): React.ReactElement<Props> {
   useEffect((): void => {
     // No unsub, global context - destroyed on app close
     api.isReady.then((): void => {
-      let lastHeaders: HeaderExtended[] = [];
+      let lastHeaders: HeaderExtendedWithMapping[] = [];
       let lastBlockAuthors: string[] = [];
       let lastBlockNumber = '';
+      const isAuthorMapping: boolean = typeof (api.query.authorMapping?.mapping) === 'function';
```
jacogr, line 49:
```suggestion
      const isAuthorMapping = isFunction(api.query.authorMapping?.mapping);
```

jacogr (commented):
LGTM.

One small comment. 

In addition, would just drop the yarn.lock changes (have not looked at what it contains)

packages/react-query/src/BlockAuthors.tsx
joelamouche: I get `ReferenceError: isFunction is not defined
    at eval (webpack://@polkadot/apps/../react-query/src/BlockAuthors.tsx?:54)`

jacogr:
May just be missing an import (have not looked at CI logs), but expect on line 10 -

`import { formatNumber, isFunction } from '@polkadot/util';`

(sorry, counldn't proposes since the line is not changed...)

packages/react-query/src/BlockAuthors.tsx
jacogr: Ahh, see the comment, basically missing the import (which I couldn't propose) - `import { formatNumber, isFunction } from '@polkadot/util';` (line 10, just above)

joelamouche:
@jacogr fixed

jacogr (approved):
Thank you!

PR #5375, opened 2021-05-26
update acala types

qwer951123:


jacogr:
Not sure why the build is failing, all the pieces looks to be there. Will click through and look at the logs.

packages/apps-config/src/api/spec/index.ts
```diff
@@ -97,7 +97,7 @@ const spec: Record<string, OverrideBundleDefinition> = {
   Pangolin: pangolin,
   VLN: vln,
   'VLN-PC': vlnrococo,
-  acala,
+  ...acala,
```
jacogr, line 100: The issue is that above you export an `OverrideBundleDefinition` and here you expand it. This is why the build fails.

qwer951123: fixed the type assertion

jacogr (approved):
Thank you.

PR #5384, opened 2021-05-28
add kylin testnetwork

samelamin:
adding kylin testnet

### DIFF
--- a/packages/apps-config/src/endpoints/testing.ts
+++ b/packages/apps-config/src/endpoints/testing.ts
@@ -193,6 +193,13 @@ export function createTesting (t: TFunction): LinkOption[] {
       }
     },
     {
+      info: 'kylin',
+      text: t('testnet.kylin-node.co.uk', 'Kylin Testnet', { ns: 'apps-config' }),
+      providers: {
+        'Kylin Network': 'wss://testnet.kylin-node.co.uk'
+      }
+    },
+    {
       info: 'litentry',
       text: t('rpc.test.litentry', 'Litentry Testnet', { ns: 'apps-config' }),
       providers: {
--- a/packages/apps-config/src/ui/colors.ts
+++ b/packages/apps-config/src/ui/colors.ts
@@ -256,7 +256,7 @@ export const nodeColors: Record<string, string> = [
   ['GEEK', nodeGeek],
   ['IpseTestnet', nodeIpse],
   ['Konomi Collator', nodeKonomi],
-  ['Kylin Node', nodeKylin],
+  ['Kylin Collator', nodeKylin],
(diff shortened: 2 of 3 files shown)


packages/apps-config/src/endpoints/testing.ts
```diff
@@ -192,6 +192,13 @@ export function createTesting (t: TFunction): LinkOption[] {
         'KILT Protocol': 'wss://full-nodes.kilt.io:9944/'
       }
     },
+    {
+      info: 'kylin',
+      text: t('testnet.kylin-node.co.uk', 'Kylin Testnet', { ns: 'apps-config' }),
+      providers: {
+        'Kylin Network': 'wss://testnet.kylin-node.co.uk'
+      }
+    },
```
jacogr, line 201: The same params are used here - https://github.com/polkadot-js/apps/blob/master/packages/apps-config/src/endpoints/testingRelayRococo.ts#L245-L252

Is this a parcahain on Rococo or a stand-alone network?

samelamin:
This is meant to be a standalone network with a different url

jacogr:
All ok, see the `rpc` vs `testnet` indicator now... (Although neither seems to load for me)

samelamin:
Hi @jacogr Sorry I didnt really go into alot of detail on this one

What we are trying to do is deploy a secondary test network that our community can start running validators on. I assume this is the correct place to add it

I am not sure why the url arent resolving for you. But I redeployed both load balancers and tested both urls and seem to be fine

https://testnet.kylin-node.co.uk
https://rpc.kylin-node.co.uk

samelamin:
Also another question for you Jacogr, I noticed that when I connect to my current node in production the css colour I set isn't taken affect 

Do I need to make CSS changes on my node, because I was under the impression the colours I set here should automatically take affect when connecting to the node via the Polkadot js app. 

Forgive me because I am not a front end developer but I don't see any obvious errors on the console that might indicate I did not set something up right

samelamin:
I think I figured out the issue, it's my node that was named incorrectly! 

> Also another question for you Jacogr, I noticed that when I connect to ourcurrent node in production the css colour I set isn't taken affect
> 
> Do I need to make CSS changes on my node, because I was under the impression the colours I set here should automatically take affect when connecting to the node via the Polkadot js app.
> 
> Forgive me because I am not a front end developer but I don't see any obvious errors on the console that might indicate I did not set something up right
> 
> Could you help me understand where the styles are being set and if I'm lucky why ours isn't?
> 
> My suspicion is because I only added a logo and what I really need is an SVG file that describes the UI. Atleast that's what I can tell from the most recent pr to "fix colours" by another project

jacogr (approved):
Thank you.

PR #5391, opened 2021-05-30
use selected theme in the validator stats page

gngchrs:
The validator stats page was not displayed correctly while in Dark theme, making it unreadable. (See attached before and after screenshots). This Pull request attempts to fix that, and I'd like some feedback. 
Thanks.

#### Before PR Changes (with Dark theme). I've annotated the chart titles with red rectangles which were not visible.
Screenshot 2021-05-30 at 08 45 59

#### After PR Changes (With Dark theme) (actual fix)
Screenshot 2021-05-30 at 08 23 28

#### After PR Changes (With Light theme)
Screenshot 2021-05-30 at 08 24 09

### DIFF
--- a/packages/page-staking/src/Query/Validator.tsx
+++ b/packages/page-staking/src/Query/Validator.tsx
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import type { Props } from './types';
+import type { ThemeProps } from '@polkadot/react-components/types';
 
 import React from 'react';
 import styled from 'styled-components';
@@ -28,11 +29,11 @@ function Validator ({ className = '', validatorId }: Props): React.ReactElement<
   );
 }
 
-export default React.memo(styled(Validator)`
+export default React.memo(styled(Validator)(({ theme }: ThemeProps) =>`
   .staking--Chart {
-    background: white;
+    background: ${theme.theme};
     border: 1px solid #eeecea;
     border-radius: 0.25rem;
     padding: 1rem 1.5rem;
   }
-`);
+`));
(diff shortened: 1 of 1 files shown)


jacogr (approved):
Brilliant! Thanks an absolute million for this.

packages/page-staking/src/Query/Validator.tsx
```diff
@@ -28,11 +29,11 @@ function Validator ({ className = '', validatorId }: Props): React.ReactElement<
   );
 }
 
-export default React.memo(styled(Validator)`
+export default React.memo(styled(Validator)(({ theme }: ThemeProps) =>`
   .staking--Chart {
-    background: white;
+    background: ${theme.theme};
```
jacogr, line 34: Just wondering if this should not rather be the table background? (Still has the same effect, but aligns closer wit the initial where the background is different from the page background)

jacogr:
CI is unhappy with some small stuff `yarn lint` locally should fix them both - https://github.com/polkadot-js/apps/pull/5391/checks?check_run_id=2703569061#step:3:486

packages/page-staking/src/Query/Validator.tsx
gngchrs: Ok. I'll update and see the result.

Thanks.

gngchrs: @jacogr I updated the PR to use the table variables instead, and this is the new look.

Screenshot 2021-05-30 at 10 51 07

jacogr (approved):
Thank you! These are the best PRs - something that bugs somebody is fixed by them. 

I really appreciate it.

PR #5417, opened 2021-06-02
added Genshiro as Kusama parachain; fixed colors and logos

alexdniep:


packages/apps-config/src/endpoints/productionRelayKusama.ts
```diff
@@ -71,6 +71,13 @@ export function createKusama (t: TFunction): EndpointOption {
           Encointer: 'wss://canary.encointer.org'
         }
       },
+      {
+        info: 'genshiro',
+        text: t('rpc.test.equilibriumtestnet', 'Genshiro', { ns: 'apps-config' }),
+        providers: {
+          Equilibrium: 'wss://testnet.equilibrium.io'
```
jacogr, line 78: Is this correct? Genshiro does indeed look like a real network, not a testnet, so the URL does have me a bit confused.

alexdniep: Production node will be deployed soon, so we decided to do in the same manner as current projects do(eg. https://github.com/polkadot-js/apps/blob/master/packages/apps-config/src/endpoints/productionRelayKusama.ts#L47 and https://github.com/polkadot-js/apps/blob/master/packages/apps-config/src/endpoints/testing.ts#L52)

jacogr (approved):
Thank you.

PR #5436, opened 2021-06-03
update paraid for kylin

samelamin:
We had to deregister 2000 due to the wrong wasm blob, but I couldn't register it again

So updating to 2013

Shame I lost the special number!


jacogr (approved):

jacogr:
This is where a shell runtime (like the Parity team used for Statemint) is handy. It is a bare-bones instance and it only allows upgrading. So cannot do much bad with it... well, until upgraded that is :)

samelamin:
You have to share the statemint info with me, this sounds very useful!

PR #5446, opened 2021-06-04
[OAK Network] Update endpoint to 'wss://rpc.testnet.oak.tech'

chrisli30:
Hi there,

No rush, but we think it clearer adding a "rpc." prefix to our API endpoint.

The old and the new endpoints both work for now, but we will use the new from this PR onward.

Thanks.


jacogr (approved):

chrisli30:
Thanks for the quick update, Jaco!

PR #5449, opened 2021-06-04
Add thales network support

bit-god:
support thales network


bit-god:
sadly. 

maybe we copy some code from other's repo, but these are just used for ethereum & evm account,  our chain code will refactor totally.

PR #5411, opened 2021-06-02
Adding KlugDossier

sbnair:

PR #5456, opened 2021-06-04
Add KlugDossier

sbnair:


packages/apps-config/src/endpoints/testing.ts
```diff
@@ -199,6 +199,13 @@ export function createTesting (t: TFunction): LinkOption[] {
         'Kylin Network': 'wss://testnet.kylin-node.co.uk'
       }
     },
+    {
```
jacogr, line 202: This needs to be before Kylin.

jacogr:
What is different to #5411? Would suggest to close that one if this one replaces it.

packages/apps-config/src/endpoints/testing.ts
sbnair: Ok, done the changes in next commit.

jacogr (approved):

PR #5511, opened 2021-06-08
Adds Moonriver

crystalin:
Adds Moonriver parachain under Kusama


packages/apps-config/src/endpoints/productionRelayKusama.ts
```diff
@@ -157,6 +157,14 @@ export function createKusama (t: TFunction): EndpointOption {
         providers: {
           StakeTechnologies: 'wss://rpc.shiden.plasmnet.io'
         }
+      },
+      {
+        info: 'moonriver',
+        paraId: 2023,
```
jacogr, line 163:
```suggestion
        isUnreachable: true,
        paraId: 2023,
```

Extra flag until RPC is up. (More-or-less when WASM is available, if reachable then)

packages/apps-config/src/ui/logos/index.ts
```diff
@@ -69,6 +69,7 @@ import nodeMath from './nodes/math.svg';
 import moonbeam from './nodes/moonbeam.png';
 import nodeMoonrock from './nodes/moonrock.png';
 import moonshadow from './nodes/moonshadow.png';
+import moonriver from './nodes/moonriver.svg';
```
jacogr, line 72: Attached file is `.png` - this is why CI fails (had me confused for a second)

crystalin:
Thank you for looking at it @jacogr .
I updated it

PR #5557, opened 2021-06-11
Register Opportunity Test Network on PolkadotJS apps

billjhlee:
PR for registration of Opportunity Test Network on PolkadotJS apps.
Opportunity is Standard Protocol's Test Network.

Changes:

- add packages/apps-config/src/ui/logos/nodes/opportunity.png
- update packages/apps-config/src/api/spec/index.ts
- update packages/apps-config/src/api/spec/opportunity.ts
- update packages/apps-config/src/endpoints/testing.ts
- update packages/apps-config/src/ui/colors.ts
- update packages/apps-config/src/ui/logos/index.ts
- update packages/apps/public/locales/en/apps-config.json
- update packages/apps/public/locales/ko/translation.json

PR #5574, opened 2021-06-12
Update opportunity types

hskang9:
resolves #5557

### DIFF
--- a/packages/apps-config/src/api/spec/opportunity.ts
+++ b/packages/apps-config/src/api/spec/opportunity.ts
@@ -38,7 +38,15 @@ const definitions: OverrideBundleDefinition = {
           ]
         },
         AmountOf: 'Amount',
-        Amount: 'i128'
+        Amount: 'i128',
+        DataVersion: 'u64',
+        RequestIdentifier: 'u64',
+        SpecIndex: 'Vec<u8>',
+        CDP: {
+          liquidation_fee: ('Balance', 'Balance'),
+          max_collateraization_rate: ('U256', 'U256'),
+          stability_fee: ('Balance', 'Balance')
+        }
       }
     }
   ]
--- a/packages/apps-config/src/api/spec/opportunity.ts
+++ b/packages/apps-config/src/api/spec/opportunity.ts
@@ -43,9 +43,9 @@ const definitions: OverrideBundleDefinition = {
         RequestIdentifier: 'u64',
         SpecIndex: 'Vec<u8>',
         CDP: {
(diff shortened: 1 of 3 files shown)


packages/apps-config/src/api/spec/opportunity.ts
```diff
@@ -38,7 +38,15 @@ const definitions: OverrideBundleDefinition = {
           ]
         },
         AmountOf: 'Amount',
-        Amount: 'i128'
+        Amount: 'i128',
```
jacogr, line 41: Not sure I see `CurrencyIdOf` (mentioned in https://github.com/polkadot-js/apps/issues/5571) - unless removed now on the chain itself.

jacogr: Yes, that is exactly it - the type doesn't appear in the type definitions, hence the question.

hskang9: do you mean the type definition [here](https://github.com/polkadot-js/apps/blob/fad1f3da1076e085cfaa35133adf573114912f0f/packages/apps-config/src/api/spec/opportunity.ts#L22) or somewhere else? I think CurrencyIdOf in our runtime is from the code [here](https://github.com/digitalnativeinc/standard-substrate/blob/4a2f50689975d2a723b12c864e3f8233ed141e55/orml/currencies/src/lib.rs#L88)

hskang9: > Not sure I see `CurrencyIdOf` (mentioned in #5571) - unless removed now on the chain itself.

This PR includes `CurrencyIdOf`, and it passed `yarn ci:chainTypes` command.

hskang9: > Not sure I see `CurrencyIdOf` (mentioned in #5571) - unless removed now on the chain itself.

Ah...now I understand. I set the testing unreachable to check if there are outdated types.

packages/apps-config/src/endpoints/testing.ts
```diff
@@ -263,6 +263,7 @@ export function createTesting (t: TFunction, firstOnly?: boolean): LinkOption[]
     },
     {
       info: 'opportunity',
+      isUnreachable: true,
```
jacogr, line 266:
```suggestion
```

Should be fine without.

jacogr (approved):
Thank you. Should fix https://github.com/polkadot-js/apps/issues/5577

PR #5576, opened 2021-06-13
Add types for beresheet runtime

raykyri:
We're starting to use a separate runtime for beresheet.


jacogr:
You probably want a specific entry for the spec -> types mapping, e.g. somewhere here - https://github.com/polkadot-js/apps/blob/master/packages/apps-config/src/api/spec/index.ts#L134

As it stands, it is added, but won't be used.

jacogr:
Once it pulls in the types correctly, should fix https://github.com/polkadot-js/apps/issues/5626

raykyri:
Thanks for the tip, just getting back to this because we pushed the new Beresheet runtime yesterday.

PR #5522, opened 2021-06-09
Fix Payouts -> My Validators toggle on page refresh

wachulski:
Before: after page refresh and before accounts were loaded
validators collection was empty which resulted in 'My validators'
being disabled.

Now: toggle accessibility relies upon validators refresh and not just
single initial assignment.

PR #5583, opened 2021-06-14
Payouts: 'My validators' set correctly on page refresh

wachulski:
Currently if you refresh the Payouts page 'My validators' are not selected automatically.
This PR improves this as when the accounts (own validators) are loaded/available then 'My validators' are selected (if own validators exist).
This is to make the behaviour same as when navigating to Payouts from other staking area tabs.

Note: initial review done by @krzysztof-jelski in https://github.com/polkadot-js/apps/pull/5522


krzysztof-jelski (approved):

packages/page-staking/src/Payouts/index.tsx
```diff
@@ -150,11 +150,19 @@ function getOptions (api: ApiPromise, eraLength: BN | undefined, historyDepth: B
...
   const { t } = useTranslation();
   const { api } = useApi();
-  const [hasOwnValidators] = useState(() => ownValidators.length !== 0);
-  const [myStashesIndex, setMyStashesIndex] = useState(() => (isFunction(api.tx.staking.payoutStakers) && hasOwnValidators) ? 0 : 1);
+  const hasOwnValidators = useMemo(() => ownValidators.length !== 0, [ownValidators]);
+  const [myStashesIndex, setMyStashesIndex] = useState(() => getMyStashesIndex(api, hasOwnValidators));
+
+  useEffect(() => {
+    setMyStashesIndex(getMyStashesIndex(api, hasOwnValidators));
+  }, [api, hasOwnValidators]);
```
jacogr, line 165:
```suggestion
  }, [api, hasOwnValidators]);

```

Weirdly the linter didn't complain here...

packages/page-staking/src/Payouts/index.tsx
```diff
@@ -150,11 +150,19 @@ function getOptions (api: ApiPromise, eraLength: BN | undefined, historyDepth: B
   return [{ text: '', value: 0 }];
 }
 
+function getMyStashesIndex (api: ApiPromise, hasOwnValidators: boolean) {
+  return (isFunction(api.tx.staking.payoutStakers) && hasOwnValidators) ? 0 : 1;
+}
```
jacogr, line 155:
```suggestion
function getMyStashesIndex (hasOwnValidators: boolean) {
  return hasOwnValidators ? 0 : 1;
}
```

Removing support for ancient payouts in https://github.com/polkadot-js/apps/pulls (not sure this needs to be an own function)

jacogr (commented):
Actually not quite convinced on the need for this - so what will happen now is that it loads, no validators and then mid-way would swap to "Validators". And then would need to swap back.

Validators I'm also not 100% sure about - there are some (were some?) issues that when a validator drops out of the set, it actually doesn't display. Not sure if this has been fixed, however it has caused myself to miss payouts in the past actually... (It certainly is faster/most optimized)

krzysztof-jelski:
Current version (without the PR) is really annoying. If you want to make a payout through the UI, you need to open the "Account actions" tab (or possibly some other) and then switch to "Payouts". No other way to see if there is a payout for your validator. Of course I'm assuming loading/refreshing the page directly (which for me happens quite often, as I have the extension turned off by default; enabling it triggers a page refresh).

krzysztof-jelski:
We may consider setting the "My stashes" option as the default. Then we wouldn't have the mid-way swap after the validators are detected. Or just not select/enable any of the options until whole loading is complete.

jacogr:
The idea was to make "validators" default for, well validators, since it has a much, much faster loading time. With some caveats, i.e. if chilled it won't show up in there (no prefs), no breakdown for all exposures, ...

Obviously this is not great, since we first need to figure out if there are validators - which in itself is not that bad to determine, however there is a truckload of queries at the same time, so stuff gets very delayed and it is not quite snappy to say the least... (Here things should actually improve a bit, instead of the keys/entries there are PRs open to expose total-num queries for the nominators/validators)

I'm just not convinced at all on swapping the selection mid-way here. The disabling may actually work - there would be a delay before it can start, however at the same time it will also not kick off yet more stuff while there is already a lot going on. I guess just need some way of indicating that it is "pre-retrieval waiting"

packages/page-staking/src/Payouts/index.tsx
wachulski: sure

packages/page-staking/src/Payouts/index.tsx
wachulski: yeah, no it adds no gain - will inline

wachulski:
I'll disable both till it loads.

wachulski:
@jacogr - so, just to clarify/confirm - you suggest that we disable the toggle group ('My validators', 'My stashes') till the accounts are loaded (and specifically the owned validators are passed to the component) and also wait with loading rewards till then. Now, when the accounts are loaded and owned validators come in (if any, could be empty still), we unblock the toggle and start loading rewards. Is that correct?

jacogr:
That should work - not sure how the UX is on that. So it has some benefits -

- it forces you to select (if you want to change)
- it has all the info to make the correct call
- it most probably helps performance since we are not trying to do more

We may also want to store the last selection here, i.e. if a user previously used validators, it will select that. If they previously used stashes, it will select that. (Obviously depending if the selection can be made, i.e. validators could be an option yesterday, but not today anymore) So we can optimize the overall UX a bit, even if it is slower to start at this point.

packages/page-staking/src/Payouts/index.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+
+    it('when some own validators, but after accounts load and rerender, then `My validators` is selected',
+      async () => {
+        payoutsPage.renderPage();
+        await payoutsPage.expectMyValidators({ isDisabled: true, isSelected: false });
+        payoutsPage.renderPage([someOwnedValidator]);
+        await payoutsPage.expectMyValidators({ isDisabled: false, isSelected: true });
+      });
+
+    it('when loading own era rewards, then `My validators` and `My stashes` are disabled', async () => {
```
wachulski, line 60: I'm unsure if that's the best behaviour, to disable changes on e.g. long loading. What do you think?

wachulski:
> That should work - not sure how the UX is on that. So it has some benefits -
> (...)

I just hid the Payouts till we know about own validators.

When it comes to saving selection preferences - I'd make that one its own issue/PR.

Also - I see there is room for improvement. E.g. Having sth like `GetTrackedCall` or applying `isLoading` to some hook return structures would help improve UX overall (hiding when no data yet or blocking queries till some dep data arrives).

//cc @jacogr

packages/page-staking/src/index.tsx
```diff
@@ -125,10 +125,10 @@ function StakingApp ({ basePath, className = '' }: Props): React.ReactElement<Pr
       />
       <Switch>
         <Route path={`${basePath}/payout`}>
-          <Payouts
+          {areAccountsLoaded && !isLoadingStashes && <Payouts
```
jacogr, line 128: I actually just did something similar here - https://github.com/polkadot-js/apps/pull/5674/files#diff-f92441f5f0aafb2ba6f441b16b386be924463f6c3fd4d373a41ee5da1e1f4e2dR84

(basically in that case the route is not available since we don't have any stashes)

The right way is indeed, like above to hide the tab, not to hide the content since then it appears unresponsive for however long it takes.

packages/react-hooks/src/useOwnStashInfos.ts
```diff
@@ -94,11 +104,16 @@ export function useOwnStashInfos (): StakerState[] | undefined {
...
   return useMemo(
-    () => ownStashes && queried && ownStashes.length === Object.keys(queried).length
-      ? ownStashes
-        .filter(([stashId]) => queried[stashId])
-        .map(([stashId]) => getStakerState(stashId, allAccounts, queried[stashId]))
-      : undefined,
-    [allAccounts, ownStashes, queried]
+    () => {
+      return {
+        isLoadingStashes: isLoading,
```
jacogr, line 109: Not sure this is needed - if 0, then nothing is loaded and we actually don't want/need payouts), it > 0 length, it is loaded. So would just forgo this.

packages/react-hooks/src/useOwnStashes.ts
```diff
@@ -31,18 +31,18 @@ function getStashes (allAccounts: string[], ownBonded: Option<AccountId>[], ownL
...
-  const { allAccounts, hasAccounts } = useAccounts();
+  const { allAccounts, areAccountsLoaded, hasAccounts } = useAccounts();
   const { api } = useApi();
   const ownBonded = useCall<Option<AccountId>[]>(hasAccounts && api.query.staking?.bonded.multi, [allAccounts]);
   const ownLedger = useCall<Option<StakingLedger>[]>(hasAccounts && api.query.staking?.ledger.multi, [allAccounts]);
 
   return useMemo(
-    () => hasAccounts
-      ? ownBonded && ownLedger
+    () => areAccountsLoaded && ownBonded !== undefined && ownLedger !== undefined
```
jacogr, line 40: Not sure what this specifically addresses? What behavior is different here? ... I'm probably slow today, so has not dawned on me by just looking :)

packages/page-staking/src/index.tsx
wachulski: sure, definitely better/proper, thx

packages/react-hooks/src/useOwnStashes.ts
wachulski: Previously when accounts were not loaded it saw this as `[]` instead of undefined.
But yeah - I collapsed `ownBonded !== undefined && ownLedger !== undefined` to the prev version

packages/react-hooks/src/useOwnStashInfos.ts
wachulski: I followed `useOwnEraRewards` return structure. But I understand the trick of undefined vs [] to indicate loading state. I'd keep this for clean communication of intention, as `areAccountsLoaded` or that one in rewards.

packages/page-staking/src/Payouts/index.spec.ts
wachulski: @jacogr - pls confirm if this makes sense. I'm not sure as I wasn't exposed to big data loads on those pages :)

jacogr (commented):
I am struggling with this one at this point. So the original intent was "don't start loading when we don't have the correct info".

At this point as it stands payout are not available when we have not retrieved the stashes, so the user, once payouts are there, can always make a click. The reason it is hidden is because of other issues, e.g. with no stashes at all, payouts are completely non-applicable (and in some cases just ends up loading forever due to this)

So at this point, I'm not sure the problem this PR sets out to solve is applicable at all (anymore). I guess the "hiding" is a low-tech solution.

(There are obviously a number of enhancements that did come out from the discussions here, e.g. saving state for ears and selection per chain)

krzysztof-jelski:
For me the biggest problem is still that the UI state is wrong after refreshing the page (or accessing the payouts URL directly). This PR fixes it.

jacogr:
Low tech - 

if (!hasOwnStashes) return null;

that way if going to the page directly it won’t try and load and we can skip any major updates?

wachulski:
> if (!hasOwnStashes) return null;
> 
> that way if going to the page directly it won’t try and load and we can skip any major updates?

@jacogr - do you want sth like this:

```
const { isLoadingStashes, hasOwnStashes, ownStashes } = useOwnStashInfos();
// which could be even collapsed to only hasOwnStashes (boolean | undefined) + ownStashes 
// or even only ownStashes of array|undefined ?
```

Also, I understand that your `if (!hasOwnStashes) return null;` would then disable loading the page on 1) when there are no stashes for the user 2) when stashes are being loaded and we don't know whether there are any.

Also, where to put that line then?

wachulski:
@jacogr -> this is accruing conflicts with the master branch more and more.
Could we merge it or I should close it?

jacogr:
Closing. Replaced by alternative.

PR #5597, opened 2021-06-15
Feature/update socket url

Xavi-vmo:


jacogr:
Closing this outright - it removes all the endpoints we have been adding. Additionally, prefer not to have unknown workflows in here, it cannot be managed or checked.

PR #5607, opened 2021-06-15
updated genshiro parachain url and id

alex-nax:


jacogr (approved):

packages/apps-config/src/endpoints/productionRelayKusama.ts
```diff
@@ -92,11 +92,11 @@ export function createKusama (t: TFunction): EndpointOption {
       },
       {
         info: 'genshiro',
-        homepage: 'https://equilibrium.io',
-        isUnreachable: true,
```
jacogr, line 96: I generally only enable these when -

- the url is reachable (i.e. RPC node running), and
- the wasm is registered on-chain (i.e. it appears in the parathread list)

Happy to do so now if the wasm is going soon.

Veniamin: Wasm is on the way and will be registered today

PR #5608, opened 2021-06-15
Fix author display on Moonbeam

JoshOrndorff:
This PR updates the storage value used to retreive Moonbeam's author account to reflect the changes from https://github.com/PureStake/moonbeam/pull/461

cc @joelamouche @crystalin


JoshOrndorff:
Actually, this isn't working @joelamouche can you please take this over?

joelamouche:
In what networks is this implemented? Not in moonshadow nor moonriver right? So we need to support both before the transition?

joelamouche:
OK so it seems moonshadow is using a call called "authorIds" whereas the other ones use mappingWithDeposit, correct?

joelamouche:
closing in favor of https://github.com/polkadot-js/apps/pull/5617

PR #5617, opened 2021-06-16
Fix author mapping call

joelamouche:
Change of function name for author mapping
- Moonshadow currently uses authorIds
- New version will use authorMapping

### DIFF
--- a/packages/react-query/src/BlockAuthors.tsx
+++ b/packages/react-query/src/BlockAuthors.tsx
@@ -46,7 +46,7 @@ function BlockAuthorsBase ({ children }: Props): React.ReactElement<Props> {
       let lastHeaders: HeaderExtendedWithMapping[] = [];
       let lastBlockAuthors: string[] = [];
       let lastBlockNumber = '';
-      const isAuthorIds = isFunction(api.query.authorMapping?.authorIds);
+      const isAuthorIds = isFunction(api.query.authorMapping?.authorIds); // TODO-MOONBEAM reevaluate in a month: 07/16/21
       const isAuthorMappingWithDeposit = isFunction(api.query.authorMapping?.mappingWithDeposit);
 
       // subscribe to all validators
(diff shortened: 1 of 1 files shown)


joelamouche:
or @crystalin

crystalin (approved):

joelamouche:
ready for review @jacogr

JoshOrndorff (approved):
Thanks for doing this @joelamouche

packages/react-query/src/BlockAuthors.tsx
```diff
@@ -61,9 +62,16 @@ function BlockAuthorsBase ({ children }: Props): React.ReactElement<Props> {
...
-          } else if (isAuthorMapping && lastHeader.digest.logs && lastHeader.digest.logs[0] && lastHeader.digest.logs[0].isConsensus && lastHeader.digest.logs[0].asConsensus[1]) {
+          } else if (isAuthorMappingWithDeposit && lastHeader.digest.logs && lastHeader.digest.logs[0] && lastHeader.digest.logs[0].isConsensus && lastHeader.digest.logs[0].asConsensus[1]) {
             // Some blockchains such as Moonbeam need to fetch the author accountId from a mapping
-            thisBlockAuthor = (await api.query.authorMapping.mapping(lastHeader.digest.logs[0].asConsensus[1])).toString();
+            thisBlockAuthor = ((await api.query.authorMapping.mappingWithDeposit(lastHeader.digest.logs[0].asConsensus[1])).toHuman() as {
+              account: string;
+              deposit: string;
+            }).account;
+            lastHeader.authorFromMapping = thisBlockAuthor;
+          } else if (isAuthorIds && lastHeader.digest.logs && lastHeader.digest.logs[0] && lastHeader.digest.logs[0].isConsensus && lastHeader.digest.logs[0].asConsensus[1]) {
```
jacogr, line 72: Not against both options, however...

There are tons and tons of stuff in the UI that supports ancient, current and future and then gets forgotten about. (See for instance the open staking payouts PR, cruft hang around for ages). In this case, we just need a plan to come and cleanup when there is a "everything on this". All code carries a cost and in this case it is slightly worse since I don't have any idea on what is current and what is future :)

jacogr (commented):
Just one small comment/question on when to drop at least some of this stuff. (It is out of hand everywhere...)

packages/react-query/src/BlockAuthors.tsx
joelamouche: I understand...
So here is what I suggest: I add `// TODO-MOONBEAM reevaluate in a month: 07/16/21` and I set myself a reminder in a month.
And every time I add something that will get deprecated I add TODO-MOONBEAM with a date that way you can keep track of who needs to cleanup

jacogr: Happy with the TODO. (Just want to make sure it has a checking every now and again as to status)

jacogr (approved):
Thank you.

Issue #476, opened 2018-11-29
Header format failure with local dev chain on substrate master


jnaviask:
Running in Chrome both locally hosted and via https://polkadot.js.org/apps/#/explorer gives the following error:
```
index.js:128 Error: Error formatting '[object Object]' as 'Header': Expected hex value to convert, found '[object Object]'
    at f (format.js:50)
    at r.map (format.js:69)
    at Array.map (<anonymous>)
    at t.default (format.js:61)
    at t.default (output.js:41)
    at t.default.formatResult (index.js:140)
    at Object.s [as callback] (index.js:103)
    at t.default.onSocketMessageSubscribe.e [as onSocketMessageSubscribe] (index.js:127)
    at WebSocket.onSocketMessage.e (index.js:82)
```

Steps to reproduce:
1. `cargo build` with current substrate master (https://github.com/paritytech/substrate/commit/01ff6f0034f84493db7d8b46b74a997600945678 at time of issue open)
2. `cargo run -- --dev`.
3. Open https://polkadot.js.org/apps/#/explorer. You should see it connect to a node, but no block data.
4. Check site console for error.

jacogr:
Master only supports poc-2. For substrate master support, use https://polkadot.js.org/apps/next/

"next" indicating the "next" branch - will be merged into master as soon as Substrate hits 1.0-beta and Polkadot poc-3 is there.

jacogr:
EDIT: Just to note, you can change the node it connects to under "Settings" - by default it connects to BBQ, but just swapping there connects to a local node. (I have updated master README as well with a note to the link)

PR #5644, opened 2021-06-21
Integrate substrate connect with Apps

wirednkod:
### **Summary**:
This PR is introducing the (experimental) light client (smoldot) into the Polkadot JS apps throush `substrate-connect`.

### **Visible Changes**: 
A new "server" is introduced under 'Live Relays & Parachains', 'Polkadot' and 'Kusama' and under 'Test Relays & Parachains', 'Westend' as can be seen below:

PolkadotJS substrate connect Sidebar
  
### **Description**: 
Substrate connect nodes can be connected either through in-browser light client or through extension.

### **Development process**: 
In order to make this as less intrusive as possible the following steps took place:
1. A new type is introduced called 'substrate-connect' and default naming the existing to 'json-rpc' (see [PR #476](https://github.com/polkadot-js/ui/pull/476))
2. Following the existing implementation of using the URL parameter `rpc` for existing nodes (e.g. _`?rpc=wss%3A%2F%2Frpc.polkadot.io#/explorer`_) - a new url param is introduced called `sc` (from substrate-connect) that has the name of the node and the characteristic `-substrate-connect` following (e.g. `?sc=kusama-substrate-connect#/explorer`). This way it is  easier a differentiation between the two, without altering existing implementation.
3. Based on api type passed in early stage of the app (`Api.tsx`) a choice is made on whether substrate-connect's Detector will be used or existing implementation).
4. Based on new type introduced (1. above), a new function is introduced (`packages/apps-config/src/endpoints/util.ts`) for endpoints that receives the url and the type (`'json-rpc` or `substrate-connect` - defaults to first if none given) and returns the url type, reducing this way duplication of code and typos).
5. All fallback methods (e.g. initSettings.ts) are updated as per bullets here.
6. Finally all endpoints are updated using function mentioned on bullet 4 and  3 new endpoints for substrate-connect are introduced as mentioned on **Visible Changes** above.


packages/react-api/package.json
```diff
@@ -10,5 +10,8 @@
     "@polkadot/extension-dapp": "^0.38.5",
     "@polkadot/x-rxjs": "^6.9.1",
     "fflate": "^0.7.0"
+  },
+  "devDependencies": {
```
raoulmillais, line 14: This should be a proper dependency not a dev dependency

wirednkod: Done. Thank you

packages/apps-config/src/endpoints/productionRelayKusama.ts
```diff
@@ -20,10 +21,11 @@ export function createKusama (t: TFunction): EndpointOption {
...
     providers: {
-      Parity: 'wss://kusama-rpc.polkadot.io',
-      OnFinality: 'wss://kusama.api.onfinality.io/public-ws',
-      'Patract Elara': 'wss://kusama.elara.patract.io',
-      Pinknode: 'wss://rpc.pinknode.io/kusama/explorer'
+      Parity: createProviderUrl('wss://kusama-rpc.polkadot.io'),
+      OnFinality: createProviderUrl('wss://kusama.api.onfinality.io/public-ws'),
+      'Patract Elara': createProviderUrl('wss://kusama.elara.patract.io'),
+      Pinknode: createProviderUrl('wss://rpc.pinknode.io/kusama/explorer'),
+      'light client': createProviderUrl('kusama-substrate-connect', 'substrate-connect')
```
jacogr, line 28: Instead of using `createProviderUrl` why not simply push through only `kusama-substrate-connect` (ends with the `-substrate-connect` as used elsewhere) or something like `substrate-connect://kusama-substrate-connect` and then handling it in actual expansion? 

The light client/ws tests below then just filters on `startWith('substrate-connect')` or `endsWith('-substrate-connect')` depending on what is used.

packages/react-api/src/Api.tsx
```diff
@@ -186,40 +188,76 @@ function Api ({ children, store, url }: Props): React.ReactElement<Props> | null
...
+            });
+        }).catch(console.error);
+      } catch (err) {
+        console.error(err);
+      }
+    }
+
+    if (apiType?.type === 'substrate-connect' as EndpointType) {
+      // eslint-disable-next-line no-void
+      void substrateConnectApi(apiType.param);
```
jacogr, line 231:
```suggestion
     substrateConnectApi(apiType.param).catch(console.error);
```

Hate not handling the promises... just logging is 100% fine, but at least can see when something goes belly up.

packages/react-api/src/Api.tsx
```diff
@@ -186,40 +188,76 @@ function Api ({ children, store, url }: Props): React.ReactElement<Props> | null
...
+            .then(setExtensions)
+            .catch(console.error);
+
+          loadOnReady(api, injectedPromise, store, types)
+            .then(setState)
+            .catch((error): void => {
+              console.error(error);
+
+              setApiError((error as Error).message);
+            });
```
jacogr, line 222: All this seems mostly duplicated - which is going to create an issue when things are moved around (yet again, these happen sadly too frequently here...)

packages/react-api/src/Api.tsx
```diff
@@ -186,40 +188,76 @@ function Api ({ children, store, url }: Props): React.ReactElement<Props> | null
...
-          console.error(error);
-
-          setApiError((error as Error).message);
-        });
-    });
-
-    setIsApiInitialized(true);
-  // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, []);
+    async function substrateConnectApi (network: string) {
```
jacogr, line 197: The issue with the free-standing function is that is should be hoisted to file-level - it gets re-created each render.

jacogr: ... if the duplication is removed as we have it now, it should actually sort itself out in one way or another

jacogr (commented):
Nice to have this. some small comments all around.

jacogr:
As for the test failure here - https://github.com/polkadot-js/apps/pull/5644/checks?check_run_id=2875618334#step:3:518

Adapt this - https://github.com/polkadot-js/apps/blob/master/jest.config.cjs#L28

packages/apps-config/src/endpoints/productionRelayKusama.ts
wirednkod: That would be a solution but isn't it extremely "hack-y"? Adding the "type" as a parameter, allows for different kinds to be integrated in the future - meaning: existing (and default): `type: 'json-rpc'`, light-client of substrate: `type: 'substrate-connect'`, some other awesome light client that does not exist yet: `type: 'awesome-light'`. It is better I think than string parsing. In addition light client/ws tests are already fixed based on the string (almost) as you propose

wirednkod:
> As for the test failure here - https://github.com/polkadot-js/apps/pull/5644/checks?check_run_id=2875618334#step:3:518
> 
> Adapt this - https://github.com/polkadot-js/apps/blob/master/jest.config.cjs#L28

Im not sure I understand what exactly to do here. Can you please elaborate?

packages/apps-config/src/endpoints/productionRelayKusama.ts
raoulmillais: We had a long debate about this and thought that overloading the idea of a URL with a custom non-standard scheme was a hack since it would be surprising for people unfamiliar with the code and didn't really accurately reflect what was going on.  In the case where you connect with a light client you are not really using a URL at all - it's conceptually completely different.  Furthermore we are going to need to extend how substrate-connect is configured to support parachains (coming soon!) If we want to add support for that to polkadotjs apps too then there will no sensible way to capture that extra configuration in a URL, but it could neatly be added to the new `Endpoint` type.

wirednkod:
> As for the test failure here - https://github.com/polkadot-js/apps/pull/5644/checks?check_run_id=2875618334#step:3:518
> 
> Adapt this - https://github.com/polkadot-js/apps/blob/master/jest.config.cjs#L28

Worked like a charm. @jacogr any suggestions on [this error](https://github.com/polkadot-js/apps/pull/5644/checks?check_run_id=2876719599#step:3:511)?

I thought that it needs to run with `experimental-vm-modules` option but from [here](https://github.com/polkadot-js/apps/pull/5644/checks?check_run_id=2876719599#step:3:511) it seems to do use it.

jacogr:
Umpf. I've had endless issues with `import.meta` and eventually gave up. (Mine was related to testing, but more actually with the fact that I needed CJS and ESM and there was just no way to have fallbacks) 

This seems to have some comments on it - https://stackoverflow.com/questions/64961387/how-to-use-import-meta-when-testing-with-jest

packages/apps-config/src/endpoints/productionRelayKusama.ts
jacogr: I hear you, but making calls here is not the way to go. If you want a specific extension, then add it instead of wrapping everything. (This already conflicts with PRs that have been subsequently opened as it stands and just gets worse since it has a lot of churn)

Even as it is specified, it is utterly confusing. So I you want something specific, then rather be explicit on the singular items instead of wrapping all the ones that already do work. So you can make it a `string | SomeType` and just for the handful, specific `SomeType`.

raoulmillais: Yeah using a union type makes total sense. You get the advantage of not messing with existing URLs and the advantage of a richer type for substrate connect

wirednkod: > So you can make it a `string | SomeType` and just for the handful, specify `SomeType` to whatever it may be.

I also agree with that. I will proceed with all needed changes for that

tomaka:
For what it's worth, the `import.meta` usage is here:

https://github.com/paritytech/smoldot/blob/b3d71d7d6b6e86bb54e0ed0cc39ca331682b1066/bin/wasm-node/javascript/src/index.js#L44-L50

While I've wrote it this way in order to have maximum compatibility, there might also be a better alternative way to spawn a worker?

jacogr:
@tomaka The `import.meta` usage is fully sane as linked, it is certainly the way to do things. The problem here is that Jest (the test framework) atm has one foot in the door with proper ESM support, it is still "experimental". 

Here seems to be something that hacks in support for Jest (found via another issue on Google) -

https://github.com/snowpackjs/create-snowpack-app/commit/c84d51bf5d10db82d6ff459dc9618710ea72f293#diff-93c47b837494a218df311b047db676e584f7a3307ed4c630e1c77dda7a404299

wirednkod:
> [snowpackjs/create-snowpack-app@c84d51b#diff-93c47b837494a218df311b047db676e584f7a3307ed4c630e1c77dda7a404299](https://github.com/snowpackjs/create-snowpack-app/commit/c84d51bf5d10db82d6ff459dc9618710ea72f293#diff-93c47b837494a218df311b047db676e584f7a3307ed4c630e1c77dda7a404299)

@jacogr I tried both solutions provided but I have nothing. With the latter quoted above babel-ject returns error (`TypeError: babelJest.createTransformer is not a function`). This `import.meta.url` starts to be really hard to resolve 😞

jacogr:
For the tests, we may as well just stub that import to resolve to a mock that won’t have the issues. I have done it before, just cannot quite remember “how”. 

It will need to be resolved one way or another at some point, but as of now it is just a PITA so would just not have it (in tests)

raoulmillais:
New Jest now has ESM support - but the new release has some caveats: https://jestjs.io/blog/2021/05/25/jest-27

wirednkod:
> New Jest now has ESM support - but the new release has some caveats: https://jestjs.io/blog/2021/05/25/jest-27

apps has Jest updated to ^27.0.4. The `import.meta.url` issue though AFAIK is irrelevant.

raoulmillais:
> > New Jest now has ESM support - but the new release has some caveats: https://jestjs.io/blog/2021/05/25/jest-27
> 
> apps has Jest updated to ^27.0.4. The `import.meta.url` issue though AFAIK is irrelevant.

It's not irrelevant because `import.meta.url` is part of the ESM spec. Shame it doesn't work though

wirednkod:
> It's not irrelevant because `import.meta.url` is part of the ESM spec. Shame it doesn't work though

Oh sorry if I got misunderstood - By "irrelevant" I actually tried to say is that even though jest 27 seems to fully support ESM there are still some pending issues with `import.meta.url`. Im not sure if [this](https://github.com/facebook/jest/issues/11167) gives any clue to the progress of ESM migration on Jest

jacogr:
I think the issue with Jest is the config - it uses babel-jest however it uses the cjs version for the compilation of dependencies. So you could try and change the import here https://github.com/polkadot-js/apps/blob/master/babel.config.cjs from `require('@polkadot/dev/config/babel-config-cjs.cjs')` to `require('@polkadot/dev/config/babel-config-esm.cjs')` but it may as well break other stuff. 

If it does break/not work it is really not worth chasing down... I've lost track of the amount of hours I've spent getting Jest to behave across repos, so wouldn't want somebody else to waste time. So if it does fix, great, if not, would just stub the import and be done with it. (The tests really are not comprehensive enough for the UI that it would make any real difference - unlike other repos which are properly tests, the apps UI is my side-project so I can click on stuff in the blockchain)

wirednkod:
Thank you Jaco. I was thinking about making that change (one of the things I saw during the few hours I have chasing the error). I will see if that will make things work and not break too many other stuff.

Concerning this:
> would just stub the import and be done with it. 

Can you please elaborate bit more what you mean by `stubbing` the import? I dont need an actual example - just cannot put my mind to sync with your words

jacogr:
https://jestjs.io/docs/manual-mocks#mocking-node-modules

wirednkod:
> https://jestjs.io/docs/manual-mocks#mocking-node-modules

I dont think this can apply here. To explain: The `import.meta.url` error that is received as a test error exists in smoldot package, which is included in substrate-connect and initiated by the `SmoldotProvider` called from `Detector` in `Api.tsx`.

Error details:
```
FAIL  packages/page-bounties/src/Bounties.spec.ts
  ● Test suite failed to run

    Jest encountered an unexpected token

    Jest failed to parse a file. This happens e.g. when your code or its dependencies use non-standard JavaScript syntax, or when Jest is not configured to support such syntax.

    Out of the box Jest supports Babel, which will be used to transform your files into valid JS based on your Babel configuration.

    By default "node_modules" folder is ignored by transformers.

    Here's what you can do:
     • If you are trying to use ECMAScript Modules, see https://jestjs.io/docs/ecmascript-modules for how to enable it.
     • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
     • If you need a custom transformation specify a "transform" option in your config.
     • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.

    You'll find more details and examples of these config options in the docs:
    https://jestjs.io/docs/configuration
    For information about custom transformations, see:
    https://jestjs.io/docs/code-transformation

    Details:

    /forks/apps/node_modules/smoldot/src/index.js:56
      const worker = new _compatNodejs.Worker(new URL('./worker.js', import.meta.url));
                                                                            ^^^^
    SyntaxError: Cannot use 'import.meta' outside a module

      at Runtime.createScriptFromCode (node_modules/jest-runtime/build/index.js:1479:14)
      at Object.<anonymous> (node_modules/@substrate/connect/src/SmoldotProvider/SmoldotProvider.ts:11:1)
```

My assumption of the error that occurs in `packages/page-bounties/src/Bounties.spec.ts` (if i have understood the structure of testing correct) inits from `packages/page-bounties/src/Bounties.slow.spec.ts` file that renders the `<Api .... />` but I cannot really grasp how to `stub` either the `Detector` or the `<Api />` from the documentation provided above (Note: I am klms away from even being close of considering of being anything but a jest expert 🤕 )

jacogr:
Join the club of not being a Jest developer - it just makes me swear loudly at the best of times.

According to the docs, adding a `__mocks__/@substrate/connect.js` file is enough. The file would expose how you want it to look mocked.

From the docs -

"Scoped modules (also known as scoped packages) can be mocked by creating a file in a directory structure that matches the name of the scoped module. For example, to mock a scoped module called @scope/project-name, create a file at __mocks__/@scope/project-name.js, creating the @scope/ directory accordingly."

wirednkod:
> it just makes me swear loudly

Can believe that - just this - fixed it. it just made me swear loudly... Thank you @jacogr 
Locally tests passed..

jacogr:
Sorry for the slight delay, I have it top-of-the-list to go through through today so we can get this in.

packages/apps-config/src/endpoints/productionRelayKusama.ts
```diff
@@ -23,7 +24,8 @@ export function createKusama (t: TFunction): EndpointOption {
       Parity: 'wss://kusama-rpc.polkadot.io',
       OnFinality: 'wss://kusama.api.onfinality.io/public-ws',
       'Patract Elara': 'wss://kusama.elara.patract.io',
-      Pinknode: 'wss://rpc.pinknode.io/kusama/explorer'
+      Pinknode: 'wss://rpc.pinknode.io/kusama/explorer',
+      'light client': createProviderUrl('kusama-substrate-connect', 'substrate-connect')
```
jacogr, line 27: Now that I see this, we may have a small issue now. So the provider URLs are actually selected randomly (based on feedback, i.e. it spreads the loads between the providers). So this may now be randomly selected and based on what the user has available, may not work.

... will continue reading through, this may be addressed later on.

packages/apps-config/src/endpoints/util.ts
```diff
@@ -52,9 +54,12 @@ export function expandEndpoint (t: TFunction, { dnslink, genesisHash, homepage,
     .map(([host, value], index): LinkOption => ({
       ...base,
       dnslink: index === 0 ? dnslink : undefined,
+      isLightClient: typeof value === 'string' ? false : value.type === 'substrate-connect' as EndpointType,
       isRelay: false,
-      textBy: t('rpc.hosted.by', 'via {{host}}', { ns: 'apps-config', replace: { host } }),
-      value
+      textBy: typeof value === 'string'
```
jacogr, line 59: Prefer not using `typeof value`, `isString(value)` would do the job. (No difference, just a helper, everywhere where used here)

packages/apps-config/src/endpoints/util.ts
```diff
@@ -78,3 +83,7 @@ export function expandEndpoint (t: TFunction, { dnslink, genesisHash, homepage,
 export function expandEndpoints (t: TFunction, input: EndpointOption[], firstOnly?: boolean): LinkOption[] {
   return input.sort(sortLinks).reduce((result: LinkOption[], input) => result.concat(expandEndpoint(t, input, firstOnly)), []);
 }
+
+export function createProviderUrl (url: string, inpType = 'json-rpc'): Endpoint {
+  return { param: url, type: inpType as EndpointType };
+}
```
jacogr, line 89:
```suggestion
export function createProviderUrl (param: string, type: EndpointType): Endpoint {
  return { param, type };
}
```

Where we use this we are explicit, so rather be explicit here as well. (Plus since we don't have casting, it is slightly safer)

packages/apps/src/Endpoints/index.tsx
```diff
@@ -43,13 +44,21 @@ function isValidUrl (url: string): boolean {
   );
 }
 
+function getApiType (apiUrl: string): Endpoint {
+  if (apiUrl.includes('-substrate-connect')) {
+    return { param: apiUrl, type: 'substrate-connect' as EndpointType };
+  }
+
+  return { param: apiUrl, type: 'json-rpc' as EndpointType };
+}
```
jacogr, line 53:
```suggestion
function getApiType (param: string): Endpoint {
  if (apiUrl.includes('-substrate-connect')) {
    return { param, type: 'substrate-connect' as EndpointType };
  }

  return { param, type: 'json-rpc' as EndpointType };
}
```

The casting `as EndpointType` should really not be needed. Can see why, but TS should be slightly more clever here...

packages/apps-config/src/endpoints/util.ts
```diff
@@ -52,9 +54,12 @@ export function expandEndpoint (t: TFunction, { dnslink, genesisHash, homepage,
     .map(([host, value], index): LinkOption => ({
       ...base,
       dnslink: index === 0 ? dnslink : undefined,
+      isLightClient: typeof value === 'string' ? false : value.type === 'substrate-connect' as EndpointType,
```
jacogr, line 57:
```suggestion
      isLightClient: typeof value === 'string' ? false : value.type === 'substrate-connect' as EndpointType,
```

Not sure why the `as EndpointType` is needed here. Really should not be.

wirednkod:
> Sorry for the slight delay, I have it top-of-the-list to go through through today so we can get this in.

No worries - I'll resolve the conflicts as well to make this little bit easier for you

packages/apps/src/Endpoints/index.tsx
```diff
@@ -216,21 +225,29 @@ function Endpoints ({ className = '', offset, onClose }: Props): React.ReactElem
   const _onApply = useCallback(
     (): void => {
       settings.set({ ...(settings.get()), apiUrl });
-
-      window.location.assign(`${window.location.origin}${window.location.pathname}?rpc=${encodeURIComponent(apiUrl)}${window.location.hash}`);
+      window.location.assign(`${window.location.origin}${window.location.pathname}${getApiType(apiUrl).type === 'substrate-connect' as EndpointType ? '?sc=' : '?rpc='}${encodeURIComponent(apiUrl)}${window.location.hash}`);
```
jacogr, line 228: One more `as EndpointType`, something is weird with the typing somewhere...

packages/apps/src/Endpoints/index.tsx
```diff
@@ -216,21 +225,29 @@ function Endpoints ({ className = '', offset, onClose }: Props): React.ReactElem
...
 
+  const isSwitchDisabled = (): boolean => {
+    if (!hasUrlChanged) { return true; }
+
+    if (getApiType(apiUrl).type === 'substrate-connect' as EndpointType) { return false; }
+
+    if (isUrlValid) { return false; }
+
+    return true;
+  };
```
jacogr, line 243:
```suggestion
  function isSwitchDisabled (): boolean => {
    if (!hasUrlChanged) { 
      return true;
    } else if (getApiType(apiUrl).type === 'substrate-connect') { 
      return false
    } else if (isUrlValid) { 
      return false;
    }

    return true;
  };
```

Always use the `function` form and try to not inline returns. However, this is slightly problematic since this function is re-created each render. So really would like to avoid that at all costs.

packages/apps/src/initSettings.ts
```diff
@@ -7,9 +7,18 @@ import store from 'store';
 import { createWsEndpoints } from '@polkadot/apps-config';
 import { extractIpfsDetails } from '@polkadot/react-hooks/useIpfs';
 import { settings } from '@polkadot/ui-settings';
+import { Endpoint, EndpointType } from '@polkadot/ui-settings/types';
 import { assert } from '@polkadot/util';
 
-function getApiUrl (): string {
+function networkOrUrl (apiType: Endpoint): void {
+  if (apiType.type === 'json-rpc' as EndpointType) {
```
jacogr, line 14: Some more `as EndpointType` - really need to see why we need to do this. A single place to coerce when we already know would be fine, but we have it everywhere when working with this type.

packages/apps/src/initSettings.ts
```diff
@@ -24,7 +33,15 @@ function getApiUrl (): string {
 
     assert(url.startsWith('ws://') || url.startsWith('wss://'), 'Non-prefixed ws/wss url');
 
-    return url;
+    return { param: url, type: 'json-rpc' as EndpointType };
```
jacogr, line 36: As above, casting should not be needed at all. (Last comment on this, so we should do a search for all, remove it and fix why it breaks... if explicitly needed at a single point, all ok.)

jacogr (commented):
Some initial comments on the first part.

packages/apps/src/Endpoints/index.tsx
wirednkod: Will remove from everywhere

packages/apps-config/src/endpoints/productionRelayKusama.ts
jacogr: Indeed, this will be problematic - https://github.com/polkadot-js/apps/blob/master/packages/apps/src/Endpoints/Network.tsx#L28

For the time being I am guessing we need to filter. Or, if they have light-client capabilities, select from the light-client list only. And it not (or no light clients available), select from the other "normal" list.

packages/apps/src/Endpoints/index.tsx
```diff
@@ -118,6 +127,16 @@ function loadAffinities (groups: Group[]): Record<string, string> {
...
 
+function isSwitchDisabled (hasUrlChanged: boolean, apiType: EndpointType, isUrlValid: boolean): boolean {
+  if (!hasUrlChanged) { return true; }
+
+  if (apiType === 'substrate-connect') { return false; }
+
+  if (isUrlValid) { return false; }
+
+  return true;
+}
```
jacogr, line 138:
```suggestion
function isSwitchDisabled (hasUrlChanged: boolean, apiType: EndpointType, isUrlValid: boolean): boolean {
  if (!hasUrlChanged) { 
    return true;
  } else if (apiType === 'substrate-connect') { 
    return false;
  } else if (isUrlValid) { 
    return false;
  }

  return true;
}
```

packages/apps/src/Endpoints/index.tsx
```diff
@@ -230,7 +247,7 @@ function Endpoints ({ className = '', offset, onClose }: Props): React.ReactElem
       button={
         <Button
           icon='sync'
-          isDisabled={!(hasUrlChanged && isUrlValid)}
+          isDisabled={isSwitchDisabled(hasUrlChanged, getApiType(apiUrl).type, isUrlValid)}
```
jacogr, line 250: Something like -

```js
const canSwitch = useMemo(
  () => isSwitchDisabled(hasUrlChanged, getApiType(apiUrl).type, isUrlValid),
  [hasUrlChanged, apiUrl, isUrlValid]
);
```

i.e. not sure we really need to call continuously

packages/apps/src/Endpoints/index.tsx
```diff
@@ -255,7 +272,7 @@ function Endpoints ({ className = '', offset, onClose }: Props): React.ReactElem
             <div className='endpointCustomWrapper'>
               <Input
                 className='endpointCustom'
-                isError={!isUrlValid}
+                isError={!isUrlValid || getApiType(apiUrl).type === 'json-rpc'}
```
jacogr, line 275: This should actually not show when light is selected at all or it should fallback to the previous value that was/is there -

```js
{getApiType(apiUrl).type === 'json-rpc' && (
  <div className='endpointCustomWrapper'>...
)}
```

The reasoning is that we now create error inputs instead of just checking what the user has.

packages/apps/src/Endpoints/index.tsx
```diff
@@ -270,7 +287,7 @@ function Endpoints ({ className = '', offset, onClose }: Props): React.ReactElem
                 : <Button
                   className='customButton'
                   icon='save'
-                  isDisabled={!isUrlValid || isKnownUrl}
+                  isDisabled={!isUrlValid || isKnownUrl || getApiType(apiUrl).type === 'json-rpc'}
```
jacogr, line 290: This goes with the above - so either hide completely (don't fallback to previous) or show the previous valid.

packages/apps/src/Root.tsx
```diff
@@ -44,8 +44,8 @@ function Root ({ store }: Props): React.ReactElement<Props> {
       <ThemeProvider theme={theme}>
         <Queue>
           <Api
+            apiType={settings.apiType }
```
jacogr, line 47:
```suggestion
            apiType={settings.apiType}
```

packages/apps/src/overlays/Connecting.tsx
```diff
@@ -5,12 +5,25 @@ import React from 'react';
...
+function networkOrUrl (apiType: Endpoint): string {
+  if (apiType.type === 'json-rpc') {
+    return apiType.param;
+  }
+
+  if (apiType.type === 'substrate-connect') {
+    return apiType.param;
+  }
+
+  return '';
```
jacogr, line 22: Not sure why is should/would get to this point? There are only 2 options?

packages/react-api/src/Api.tsx
```diff
@@ -186,22 +188,13 @@ function Api ({ children, store, url }: Props): React.ReactElement<Props> | null
   const [extensions, setExtensions] = useState<InjectedExtension[] | undefined>();
 
   const value = useMemo<ApiProps>(
-    () => ({ ...state, api, apiError, apiUrl: url, extensions, isApiConnected, isApiInitialized, isWaitingInjected: !extensions }),
-    [apiError, extensions, isApiConnected, isApiInitialized, state, url]
+    () => ({ ...state, api, apiError, apiUrl: apiType?.param, extensions, isApiConnected, isApiInitialized, isWaitingInjected: !extensions }),
+    [apiError, extensions, isApiConnected, isApiInitialized, state, apiType?.param]
```
jacogr, line 192:
```suggestion
    [apiError, extensions, isApiConnected, isApiInitialized, state, apiType]
```

The Babel compilation always makes these a bit more hairy than it should be, so keeping to the base actually yields less suprises... :)

packages/react-api/src/Api.tsx
```diff
@@ -212,14 +205,49 @@ function Api ({ children, store, url }: Props): React.ReactElement<Props> | null
...
+
+    const types = getDevTypes();
 
-    setIsApiInitialized(true);
-  // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, []);
+    if (apiType?.type === 'substrate-connect') {
+      // eslint-disable-next-line no-void
+      substrateConnectApi(apiType.param).catch(console.error);
+    } else if (apiType?.type === 'json-rpc') {
```
jacogr, line 235: This is a bit of an issue - since `apiType` is optional, we actually could end up where we do nothing now. contrast this to the original where it always constructed an API instance.

So only check for light, if not, always follow the URL path.

jacogr (commented):
Hopefully the last batch of comments...

packages/apps-config/src/endpoints/productionRelayKusama.ts
wirednkod: > For the time being I am guessing we need to filter. Or, if they have light-client capabilities, select from the light-client list only. And it not (or no light clients available), select from the other "normal" list.

I would remove the light-clients for now (as it is experimental and user should know that chooses it).
It should always be randomly selected from the "normal" list only - and user can manually switch to light client if he/she wishes to do so

packages/apps/src/overlays/Connecting.tsx
wirednkod: This is just to satisfy TS. If removed then error appears:
```
Function lacks ending return statement and return type does not include 'undefined'.ts(2366)
```
Even adding undefined on the return types (and removing that) then different TS warning appears mentioning:
```
Not all code paths return a value.ts(7030)
```

packages/react-api/src/Api.tsx
wirednkod: I would actually make it mandatory, as few lines below the WSProvider cannot be created without it (and type is not optional in apiType):
```
const provider = new WsProvider(apiType.param);
```
Afair apiType should always be provided from now on in the Endpoint type format

jacogr: Happy to have it mandator and remove all the `apiType?.` checks everywhere.

packages/apps/src/overlays/Connecting.tsx
jacogr: So the issue is that to satisfy the linter, we shouldn't put in code that is "weird". So either -

```js
if (type === 'substrate-connect') {
  return 'something';
}

// this is the only other case, so return
return 'rpc-here-we-go';
```

jacogr: ... there is something where we can assert never in the default path and TS will check if all are followed, just cannot get an example off-hand now.

wirednkod: Well I rechecked the code and (after we changed to param) I dont think there is any need of the `networkOrUrl` function at all.
We could just:
```
const wsUrl = settings.apiType.param;
```

packages/apps/src/Endpoints/Network.tsx
```diff
@@ -25,7 +25,11 @@ function NetworkDisplay ({ apiUrl, className = '', setApiUrl, value: { icon, isC
   );
 
   const _selectUrl = useCallback(
-    () => setApiUrl(name, providers[Math.floor(Math.random() * providers.length)].url),
+    () => {
+      const filteredProviders = providers.filter((p) => !p.name.startsWith('light client'));
```
jacogr, line 29: Not complaining, but we should pass through the `isLightClient` thing here. Can be done later... but probably jut need a comment to make sure we pass that through the chain somehow.

wirednkod: > Not complaining, but we should pass through the `isLightClient` thing here. Can be done later... but probably jut need a comment to make sure we pass that through the chain somehow.

You want to add it in this PR or open an issue for later on?

jacogr: We can just open an issue.

package.json
```diff
@@ -74,6 +74,7 @@
     "@pinata/sdk": "^1.1.23",
     "@polkadot/dev": "^0.62.43",
     "@polkadot/ts": "^0.4.1",
+    "@substrate/connect": "0.3.10",
```
jacogr, line 77:
```suggestion
```

Added it in https://github.com/polkadot-js/apps/pull/5644 (have too many breakages each time there is a merge...)

packages/react-api/package.json
```diff
@@ -9,6 +9,7 @@
     "@polkadot/api": "^4.16.2",
     "@polkadot/extension-dapp": "^0.38.7",
     "@polkadot/x-rxjs": "^6.10.1",
+    "@substrate/connect": "0.3.10",
```
jacogr, line 12:
```suggestion
```

Added it in https://github.com/polkadot-js/apps/pull/5644 (have too many conflicts each time there is a merge...)

jacogr (commented):
Just merge in master after https://github.com/polkadot-js/apps/pull/5644 goes in 

Have too many breakages each time there is a merge... since we have a conflict each time there is a merge and it just wastes your time. (Should have been done days ago, my humble apologies for only thinking of it now)

packages/apps/src/Endpoints/Network.tsx
wirednkod: Opened [this issue](https://github.com/polkadot-js/apps/issues/5734).

Is there anything else pending for merging this?

wirednkod:
> Just merge in master after #5733 goes in
> 
> We have a conflict each time there is a merge and it just wastes your time. (Should have been done days ago, my humble apologies for only thinking of it now)

@jacogr  This is not correct. You should not use the curet (^) with substrate connect for now.
Instead of `^0.3.10` it should be `0.3.10`. 
This is due to the whole `ts-jest` that we discussed before

jacogr:
Ummmm... that is going to be very problematic. I don't have specific rules as to "what not to bump" and try to bump all 3rd party blockchain deps at least weekly. So this is bound to break horribly as soon as updates are made.

wirednkod:
> Ummmm... that is going to be very problematic. I don't have specific rules as to "what not to bump" and try to bump all 3rd party blockchain deps at least weekly. So this is bound to break horribly as soon as updates are made.

We are trying to figure out some fixes - will not stay "not-bumped" for long for sure. Just for now (a solution is for example - when build auto-add extensions on files (or something like that)

jacogr (commented):
Generally looks good. 2 small comments remaining -

- will just play through now to ensure the normal operation is still all-ok (I know it has been tested, but also don't want a resulting fallout)
- jus remove the yarn.lock changes, not sure why there are still changes (the required packages are in master)

wirednkod:
I am not sure why the CIs started to break (I don't understand the error but something seems to be wrong with .yarn).
@jacogr any insights on this?

jacogr:
It is a yarn.lock merge issue. Just copy the yarn.lock from master and replace it here. There should be no updates needed to it at all.

jacogr:
Played through. It seems to do what it says on the tin, which is great.

There is a small issue - when the user swaps to a light client (and he has nothing), the screen just goes white and stays that way. We need some of fallback for that.

Apart from that, seems great.

wirednkod:
> There is a small issue - when the user swaps to a light client (and he has nothing)

What do you mean by "He has nothing" ?

> EDIT: We also need somewhere to prompt people how and what to add. (But can do that as a follow-up, not even sure how/where and what something like that would look like)

There are for now the substrate-connect [landing-page[(https://paritytech.github.io/substrate-connect/) and [api docs](https://paritytech.github.io/substrate-connect/api/) - Im not sure if that is what we could use though

jacogr:
So this is what happens for me...

image

Just dead as soon as I sawp.

wirednkod:
@jacogr This is a little tricky.
When you are testing in localhost - then the SmoldotProvider (the one that runs the light client in the opened tab and not the extension) will never run correctly. You need to make an actual build of the apps and then navigate into the build dir and run  an http-server in order to see it work without the extension (simulates the actual static files on server in a way)

Steps:
a) install globally an http-server (`npm install http-server -g`)
b) Make a build of the apps (`cd ./apps && yarn build`)
c) Navigate in the build dir (`$ apps > cd packages/apps/build`)
d) Run the server (`http-server`)

Please try to test with [Westend](http://192.168.1.14:8080/?sc=westend-substrate-connect#/explorer) as I am not sure why Polkadot/Kusama not always respond (I'll cc @tomaka here)

You should see something like the lines below running in your console:
```
Chain of light client is = westend
content_script.js:159 script added
main.98207691.js:2 [smoldot_js] Loaded chain spec #0: Westend
main.98207691.js:2 [smoldot_js] Genesis hash of Westend is: 0xe143…423e
main.98207691.js:2 [smoldot_js] Using light checkpoint starting at #5509229
main.98207691.js:2 [smoldot_js] Initialization complete
main.98207691.js:2 [network] Connected to 12D3KooWR3UGwwSP5wdBMk2JXXuzXoscPSudv8hmQkzfZTBzSbeE
main.98207691.js:2 [network] Connected to 12D3KooWQeuYtdKUgC9A2wNchxQDXWHbotZWUHxnraBwiV15GH4K
main.98207691.js:2 [sync-verify] GrandPa warp sync finished to #6350101
polkadot.02.8b5b3183.js:1 web3Enable: Enabled 0 extensions: 
polkadot.02.8b5b3183.js:1 web3Accounts: Found 0 addresses: 
main.98207691.js:19 chain: Westend (Live), {"ss58Format":42,"tokenDecimals":[12],"tokenSymbol":["WND"]}
```

tomaka:
> Polkadot/Kusama not always respond (I'll cc @tomaka here)

To give a little overview:

Since smoldot is a light client, it doesn't store the state of the chain. Instead, it queries it from other nodes.
PolkadotJS doesn't really behave properly when it asks for the storage value through a JSON-RPC request, and an error is returned. The UI more or less stops working.
Unfortunately, while smoldot is far away from the head of the chain, storage requests over the network typically fail because nodes prune old blocks.

In order to circumvent that, we only report substrate-connect as "connected" when it has synced to the head of the chain. Once at the head of the chain, all storage queries normally succeed.

This works well for Westend. For Kusama and Polkadot, unfortunately, there is only one "WebSocket secure" node to connect to at the moment for each chain (cc https://github.com/paritytech/devops/issues/948).
This node is unfortunately often full.

When the node is full, smoldot can't sync to the head of the chain at all, meaning that no "connected" event is reported to PolkadotJS.

It should work when using the substrate-connect browser extension, as this unlocks more nodes to connect to.

jacogr:
The point is that I'm still in white-screen mode on a local server -

image

So, the issue is that people are going to click this, get exactly the same and have no way of recovering.

image

The only reason I could recover here to swap is because there is a bug where a "remove params" and refreshing doesn't actually load the light client again.

wirednkod:
The white screen has to do with the `await detect.connect` that [exists in packages/react-api/src/Api.tsx](https://github.com/polkadot-js/apps/pull/5644/files#diff-6aaa8c30f4959eafef04b2c2e7150afe93717aeb195b604a35cc76737f63cfe3R218).

Due to the fact that this exists in a Component (`<Api />`) prior to `<Apps />`, the await (expecting Promise<ApiPromise>) is blocking the React flow to "draw" the components and expect Api to become ready ( the code freezes until Promise of detector is resolved - so flow never reaches <Apps> thus never draws the "waiting" state - instead we see a white page)

I will be working on this to resolve it asap - if there is something that can be used (e.g. useApi) and I am not aware of please for your feedback @jacogr 

cc @raoulmillais @tomaka

jacogr:
catch-22 - the `useApi` uses the context, so actually uses the `<Api />` wrapper. There should be a way with minimal changes to get it to behave inside the provider.

wirednkod:
@jacogr After further investigation and re-tries it was fairly easy to fix the "await" part but due to the fact that api was undefined a huge amount of fixes should take place across all Apps codebase in order to make sure that there will be no breaking.

Instead we proceed with a workaround on the substrate-connect side that would resolve the issue and make the Apps code clearer. `Substrate-connect` now, through the `Detector` besides the `connect()` it now provides one, called `provider()` that, depending on whether the user has the substrate connect extension installed or not, returns a provider of ProviderInterface type, that is used by passed into the ApiPromise.

packages/page-accounts/src/CreateAccount.slow.spec.tsx
```diff
@@ -14,15 +14,18 @@ import { Api } from '@polkadot/react-api';
...
+  const apiType: Endpoint = { param: `ws://127.0.0.1:${SUBSTRATE_PORT}`, type: 'json-rpc' };
 
   return render(
     <MemoryRouter>
       <ThemeProvider theme={lightTheme}>
-        <Api store={memoryStore}
-          url={`ws://127.0.0.1:${SUBSTRATE_PORT}`}>
+        <Api
+          apiType={apiType}
+          store={memoryStore}>
```
jacogr, line 28:
```suggestion
          store={memoryStore}
        >
```

packages/page-bounties/src/Bounties.slow.spec.tsx
```diff
@@ -17,18 +17,22 @@ import { MemoryStore } from '@polkadot/test-support/keyring';
...
+
   return render(
     <Suspense fallback='...'>
       <MemoryRouter>
         <ThemeProvider theme={lightTheme}>
-          <Api store={memoryStore}
-            url={`ws://127.0.0.1:${SUBSTRATE_PORT}`}>
+          <Api
+            apiType={apiType}
+            store={memoryStore}>
```
jacogr, line 35:
```suggestion
            store={memoryStore}
          >
```

jacogr (commented):
Seems sane after resolving the conflicts. Will just do a final play-through.

jacogr (approved):
Nice! Seems to do exactly what it says on the tin, this is an excellent addition. 

Thank you.

PR #5645, opened 2021-06-21
fix Bifrost connect bug

awesomepan:


packages/apps-config/src/endpoints/productionRelayKusama.ts
```diff
@@ -55,7 +55,6 @@ export function createKusama (t: TFunction): EndpointOption {
       },
       {
         info: 'bifrost',
-        isUnreachable: true, // https://github.com/polkadot-js/apps/issues/5619 (duplicated in testing networks)
```
jacogr, line 58: The issue I see with this one and the one in `testing` is that one is a test network and one is a production parachain. Both RPCs are exactly the same, which cannot be... So the incorrect one (where there is no available RPC as of yet), needs to stay "unreachable" until there is an RPC deployed and the endpoint updated.

I'm assuming this is the parachain that stays `isUnreachable` for the time being. it doesn't make sense to connect to what you believe is the parachain and then getting to a different network.

awesomepan: updated, pls review, thx

jacogr (approved):
Thank you.

PR #5648, opened 2021-06-21
Update Zeitgeist values

c410-f3r:


lsaether (approved):

jacogr (approved):

c410-f3r:
`
    Error: ENOENT: no such file or directory, open '/tmp/tmp-2111-kllNVuoe76E7/1'
`

Error doesn't seem related

jacogr:
The above is not the actually failure, it fails here - https://github.com/polkadot-js/apps/pull/5648/checks?check_run_id=2876761488#step:3:946

c410-f3r:
Not complete sure but it seems that all Rococo and Westend parachains are being grouped together so Zeitgeist will always be placed after Westend.

c410-f3r:
> The above is not the actually failure, it fails here - https://github.com/polkadot-js/apps/pull/5648/checks?check_run_id=2876761488#step:3:946 (the above is ignored, i.e. doesn't lead to a failure, as horrible as it looks)

Thanks for pointing out the issue

jacogr:
No issues, the above error tripped me up a couple of times. (Really need to find a way to silence/ignore that one)

packages/apps-config/src/endpoints/index.ts
```diff
@@ -31,7 +31,16 @@ export function createWsEndpoints (t: TFunction, firstOnly?: boolean): LinkOptio
       textBy: '',
       value: ''
     },
-    ...createTestingRelays(t, firstOnly),
+    ...createRococoRelays(t, firstOnly),
+    {
+      isDisabled: false,
+      isHeader: true,
+      isSpaced: true,
+      text: t('rpc.header.live', 'Live networks', { ns: 'apps-config' }),
```
jacogr, line 39:
```suggestion
      text: t('rpc.header.westend.relay', 'Westend test & parachains', { ns: 'apps-config' }),
```

Would just appreciate an update on line 30 as well to `Rococo test & parachains` (Sadly cannot get to it via the GH editor)

Thanks fo r this split, it was meant to be done when Polkadot has at least parathreads.

c410-f3r: Done

PR #5670, opened 2021-06-23
Fix docker build script

sorpaas:
The current Dockerfile script doesn't seem to be able to build successfully from the ground up. This PR fixes it:

* Install `ca-certificates`. Without it, it seems that every SSL requests would fail -- nodesource can't be fetched, npm can't install.
* Remove duplicate `nodejs` and `npm` package installation. `nodejs` comes from `nodesource` which includes the `npm` executable, and `npm` comes from Ubuntu. They conflict.


jacogr (approved):
Thank you very much.

PR #5675, opened 2021-06-24
Support for new features in equilibrium testnet

alex-nax:


yarn.lock
```diff
@@ -140,9 +140,9 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@babel/helper-create-class-features-plugin@npm:^7.14.5":
-  version: 7.14.5
-  resolution: "@babel/helper-create-class-features-plugin@npm:7.14.5"
+"@babel/helper-create-class-features-plugin@npm:^7.14.5, @babel/helper-create-class-features-plugin@npm:^7.14.6":
```
jacogr, line 143: Rather only bump your package instead of all. Specifically stuff like babel really needs a local test first before bumping - despite patches, have had issues in the past.

So simplest is just -

- bump version in package.json
- yarn

PR #5760, opened 2021-07-04
bump acala

xlc:
pull in https://github.com/AcalaNetwork/acala.js/pull/171


jacogr:
You can add all the instances of the council into the array. As rightly said, atm only the first will be used, atm it will just always grab the first. (However also having all makes it easier to get and test for multiples)

jacogr (approved):
Thank you.

PR #5766, opened 2021-07-05
Add chainType to chainInfo

joelamouche:
In order to be recognized by the extension as an ethereum chain

### DIFF
--- a/packages/page-settings/src/types.ts
+++ b/packages/page-settings/src/types.ts
@@ -3,6 +3,9 @@
 
 import type { MetadataDef } from '@polkadot/extension-inject/types';
 
+type ChainType ='substrate' | 'ethereum';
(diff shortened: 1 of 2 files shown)


packages/page-settings/src/types.ts
```diff
@@ -3,6 +3,9 @@
 
 import type { MetadataDef } from '@polkadot/extension-inject/types';
 
+type ChainType ='substrate' | 'ethereum';
```
jacogr, line 6:
```suggestion
type ChainType = 'substrate' | 'ethereum';
```

jacogr (commented):
One small comment, seems 100% apart from that.

jacogr (approved):
Thank you.

PR #5774, opened 2021-07-06
Fix crab, pangolin types

WoeOm:
1. https://github.com/polkadot-js/apps/issues/5726
2. Merged the Darwinia, Crab, Pangolin network types.

### DIFF
--- a/packages/apps-config/src/api/spec/index.ts
+++ b/packages/apps-config/src/api/spec/index.ts
@@ -17,6 +17,7 @@ import centrifugeChain from './centrifuge-chain';
 import chainx from './chainx';
 import clover from './clover';
 import cloverRococo from './clover-rococo';
+import crab from './crab';
 import crust from './crust';
 import testPara from './cumulus-test-parachain';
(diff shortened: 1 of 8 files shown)


packages/apps-config/src/api/spec/index.ts
```diff
@@ -99,13 +97,11 @@ import zero from './zero';
 
 // NOTE: The mapping is done from specName in state.getRuntimeVersion
 const spec: Record<string, OverrideBundleDefinition> = {
-  Crab: crab,
-  Darwinia: darwinia,
+  ...darwinia,
```
jacogr, line 100: I really would prefer an explicit mapping here - it is indeed more of a pain, however it also ensures that no other keys are overridden unintentionally. You can still import the bundle from the single file, i.e.

```js
import { Crab, Darwinia } from './darwinia';

...
{
...
  Crab,
  Darwinia,
...
}
```

WoeOm: Has changed explicit mapping.

PR #5787, opened 2021-07-07
integritee parathread launch branding update

brenzi:
we'll be waiting for this to be merged to register our parathread on kusama


jacogr (approved):
Thank you.

PR #5795, opened 2021-07-08
chore: Update semgrep.yml with deployment id

ltfschoen:
Some projects may use a fork of polkadot-js/apps. They cannot use deployment id 1395. They need to setup their own semgrep configuration that has a unique deployment id and token to pass CI checks. Suggest having the associated deployment id also stored in secrets of the Github repo. So for polkadot-js/apps an additional secret name of `SEMGREP_APP_DEPLOYMENT_ID ` with value of `1395` would be added to https://github.com/polkadot-js/apps/settings/secrets/actions, whilst those using a fork would create their own Semgrep name and api token here https://semgrep.dev/manage/settings and add it to their Github settings at https://github.com/<FORK_USER>/apps/settings/secrets/actions

### DIFF
--- a/.github/workflows/semgrep.yml
+++ b/.github/workflows/semgrep.yml
@@ -15,4 +15,4 @@ jobs:
         with:
           auditOn: push
           publishToken: ${{ secrets.SEMGREP_APP_TOKEN }}
-          publishDeployment: 1395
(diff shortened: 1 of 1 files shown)


cgroeschel (approved):

jacogr:
This should be handled as a change in the forks themselves. Have no issues with teams maintaining forks (and actually encourage it for their specific uses), as long as it doesn't interfere with the setup here, which this does.

PR #5798, opened 2021-07-09
Update Parachain ID of Zeitgeist

c410-f3r:
My apologies for the confusion. The last deployed WASM blob was non-deterministic due to the fact that the chain specification wasn't using a generated file and this situation could lead to bad future network updates.


jacogr (approved):
All fine. That is why Rococo is there so the same mistakes are not made elsewhere.

c410-f3r:
Thank you!

PR #5855, opened 2021-07-20
Upgrade @substrate/connect@0.3.16

raoulmillais:
We deployed 7 more bootnodes which listen on TLS over the weekend for both kusama and polkadot. This should make connecting to polkadot and kusama much more reliable with substrate connect.  This PR bumps the version of @susbtrate/connect to pick up the new chainspecs that include those new bootnodes.


jacogr (approved):
Thank you.

PR #5858, opened 2021-07-20
Add Aleph Zero testnet

timorl:
We  have been running the testnet for a while already, hosting our own version of polkadot.js, and it would be nice to be included in the official version too.


jacogr (approved):
Thank you.

PR #5866, opened 2021-07-22
Add summary bar for accounts view

vitaliyworks:


vitaliyworks:
Hi @jacogr !

I wanted to get your opinion on the current state of my PR.
Specifically regarding these topics:

* CardSummary styling

  I used CardSummary component for the summary so that it is consistent with the other views.
  Here is the result:
Screenshot 2021-07-22 at 12 38 20

  But the target design has the summary styled differently:  
Screenshot 2021-07-22 at 12 38 31

  I am not sure if this PR should style the CardSummary, or a separate one.
  What do you think?

* "0.0000" vs "--" vs `display: none`

  Current implementation shows summary for all balances, even if most of them are zero.
  But what is best way to handle this?

* Code duplication

  Most of the new calculations are small, but still, they are copy-pasted from other places.
  Deduplication is left out in this PR with the intention of making a following PR that handles it,
  so that the diff is easier on the eyes.

  Is this OK?

jacogr:
UI consistency is important. So the current use is fine. 

We can remove zero values, I would not go to a “—“ display since that may indicate “loading”. 

Small steps are 100% fine, don’t need to get all at once.

krzysztof-jelski:
@jacogr do we have a component that could use to show the "summary-breakdown" relationship? In the designs we have this box with "Total locked", that is further broken down into "Redeemable", "Bonded" and "Unbonding". Though there might be a component that would serve a similar purpose. Hopefully not the `Expander`, though 😄 .

jacogr:
No. However the summary components can render and content you pass.

vitaliyworks:
Thanks for the feedback @jacogr !
In response, I made cards with empty balance to not show up and removed the WIP status.
We plan to make follow up PRs that will style the summary bar further.

PS. sorry for the force push, I just now learned it is not appreciated.

vitaliyworks:
Moved this PR so that it is based on local branch.
New PR: #5876

PR #5887, opened 2021-07-28
Add chain type to metadata

joelamouche:
Both in the QR content and in the metadata UI.
Chaintype is either ethereum or substrate.
Necessary for the mobile app to create correct address type (as per paritytech/parity-signer#788)


jacogr (approved):
Thank you.

PR #5890, opened 2021-07-29
Add summary bar tests

vitaliyworks:


packages/page-accounts/src/Accounts/index.spec.ts
```diff
@@ -50,6 +51,13 @@ describe('Accounts page', () => {
 
       expect(noAccountsMessage).not.toBeNull();
     });
+
+    it('no summary displays', () => {
```
krzysztof-jelski, line 55:
```suggestion
    it('no summary is displayed', () => {
```

packages/page-accounts/src/Accounts/index.spec.ts
```diff
@@ -50,6 +51,13 @@ describe('Accounts page', () => {
...
+      accountsPage.renderPage([]);
```
krzysztof-jelski, line 56: This line can be moved do a `beforeEach` in `when no accounts` describe/context block.

packages/react-hooks/src/useStakingInfo.ts
```diff
@@ -3,7 +3,8 @@
 
 import type { DeriveStakingAccount } from '@polkadot/api-derive/types';
 
-import { useApi, useCall } from '@polkadot/react-hooks';
+import { useApi } from './useApi';
```
krzysztof-jelski, line 6: The convention is to keep local imports when importing from current package.

packages/page-accounts/src/Accounts/index.spec.ts
vitaliyworks: Indeed

vitaliyworks: Done, thanks 👍

packages/page-accounts/src/Accounts/index.spec.ts
```diff
@@ -115,4 +212,70 @@ describe('Accounts page', () => {
 
     return balanceOf(amountInt.toString() + decimalsPadded);
   };
+
+  const defaultStakingAccounts = [
```
krzysztof-jelski, line 216:
```suggestion
  const accountsWithStaking = [
```

packages/page-accounts/src/Accounts/index.spec.ts
```diff
@@ -115,4 +212,70 @@ describe('Accounts page', () => {
...
+        freeBalance: balance(200),
+        lockedBalance: balance(750),
+        reservedBalance: balance(150)
+      },
+      staking: {
+        redeemable: balance(3000),
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+        stakingLedger: {
+          active: {
+            unwrap: () => balance(60)
```
krzysztof-jelski, line 261: Should be fairly easy to create an `Option` instance. Sth along the lines of:
`new TypeRegistry().createType('Option<Balance>', new TypeRegistry().createType('Balance', new BN(number)));`
or https://github.com/polkadot-js/api/blob/master/packages/api-derive/test/bountyFactory.ts#L30.

Basically the big question is whether we keep creating the `TypeRegistry` for each creation method or reuse a single instance across all tests.

packages/page-accounts/src/Accounts/index.spec.ts
```diff
@@ -102,6 +107,98 @@ describe('Accounts page', () => {
...
+
+      const summary = await screen.findByTestId(/card-summary:(total )?bonded/i);
+
+      expect(summary).toHaveTextContent(expectedText);
+    });
+
+    it('displays unbonding summary', async () => {
+      accountsPage.renderPage(defaultStakingAccounts);
+
+      // Sum of "unlocking" for each account
```
krzysztof-jelski, line 183: Most of the tests have explaining comments, but not all. I guess we can remove all comments, as the logic is pretty simple. No strong opinion, though.

My preferred way would be to see all relevant data (and only relevant data) in each test case. Eg. if we sum only the "unlocking" part, we render with `merge(accountsWithStaking, [{unlocking: 500}, {unlocking:1000}])`. But I understand that could be too bit too much for some of the cases we have here.

krzysztof-jelski (approved):
Looks good overall. The arrange part in summary test cases can use a little refactoring.

packages/page-accounts/src/Accounts/index.spec.ts
vitaliyworks: Oh yeah, this will look great

vitaliyworks: Updated as per comment. I think it is an improvement overall

packages/page-accounts/src/Accounts/index.spec.ts
vitaliyworks: Done. I did not create a `TypeRegistry` just yet because we create new ones in other places too

packages/page-accounts/src/Accounts/index.spec.ts
wachulski: Yup, I agree though we need to be careful with those opinionated default shortcuts, as if we put to much into them in a rigid manner, then it may be hard later to unwind defaults to have e.g. account addresses + a portion of default facilities in order to easily code unit tests for sorting features. Just FYI as we cannot know it upfront and be prepared. Just a word of caution.

Great suggestion + realization :) Thanks!

wachulski:
I hereby approve :)

PR #5893, opened 2021-07-29
Style tags

MiZiet:
hi, this will change look of tags in app according to the ui grant design.

image
image
image
image

PR #5894, opened 2021-07-29
Account row collapsible row details

wachulski:
Collapsible account details info containing:
- tags (removed Tags column)
- balances details

After:
image

Before:
image

In upcoming PRs:
- [x] improve Tags look: https://github.com/polkadot-js/apps/pull/5893
- [x] add Tags headers: after https://github.com/polkadot-js/apps/pull/5893 gets merged


packages/page-accounts/src/Accounts/Account.tsx
```diff
@@ -355,283 +357,340 @@ function Account ({ account: { address, meta }, className = '', delegation, filt
...
-                color='orange'
-                hover={t<string>('This is a development account derived from the known development seed. Do not use for any funds on a non-development network.')}
-                icon='wrench'
-              />
-            )
-            : (
-              <Badge
-                color='orange'
-                hover={
+    <>
```
wachulski, line 370: Dear reviewer - because of this change, all the code of a TR got indented, so it is shows as changed. Be notified as you may want to optimize diff showing for this PR in github (Split, no whitespaces)

packages/page-accounts/test/pages/accountsPage.tsx
```diff
@@ -25,12 +25,48 @@ let queueExtrinsic: (value: PartialQueueTxExtrinsic) => void;
...
+    const balanceExpected = formatBalance(expected, { decimals: 12, withUnit: true });
+
+    expect(balanceActual).toHaveTextContent(balanceExpected);
+  }
+
+  async assertBalancesDetails (expected: {name: string, amount: Balance}[]): Promise<void> {
+    for (const expectedBalanceDetailsItem of expected) {
+      const labelElement = await within(this.detailsRow).findByText(expectedBalanceDetailsItem.name);
+      const balanceElement = labelElement.nextSibling;
+      const amount = formatBalance(expectedBalanceDetailsItem.amount, { decimals: 12, withUnit: true });
```
vitaliyworks, line 49: I think we need `forceUnit: '-'` here as well because we it do in `FormatBalance`[1], which is what displays in tests. If we skip it, then the text content will differ for some amounts; I encountered this in my own tests, but don't remember how to reproduce (which values give different results when this param is not specified).
Since I use this code as well, maybe we should also refactor it into a separate function, like `defaultFormatBalance`?

[1] https://github.com/polkadot-js/apps/blob/master/packages/react-query/src/FormatBalance.tsx#L60

packages/react-query/src/FormatBalance.tsx
```diff
@@ -84,17 +86,17 @@ function FormatBalance ({ children, className = '', format, formatIndex, isShort
...
+      {label ? <>{label}&nbsp;</> : ''}<span className='ui--FormatBalance-value'
+        data-testid='balance-summary' >{
           valueFormatted
             ? splitFormat(valueFormatted, labelPost, isShort)
             : value
               ? value === 'all'
                 ? t<string>('everything{{labelPost}}', { replace: { labelPost } })
                 : applyFormat(value, formatInfo, withCurrency, withSi, isShort, labelPost)
-              : `-${labelPost || ''}`
+              : labelPost instanceof String
```
vitaliyworks, line 97: Is capitalisation of String here intentional? I think it should never be the case that `labelPost` is instance of `String` because String is not primitive type. See this: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/instanceof#using_instanceof_with_string
This is why we had to cast `labelPost` to string in the line below, when normally the cast would not be required after the type check.

Could we do `typeof(labelPost) === string ? "-" + labelPost : labelPost` here instead?

vitaliyworks (commented):
I noted minor things, but overall, the patch looks good 👍

packages/react-query/src/FormatBalance.tsx
wachulski: interesting, let me check....

packages/page-accounts/test/pages/accountsPage.tsx
wachulski: That was great advice! Thx! For now I refactored outside the class, to a common fn in AccountPage and AccountRow.

packages/react-query/src/FormatBalance.tsx
wachulski: TSC is not happy with what you provided:
```
TS2693: 'string' only refers to a type, but is being used as a value here.
```
also linter: `ESLint: Operands of '+' operation must either be both strings or both numbers. Consider using a template literal. (@typescript-eslint/restrict-plus-operands)`

wachulski: But definitely sth is around here. It didn't look beauty after I introduced it and yes, I hoped someone would rant here :) Let's leave this discussion open.

packages/page-accounts/src/Accounts/Account.tsx
```diff
@@ -86,8 +87,9 @@ const transformRecovery = {
   transform: (opt: Option<RecoveryConfig>) => opt.unwrapOr(null)
 };
 
-function Account ({ account: { address, meta }, className = '', delegation, filter, isFavorite, proxy, setBalance, toggleFavorite }: Props): React.ReactElement<Props> | null {
+function Account ({ account: { address, meta }, className = '', delegation, filter, isEven, isFavorite, proxy, setBalance, toggleFavorite }: Props): React.ReactElement<Props> | null {
   const { t } = useTranslation();
+  const [isExpanded, setIsExpanded] = useState(false);
```
krzysztof-jelski, line 92:
```suggestion
  const [isExpanded, toggleIsExpanded] = useToggle();
```
Then you can get rid of `handleOnExpandIconClick` and call `toggleIsExpanded` directly, as it's memoized in the hook. I believe similar change can be made in `Bounty.tsx`.

packages/react-components/src/Table/Head.tsx
```diff
@@ -124,6 +124,11 @@ export default React.memo(styled(Head)`
     &.start {
       text-align: left;
     }
+
+    &.balances {
+      text-align: right;
+      padding-right: 218px;
```
krzysztof-jelski, line 130: Avoid using `px` in favour of `rem`. Also, I wonder if this styling could possibly go closer to Accounts.

packages/page-accounts/src/Sidebar/Balances.tsx
```diff
@@ -21,17 +21,12 @@ function Balances ({ address, className }: Props): React.ReactElement<Props> | n
...
-          &nbsp;
-          {t<string>('balance')}
-        </div>
+        {t<string>('balance')}
       </div>
       <AddressInfo
         address={address}
         className='balanceExpander'
         withBalance={WITH_BALANCE}
-        withBalanceToggle
```
krzysztof-jelski, line 34: Not sure those changes go with this PR. Seems unrelated to Account row changes.

packages/react-components/src/Table/Head.tsx
wachulski: @MiZiet

packages/page-accounts/src/Sidebar/Balances.tsx
wachulski: @MiZiet - does your styling change require resigning from toggle'able address info in other places?

MiZiet: I can revert changes in sidebar. I've styled balances in details row and in sidebar. I agree with @krzysztof-jelski that sidebar changes shouldn't be in this pr.

packages/react-query/src/FormatBalance.tsx
vitaliyworks: That warning is interesting. Have you quoted the word string?

packages/page-accounts/test/hooks/default.ts
```diff
@@ -5,11 +5,13 @@ import type { DeriveBalancesAll, DeriveStakingAccount } from '@polkadot/api-deri
 
 import BN from 'bn.js';
 
+import { UseAccountInfo } from '@polkadot/react-hooks/types';
```
jacogr, line 8:
```suggestion
import type { UseAccountInfo } from '@polkadot/react-hooks/types';
```

packages/react-components/src/AddressInfo.tsx
```diff
@@ -353,7 +363,8 @@ function createBalanceItems (formatIndex: number, lookup: Record<string, string>
       <React.Fragment key={7}>
         <Label label={t<string>('unbonding')} />
         <div className='result'>
-          <StakingUnbonding stakingInfo={stakingInfo} />
+          <StakingUnbonding iconPosition='right'
+            stakingInfo={stakingInfo} />
```
jacogr, line 367:
```suggestion
          <StakingUnbonding
            iconPosition='right'
            stakingInfo={stakingInfo}
          />
```

packages/react-components/src/StakingUnbonding.tsx
```diff
@@ -59,10 +60,13 @@ function StakingUnbonding ({ className = '', stakingInfo }: Props): React.ReactE
 
   return (
     <div className={className}>
-      <Icon
-        icon='clock'
-        tooltip={trigger}
-      />
+      { iconPosition === 'left' && (
```
jacogr, line 63:
```suggestion
      {iconPosition === 'left' && (
```

packages/react-components/src/StakingUnbonding.tsx
```diff
@@ -86,18 +90,30 @@ function StakingUnbonding ({ className = '', stakingInfo }: Props): React.ReactE
         ))}
         trigger={trigger}
       />
+      { iconPosition === 'right' && (
```
jacogr, line 93:
```suggestion
      {iconPosition === 'right' && (
```

packages/react-query/src/FormatBalance.tsx
jacogr, line 97:
```suggestion
              : isString(labelPost)
```

Which may pickup `String` as well (so `.toString()` could be required below)

packages/react-query/src/FormatBalance.tsx
```diff
@@ -84,17 +86,17 @@ function FormatBalance ({ children, className = '', format, formatIndex, isShort
...
+                ? `-${labelPost as string || ''}`
```
jacogr, line 98:
```suggestion
                ? `-${labelPost}`
```

packages/react-query/src/FormatBalance.tsx
wachulski: /cc @vitaliyworks :)

vitaliyworks: I still think that `typeof(labelPost) === 'string'` should be used because `labelPost` cannot be of type `String`

packages/page-accounts/test/hooks/default.ts
MiZiet: ✅

packages/react-components/src/StakingUnbonding.tsx
MiZiet: ✅

packages/react-components/src/StakingUnbonding.tsx
MiZiet: ✅

packages/react-components/src/AddressInfo.tsx
MiZiet: ✅

packages/page-accounts/src/Sidebar/Balances.tsx
MiZiet: ✅

packages/react-components/src/Table/Head.tsx
MiZiet: Changed px to rem.

packages/react-query/src/FormatBalance.tsx
jacogr: Since you pass in a string, the isString() should do the right thing. Plus, it aligns with the idioms used in polkadot-js.

packages/page-accounts/src/Accounts/Account.tsx
MiZiet: ✅

packages/react-query/src/FormatBalance.tsx
MiZiet: ✅

packages/react-query/src/FormatBalance.tsx
MiZiet: ✅

packages/page-accounts/src/Accounts/index.spec.ts
```diff
@@ -332,33 +313,65 @@ describe('Accounts page', () => {
...
+    return format(balance(amount));
+  }
+
+  const defaultAddresses = [
+    aliceAddress,
+    bobAddress,
+    charlieAddress
+  ];
+
+  function renderAccountsWithDefaultAddresses (...overrides: AccountOverrides[]): void {
```
krzysztof-jelski, line 348: The three `render...` methods here look pretty much alike. Maybe one could be "base" and other would call it with different params? Or perhaps the `for` loops could be replaced with `.map` on the arrays.

packages/page-accounts/src/Accounts/index.tsx
```diff
@@ -134,21 +134,6 @@ function Overview ({ className = '', onStatusChange }: Props): React.ReactElemen
     []
   );
 
-  const footer = useMemo(() => (
```
krzysztof-jelski: Yeah, totally! 👍

packages/page-accounts/src/Accounts/index.spec.ts
ekowalsk: ✅

krzysztof-jelski (approved):

krzysztof-jelski:
@jacogr when you get down to reviewing PRs, it would be perfect if you start with this one. A couple of other developments depend on it.

jacogr (approved):
Lgtm. Thank you.

PR #5895, opened 2021-07-29
Remove code duplication in Create and Derive modals

vitaliyworks:
This PR removes duplications between `modals/Create.tsx` and `/modals/Derive.tsx` files.
Functionality is not changed.
UI has one change - derive account window also got "hints" for passwords and name.


packages/page-accounts/src/modals/Create.tsx
```diff
@@ -178,9 +152,9 @@ function Create ({ className = '', onClose, onStatusChange, seed: propsSeed, typ
   const [step, nextStep, prevStep] = useStepper();
   const [isBusy, setIsBusy] = useState(false);
   const [{ isNameValid, name }, setName] = useState({ isNameValid: false, name: '' });
-  const [{ isPasswordValid, password }, setPassword] = useState({ isPasswordValid: false, password: '' });
+  const [{ isPassValid, password }, setPassword] = useState({ isPassValid: false, password: '' });
```
wachulski, line 155: why rename?

packages/page-accounts/src/modals/CreateAccountInputs.tsx
```diff
@@ -0,0 +1,61 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+import React, { useCallback } from 'react';
+
+import { Input, Modal } from '@polkadot/react-components';
+import { useTranslation } from '@polkadot/react-components/translate';
+
+import PasswordInput from './PasswordInput';
+
+interface AccountPass {
```
wachulski, line 11: AccountPassword?
AccountPass may indicate a successful operation state on Account instance

packages/page-accounts/src/modals/Create.tsx
```diff
@@ -134,35 +132,11 @@ function updateAddress (seed: string, derivePath: string, seedType: SeedType, pa
...
-    InputAddress.setLastValue('account', address);
-
-    if (getEnvironment() === 'web') {
-      downloadAccount(result);
-    }
-  } catch (error) {
-    status.status = 'error';
-    status.message = (error as Error).message;
-  }
+  const getResult = () =>
```
wachulski, line 136: this actually produces side effects which constitute the main operation of adding an account to the keyring. thus
I'd avoid `get....` and rather go for `commitAccount` or `pushAccountToKeyring` etc. any ideas?

packages/page-accounts/src/modals/Derive.tsx
```diff
@@ -54,29 +55,15 @@ function deriveValidate (suri: string, pairType: KeypairType): string | null {
 }
 
 function createAccount (source: KeyringPair, suri: string, name: string, password: string, success: string, genesisHash?: string): ActionStatus {
-  // we will fill in all the details below
-  const status = { action: 'create' } as ActionStatus;
-
-  try {
+  const getResult = () => {
```
wachulski, line 58: same consideration as in `Create`

packages/page-accounts/src/modals/Derive.tsx
```diff
@@ -85,16 +72,15 @@ function Derive ({ className = '', from, onClose }: Props): React.ReactElement {
...
   const [{ address, deriveError }, setDerive] = useState<DeriveAddress>({ address: null, deriveError: null });
   const [isConfirmationOpen, toggleConfirmation] = useToggle();
   const [{ isLocked, lockedError }, setIsLocked] = useState<LockState>({ isLocked: source.isLocked, lockedError: null });
-  const [{ isNameValid, name }, setName] = useState({ isNameValid: false, name: '' });
-  const [{ isPassValid, password }, setPassword] = useState({ isPassValid: false, password: '' });
-  const [{ isPass2Valid, password2 }, setPassword2] = useState({ isPass2Valid: false, password2: '' });
   const [{ isRootValid, rootPass }, setRootPass] = useState({ isRootValid: false, rootPass: '' });
   const [suri, setSuri] = useState('');
   const debouncedSuri = useDebounce(suri);
-  const isValid = !!address && !deriveError && isNameValid && isPassValid && isPass2Valid;
```
wachulski, line 97: why a logical change? not used?

wachulski (commented):
good content, some minor remarks, let's discuss them
- [ ] do you think there is a cost-effective space for retrofitting unit tests?

packages/page-accounts/src/modals/Create.tsx
vitaliyworks: Accidentally, sorry. Reverting...

packages/page-accounts/src/modals/CreateAccountInputs.tsx
vitaliyworks: Good call, thanks

packages/page-accounts/src/modals/Create.tsx
vitaliyworks: Hmm, `commitAccount` sounds good. I will go for it

packages/page-accounts/src/modals/Derive.tsx
vitaliyworks: Ack.

packages/page-accounts/src/modals/Derive.tsx
vitaliyworks: Actually the logic is not changed, but this not obvious because the same name is used for the variable that works differently now. `isPassValid` is handled by `CreateAccountInputs` is such a way that it is equivalent to the older `isPassValid && isPass2Valid` expression. The second password is handled internally by the `CreateAccountInputs` component, just as it was handled in the Create modal. Do you agree or did I miss something?

vitaliyworks: By "logic is not changed" I meant that functionally this works the same as before. And yes, the second password is not used here as it is handled internally by the `CreateAccountInputs`

vitaliyworks:
Thanks @wachulski :) Some tests would be nice, something similar to what we did for Accounts page? Although they would be especially nice prior to refactoring xD But as for efficiency, I think this is not as critical as sorting feature. How about I assign tests for myself but do the sorting first?

packages/page-accounts/src/modals/Derive.tsx
wachulski: got it, thx

wachulski (approved):

wachulski:
> Thanks @wachulski :) Some tests would be nice, something similar to what we did for Accounts page? Although they would be especially nice prior to refactoring xD But as for efficiency, I think this is not as critical as sorting feature. How about I assign tests for myself but do the sorting task first?

I added a note to our backlog. Let's cover it if time permits.

packages/page-accounts/src/modals/Create.tsx
```diff
@@ -134,35 +132,11 @@ function updateAddress (seed: string, derivePath: string, seedType: SeedType, pa
...
-
-    if (getEnvironment() === 'web') {
-      downloadAccount(result);
-    }
-  } catch (error) {
-    status.status = 'error';
-    status.message = (error as Error).message;
-  }
+  const commitAccount = () =>
+    keyring.addUri(getSuri(seed, derivePath, pairType), password, { genesisHash, isHardware: false, name, tags }, pairType === 'ed25519-ledger' ? 'ed25519' : pairType);
```
jacogr, line 137: `useCallback`

packages/page-accounts/src/modals/CreateAccountInputs.tsx
```diff
@@ -0,0 +1,61 @@
...
+        onEnter={_onCommit}
+        placeholder={t<string>('new account')}
+        value={name}
+      />
+    </Modal.Columns>
+    <PasswordInput
+      onChange={_onChangePass}
+      onEnter={_onCommit}
+    />
+  </>);
```
jacogr, line 58:
```suggestion
  return (
    <>
      <Modal.Columns hint={t<string>('The name for this account and how it will appear under your addresses. With an on-chain identity, it can be made available to others.')}>
        <Input
          className='full'
          help={t<string>('Name given to this account. You can edit it. To use the account to validate or nominate, it is a good practice to append the function of the account in the name, e.g "name_you_want - stash".')}
          isError={!isNameValid}
          label={t<string>('name')}
          onChange={_onChangeName}
          onEnter={_onCommit}
          placeholder={t<string>('new account')}
          value={name}
        />
      </Modal.Columns>
      <PasswordInput
        onChange={_onChangePass}
        onEnter={_onCommit}
      />
    </>
  );
```

packages/page-accounts/src/util.tsx
```diff
@@ -17,6 +20,39 @@ export function createMenuGroup (key: string, items: (React.ReactNode | false |
     : null;
 }
 
+export function downloadAccount ({ json, pair }: CreateResult): void {
+  const blob = new Blob([JSON.stringify(json)], { type: 'application/json; charset=utf-8' });
+
+  FileSaver.saveAs(blob, `${pair.address}.json`);
+}
```
jacogr, line 27:
```suggestion
export function downloadAccount ({ json, pair }: CreateResult): void {
  FileSaver.saveAs(
    new Blob([JSON.stringify(json)], { type: 'application/json; charset=utf-8' }),
    `${pair.address}.json`
  );
}
```

packages/page-accounts/src/modals/Create.tsx
vitaliyworks: But this code is evaluated conditionally - only on click, so can't use react hooks here, including `useCallback`. Or am I missing something?

jacogr: You are right, sorry it is inside an external function.

packages/page-accounts/src/modals/Create.tsx
```diff
@@ -395,22 +359,12 @@ function Create ({ className = '', onClose, onStatusChange, seed: propsSeed, typ
...
-              placeholder={t<string>('new account')}
-              value={name}
-            />
-          </Modal.Columns>
-          <PasswordInput
-            onChange={_onPasswordChange}
-            onEnter={_onCommit}
-          />
+          <CreateAccountInputs
+            _onCommit={_onCommit}
```
jacogr, line 363:
```suggestion
            onCommit={_onCommit}
```

packages/page-accounts/src/modals/CreateAccountInputs.tsx
```diff
@@ -0,0 +1,63 @@
...
+interface AccountPassword {
+  password: string;
+  isPasswordValid: boolean;
+}
+
+interface Props {
+  name: AccountName;
+  setName: (value: AccountName) => void;
+  setPassword: (value: AccountPassword) => void;
+  _onCommit: () => void;
```
jacogr, line 25:
```suggestion
  onCommit: () => void;
```

As above, not sure why this.

packages/page-accounts/src/modals/Derive.tsx
```diff
@@ -230,34 +201,12 @@ function Derive ({ className = '', from, onClose }: Props): React.ReactElement {
...
-                  className='full'
-                  help={t<string>('Verify the password entered above.')}
-                  isError={!isPass2Valid}
-                  label={t<string>('password (repeat)')}
-                  onChange={_onChangePass2}
-                  onEnter={_onCommit}
-                  value={password2}
-                />
+                <CreateAccountInputs
+                  _onCommit={_onCommit}
```
jacogr, line 205:
```suggestion
                  onCommit={_onCommit}
```

jacogr (commented):
Think it looks sane, some small comments.

packages/page-accounts/src/modals/Create.tsx
MiZiet: ✅

packages/page-accounts/src/modals/CreateAccountInputs.tsx
MiZiet: ✅

packages/page-accounts/src/modals/Derive.tsx
MiZiet: ✅

jacogr:
Needs resolving.

MiZiet:
@jacogr Done ✅

jacogr (approved):

PR #5909, opened 2021-08-02
Rebrand polkaBTC to interBTC

savudani8:


jacogr (approved):

savudani8:
It seems like there is a problem with using the latest version of the custom types, I'm getting this on my local: `FATAL: Unable to initialize the API: createType(CurrencyId):: DoNotConstruct: Cannot construct unknown type CurrencyId`. However if I manually copy-paste the same definitions, it works

packages/apps-config/src/api/spec/index.ts
```diff
@@ -120,7 +120,6 @@ const spec: Record<string, OverrideBundleDefinition> = {
   'bifrost-parachain': bifrostParachain,
   'bitcountry-node': bitcountry,
   'bitcountry-parachain': bitcountryParachain,
-  'btc-parachain': polkabtc,
```
jacogr, line 123: Possibly this - on non-upgraded chains (i.e. where the spec_name has not been changed), the types won't be injected at all. It will only match the new name below.

Would suggest leaving this line in. (just `: interbtc`

savudani8: Thank you! We're soon to re-genesis and are in this in-between phase with the chainspec names

packages/apps-config/src/ui/colors.ts
```diff
@@ -236,8 +238,6 @@ export const chainColors: Record<string, string> = Object.entries({
   PHOENIX: chainPhoenix,
   Plasm: chainPlasm,
   'Plasm PC2': chainPlasm,
-  PolkaBTC: chainPolkabtc,
```
jacogr, line 239: This is possibly the same - just needs to be kept for the time being.

savudani8: The problem was that there are two versions of the interbtc-parachain: one with `aura-grandpa` consensus (for testing) and one with `cumulus`. I included both in `api/spec/index.ts` and it works now - I think it's fine to remove the PolkaBTC configs

jacogr: All ok. The above is just for the color matching.

Issue #5910, opened 2021-08-02
Adapt block time to non-babe parachains


joelamouche:
## Problem

Currently, block time is computed this way:

```
const blockTime = (
        a.consts.babe?.expectedBlockTime ||
        a.consts.difficulty?.targetBlockTime ||
        a.consts.timestamp?.minimumPeriod.muln(2) ||
        DEFAULT_TIME
      );
```

But for parachains that dont use `babe` nor `difficulty`, and for which `timestamp?.minimumPeriod` doesnt make sens, we need to specify a value.

## Proposal

If the parachainSystem pallet is there, set the blocktime to 12000

joelamouche:
@bkchr what do you think

jacogr:
As the last (prior to default) fallback, sounds sane.

joelamouche:
I would put it before `a.consts.timestamp?.minimumPeriod.muln(2)` because our timestamp.minimumPeriod is not consistant with block time and is set to zero

jacogr:
Don’t believe that is right. Can certainly add a zero check, but no assumptions at this point for parachain fallbacks. Since it means rework.

bkchr:
> I would put it before `a.consts.timestamp?.minimumPeriod.muln(2)` because our timestamp.minimumPeriod is not consistant with block time and is set to zero

Why?

joelamouche:
Sorry I meant it's set to one

jacogr:
It is supposed to be blocktime / 2 for Aura -

```rust
		/// The minimum period between blocks. Beware that this is different to the *expected* period
		/// that the block production apparatus provides. Your chosen consensus system will generally
		/// work with this to determine a sensible block time. e.g. For Aura, it will be double this
		/// period on default settings.
		#[pallet::constant]
		type MinimumPeriod: Get<Self::Moment>;
```

So for aura parachains, should be 6000

JoshOrndorff:
> It is supposed to be blocktime / 2 for Aura

That's true. But Moonbeam is not using Aura.

>  >  timestamp.minimumPeriod is not consistant with block time and is set to <del>zero</del> one
>
> Why?

The Moonbeam runtime is used in multiple different contexts. The most obvious one, and the one @joelamouche is trying to fix here is the parachain context. However our runtime is also used in an instant-sealing node for unit tests, and for development (both our own internal development, and smart contract devs targeting Moonbeam). There is no reason that blocks can't come in very close succession in the instant sealing node.

[...]

PR #5913, opened 2021-08-03
Discard invalid configuration minimumPeriod values

jacogr:
Closes https://github.com/polkadot-js/apps/issues/5910

### DIFF
--- a/packages/react-hooks/src/useBlockTime.ts
+++ b/packages/react-hooks/src/useBlockTime.ts
@@ -4,19 +4,22 @@
 import type { ApiPromise } from '@polkadot/api';
 import type { Time } from '@polkadot/util/types';
 
-import BN from 'bn.js';
 import { useMemo } from 'react';
 
-import { BN_ONE, extractTime } from '@polkadot/util';
+import { BN, BN_ONE, BN_THOUSAND, BN_TWO, bnToBn, extractTime } from '@polkadot/util';
 
 import { useTranslation } from './translate';
 import { useApi } from './useApi';
 
 type Result = [number, string, Time];
 
-const DEFAULT_TIME = new BN(6000);
+const DEFAULT_TIME = new BN(6_000);
 
-export function useBlockTime (blocks = BN_ONE, apiOverride?: ApiPromise | null): Result {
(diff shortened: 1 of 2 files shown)


packages/react-hooks/src/useBlockTime.ts
```diff
@@ -4,33 +4,45 @@
...
       const blockTime = (
+        // Babe
         a.consts.babe?.expectedBlockTime ||
-        a.consts.difficulty?.targetBlockTime ||
-        a.consts.timestamp?.minimumPeriod.muln(2) ||
-        DEFAULT_TIME
+        // POW, eg. Kulupu
+        a.consts.difficulty?.targetBlockTime || (
+          // Check against threshold to determine value validity
+          a.consts.timestamp?.minimumPeriod.gt(THRESHOLD)
```
jacogr, line 35:
```suggestion
          a.consts.timestamp?.minimumPeriod.gte(THRESHOLD)
```

joelamouche:
Thanks :)

packages/react-hooks/src/useBlockTime.ts
```diff
@@ -4,33 +4,45 @@
...
 import { useTranslation } from './translate';
 import { useApi } from './useApi';
 
 type Result = [number, string, Time];
 
-const DEFAULT_TIME = new BN(6000);
+const DEFAULT_TIME = new BN(6_000);
 
-export function useBlockTime (blocks = BN_ONE, apiOverride?: ApiPromise | null): Result {
+// Some chains incorrectly use these, i.e. it is se to values such as 0 or even 2
```
JoshOrndorff, line 18: Thank you for fixing Moonbeam;s issue.

FWIW, I don't think our use is incorrect. The field is called `MinimumPeriod` not `ExpectedPeriod` or `HalfOfTheExpectedPeriod`, and the docs also say `Beware that this is different to the *expected* period
		/// that the block production apparatus provides.`

crystalin:
Awesome @jacogr , that clearly improves our user experience on PolkadotJs. I still wonder why we never thought about it before :)

PR #5923, opened 2021-08-05
Add sorting for accounts view

vitaliyworks:
Add sorting by `parent`, `date`, `type` and by `total balance`.
This patch includes optimisations that prevent account rows reinitialization after sorting.


packages/page-accounts/src/Accounts/index.tsx
```diff
@@ -71,23 +72,42 @@ function Overview ({ className = '', onStatusChange }: Props): React.ReactElemen
   });
   const isLoading = useLoadingDelay();
 
+  const _sort = (header: SortCategory) => {
+    setSortBy(({ sortBy, sortFromMax }) =>
+      ({ sortBy: header,
+        sortFromMax: sortBy === header ? !sortFromMax : true }));
+  };
+
+  const sortableHeader = (header: SortCategory, clas?: string) => [
```
wachulski, line 81: class?

wachulski: htmlClass?

packages/page-accounts/src/util.tsx
```diff
@@ -27,23 +28,91 @@ function expandList (mapped: SortedAccount[], entry: SortedAccount): SortedAccou
...
+// This function does not sort inplace.
+function stableSort<T> (elements: T[], cmp: (a: T, b: T) => number): T[] {
+  const indexedElements: [T, number][] = elements.map((x, index) => [x, index]);
+  const stableCmp = (a: [T, number], b: [T, number]) =>
+    cmp(a[0], b[0]) || a[1] - b[1];
+
+  indexedElements.sort(stableCmp);
+
+  return indexedElements.map(([x]) => x);
+}
```
wachulski, line 70: this is Accounts-free/agnostic. What about extracting to a sep util file?
benefits: + sharing in other areas, + more explicit -> easier to replace with some other algorithm in the future if needed.

packages/page-accounts/src/Accounts/index.tsx
```diff
@@ -60,6 +60,7 @@ function Overview ({ className = '', onStatusChange }: Props): React.ReactElemen
   const [filterOn, setFilter] = useState<string>('');
   const [sortedAccountsWithDelegation, setSortedAccountsWithDelegation] = useState<SortedAccount[] | undefined>();
   const [{ sortedAccounts, sortedAddresses }, setSorted] = useState<Sorted>({ sortedAccounts: [], sortedAddresses: [] });
+  const [{ sortBy, sortFromMax }, setSortBy] = useState<{ sortBy: SortCategory, sortFromMax: boolean }>({ sortBy: 'date', sortFromMax: true });
```
wachulski, line 63: could you extract some const like DEFAULT_ACCOUNTS_SORT or similar?

packages/page-accounts/src/util.tsx
wachulski: what about testing this little functionality? just a bunch of basic ones?

wachulski (commented):
Thanks for thinking it through.
What about unit testing of that little stable sort facility?
What about unit testing Accounts table with accounts rows with sorting?

packages/page-accounts/src/util.tsx
vitaliyworks: Sure. Moving this away into a separate file and adding some tests.

packages/page-accounts/src/Accounts/index.tsx
vitaliyworks: Sure will do

vitaliyworks:
Thanks, I will spent some more time to thing how to optimise this.
Will add tests for sorting function.
But tests for the UI would probably need to be rewritten once the sorting changes, so I will delay those, OK?

packages/page-accounts/src/Accounts/index.tsx
vitaliyworks: I did not wanted to use `class` because it could conflict with the keyword. But `htmlClass` is the best here

packages/page-accounts/src/Accounts/index.tsx
```diff
@@ -225,23 +269,29 @@ function Overview ({ className = '', onStatusChange }: Props): React.ReactElemen
...
+            onChange={(item: SortCategory) => setSortBy({ sortBy: item, sortFromMax })}
+            options={sortCategory.map((x) => ({ text: x, value: x }))}
+          />
+          <Button
+            icon='arrows-alt-v'
+            onClick={() => setSortBy({ sortBy, sortFromMax: !sortFromMax })}
+          />
+        </section>
+      </SummaryBox>
+
```
jacogr, line 287:
```suggestion
```

packages/page-accounts/src/Accounts/index.tsx
```diff
@@ -225,23 +269,29 @@ function Overview ({ className = '', onStatusChange }: Props): React.ReactElemen
...
-            filter={filterOn}
-            isFavorite={isFavorite}
-            key={`${index}:${account.address}`}
-            proxy={proxies?.[index]}
-            setBalance={_setBalance}
-            toggleFavorite={toggleFavorite}
-          />
-        ))}
+        {!isLoading &&
+          sortedAccounts.map(({ address }) => accountComponets[address])}
```
jacogr, line 294:
```suggestion
          sortedAccounts.map(({ address }) => accountComponents[address])}
```

packages/page-accounts/src/stableSort.ts
```diff
@@ -0,0 +1,21 @@
...
+ * @returns New sorted array.
+ */
+export default function<T> (elements: T[], cmp: (a: T, b: T) => number): T[] {
+  const indexedElements: [T, number][] = elements.map((x, index) => [x, index]);
+  const stableCmp = (a: [T, number], b: [T, number]) =>
+    cmp(a[0], b[0]) || a[1] - b[1];
+
+  indexedElements.sort(stableCmp);
+
+  return indexedElements.map(([x]) => x);
```
jacogr, line 20:
```suggestion
return [...elements].sort(cmp);
```

Rather than mapping twice and sorting, this only creates a single array.

packages/page-accounts/src/Accounts/index.tsx
```diff
@@ -31,18 +32,26 @@ interface Balances {
...
+
+const DEFAULT_SORT_CONTROLS: SortControls = { sortBy: 'date', sortFromMax: true };
+
 const STORE_FAVS = 'accounts:favorites';
 
+const SortDropdown = React.memo(styled(Dropdown)`
+  max-width: 185px;
+  align: flex start;
+  margin-left: -28px;
+`);
```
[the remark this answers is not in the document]
vitaliyworks: ✅

packages/page-accounts/src/stableSort.ts
vitaliyworks: I was trying to not use bare `.sort()` because I thought that it is not stable, but the ecmascript spec changed recently and now it is stable: https://v8.dev/features/stable-sort
So this function is not actually needed, I will remove it.

ekowalsk:
Hi @jacogr, when you get down to reviewing pr's could you have a look at this one? Thank you :)

jacogr:
Needs resolving.

jacogr:
Needs resolving :)

ekowalsk:
✅  Resolved

packages/react-components/src/CryptoType.tsx
```diff
@@ -1,14 +1,14 @@
 // Copyright 2017-2021 @polkadot/react-components authors & contributors
 // SPDX-License-Identifier: Apache-2.0
 
-import type { AccountId, AccountIndex, Address } from '@polkadot/types/interfaces';
+import type { AccountIdIsh } from '../../page-accounts/src/util';
```
jacogr, line 4: We should not import from here.

- Generally (this is broken in a few places) we don't want to import cross-pages. 
- Since react-components is top-most level, don't want ti import from a page
- Where imports are used, it should be via `@polkadot/app-accounts`

In this case move the type to react-components and app-accounts can import it. (Or even duplicate)

ekowalsk: ✅ I hope I found all of the bad imports

jacogr (approved):
Seems solid.

MiZiet:
@jacogr could you merge it? Thanks!

PR #5924, opened 2021-08-05
Remove sui popup

MiZiet:


packages/react-components/src/Popup/PopupWindow.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+import React from 'react';
+import { createPortal } from 'react-dom';
+import styled, { css } from 'styled-components';
+
+import { useElementPosition } from '@polkadot/react-hooks';
+
+import { getPosition } from './utils';
+
+export const POINTER_OFFSET = 14 * 0.8;
```
wachulski, line 14: Could you put a line of docs - why such values?

packages/react-components/src/Popup.tsx
```diff
@@ -1,34 +1,122 @@
...
+
+interface WindowProps {
+  position: 'left' | 'middle' | 'right'
+}
+
+const PopupWindow = styled.div<WindowProps>`
+  position: absolute;
+  margin-top: 0.7rem;
+  top: 100%;
+  //bottom: 0;
```
wachulski: not needed?

wachulski (approved):
LGTM

packages/react-components/src/Popup/PopupWindow.tsx
MiZiet: ✅

packages/react-components/src/Popup.tsx
MiZiet: I think it's not in the code anymore, you must have looked at some old commit

packages/react-components/src/Popup/PopupWindow.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+import React from 'react';
+import { createPortal } from 'react-dom';
+import styled, { css } from 'styled-components';
+
+import { useElementPosition } from '@polkadot/react-hooks';
+
+import { getPosition } from './utils';
+
+export const PopupWindow: React.FC<PopupWindowProps> = ({ children, position, setRef, triggerPosition }) => {
```
jacogr, line 14:
```suggestion
function PopupWindow({ children, position, setRef, triggerPosition }: PopupWindowProps) {
```

packages/react-components/src/Popup/PopupWindow.tsx
```diff
@@ -0,0 +1,90 @@
...
+      ref={setRef}
+      windowPosition={getPosition(triggerPosition, position, triggerPosition.verticalPosition, windowPosition)}
+    >
+      {children}
+    </Window>,
+    document.body
+  );
+};
+
+const Window = styled.div<WindowProps>`
```
jacogr, line 30: Don't use this form in styled-components, rather follow the same pattern as used elsewhere, i.e. div + className + styles on React.memo

packages/react-components/src/Popup/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+          />
+        )}
+      </div>
+    </PopupWrapper>
+  );
+};
+
+export default React.memo(Popup);
+
+const PopupWrapper = styled.div`
```
jacogr, line 58: Include styles alongside React.memo

packages/react-components/src/styles/semantic.ts
```diff
@@ -171,23 +171,6 @@ export default (_theme: ThemeDef): string => `
     display: flex !important;
   }
 
-  .ui.popup {
```
jacogr, line 174: +1

packages/react-hooks/src/useElementPosition.ts
```diff
@@ -0,0 +1,43 @@
...
+  height: number,
+  verticalPosition: 'bottom' | 'top'
+}
+
+export function useElementPosition (ref: React.MutableRefObject<HTMLElement | undefined | null>): ElementPosition | undefined {
+  const [elementPosition, setElementPosition] = useState<ElementPosition>();
+
+  const windowSize = useWindowSize();
+
+  const scrollY = useScroll();
```
jacogr, line 24:
```suggestion
  const [elementPosition, setElementPosition] = useState<ElementPosition>();
  const windowSize = useWindowSize();
  const scrollY = useScroll();
```

packages/react-hooks/src/useScroll.ts
```diff
@@ -0,0 +1,26 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+import { useEffect, useState } from 'react';
+
+export function useScroll (): number {
+  const [scrollY, setScrollY] = useState(0);
+
+  function logit () {
+    setScrollY(window.pageYOffset);
+  }
```
jacogr, line 11: useCallback

jacogr (commented):
It is great having these component replaced. Some small comments.

packages/react-hooks/src/useScroll.ts
MiZiet: ✅

packages/react-hooks/src/useElementPosition.ts
MiZiet: ✅

packages/react-components/src/Popup/index.tsx
MiZiet: ✅

packages/react-components/src/Popup/PopupWindow.tsx
MiZiet: ✅

packages/react-components/src/Popup/PopupWindow.tsx
MiZiet: ✅

jacogr (commented):
The placing here is really funny at points. Sometimes it works, sometimes it doesn't seem to. See below.

image

Wondering if a horizonal arrow and placing it to the left is not more sane. Instead of trying to move it above/below, simplify?

Additionally would make the triangle slightly larger.

From a style perspective, it is still just as ugly (don't take this as a criticism) as the old SUI version, but after this we have better control so can make it fit in better, which is absolutely great.

jacogr:
Yes, the calcs are indeed an issue, hence the question if we maybe won't be better suited moving it horizontally to the left. (Not a request, rather a suggestion _if_ it makes it easier) Will play through now.

jacogr (approved):
lgtm.

PR #5935, opened 2021-08-07
Added Pinknode Endpoints

Capeguy:


jacogr:
Not sure I want these re-enabled - last round all were down for >24hrs and their monitoring didn't pick it up. Since the apps UI randomly selects endpoints (assuming all are actually there), this created a support nightmare on my side.

Capeguy:
Hey @jacogr, I've been actively monitoring those endpoints since the last time you flagged them out. You can check the GitHub actions I've got running every 15 minutes here: https://github.com/Pinknode/apps/actions
Previously when those failed, we were on old infrastructure and since then we have moved to new ones. Do let me know what I can do to make this work.

Fergiefur (approved):

jacogr:
Needs resolving with master.

A couple of throw away comments -

1. We can enable this provider on the non high-traffic chains where the various teams are masters of their destiny. So anything apart from the Kusama & Polkadot relay chains. If that behaves satisfactory, we can bring on Polkadot and Kusama.

2. GH actions is a monitoring solution - the reason why I use it is not to monitor, but as a quick-and-dirty solution to detect where things _may_ have gone wrong. It is not the way I manage my own infrastructure, but rather only here since it allows linking from PRs.

3. Ensure that the light client solution (where applicable) is always kept last.

Capeguy:
Hi @jacogr,

Have updated based on your comments.

jacogr (approved):

PR #5947, opened 2021-08-13
Display parent account info above account name

ekowalsk:
image
image
image


krzysztof-jelski:
Styling remark: in the designs the account name is aligned all the time with middle of identicon. Would be good if we could preserve that.
image

packages/page-accounts/src/Accounts/Account.tsx
```diff
@@ -472,7 +491,11 @@ function Account ({ account: { address, meta }, className = '', delegation, filt
           )}
         </td>
         <td className='address'>
-          <AddressSmall value={address} />
+          <AddressSmall
+            parentAddress={(typeof meta.parentAddress === 'string') ? meta.parentAddress : undefined}
```
jacogr, line 495:
```suggestion
            parentAddress={meta.parentAddress || undefined}
```

But even the `|| undefined` should not be needed, these should just be string or undefined (if null, should also be all ok)

packages/react-components/src/AddressSmall.tsx
```diff
@@ -6,6 +6,9 @@ import type { AccountId, Address } from '@polkadot/types/interfaces';
 import React from 'react';
 import styled from 'styled-components';
 
+import ParentAddress from '@polkadot/react-components/ParentAddress';
+import { toShortAddress } from '@polkadot/react-components/util';
```
jacogr, line 10:
```suggestion
import ParentAddress from './ParentAddress';
import { toShortAddress } from './util';
```

We are in the same package.

packages/react-components/src/ParentAddress.tsx
```diff
@@ -0,0 +1,42 @@
...
+      </AccountName>
+    </div>
+  );
+}
+
+export default React.memo(styled(ParentAddress)`
+    display: flex;
+    align-items: center;
+    color: #8B8B8B;
+    font-size: 0.75rem;
```
jacogr, line 35:
```suggestion
  align-items: center;
  color: #8B8B8B;
  display: flex;
  font-size: 0.75rem;
```

jacogr (commented):
Love the display.

packages/page-accounts/src/Accounts/Account.tsx
ekowalsk: ✅

packages/react-components/src/AddressSmall.tsx
ekowalsk: ✅

packages/react-components/src/ParentAddress.tsx
ekowalsk: ✅

jacogr (approved):
Thank you.

PR #5966, opened 2021-08-18
Add max transfer bound to transfer modal

maciejnems:
Our users are complaining about accidentally making transfers that together with a fee are bigger than the account balance. Also currently it is inconvenient to check if a transfer together with a fee is lower than the balance. Because of that we thought about preventing users from making invalid transfers.

This PR passes `maxValue` prop for `InputBalance` set to `maxTransfer`. Now, when user tries to make a transfer that is too big the transfer button will be locked and the input field will be red. This will stop users from making this type of mistake and loosing fee in a process. Also now it is easier to check if a transfer is valid.

Screenshot from 2021-08-18 13-18-08


jacogr:
Seems sane as it stands, just needs a small fix to the linting - props needs to be alphabetical. (`yarn lint` locally should just fix it)

jacogr (approved):

PR #5975, opened 2021-08-19
Account row collapsible row details

ekowalsk:


jacogr (commented):
The horizontal spacing has been resolved, this is great.

We still have a density problem in staking. On this 13 rows takes up the same space as 20 rows on old.

ekowalsk:
The density problem has been resolved. Now also the Council looks as dense as previously - is it ok?

PR #6003, opened 2021-08-25
Style flags

ekowalsk:
image
image
image
image
image

### DIFF
--- a/packages/react-components/src/Flag.tsx
+++ b/packages/react-components/src/Flag.tsx
@@ -16,7 +16,7 @@ interface FlagProps {
 function Flag ({ className = '', color, label }: FlagProps): React.ReactElement<FlagProps> {
   return (
     <Tag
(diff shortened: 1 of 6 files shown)


packages/react-components/src/Flag.tsx
```diff
@@ -0,0 +1,66 @@
...
+      label={label}
+      size='tiny'
+    />
+  );
+}
+
+export default React.memo(styled(Flag)`
+  border-radius: 0 0.25rem 0.25rem 0;
+  padding: 0.5833em 1.25em 0.5833em 1.5em;
+  font-size: 0.78571429rem;
```
jacogr, line 31: This sizing should really be applied to tags as well. Same height, same font. (Have not checked the font now, but at least the height s is ms different - not for this PR)

jacogr (approved):

PR #6012, opened 2021-08-27
Change buttons position in account table

MiZiet:
When applied these changes will:
- move the three dots button next to the row expander to keep the consistency with the rest of the app (we have the same buttons order in the bounties-page)

image


jacogr (changes requested):
Not quite on-board with this. Perhaps the bounties needs adjustment - the actions pertinent to the user should be next to each other, not spread with links in-between.

MiZiet:
@jacogr so we should keep the three dot button and the send button next to each other? Should we leave it as it is or move the links to the left?

jacogr:
Move in the links left (everywhere) and ensure that any actions are last, it probably the sane approach.

MiZiet:
I've moved the links to the left in accounts and in the address-book.

jacogr:
Just push a change to .123trigger in the root (incrementing the first counter)  not sure why GH decided to not run CI

MiZiet:
@jacogr Done ✅

jacogr (approved):

PR #6024, opened 2021-08-29
Added Dwellir RPC endpoints for Kusama and Polkadot relay chains

Maharacha:
[Dwellir](https://dwellir.com/) runs RPC nodes for both Polkadot and Kusama relay chains. We want to contribute by making them public available!

### DIFF
--- a/packages/apps-config/src/endpoints/productionRelayKusama.ts
+++ b/packages/apps-config/src/endpoints/productionRelayKusama.ts
@@ -23,6 +23,7 @@ export function createKusama (t: TFunction): EndpointOption {
       Parity: 'wss://kusama-rpc.polkadot.io',
       OnFinality: 'wss://kusama.api.onfinality.io/public-ws',
       'Patract Elara': 'wss://pub.elara.patract.io/kusama',
+      Dwellir: 'wss://kusama-rpc.dwellir.com',
       'light client': 'light://substrate-connect/kusama'
       // Pinknode: 'wss://rpc.pinknode.io/kusama/explorer' // https://github.com/polkadot-js/apps/issues/5721
     },
--- a/packages/apps-config/src/endpoints/productionRelayPolkadot.ts
+++ b/packages/apps-config/src/endpoints/productionRelayPolkadot.ts
@@ -23,6 +23,7 @@ export function createPolkadot (t: TFunction): EndpointOption {
       Parity: 'wss://rpc.polkadot.io',
       OnFinality: 'wss://polkadot.api.onfinality.io/public-ws',
       'Patract Elara': 'wss://pub.elara.patract.io/polkadot',
+      Dwellir: 'wss://polkadot-rpc.dwellir.com',
       'light client': 'light://substrate-connect/polkadot'
       // Pinknode: 'wss://rpc.pinknode.io/polkadot/explorer' // https://github.com/polkadot-js/apps/issues/5721
     },


packages/apps-config/src/endpoints/productionRelayKusama.ts
```diff
@@ -23,6 +23,7 @@ export function createKusama (t: TFunction): EndpointOption {
       Parity: 'wss://kusama-rpc.polkadot.io',
       OnFinality: 'wss://kusama.api.onfinality.io/public-ws',
       'Patract Elara': 'wss://pub.elara.patract.io/kusama',
+      '🛡 DWELLIR 🛡': 'wss://kusama-rpc.dwellir.com:9977',
```
jacogr, line 26:
```suggestion
      DWELLIR: 'wss://kusama-rpc.dwellir.com:9977',
```

packages/apps-config/src/endpoints/productionRelayPolkadot.ts
```diff
@@ -23,6 +23,7 @@ export function createPolkadot (t: TFunction): EndpointOption {
       Parity: 'wss://rpc.polkadot.io',
       OnFinality: 'wss://polkadot.api.onfinality.io/public-ws',
       'Patract Elara': 'wss://pub.elara.patract.io/polkadot',
+      '🛡 DWELLIR 🛡': 'wss://polkadot-rpc.dwellir.com:9966',
```
jacogr, line 26:
```suggestion
      DWELLIR: 'wss://polkadot-rpc.dwellir.com:9966',
```

jacogr (commented):
Prefer to stay away from non-ascii

jacogr (commented):
What monitoring do you have in-place? 

I have had issues in the past with providers not being able to keep the infrastructure available, which ended up creating a support nightmare for me. Since these are really high-traffic chains, it needs to be handled appropriately.

(Endpoints are randomly selected for users, so when issues do happen it does create _slight_ issues on my support queue)

Maharacha:
> What monitoring do you have in-place?
> 
> I have had issues in the past with providers not being able to keep the infrastructure available, which ended up creating a support nightmare for me. Since these are really high-traffic chains, it needs to be handled appropriately.
> 
> (Endpoints are randomly selected for users, so when issues do happen it does create _slight_ issues on my support queue)

To monitor the system resources, we use Grafana. We also have external monitoring that monitors all our services on our sites. For notifications on our alarms we use PagerDuty. Regarding availability, we have three sites located at different places in Sweden, where we can move our services between.

Can you give us a hint on how the much the traffic will be when we are added?

packages/apps-config/src/endpoints/productionRelayKusama.ts
Maharacha: Ok! We will change the ports to standard web ports as well. Some firewalls might block these ports.. Update will come.

jacogr:
Honestly don’t know about the load - I don’t have access to infrastructure to check. But since endpoints gets selected randomly, you can more-or-less expect 1/N connections to your services. (Where N is the number of providers available)

Maharacha:
We have updated the name now. We have also changed to standard web ports. Ready to merge from our side!

erik78se:
As a followup on your question about monitoring, we perform service monitoring remotely, performing a remote rpc call and an alarm is triggered if it takes more than 5 seconds to get a valid result.

We are going to implement automated failover for the service using some GSLB which will require no actions from your side.

jacogr (approved):

Maharacha:
@jacogr We are writing proposal

Maharacha:
@jacogr We are writing Treasury Proposals for Kusama and Polkadot RPC. We have now been told that we should wait with merging until the proposal is approved. And that we should do two separate PR's, one for Kusama and one for Polkadot, since there will be two different proposals. Can you revert this and I will create new PR's?

rrtti:
Its fine if its merged on Kusama: the proposal will be for Kusama for now, the team cna show the nodes are already running and focus the efforts on the improvements they want to achieve for the service - as you prefer.

PR #6029, opened 2021-08-30
make heiko reachable

GopherJ:
parallel heiko is now reachable: https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Fheiko-rpc-0.parallel.fi


jacogr (approved):

jacogr:
Seems good, failing. Please check CI.

GopherJ:
@jacogr strange it looks ok and I changed only `isUnreachable` which shouldn't warn about sorting, could you give some hints?

jacogr:
`isDisabled` and `isUnreachable` is not being checked for sorting. So this means when enabled again, it may yields issues since it could sit in the wrong place.

GopherJ:
@jacogr kist found that the order is by `text` instead of `info`

jacogr (approved):

PR #6049, opened 2021-09-01
Account sidebar - single button for editing name and tags

ekowalsk:
image
image
image
image


jacogr (commented):
I would suggest instead of the "owned" (which is not a button), removing it and placing the edit button there - with the same styling and the send.

Like the idea, it certainly helps.

ekowalsk:
> I would suggest instead of the "owned" (which is not a button), removing it and placing the edit button there - with the same styling and the send.
> 
> Like the idea, it certainly helps.

Ok, so the Owned should be removed completely from the sidebar and not displayed elsewhere? And only the remove button should be displayed if the account is not owned?

jacogr:
Yes, I think that should do the trick - open for other suggestions?

ekowalsk:
I've just added a commit with styling, so the result is as in the screenshots

jacogr (commented):
Some small comments.

ekowalsk:
✅

packages/react-components/src/styles/index.ts
```diff
@@ -498,6 +498,10 @@ export default createGlobalStyle<Props & ThemeProps>(({ theme, uiHighlight }: Pr
     }
   }
 
+  .opacity-70 {
```
krzysztof-jelski, line 501: Is this actually used anywhere?

packages/react-components/src/Tags.tsx
```diff
@@ -18,10 +18,11 @@ interface Props {
   onToggleIsEditing?: () => void;
   onSave?: () => void;
   value: string[];
+  withEditButton?: boolean;
```
krzysztof-jelski, line 21: Is there any place we still use the edit button? Can't find it.

Even if we do, I feel like we have to many props. Should be enough to have `isEditing` (for showing inputs) and `isEditable` (for showing button).

packages/react-components/src/styles/index.ts
ekowalsk: it's applied to flags when editing 
image

krzysztof-jelski: Not really, looks like it's styled directly there: https://github.com/polkadot-js/apps/pull/6049/files#diff-288169aeb02a8eaf9159637d338b805e1fa0f48492cc2d5abcac2b72ad0f4747R214.

packages/page-accounts/test/pages/accountsPage.tsx
```diff
@@ -30,59 +33,11 @@ class NotYetRendered extends Error {
 }
 
 // utility wrapper over an account item in accounts table, serves basic assertions about an account row
```
krzysztof-jelski, line 35: This comment belongs with the extracted `AccountRow`.

packages/page-accounts/src/Accounts/index.spec.ts
```diff
@@ -299,6 +301,178 @@ describe('Accounts page', () => {
...
+          });
+        });
+
+        describe('tags', () => {
+          beforeEach(async () => {
+            await sideBar.setTag(defaultTag);
+          });
+
+          it('within sidebar', async () => {
+            await waitFor(() => sideBar.tagsEqual(defaultTag));
```
krzysztof-jelski, line 349: Lines like this don't test anything. `waitFor` wait for the callback not to throw, or for the promise not to reject. It does not check the boolean result from the promise.

packages/page-accounts/src/Accounts/index.spec.ts
```diff
@@ -299,6 +301,178 @@ describe('Accounts page', () => {
...
+          await waitFor(() => sideBar.accountNameEquals('ALICE'));
+
+          sideBar = await openSidebarForAccountRow(1);
+          await sideBar.assertAccountName('BOB');
+        });
+      });
+
+      afterEach(() => {
+        keyring.forgetAccount(injectedAddress);
+        keyring.forgetAddress(injectedAddress);
```
krzysztof-jelski, line 467: Those lines trigger a `useEffect`, that updates some react components. This raises a warning in tests, as we don't tell react we're expecting it. We should either (1) call [cleanup](https://testing-library.com/docs/react-testing-library/api#cleanup) before those lines or (2) wrap them in `act` to tell react component updates are OK here.

packages/page-accounts/test/pages/accountsPage.tsx
ekowalsk: ✅

packages/page-accounts/src/Accounts/index.spec.ts
ekowalsk: I removed those test as I couldn't test tags

packages/react-components/src/styles/index.ts
ekowalsk: ✅

packages/page-accounts/src/Sidebar/Sidebar.tsx
```diff
@@ -1,55 +1,44 @@
...
-  );
-
-  const _onUpdateName = useCallback(
-    (): void => {
-      onSaveName();
-      onUpdateName && onUpdateName();
-    },
-    [onSaveName, onUpdateName]
-  );
+  const ref = useRef<HTMLDivElement>();
```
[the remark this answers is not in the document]
ekowalsk: ✅

packages/react-components/src/Tags.tsx
ekowalsk: It is also used in Contracts and there we don't provide button for editing like in sidebar. 
image

packages/page-accounts/src/Sidebar/AccountMenuButtons.tsx
```diff
@@ -0,0 +1,141 @@
...
+    display: flex;
+    flex-direction: row;
+    justify-content: space-between;
+    margin-bottom: 0;
+
+    .ui--Button {
+      width: 100%;
+      min-width: max-content;
+    }
+  }
```
jacogr, line 140: Not fond of this layout, the buttons here are different to the rest of the UI, i.e. they take up 50% of space each. I would suggest just sticking the the normal group layout.

jacogr (commented):
One small comment on the button layout, i.e. being different than elsewhere.

Apart from that, does what it says on the tin.

ekowalsk:
The layout was changed, currently it looks like that:
image
image
image

jacogr (approved):
Thank you.

PR #6089, opened 2021-09-07
Extract input for filtering by name or tags

ekowalsk:
image


packages/page-accounts/src/Accounts/index.tsx
```diff
@@ -285,11 +285,13 @@ function Overview ({ className = '', onStatusChange }: Props): React.ReactElemen
             isCircular
             onClick={onSortDirectionChange()}
           />
+          <div className='filter'>
+            {filter}
+          </div>
```
MiZiet, line 290: I'm not sure if we should wrap it into that div. In `filter` it already has a `div.filter-tags` so we should style it there or delete it.

ekowalsk: ✅

MiZiet (approved):
LGTM

krzysztof-jelski:
Shouldn't we do it for all the tables across whole app?

jacogr:
We should - however in this case, probably makes sense to do it at one place (getting it right) and then have a follow-up PR that tackles the rest?

ekowalsk:
I'm not sure if I get it correctly what exactly should be included in this pr - should the filtering be somehow improved for the Accounts? And as I understand you would like to apply the collapsible row for the AddressBook within a follow-up pr? Or within that one?

jacogr:
My suggestion is to -

- leave the other filters for a follow-up PR
- same with address balances

Always easier to review and test less.

krzysztof-jelski:
@jacogr OK, so perhaps you can approve & merge this one, and we'll create follow-up PRs as suggested.

packages/page-accounts/src/Accounts/index.tsx
```diff
@@ -306,12 +306,12 @@ function Overview ({ className = '', onStatusChange }: Props): React.ReactElemen
 
 export default React.memo(styled(Overview)`
   .filter--tags {
-    .ui--Dropdown {
-      padding-left: 0;
+    width: 17.857rem;
```
jacogr, line 309: Is there a reason why this doesn't take up the available space? It is always quite tricky with these (the mock-ups do it nicely, ie. these are in the same line as the buttons, assuming it takes up space), however as it is here it is a bit of a "random" (not the right word, but hopefully drives the point across) sizing.

ekowalsk:
OK, I've changed the width of filter input so that it looks like at the mock-ups, and I've moved group buttons to the same line (as at the designs)

jacogr (approved):

PR #6091, opened 2021-09-07
Remove Patract from avaiable Polkadot Nodes

shawntabrizi:
I have had consistent issues when connecting to Patract node when going on the Polkadot JS Apps page.

image

I suggest we remove them for now, until they can prove their nodes are more stable.


jacogr:
They are in Asia - they never appear in the list of unavailable nodes, eg. https://github.com/polkadot-js/apps/issues/6081

yuelipeng:
It may have some disconnection when reaching the bandwidth limit of k8s, we are increasing the limit and fix disconnection bug.

PR #6095, opened 2021-09-08
[integritee] add parachain rpc node which is online

brenzi:
not sure if manually removing the unreachable flag is how it should be done. Anyway, that node is reachable now. Let me know


jacogr (approved):
All perfect.

PR #6096, opened 2021-09-08
Style identity and balance sections in sidebar

MiZiet:
If applied these changes will:
- style identity section in sidebar,
- style balances section in sidebar.


packages/page-accounts/src/Sidebar/Sidebar.spec.tsx
```diff
@@ -0,0 +1,128 @@
...
+import { keyring } from '@polkadot/ui-keyring';
+
+import { AccountOverrides } from '../../test/hooks/default';
+import { AccountsPage } from '../../test/pages/accountsPage';
+
+const anAccount = (): AccountOverrides => ({});
+const aliceAddress = '5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY';
+const bobAddress = '5FHneW46xGXgs5mUiveU4sbTyGBzmstUspZC92UhjJM694ty';
+
+describe('Sidebar, subs list', () => {
```
MiZiet, line 16: Once #6049 is merged we can refactor these tests.

packages/page-accounts/src/Sidebar/Identity.tsx
```diff
@@ -24,20 +26,17 @@ function Identity ({ address, identity }: Props): React.ReactElement<Props> | nu
   const { isRegistrar, registrars } = useRegistrars();
   const [isJudgementOpen, toggleIsJudgementOpen] = useToggle();
   const subs = useCall<[BalanceOf, AccountId[]]>(api.query.identity?.subsOf, [address])?.[1];
+  const [isOpen, toggleIsOpen] = useToggle();
```
krzysztof-jelski, line 29:
```suggestion
  const [isSubsModalOpen, toggleSubsModalOpen] = useToggle();
```

packages/page-accounts/src/Sidebar/Sidebar.spec.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+    const sidebar = await screen.findByTestId('sidebar');
+    const identitySection = await within(sidebar).findByTestId('identity-section');
+
+    const subs = within(identitySection).queryAllByText('sub');
+
+    expect(subs).toHaveLength(0);
+  });
+
+  it('displays amount of subs', async () => {
```
krzysztof-jelski, line 45:
```suggestion
  it('displays count of subs', async () => {
```
Count or number, amount goes better with money or uncountable nouns.

krzysztof-jelski:
For some accounts, the layout breaks:
image

packages/page-accounts/src/Sidebar/SubAccountsModal.tsx
```diff
@@ -0,0 +1,69 @@
...
+  const { t } = useTranslation();
+
+  return (
+    <Modal
+      className={className}
+      header={t<string>('sub-identities')}
+      onClose={onClose}
+      size='tiny'
+    >
+      <Modal.Content className='content'>
```
krzysztof-jelski, line 28: It would be perfect to have just the content scrollable, with the header pinned to top. But perhaps that's a separate improvement to the new modal.

packages/page-accounts/src/Sidebar/Identity.tsx
MiZiet: ✅

packages/page-accounts/src/Sidebar/Sidebar.spec.tsx
MiZiet: ✅

packages/page-accounts/src/Sidebar/SubAccountsModal.tsx
MiZiet: I've set only content to be scrollable but only in this modal. But if it's needed we can make another pr with setting this to all modals.

MiZiet:
> For some accounts, the layout breaks:
> image

I've set greater width there, it should be ok now.

MiZiet:
Styled multisig section

ekowalsk:
💪  Looks very good

ekowalsk (approved):

MiZiet:
hi @jacogr, maybe you can give it a try and play with it?

this is how it looks like:
image
image

with current look we still have to click on dropdown to see all subs, which also causes some problem like jumping subs label or whole content jumping due to scrollbar appearing.

jacogr:
Have played through - the issue I have with it is that modal from modal is normally not a good practice. Sadly, it gets done in a couple (less than a handful) of places.

In this case, there is a work-around, so we don't need to wade into this approach. 

- we can expand sections
- we can scroll containers

ekowalsk (approved):

MiZiet:
@jacogr so i've removed the subs modal and added scroll:

image

What do you think?

jacogr:
The above is fine, but would make it an expander - specifically for situations like these where it is just, well, a lot.

Since we merged the address styling, some conflicts snuck in here.

MiZiet:
hi @jacogr, I've resolved conflicts and wrapped subs into an expander (when their number is greater than the set threshold).  
What do you think?

jacogr (approved):
Thank you!

PR #6113, opened 2021-09-12
fix vote input bug

awesomepan:
The default value is always changed,  the user cannot type the desired value


awesomepan:
@jacogr  Hello Jaco,  can you help review it, thank you.

PR #6116, opened 2021-09-13
Added Polkadot RPC endpoint, hosted by Dwellir

Maharacha:
Waiting for proposal.

### DIFF
--- a/packages/apps-config/src/endpoints/productionRelayPolkadot.ts
+++ b/packages/apps-config/src/endpoints/productionRelayPolkadot.ts
@@ -24,7 +24,7 @@ export function createPolkadot (t: TFunction): EndpointOption {
       OnFinality: 'wss://polkadot.api.onfinality.io/public-ws',
       'Patract Elara': 'wss://pub.elara.patract.io/polkadot',
       // 'Geometry Labs': 'wss://polkadot.geometry.io/websockets', // https://github.com/polkadot-js/apps/pull/6746
-      // Dwellir: 'wss://polkadot-rpc.dwellir.com',
+      Dwellir: 'wss://polkadot-rpc.dwellir.com',
       'light client': 'light://substrate-connect/polkadot'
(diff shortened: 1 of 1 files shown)


jacogr:
Sorry... this is the _completely_ wrong place to ask these things, however since my bots lock old PRs to prevent pesky necro-posting, please forgive me.

I've noticed the Dwellir RPCs for Kusama (unlike other providers) does not run in archive mode, i.e. trying to query old blocks fail since the state is discarded. Is this by design or an oversight?

Maharacha:
> Sorry... this is the _completely_ wrong place to ask these things, however since my bots lock old PRs to prevent pesky necro-posting, please forgive me.
> 
> I've noticed the Dwellir RPCs for Kusama (unlike other providers) does not run in archive mode, i.e. trying to query old blocks fail since the state is discarded. Is this by design or an oversight? (I suppose the question here is probably relevant before this PR does go in)

Our nodes was stuck syncing due to this bug https://github.com/paritytech/polkadot/issues/3778
The get them in sync we had to download a snapshot, which unfortunately is not an archive. A workaround might be to sync the node with an older version (0.9.8) before the bug was introduced, and then upgrade when the node is in sync. We have not tried this though, but can give it a shot if this is a big problem. Otherwise we will wait for the next release where they have fixed the bug.

jacogr:
Ahh, that makes sense. Perfect, really no need to downgrade/upgrade, the next release binary seems to be going out very soon.

Maharacha:
@jacogr Dwellir Polkadot RPC is ready to fly! Feel free to merge.

Maharacha:
> @jacogr Dwellir Polkadot RPC is ready to fly! Feel free to merge.

ping @jacogr

jacogr (approved):

jacogr (changes requested):
Needs resolving - I disabled one non-operational endpoint earlier.

Maharacha:
> Needs resolving - I disabled one non-operational endpoint earlier.

Rebased now!

jacogr (approved):

PR #6122, opened 2021-09-14
Apply collapsible row styling for addresses

ekowalsk:


packages/react-hooks/src/useCall.ts
```diff
@@ -29,7 +29,7 @@ interface QueryTrackFn {
   };
 }
 
-type TrackFn = RpcPromiseResult<AnyFunction> | QueryTrackFn;
+export type TrackFn = RpcPromiseResult<AnyFunction> | QueryTrackFn;
```
MiZiet, line 32: Can't find where do we use it

MiZiet (approved):
LGTM

jacogr (approved):

PR #6129, opened 2021-09-16
Restore slow tests

krzysztof-jelski:
Currently the slow (nightly) tests fail on first Typescript import statement (`import './Backend'` in `i18n`). Jest resolvers tries to fetch this module as `Backend.cjs` and obviously fails.

I confirmed it has nothing to do with what we do in the `globalSetup` file. Exporting an empty function gives the same result.

I didn't get to the bottom of the `jest` bug, but found a solution nonetheless. Changing module type from TypeScript to CommonJS for the `globalSetup` file apparently helps.

### DIFF
--- a/jest-slow.config.cjs
+++ b/jest-slow.config.cjs
@@ -6,6 +6,6 @@ const config = require('./jest.config.cjs');
 module.exports = {
   ...config,
   displayName: 'all-tests',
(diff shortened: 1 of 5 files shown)


jacogr (approved):
Thank you. Have been meaning to look for what feel like weeks now. 

The Jest ESM support is “iffy” at best. It will improve.

PR #6155, opened 2021-09-20
Apply the same filter and buttons style to Contacts as in Accounts

ekowalsk:
image

### DIFF
--- a/packages/react-components/src/FilterInput.tsx
+++ b/packages/react-components/src/FilterInput.tsx
@@ -1,0 +1,39 @@
+// Copyright 2017-2021 @polkadot/app-accounts authors & contributors
+// SPDX-License-Identifier: Apache-2.0
+
+import React from 'react';
+import styled from 'styled-components';
+
+import { Input } from '@polkadot/react-components/index';
+
+interface Props {
+  className?: string;
+  filterOn: string;
+  label: string;
+  setFilter: (filter: string) => void;
+}
+
+function Filter ({ className = '', filterOn, label, setFilter }: Props) {
+  return (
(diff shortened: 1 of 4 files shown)


packages/react-components/src/FilterInput.tsx
```diff
@@ -0,0 +1,43 @@
...
+interface Props {
+  className?: string;
+  filterOn: string;
+  label: string;
+  setFilter: (filter: string) => void;
+}
+
+function Filter ({ className = '', filterOn, label, setFilter }: Props) {
+  return (
+    <div className={`${className} filter--tags`}>
```
MiZiet, line 18:
```suggestion
    <div className={className}>
```
do we use `filter--tags` anywhere?

packages/react-components/src/FilterInput.tsx
```diff
@@ -0,0 +1,43 @@
...
+    :not(:only-child) {
+      margin-left: 1.5rem;
+    }
+
+    .ui--Input {
+      margin: 0;
+      height: 3.893rem;
+    }
+  }
+`);
```
MiZiet, line 43:
```suggestion
export default React.memo(styled(Filter)`
  width: 29.5rem;

  &:not(:only-child) {
    margin-left: 1.5rem;
  }

  .ui--Input {
    margin: 0;
    height: 3.893rem;
  }
`);

```

ekowalsk: ✅

MiZiet (approved):
Nice

jacogr (approved):
Nice. Apart from aligning, it removes real uglies in the code. Brilliant.

PR #6161, opened 2021-09-21
Update types for Realis.Network

Daelon02:


packages/apps-config/src/api/spec/realis.ts
```diff
@@ -30,9 +30,10 @@ const definitions: OverrideBundleDefinition = {
             'Diamond'
           ]
         },
+        String: 'Vec<u8>',
```
jacogr, line 33: Would suggest “Text” - it extends the JS string type.

Daelon02: In blockchain, we use String

jacogr: I understand. On the JS API the “Text” class maps to the Rust string type. 

(String is already a class in JS, hence the rename)

packages/apps-config/src/api/spec/realis.ts
```diff
@@ -32,7 +32,7 @@ const definitions: OverrideBundleDefinition = {
             'Diamond'
           ]
         },
-        String: 'Vec<u8>',
+        Text: 'Vec<u8>',
```
jacogr, line 35:
```suggestion
        String: 'Text',
```

The JS equivalent is called Text :)

Daelon02: Okey, we change

Daelon02: We cant re name this type, because UI throw waning about Metadata V14

Daelon02:
@jacogr Can you aprove please?

jacogr (changes requested):
Lint is still failing, so cannot approve/merge.

jacogr (approved):

PR #6208, opened 2021-09-27
Style badges as in the designs and add hovers

ekowalsk:
If applied, this will style badges in Accounts rows and add possibility to open modals from hovers 
image
image


jacogr:
Have not played as of yet, just not entirely happy with the icons not being in the same line, i.e. the fist row in the light theme design.

Apart from that, this should really give it some visual pleasantries.

ekowalsk:
The idea was, that in the first line we display badges which tooltips don't open any modal, and in the second line - badges with tooltips opening modals :)

MiZiet (approved):

jacogr (approved):
This seems spot-on.

PR #6220, opened 2021-09-30
Style redeemable balance in accounts and addresses

MiZiet:
If applied these changes will change size  of redeemable button and position redeemable amount properly when the button isn't displayed (in addresses).

image


ekowalsk (approved):

jacogr (approved):
Code looks 100% - have not verified the UI, but I trust and believe the screenshots... which, well, looks 100x better :)

Issue #6221, opened 2021-09-30
Adjust z-index for new modal & alerts


jacogr:
![image](https://user-images.githubusercontent.com/1424473/135468565-00195dd6-4abe-417d-94a4-ccb10d4d710e.png)

jacogr:
Closing. The new expandable status displays removes the need for it.

PR #6229, opened 2021-10-03
Adjust z-index for modal/status

jacogr:
Closes https://github.com/polkadot-js/apps/issues/6221


jacogr:
Doesn't work, parent can't break out of parent z-index.

PR #6244, opened 2021-10-06
Ignore newRaise

jacogr:


jacogr:
Invalid. Needs to be in `NewRaise` (if campaign not active, can `poke` it in)

PR #6247, opened 2021-10-06
Increase UI test timeouts and mark interBTC unreachable

krzysztof-jelski:
So here's another attempt to make the nightly tests pass 😄 

- I've marked interBTC unreachable as per #6230 
- I've increased the React Testing Library timeouts globally to 10 seconds. 5 seconds we had before was enough to make the test pass consistently on a laptop. Though it was a boundary value, eg. 4 seconds most often failed. As the GH actions tend to be slower than that, a global 10sec timeout for should help.

### DIFF
--- a/jest/jest-setup.ts
+++ b/jest/jest-setup.ts
@@ -3,9 +3,13 @@
 
 import '@testing-library/jest-dom';
 
+import { configure } from '@testing-library/dom';
+
 // eslint-disable-next-line @typescript-eslint/ban-ts-comment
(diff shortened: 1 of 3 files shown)


jacogr:
Can we remove interBTC for now - I need to just look at that one, I am worried they are on metadata v14 and something is going haywire.

krzysztof-jelski:
Oh, sure, no problem. Restored.

jacogr (approved):

PR #6249, opened 2021-10-06
new edgeware logo and color

Depth-Hoar:
It would be great if someone could explain how to update the beresheet testnet colors as well.


jacogr:
For a chain (based on name, not node or spec) definition, add an entry to the `chainColors` with the name -> color mapping.

PR #6463, opened 2021-11-02
re-enable dock testnet endpoint

SamHellawell:
node address was changed in previous PR, just realised its still set to disabled


jacogr (approved):

jacogr:
This doesn't have a DNS entry? - https://github.com/polkadot-js/apps/issues/6469

Going to disable until deployed correctly.

SamHellawell:
there was some issues with dns on that server, i will update to point to a more stable testnet node (we often do dev on other one, so i believe thats why it went down - it was working). sorry for this!

PR #6464, opened 2021-11-03
add parallel chain for dot auction

GopherJ:
Hi, we will participate the first batch of polkadot's auction. cc @yubo-ruan

image


jacogr (approved):
Brilliant, thank you!

PR #6473, opened 2021-11-03
Litentry parachain westend

chenzongxiong:
Add Litentry Parachain info to Westend. 

We are going to launch parachain for Westend within 6 hours. Can you merge and deploy to live production once you approved. 

Thanks in advance.


jacogr:
For some reason a yarn.lock file was included that was blindly upgraded. This broke the apps build as CI indicates. Not sure where this comes from, but cannot be merged in this state (apart from CI failing, which means a merge cannot be deployed).

jacogr (changes requested):
As commented. The code changes are fine, unfortunately the `yarn.lock` updates here are breaking the build (generally we are very specific on these via `yarn upgrade-interactive` and the local testing with the updated packages, i.e. where they affect, to not create issues)

PR #6506, opened 2021-11-06
fix crust para id

pangwa:
2002 is reserved by Clover Finance. Crust's para id is 2008


mmyyrroonn:
@pangwa Thank you! Crust ID was added by me in this PR. https://github.com/polkadot-js/apps/pull/6484. Our reserved para ID is 2008 right now. Could you please help us merge this PR to fix the issue?

Issue #4931, opened 2021-03-25
Warning for ED on accounts with enough total balance


michalisFr:
It seems the UI doesn't check the total balance of the account when checking for the Existential Deposit, but instead it checks only the transferable balance. 
![Screenshot 2021-03-25 at 2 38 21 PM](https://user-images.githubusercontent.com/45905709/112476674-d4c2e100-8d7a-11eb-9cc0-da0949762f94.png)

In the screenshot, this is my stash account, so there are bonded funds. So, although the message is correct (in the sense that the account can't be reaped) there's no risk of that, since there's still balance in it.

Similarly, if the transferable balance is below the ED (or is going to be after the tx) this message is displayed:
![Screenshot 2021-03-25 at 3 02 38 PM](https://user-images.githubusercontent.com/45905709/112477645-e658b880-8d7b-11eb-944b-35dc273530f2.png)
But of course the transaction will not fail. 

This creates two issues:
1. Users are getting confused and think they can't send funds
2. A user can't send all the transferable out unless they subtracting the fees manually

jacogr:
- the reference check is, well, on references - not balance (screenshot 1)
- You cannot transfer all the transferrable with keep-alive (screenshot 2)

michalisFr:
> the reference check is, well, on references - not balance (screenshot 1)

Well, yes of course, but you can still send the total transferable balance without the account getting reaped. I propose two changes:
1. Display the "transfer full account balance" button but rephrase it to:
"Transfer the full _transferable_ balance" without `reap the sender`
2. Change the message to "As such the account _will not_ be reaped from the state"

> You cannot transfer all the transferrable with keep-alive (screenshot 2)

True, but nothing changes if you disable "keep alive". You simply get both warnings:
![Screenshot 2021-03-26 at 10 41 51 AM](https://user-images.githubusercontent.com/45905709/112605386-eca07080-8e1f-11eb-8369-8c910af42097.png)

In this case the second warning is incorrect, especially the part where it states "As such the transfer will fail. The account needs more free funds to cover the transaction fees."

jacogr:
The second is not incorrect - if the application of fees drops the balance below ED, it fails. (Keep-alive, non-keep-alive, you always need ED + fees available to transact)

michalisFr:
Indeed, but that's true for the total balance, while the UI seems to be checking the transferable balance.

[...]

PR #6518, opened 2021-11-08
Improving UX when making balance transfers

DrW3RK:
Although the front end can stop the user from sending a transaction by checking if the balance transfer will kill the account, it sends it to the network! 

Here is an instance that shows how detrimental it is for UX - https://polkadot.subscan.io/account/15vzAirFFFWcEgF74jxLoCTQxh11caDYgBrdQpUAnSEaDkrz

With some additional code logic, this can be averted. Also, the corner case for the second issue mentioned here is not handled well - https://github.com/polkadot-js/apps/issues/4931

This PR will take some time to finish. Need to understand how things work, before any changes are pushed to the PR.


DrW3RK:
No time to follow through this effort. fail!

PR #6576, opened 2021-11-19
Update moonbeam-types-bundle to 2.0.1

joelamouche:


jacogr (approved):
Thank you. Sorry for the run-about on the deps, just had real horrors in the past so gun-shy without manual testing.

PR #6616, opened 2021-11-29
feat(gamedao): update types

2075:
hi there,
we have updated our types again :)
thanks and greetings!


jacogr:
Cannot be merged, CI fails.

2075:
hm, just one version bump, seems more like the ci just aped without reason?

jacogr:
No, CI is correct. The lockfile is not updated on CI since we need to ensure all updates are checked. Hence when when making a PR on bumps, the lockfile needs to be committed. This generally means -

1. it has been tested locally
2. we can see which packages are added
3. can actually do an in-depth check on these if needed

jacogr (approved):

PR #6651, opened 2021-12-09
Subspace Updates - Fix block author from derive

1devNdogs:
# Summary

Subspace spec file changed to support block authoring from our custom header. Adding author from solution section contained in our preDigest logs.

Thanks!


jacogr (approved):
Thank you.

PR #6657, opened 2021-12-10
Add storage deposit support in apps

statictype:
Introduced in Substrate [here](https://github.com/paritytech/substrate/pull/10082)

Done:
- storageDepositLimit = null
- rename endowment to value
- update input info text
- make value input zeroable where storageDepositLimit is present (updated nodes)


packages/page-contracts/src/Codes/Upload.tsx
```diff
@@ -83,6 +84,18 @@ function Upload ({ onClose }: Props): React.ReactElement {
     abiName && setName(abiName);
   }, [abiName, setName]);
 
+  useEffect((): void => {
+    accountId && hasStorageDeposit &&
+      api.derive.balances
+        .account(accountId)
+        .then(({ freeBalance }) => {
+          // instantiation fails when using the entire free balance
+          const limit = freeBalance.sub(freeBalance.div(new BN(10)));
```
athei, line 92: You need to substract `value` and the existential deposit here. So that the sending account stays above the existential deposit even after sending the `value` and using all of the storage deposit.

statictype:
instantiation works now on nodes with or without the latest changes, as far as I tested

i can't get `call` to work though, i keep getting this error

Screenshot 2021-12-12 at 23 08 02

Screenshot 2021-12-12 at 23 02 54

jacogr:
The above one is problematic - the RPC results changed. The API have a `__fallback` specifier that may cater for this case. But basically the error suggests it cannot decode a number (I would guess... the bn.js errors def. have something exactly like that)

packages/page-contracts/src/Codes/Upload.tsx
```diff
@@ -202,7 +202,7 @@ function Upload ({ onClose }: Props): React.ReactElement {
             <InputBalance
               help={t<string>('The balance to transfer from the `origin` to the newly created contract.')}
               isError={!isValueValid}
-              isZeroable={hasStorageDeposit}
+              isZeroable={contractAbi.constructors[constructorIndex].isPayable}
```
athei, line 204: What does `isZeroable` do if true?

statictype: it makes the input validation fail. the input gets a red border and you can't move to the next step

statictype: now i tried with the older node and even if the constructor is not payable, the transaction fails with "contract not founded" if i pass 0.
is is safe to assume that if the node supports storage deposit than the value can be 0 for non payable constructors?

athei: You have to differentiate between ABI (payable) and the node. ABI comes form ink! and is decoupled from the node version.

Old Node:
Always needs `value/endowment` in order to pay for rent. Disregard `payable`.

New Node:
Does not need any `value` **except** the constructor is `payable`. Hence you should not even show the field and use `0` when the constructor isn't payable.

ink!:
Ink constructors are implicitly **payable***. The fuck up is that this is implicit. The field is missing from ABI. What you should do is interpete the absense of the payable field as `payable = true`. We intent to make it explicit and also allow to make them not payable. The following is forward compatible:
`undefined` -> true
`false`-> false
`true` -> true


The

statictype: ok, got it! 
i am using [this](https://github.com/polkadot-js/apps/blob/b3ad01d90dd27ba0d1b3a81fc07f40f6e6aa8566/packages/page-contracts/src/Contracts/Call.tsx#L49) to check if the node has storage deposit or not. because the changes to value are related to storage deposit i wanted to double check it the same condition can be used

statictype:
@jacogr this is ready for review, i think. is there something else we can do to get this available to the users faster?

jacogr:
I'll get through the queue - I know I'm a bottlenck, it takes a while to get to stuff in the review queue. (In apps I currently have a large number of PRs > 1 month old which has not had a look, the queue should be at 0 and it runs longer than a page here, just because of not getting there. It has been quite neglected due to pressures elsewhere.)

statictype:
whenever you have some time. build should pass now.

jacogr:
On the list for today.

yarn.lock
```diff
@@ -2568,22 +2568,52 @@ __metadata:
...
-    "@polkadot/types-create": 7.0.2
+    "@polkadot/rpc-core": 7.0.3-1
+    "@polkadot/types": 7.0.3-1
+    "@polkadot/util": ^8.2.2
+    rxjs: ^7.4.0
+  checksum: 8e6d7264d89650b31e88c778b6ac3cbd979d9e0db2973d7adad01982012e6e493766575a422f6c16ca4ea07a03302b5e86ba3541c95f74ae791f039b4d903f53
+  languageName: node
+  linkType: hard
+
+"@polkadot/api-contract@npm:^7.0.3-0":
```
jacogr, line 2584: Somewhere things got mismatched between the versions here - the `./dev/all-update.sh` script will ensure there is a single version and dedupe it as well. (The lockfile here is weird with the `-0` & `-1` matching)

jacogr (commented):
lgtm code-wise, just some small comments.

The bumps here are weird, was it done manually?

packages/page-contracts/src/Contracts/Deploy.tsx
```diff
@@ -50,6 +51,7 @@ function Deploy ({ codeHash, constructorIndex = 0, onClose, setConstructorIndex
 
   const [name, isNameValid, setName] = useNonEmptyString(code && code.json.name);
   const { contractAbi, errorText, isAbiError, isAbiSupplied, isAbiValid, onChangeAbi, onRemoveAbi } = useAbi([code && code.json.abi, code && code.contractAbi], codeHash, true);
+  const isConstructorPayable = contractAbi?.constructors[constructorIndex].isPayable !== false;
```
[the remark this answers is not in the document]
statictype: the intention is to make the constructors explicitly payable / not payable in ink! 
just trying to be forward compatible. but yes, it's an unnecessary check at this point.

yarn.lock
statictype: what i did is to bump the versions in `package.json` and then run `yarn`

packages/page-contracts/src/Contracts/Deploy.tsx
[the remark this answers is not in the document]
jacogr: I understand that, but it is actually not forward-compatible at all :) So, if we want to really do this - pass the flag here https://github.com/polkadot-js/api/blob/master/packages/api-contract/src/Abi/index.ts#L93 (like we do for messages)

For now it would need to always be true. When the time comes that the constructors pass it, on the API level it would need to change into an `Option<bool>` for that fields where `isPayable: o.isNone || o.unwrap().isTrue` (to maintain backwards compat)

So not against the checks, however the tristate logic should not sit quite here. It is ok for now, but needs cleanup on the ABI level and then here.

yarn.lock
jacogr: Ahh, yes, that is the issue - they are in multiple places in the root and in the workspace level (not actually enough, most pages do still pull in transient deps).

jacogr:
Just merge in master and we should be good to go :)

jacogr (approved):
Thank you very much, this is great!

PR #6681, opened 2021-12-15
Update Rococo parachain URLs and enable teleport

PierreBesson:
This should also let users teleport ROC to Statemint/Canvas. See for example, ROC to Canvas teleports as described in the [Canvas README](https://github.com/paritytech/canvas#rococo-deployment).

However in my tests, my changes fail to bring up the Teleport tab for the Rococo network.


jacogr:
Ummm... the config looks correct to enable bi-directional teleports, so not sure why it doesn't show for you.

jacogr:
Ahhh... if Rococo was reset once again, the genesis hash needs an update. It matches to a chain on genesis.

PierreBesson:
Thanks @jacogr, I updated the Rococo genesis hash and not the teleport tab is available.

jacogr (approved):
Thank you!

PR #6713, opened 2021-12-20
Add preruntime digest

joelamouche:
Moonbeam's new runtime (1100) now uses a different digest to get the block author


jacogr (approved):
lgtm with no changes to non-Moonbeam nodes.

PR #6726, opened 2021-12-22
Remove tick trick track parachains

BulatSaif:
On 22.12.2022 rococo network was reset.  Only rococoStatemint and rococoCanvas will be present after reset.


jacogr (approved):

jacogr (approved):
Thank you. 

The genesisHash also need updating to ensure that teleports work (and that the registered chains match up and names are displayed)

PR #6727, opened 2021-12-23
Grammar fix in staking tooltip

tarrball:
Grammar fix in staking tooltip and associated translations.

An alternative fix for this could be: "..or because they willingly stop validating" (with "have" removed).


jacogr (approved):

jacogr:
Thank you!

PR #6774, opened 2022-01-03
Reactivated Dwellir Kusama RPC

Maharacha:
* We have scaled up the service to deal with the workload.
* We have improved and tuned our monitoring so that we will be alarmed if we need to scale up.


Maharacha:
@jacogr merge?

jacogr (approved):
Can merge now - I didn't want this to go into a release before being run for a while. 

The constant issues around unavailable RPCs means that if something unstable goes into a release it basically breaks the IPFS version until the next release cycle. Especially for Polkadot/Kusama really gun-shy now, there has been a lot of churn.

PR #6781, opened 2022-01-04
Clarify Teleport transaction fee message

DrW3RK:
As the destination transfer fee estimate is removed in https://github.com/polkadot-js/apps/pull/6745/ , it is resourceful to update the message mentioning that the teleported amount is subject to destination transfer fees along with the transfer fees at the source (and the balance should be above Existential Deposit, if the recipient's account is new


jacogr (approved):
Thank you.

PR #6911, opened 2022-02-01
Allow to upload a JSON file with the list of multisig signatories

jsidorenko:
This PR adds a new file upload field, so it could be possible to upload a list of multisig signatories

image

When the file gets uploaded it hides the current accounts selection row and displays found addresses in the provided JSON file.

image

### DIFF
--- a/packages/page-accounts/src/modals/MultisigCreate.tsx
+++ b/packages/page-accounts/src/modals/MultisigCreate.tsx
@@ -6,10 +6,22 @@ import type { ModalProps } from '../types';
 
 import React, { useCallback, useState } from 'react';
 
-import { Button, Input, InputAddressMulti, InputNumber, Modal } from '@polkadot/react-components';
+import {
+  AddressMini,
+  Button,
+  IconLink,
+  Input,
+  InputAddressMulti,
+  InputFile,
+  InputNumber,
+  Labelled,
+  MarkError,
+  Modal,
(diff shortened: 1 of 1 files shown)


shawntabrizi:
Would love to see a screenshot of error handling

jsidorenko:
Sure thing!
Wrong file structure:
image

Wrong address:
image

Too many addresses provided:
image

jacogr (commented):
I actually like it. Code looks great.

I have one suggestion that we rather have a toggle to go between selection and upload. It is a bit weird now where it hides the one in the upload direction, but not the other way around.

So I guess what I'm saying is that we provide 2 options here, which is great, but can already see that people select and then ask support "where do I get the file", assuming it is an "and", not an "or". A toggle between the two would make it explicit that it is an "or" operation.

jsidorenko:
Yeah, I was thinking about the toggle as well :) another idea was to put the text "OR" between the top accounts selector and the file upload field.

Where would you like to have that toggle? On top of the form or below the accounts selector?

shawntabrizi:
Look at an example of a toggle here:

image

I would place the toggle at the top right (middle) of the UX, and have that change the underlying field boxes.

jacogr (approved):
Thank you!

shawntabrizi:
👏 @jsidorenko

PR #6913, opened 2022-02-01
change width of line below the tab (active and hover)

xnorly:
The line below the Tab (both active and hover)) should have same width as the tab, otherwise it looks and feels too short.

### DIFF
--- a/packages/react-components/src/Tabs/Tab.tsx
+++ b/packages/react-components/src/Tabs/Tab.tsx
@@ -70,7 +70,7 @@ export default React.memo(styled(Tab)`
     &.tabLinkActive .tabLinkText::after {
       content: '';
       position: absolute;
-      width: 3.14rem;
+      width: 100%;
       height: 2px;
       bottom: -2px;
       left: 50%;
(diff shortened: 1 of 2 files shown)


xnorly:
# Screenshots:

### Old version:
Screenshot 2022-02-02 at 1 30 52 AM

### New Version:
Screenshot 2022-02-02 at 1 30 33 AM

jacogr:
So it basically started off as you have it here, then evolved to being exactly the length of the text (which is my preference, aka like GH mostly does it as well), then to the current shorter version as part of a partial UI redesign (where an actual UX designer looked at _some_ elements... the apps UI is not great on UI overall)

So basically, with this PR it is full circle - and it really depends on who likes what. As they say, "Beauty is in the eye of the beholder".

With that in mind, while appreciative for all code contributed, I'm not going to merge this.

xnorly:
Oh, sorry, I haven't checked out the full history of this.
Thank you for the info!

jacogr:
I really do appreciate any and all PRs, especially if they improve the experience - they make my day :) So I hate saying "no".

Issue #6828, opened 2022-01-14
Invalid Threshold for 2 members Council for external proposal


crystalin:
Usually the external proposal button is able to guess the right amount of threshold for the council vote.
However when the number in the council is 2, it fails to send the proposal as it puts a 1 threshold instead of 2 in my case.

jacogr:
It is actually incorrect for all council with even numbers, i.e. threasholds are always calculated on `>=` instead of `>` or `>=`. In the case of odd number of council, this always works correctly, in the case of even number of council members, this doesn't.

wilwade:
Looks like the recent release made this much more important. The vote button is now removed when then threshold is reached. So it is not easy to get to where you can vote 50%+1.

PR #4588, opened 2021-02-03
Create use thresholds hook and use correct thresholds for reject origin

ekowalsk:

PR #6924, opened 2022-02-04
Ensure evenly divisible thresholds are adjusted

jacogr:
Closes https://github.com/polkadot-js/apps/issues/6828
Closes https://github.com/polkadot-js/apps/pull/4588

The correct solution is probably to carry the `>` and `>=` but it certainly makes things a lot more complex. In this implementation we actually DRY things a lot. So could be re-visited possibly if really needed. (The hook in the linked PR could be interesting, instead of `useMemo` everywhere.


h4x3rotab:
A quick fix in #9419. Instead of using `ceil`, it should be `floor() + 1`.

h4x3rotab:
If you want to find a chain with even council seats, please try Phala (it has 8 members):

```
wss://phala.api.onfinality.io/public-ws
```

packages/apps-config/src/api/params/proposalThresholds.ts
```diff
@@ -52,3 +54,20 @@ export function getSlashProposalThreshold (api: ApiPromise): number {
...
+      members.length,
+      ceil + (
+        // for evenly divisible, adjust upwards
+        Math.floor(frac) === ceil
+          ? 1
+          : 0
+      )
+    )
+  );
+}
```
h4x3rotab, line 73: The function is identical to:

```javascript
function calcThreshold (members: unknown[], threshold: number): BN {
  return new BN(
    Math.min(
      members.length,
      Math.floor(members.length * threshold) + 1
    )
  );
}
```

ap211unitech:
Closing this PR as it appears to be outdated. If the problem persists, a new PR can be opened with updated details.

PR #6929, opened 2022-02-06
[WIP] unique support

### LINKED ISSUE #5397: Support uniques pallet
As per https://github.com/paritytech/substrate/pull/8813

gilescope:
Will fix #5397

So far: uniques overview seems to work well including creating and minting uniques.
Possibly setting the data is not - need to recheck that...

Might need some pointers on userInstances.ts if anyone has a spare moment?

If I get instances tab working, feels to me like a 'my uniques' tab is also needed that shows uniques associated with a particular account.


jacogr:
Going to close this on the back of the PR that went in in https://github.com/polkadot-js/apps/pull/7144 since I'm not convinced it can be merged in as it stands (since the full base is duplicated).

However, you have some great features in here that is not available in the linked PR, e.g. minting so there is quite a bit of effort remaining to make it fully featured. Those should be translatable to the master version.

PR #6939, opened 2022-02-07
Encointer teleporter kusama

brenzi:
since runtime 9016, Encointer is a trusted teleporter. This PR adds UI functionality for this

https://github.com/paritytech/polkadot/blob/41ab002d7451766324a9f314fee11c9c53314350/runtime/kusama/src/xcm_config.rs#L106-L112

### DIFF
--- a/packages/apps-config/src/endpoints/productionRelayKusama.ts
+++ b/packages/apps-config/src/endpoints/productionRelayKusama.ts
@@ -27,7 +27,7 @@ export function createKusama (t: TFunction): EndpointOption {
       'light client': 'light://substrate-connect/kusama'
       // Pinknode: 'wss://rpc.pinknode.io/kusama/explorer' // https://github.com/polkadot-js/apps/issues/5721
     },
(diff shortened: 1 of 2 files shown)


gilescope:
I just tried teleporting some assets on westend to encointer and they did not come out the other end. Does it work on westend before enabling it on kusama?

jacogr:
Although this is slightly off-topic for the PR, but somehow related... 

Yeap, I had the same experience on Westend. I was eagerly awaiting the 16 upgrade to test and gave it a go as that went up. I would certainly want to have the gremlins sorted out on Westend before bringing this into Kusama, don't want to cost people tokens.

I have not looked at the events, but my guess it there is either a versioning or config error somewhere in the pile.

brenzi:
Joe was able to teleport to encointer, but only using `V1`. My attempt to use `V0` failed. can we make js/apps use `V1` by default?

jacogr (approved):

PR #6958, opened 2022-02-09
substrate connect - upgrade to 0.5.0

wirednkod:
Latest version of Substrate-connect (0.5.0) has some several changes to the public API and exposed classes.
The changes on the integration with PolkadotJS Apps is fairly easy and straightforward.


jacogr (commented):
One tiny OCD-related comment. 

(It actually took ages to get the type vs normal imports right, well, it is correct AFAIK, but I'm probably guessing 99.5%, not sure where the 0.5% sits...)

jacogr (approved):

PR #6973, opened 2022-02-11
Fix contract trapped

statictype:
Constructors are not payable by default so the instantiation was failing with `ContractTrapped` when passing any value (endowment)
Fixes https://substrate.stackexchange.com/questions/120/polkadot-canvas-testnet-contract-deploy-error-contracts-contracttrapped

The instantiation still fails with `NewContractNotFunded` because Canvas needs an update (coming soon)
Not sure this PR is needed though, the error might be fixed with the Canvas update.


jacogr (approved):
Still believe it is needed, it shouldn't attempt to pass values through.

Thank you.

Issue #6922, opened 2022-02-04
Swap teleport to using V2 XCM


jacogr:
It is currently still using V0 which is to be deprecated soon (and V1 will follow afterwards)

joepetrowski:
I'm not entirely sure of the naming conventions, but there seem to be different versions for _messages_ and certain _data structures._ For example, this is a V2 `Instruction` but V1 is the latest version of `MultiLocation`. We know that V2 messages using V1 assets/locations are failing _now_ (not to be deprecated soon).

![image](https://user-images.githubusercontent.com/25483142/155299895-17dc5fb8-6f79-48ac-8880-1bdc027d2b55.png)

jacogr:
The naming of the types are from the metadata. V2 re-uses the V1 types, hence the module it is placed in is the same and reflects in the naming inside the metadata, aka the actual type path (which is used to construct the names).

PR #7025, opened 2022-02-23
Only use current-gen limitedTeleportAssets

jacogr:
Cleans up all the previous-runtime-detection magic that would make things like https://github.com/polkadot-js/apps/issues/6922 easier


jacogr:
Tested on Kusama <-> Statemine

PR #7032, opened 2022-02-24
Add OnFinality endpoint for Crust and make it reachable

banananeko:


jacogr (changes requested):
Is this connected as a parachain to the Polkadot relay?

It has produced 3M blocks already on both RPCs. The first parachain that went in is on 700k blocks and is actually in the Polkadot set.

These are only for parachains that are connected to the relay.

packages/apps-config/src/endpoints/productionRelayPolkadot.ts
```diff
@@ -128,16 +128,6 @@ export function createPolkadot (t: TFunction): EndpointOption {
           Composable: 'wss://rpc.composable.finance'
         }
       },
-      {
-        info: 'crustParachain',
-        homepage: 'https://crust.network',
-        paraId: 2008,
-        isUnreachable: true,
-        text: t('rpc.polkadot.crust', 'Crust', { ns: 'apps-config' }),
```
jacogr, line 136: By removing this, it won't show up here -

image

I'm not sure that is intended.

jacogr (changes requested):
As commented, I would not remove the parachain config since that has a negative impact on mapping the paraId to your name which means it won't show up as a named parathread nor in the crowdloans. 

Additionally the `isDisabled` flag should be removed from `endpoints/production` to make it active.

banananeko:
Updated. Thank you so much for your very thorough explanation!

jacogr (approved):

PR #7083, opened 2022-03-01
Add dwellir rpc provider to crab parachain on kusama

Ank4n:


jacogr (changes requested):
The live endpoint cannot be the same as the parachain endpoint, they are different chains - https://github.com/polkadot-js/apps/runs/5377798108?check_suite_focus=true#step:3:763

PR #7103, opened 2022-03-03
[encointer] remove westend

mosonyi:
We have stopped our westend ( api.westend.encointer.org ), therefore deleted from config.

### DIFF
--- a/packages/apps-config/src/endpoints/testingRelayWestend.ts
+++ b/packages/apps-config/src/endpoints/testingRelayWestend.ts
@@ -43,15 +43,6 @@ export function createWestend (t: TFunction): EndpointOption {
         },
         teleport: [-1]
       },
-      {
-        info: 'encointer',
-        paraId: 1001,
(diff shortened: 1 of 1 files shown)


jacogr (approved):
A huge thanks for this PR, generally I need to hunt and pack these stoppages via issues that are created.

Having said that, as a common-good I don't see this as a good move - you are basically removing the only test avenue available. (For instance for teleports, it was tested on this avenue before merging.) Bear this in mind for future changes (e.g. XCM v3 is on the horizon), when it comes to real funds I generally don't play cowboy unless verified, so some features may be missing in the future where I cannot verify.

jacogr (changes requested):
Tests most probably fail since the teleport destination is not removed for the relay, i.e. `[1000, 1001]` for the relay needs to change to `[1000]` only.

jacogr (approved):

brenzi:
@jacogr we will of course continue to test, but we were encouraged to move to rococo for future testing. So we didn't want infrastructure to idle around and shut down our westend infra.

PR #7121, opened 2022-03-08
fix pangolin and pangoro account error

sxlwar:
Hmm, this pr looks ugly. What is the suggested way to fix this issue?

### DIFF
--- a/packages/page-accounts/src/Accounts/Account.tsx
+++ b/packages/page-accounts/src/Accounts/Account.tsx
@@ -131,8 +131,12 @@ function Account ({ account: { address, meta }, className = '', delegation, filt
 
   useEffect((): void => {
     if (balancesAll) {
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access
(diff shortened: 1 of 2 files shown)


packages/page-accounts/src/Accounts/Account.tsx
```diff
@@ -131,8 +131,12 @@ function Account ({ account: { address, meta }, className = '', delegation, filt
 
   useEffect((): void => {
     if (balancesAll) {
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access
+      const bonded = stakingInfo?.stakingLedger.active || (stakingInfo?.stakingLedger as unknown as any)?.activeRing;
```
jacogr, line 135: Really don't want `activeRing` in here - it is not part of the Substrate base modules. (aka it is non-maintainable)

jacogr, line 135: Really don't want `activeRing` in here - it is not part of the Substrate base modules. (aka it is non-maintainable)

sxlwar: Yeah, we have fixed it on the upstream.

Does this mean that the networks in Polkadot Apps must provide the same API as the substrate base module?

Issue #6918, opened 2022-02-03
Feature request: Display information with regard to bags list nomination limits


michalisFr:
With the coming of bags list to staking in 0.9160 there are two limits with regard to staking that will be essential for users to see:

1. What is the number of nominators that will be selected from the sorted list to be applied to the election. This number is dynamic and will be:
`VoterSnapshotPerBlock - CounterForValidators`
`VoterSnapshotPerBlock` is the max number of voters, that includes validators, and is currently 22,500

2. Lowest stake in order for your nominations to be applied. It would be the stake of the last account in the ordered list.

Relevant issue for reference:
https://github.com/polkadot-js/apps/issues/6460

kianenigma:
> 2. Lowest stake in order for your nominations to be applied. It would be the stake of the last account in the ordered list.

This is already displayed in `Staking -> Targets -> min-nominated`

kianenigma:
All in all though, I think the `stats` section can use a small revamp. 

Let me break it down like this. We have 2 user groups, validators and nominators. Each have 2 variables: 

1. count, which has active and waiting sub-components. 
2. stake, which has minimum and threshold sub-components. 

With the exception that the validator group has two minimums: 
  - minimum self-stake among validators
  - minimum total stake amount validators

This leaves us with the following 4 boxes to show: 

1. nominator stake: minimum + threshold.
2. nominator count: active + total.
4. validator stake: self minimum + total minimum + threshold.
5. validator count: active + total.

This is how you calculate all of the 4 bits of information needed here:

```typescript 
const b = (x: BN): string => api.createType('Balance', x).toHuman()

// a map from all nominators to their total stake.
const assignments: Map<AccountId32, BN> = new Map();
const currentEra = (await api.query.staking.currentEra()).unwrap();
const stakers = await api.query.staking.erasStakers.entries(currentEra);
stakers.map((x) => x[1].others).flat(1).forEach((x) => {
	// @ts-ignore
	const nominator: AccountId32 = x.who;
	// @ts-ignore
	const amount: Compact<u128> = x.value;
	if (assignments.get(nominator)) {
		assignments.set(nominator, amount.toBn().add(assignments.get(nominator)!))
	} else {
		assignments.set(nominator, amount.toBn())
	}
})

// nominator stake
{
	const threshold = await api.query.staking.minNominatorBond();
	const stakes = Array.from(assignments.values());
	stakes.sort((a, b) => a.cmp(b));
	const min = stakes[0];
[...]

PR #7179, opened 2022-03-17
Nik fix staking numbers

wirednkod:


packages/page-staking/src/Overview/Summary.tsx
```diff
@@ -21,49 +20,19 @@ interface Props {
...
 
   return (
     <SummaryBox className={`${className}${!isVisible ? ' staking--hidden' : ''}`}>
       <section>
-        <CardSummary label={t<string>('validators')}>
-          {stakingOverview
-            ? <>{formatNumber(stakingOverview.validators.length)}&nbsp;/&nbsp;{formatNumber(stakingOverview.validatorCount)}</>
-            : <Spinner noLabel />
-          }
+        <CardSummary label={<>{t<string>('nominators')}<div>{t<string>('min active stake')}</div></>}>
```
jacogr, line 29: The issue with this is that it just caters for latest Polkadot/Kusama. This means it will break for other chains. There needs to be previous and current compatibility. (Otherwise my support queue fills up)

jacogr: … not all chains supported have latest and not all chains will swap to bags for some time.

wirednkod: Thank you @jaco for the response.

Is the best approach to conditionally choose old/new code based on chain (if polkadot or kusama then new else old) or based on pallette availability?
Or comment suggests that this should not happen (at least at the moment)?

jacogr: No, we don't choose path based on the chain, that would be an absolute maintenance nightmare. Rather we detect based on the functionality exposed in the runtime. e.g. if we choose to detect based on `bagsList` something like -

`api.consts.bagList.bagThresholds ? <new stuff here, we have bags> : <old stuff here, no bags>`

kianenigma: I think the title of the PR and the issue to which it links to is a bit misleading: this in itself is not adding any functionality that needs the existence of the bags-list. It merely shows who are the least staked entities in `staking.ErasStakers`, which only depends on the existence of the staking pallet.

packages/page-staking/src/useSortedTargets.ts
```diff
@@ -244,6 +247,28 @@ function extractBaseInfo (api: ApiPromise, allAccounts: string[], electedDerive:
   };
 }
 
+const b = (x: BN, api: ApiPromise): string => api.createType('Balance', x).toHuman();
```
kianenigma, line 250: I think there's already a `formatBalance` or something in the `/api/utils` that might be more ergonomic to use in this repo.

packages/page-staking/src/useSortedTargets.ts
```diff
@@ -244,6 +247,28 @@ function extractBaseInfo (api: ApiPromise, allAccounts: string[], electedDerive:
...
+
+function getMinActiveThreshold (api: ApiPromise, stakers: [StorageKey<[u32, AccountId32]>, Codec][]) {
```
kianenigma, line 252: this function seems correct to me, then my only guess is that `stakers` is not what we expect.

wirednkod: I think I understood my mistake. Will correct ASAP

wirednkod: Apparently I was mistaken - code is as in `polkadot-scripts`. Im not sure why the active min stake of nominators is so low.

jacogr (commented):
I'm not sure I "get" it. The numbers I would love to see on the overview page (number of actives) are now missing and nowhere to be found. (These numbers are on the targets page as well, with averages)

So while I really appreciate it, and certainly see a need to improve the displays, I'm not sure this is it since it (a) duplicates already available information & (b) removes other useful information.

Critically, there is a performance issue here, it burns the CPU, eating up one of my cores.

kianenigma:
> I'm not sure I "get" it. The numbers I would love to see on the overview page (number of actives) are now missing and nowhere to be found. (These numbers are on the targets page as well, with averages)

> So while I really appreciate it, and certainly see a need to improve the displays, I'm not sure this is it since it (a) duplicates already available information & (b) removes other useful information.

I disagree: The numbers that most users want to see is the answer to: 

1. How much stake I need to be an active nominator?
2. How much stake I need to be an active validator?

The count of nominators is mostly a metric that you, I and other devs are interested in, and while they should exist, they need not be in the main page. 

Our plan is to: 

1. Remove the scattered information that was hard to find before (which is why you see some stuff removed here).
2. Add these two in a smaller PR for ease of merging. 
4. Add the full matrix of information (all counters and all stakes, as [illustrated here](https://github.com/polkadot-js/apps/issues/6918#issuecomment-1068412537)) as a drop-down (or a new tab, both fine for me) that users can optionally open. Fetching all of this information is rather expensive and should be optional.


> Critically, there is a performance issue here, it burns the CPU, eating up one of my cores.

Both of the numbers that are displayed as a part of this PR can be 100% deduced from the list of validators and their backers (`ErasStakers`) which is already being fetched in this page. So while the code might need more care regarding re-using data, in principle, this has zero overhead.

wirednkod:
> Critically, there is a performance issue here, it burns the CPU, eating up one of my cores.

I think I know where the performance issue may come from. I will fix this.

jacogr:
I'm sorry to disagree here - the info for min stake is there, explicitly. It may make sense to also display it on the "actions" page, but sadly as it stands we lost a whole bunch and duplicated.

jacogr:
Thinking out, may be are trying to fit a square peg into a round hole. What about the following approach -

1. add a "stats-only" landing, that has all the info that is useful for both power and normal users, basically a big dashboard and dashboard only. (Some info will be there immediately, since query some can only be filled when the background stuff is fully loaded)
2. Rename "Overview" to "Validators" and this new page to "Overview"

This seems better than trying to shoehorn everything into the single top-bar?

jacogr:
There is no duplicates -

- the way staking is structured, for the bulk info we have a hook and pass it into each component - so they are not duplicated
- on the API layer itself, all calls are memoized, so the there are no duplicates made

wirednkod:
> @wirednkod has there been any further updates on this? WDYT?

(yeap :) - Just delay due to COVID)

I have 2 approaches for the `Overview` page and I would like some opinion design-wise @jacogr @kianenigma 
a) to follow same design as `Summary` - just bring everything in front page and organise them based on "group" (e.g. Validators, Nominators, Targets etc) as can be seen below:

image

Or b) create a "table" view for Nominators/Validators as can be seen below:

image

Opinions?

kianenigma:
not a fan of `b`, but both `a` and `c` look fine to me. 

Looking at all the stats that we have, and combining them with the new stuff that I want, I can list the following as these sections in design `a` and/or `c`: 

- Overview 
  - Epoch and Era Progress
  - Stake rates (ideal, current, inflation)
  - Returns
- Validators 
  - Count
    - Total
    - Active
    - Waiting 
  - Stake 
    - Intention threshold
    - Min active threshold 
    - Average active
- Nominators 
  - Count
    - Total
    - Active 
  - Stake
    - Intention threshold
    - Min active threshold 
- Bags 
  - Total bags
  - Total nodes 

Of the above, I think only the bags stats can stay in its own tab as well as in the overview page. Everything else can only be in the overview page.

wirednkod:
@jacogr any feedback would be highly appreciated 🙏🏽

jacogr:
Opening it, I went "wow, nice". Certainly really cool.

I will delve into the code (hopefully) today. A couple of play-through comments -

- we have lost compatibility, i.e. networks such as Dock never finishes loading everything - http://localhost:3000/?rpc=wss%3A%2F%2Fmainnet-node.dock.io#/staking (this means we assume some params that are on the current staking is everywhere, without fallbacks)

- not 100% sure I quite like the hover-highlight, it is probably my old eyes, I just find myself squinting

Will go through the code.

packages/page-staking/src/Overview/SummaryNominators.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const { t } = useTranslation();
+
+  return (
+    <Card withBottomMargin>
+      <Title>{t<string>('nominators')}</Title>
+      <SummaryBox>
+        <Section>
+          <CardSummary
+            help={t<string>('Total count of nominators.')}
+            label={t<string>('total')}
```
kianenigma, line 33:
```suggestion
            help={t<string>('Maximum number of nominator intentions.')}
            label={t<string>('maximum')}
```

packages/page-staking/src/Overview/SummaryNominators.tsx
```diff
@@ -0,0 +1,93 @@
...
+          >
+            <SpinnerWrap check={maxNominatorsCount}>
+              {formatNumber(maxNominatorsCount?.toNumber())}
+            </SpinnerWrap>
+          </CardSummary>
+        </Section>
+        <Section>
+          <CardSummary
+            help={t<string>('Electing count of nominators.')}
+            label={t<string>('electing')}
```
kianenigma, line 43:
```suggestion
            help={t<string>('Number of electing nominators.')}
            label={t<string>('electing')}
```

kianenigma: This value seems to be incorrect and should be `const.electionProviderMultiPhase.maxElectingNominators`, i.e. 22,500 for polkadot.

packages/page-staking/src/Overview/SummaryNominators.tsx
```diff
@@ -0,0 +1,93 @@
...
+              {formatNumber(nominatorActiveCount)}
+            </SpinnerWrap>
+          </CardSummary>
+        </Section>
+      </SummaryBox>
+      <SummaryBox>
+        <Section>
+          <CardSummary
+            help={t<string>('Threshold stake among intended nominators.')}
+            label={t<string>('intention thrs')}
```
kianenigma, line 65:
```suggestion
            help={t<string>('Threshold stake to intend nomination.')}
            label={t<string>('intention thrs')}
```

packages/page-staking/src/Overview/SummaryNominators.tsx
```diff
@@ -0,0 +1,93 @@
...
+}
+
+function SummaryNominators ({ targets: { maxNominatorsCount,
+  minNominatorBond,
+  nominatorActiveCount,
+  nominatorElectingCount,
+  nominatorMinActiveThreshold } }: Props) {
+  const { t } = useTranslation();
+
+  return (
```
kianenigma, line 26: we need an info banner here with a short text that says: 

"In order to receive staking rewards, and be exposed with an active validator to slash, a nominator needs to be 'active'. Compare the below thresholds with your active stake to ensure you are in the correct set between 'active' and 'intention'.

kianenigma: Ideally, we could link to polkadot wiki or at least to an edition of my staking newsletter where these roles are explained, but I supposed that might be against your policy @jacogr? LMK what you think. This is the current index of the staking updates that I compile every month, and the February one is particularly useful for this page. https://gist.github.com/kianenigma/aa835946455b9a3f167821b9d05ba376

packages/page-staking/src/Overview/SummaryNominators.tsx
wirednkod: > This value seems to be incorrect and should be `const.electionProviderMultiPhase.maxElectingNominators`, i.e. 22,500 for polkadot.

[fixed here](https://github.com/polkadot-js/apps/pull/7179/commits/fc5ba1529067e318cb07840cbf8144bcbb7f80cc)

packages/page-staking/src/Overview/SummaryNominators.tsx
jacogr: Linking is all fine - there are a number of links already, including to the wiki.

kianenigma: nit: I think the box should go inside `nominators` box 🙈

jacogr:
Will got through (hopefully final) today. Looking forward to have it in :)

kianenigma:
Can't wait to see this merged.

jacogr:
Some small issues remaining -

- http://localhost:3000/?rpc=wss%3A%2F%2Fws.azero.dev#/staking - fails (older version of staking, e.g. "Cannot read properties of undefined (reading 'maxElectingVoters')")
- On Kusama we just need to do the `(api.query.voterList || api.query.bagsList).` dance (could be either pallet)

jacogr:
The bagsList fallback won't be removed anytime reasonable, which is why we have the detection in-place, e.g. [use detected](https://github.com/polkadot-js/apps/blob/6bbff1c80fedb0efd473eafabe6564a0afdd362d/packages/page-staking/src/Bags/useBagsNodes.tsx#L45) & [detect pallet](https://github.com/polkadot-js/apps/blob/6bbff1c80fedb0efd473eafabe6564a0afdd362d/packages/page-staking/src/Bags/useQueryModule.ts#L14).

The apps UI has to lag any removals by a very long time (in the order of 9+ months before even looking at deprecations), since teams do different things. (Same goes for new features, anything old cannot just be dropped, even if it is on Kusama/Polkadot - even between these 2 there is also some lag. This just compounds with the number of different teams out there, hence we are not looking at delay between these 2, but rather across the ecosystem)

wirednkod:
> * http://localhost:3000/?rpc=wss%3A%2F%2Fws.azero.dev#/staking - fails (older version of staking, e.g. "Cannot read properties of undefined (reading 'maxElectingVoters')")

Fixed

> * On Kusama we just need to do the `(api.query.voterList || api.query.bagsList).` dance (could be either pallet - this may be fixed by merging in master if the existing hooks are used)

Im not sure I understand the `dance` that you refer to. Does not sound too correct to me - to have in every reference of bagsList also the `(api.query.voterList || api.query.bagsList)`.
In addition Kusama seems to respond fine in my PR (Im not sure about the numbers ofc) - see screen below:

image

wirednkod:
> Some small issues remaining -
> 
> * http://localhost:3000/?rpc=wss%3A%2F%2Fws.azero.dev#/staking - fails (older version of staking, e.g. "Cannot read properties of undefined (reading 'maxElectingVoters')")
> * On Kusama we just need to do the `(api.query.voterList || api.query.bagsList).` dance (could be either pallet - this may be fixed by merging in master if the existing hooks are used)

Fixed

jacogr:
Something small, another network - http://localhost:3000/?rpc=wss%3A%2F%2Fautomata.api.onfinality.io%2Fpublic-ws#/staking (some parts just keep spinning, I'm guessing because there are no nominators?)

wirednkod:
> Dock has the same issue - http://localhost:3000/?rpc=wss%3A%2F%2Fmainnet-node.dock.io#/staking (although they do have nominators, so the above one may not be caused by that)

Running this:
```
const [now, maxValidatorsCount, validators] = await Promise.all([
  api.query.timestamp.now(),
  api.query.staking.maxValidatorsCount,
  api.query.staking.maxNominatorsCount,
  api.query.staking.minNominatorBond,
  api.query.staking.minValidatorBond
]);

console.log('The current date is: ' + now);
console.log('The maximumNominators count: ' + maxNominatorsCount);
console.log('The maxValidatorsCount count: ' + maxValidatorsCount);
console.log('The minNominatorBond count: ' + minNominatorBond);
console.log('The minValidatorBond count: ' + minValidatorBond);
```

In the developer's screen of `Dock PoS Mainnet` and received for each one of them as `undefined`.
I think the case for Automata is similar - just haven't checked all Queries....

Im not sure on how to solve this. Any suggestions @jacogr @kianenigma ?

jacogr:
So to detect if not available, assuming `api.query.staking.maxValidatorsCount` (simplified) -

```js
const maxValidatorsCount = useCall(api.query.staking.maxValidatorsCount);

...
return (
  {api.query.staking.maxValidatorsCount
    ? maxValidatorsCount
      ? formatNumber(maxValidatorsCount)
      : <Spinner />
    : '-'
  }
);
```

wirednkod:
> Automata is a bit of a pesky example here: they have pallet-staking in their runtime, but pretty much unused (probably by mistake). So almost no storage item is set in there. Might be a bit difficult to handle that case.

I have resolved the Dock PoS case, but the `Automata` one is really edgy. My intention is to push 1 last commit that will solve this edge case by introducing a timer (I know it sounds scary and unorthodox but let me explain). 

This timer will exist only in the `useSortedTargets`, and once countdown is done (set to 10 seconds) then the `undefined` result will switch to null. This way we will know to stop showing loaders in the new dashboard and show a "-" instead.

In case for a peculiar reason an api call delayed more than 10 seconds, then when the result is retrieved the "-" will change to the correct value.

jacogr:
Yes, I would revert the hack, it just feels incorrect and smelly :( Don't have a better idea atm, however need to look at the specific loading stuff (with the other fix in), to have some ideas. (Not sure why here, like in the current summaries, it enforces these to be shown, so would need to look with the other fix in, which simulated the same behavior here)

wirednkod:
> Yes, I would revert the hack

Will do

> it just feels incorrect and smelly :( 

Well it is. The case though here is that it feels that they are not using the pallet correctly - I mean - I know the non-stop-turning loaders does not look good, but should UI care about the mis-configuration/usage that one does?
 
> however need to look at the specific loading stuff (with the other fix in), to have some ideas. 

Which fix you refer to?

wirednkod:
Thank you for the elaboration @jacogr .
Just to clarify my sayings: Of course UI should care about loading spinners. What I was trying to say is that Automata+ are mis-using the way staking should be used - as far as I understand. Meaning that there are no Stakers (equals to 0 that at the moment is not expected from the UI -> translates to "dear Loader, please, keep spinning". 

Having said that, and back to the "UI problem and solution", the issue I see here:
>  skips display of some values which is it cannot make sense of.

 is that we cannot be aware if an API call is finished or `undefined` is returned - as `undefined` for current implementation means "keep spinning". The "hack" (that I will revert as promised) is more of a "wait until something happens and if not stop the loader".
  
 > As soon as we have loading spinners and broken displays, we get support requests since people do use this stuff. Support requests are not great to wake up to, so really don't want to create the opportunity for additional ones in the queue. (And anything that looks like a broken UI, no matter what the reason, will generate support requests in large numbers)
 
 Could not agree more to that - Trying to figure out a solution that will not meddle WAY too much with the current code, but cannot think of one - any pointers that could help on the direction that would fit best?
 I mean - a "API responded and there is nothing to do here" which would led to `Completely skipping` or non-visualize the components would be the best - but based on the current implementation - we do not have that (unless I am missing something)

jacogr:
`undefined` is only returned when the actual query is not there. So for instance, if I do -

```js
const result = useCall(api.query.willNever.beThere)
```

It won't ever return a result for that specific one. In that case, UI-wise -

```js
<div>
  {api.query.willNever?.beThere && formatNumber(result)}
</div>
```

The staking interfaces are really all-over the show when you look at the ecosystem, so in a wide variety of cases some queries just won't be there - generally since they are "relatively new". (And for those, it should just not display the specific info).

If the query is in the runtime, it will always return a result.

wirednkod:
Hey @jacogr I tried to address the case with a simpler approach. WDYT?

cc @kianenigma

wirednkod:
@jacogr Kind reminder

kianenigma:
Another one :)

kianenigma:
I think this is still relevant a good improvement :)

IkerAlus:
@kianenigma is this PR still as relevant as it was at the time of your last comment?

kianenigma:
I think we should either: 

1. improve the staking page of the apps, as it still gets a lot of traffic, in which case I would say yes we should revise this and its parent issue/PR
2. put a banner in the staking tab warning users that this page is for power users and first time users should go to https://staking.polkadot.network (and optionally other platforms that provide 1 click staking)

cmunro01:
I agree that reverting the hack sounds like the safest and most appropriate approach, especially if it feels unreliable or inconsistent. Reviewing the specific loading behavior alongside the other fix could help clarify why the summaries are enforcing those elements. Testing the changes carefully will likely reveal the root cause and provide better direction. For reference or related discussions, checking resources like [crystallauncher.net.pl](https://crystallauncher.net.pl/) may also offer additional insights during the debugging process.

PR #7235, opened 2022-03-25
Add "Copy" icon to the User Sidebar

### LINKED ISSUE #7209: Usability suggestion in the Account page of polkadot.js.org
There is no receive button in the account page. Not obvious for first comers, or crypto beginners to see how they can send funds to their account. 

Screenshot 2022-03-22 at 17 42 35

Suggestion: Add a receive button that when clicked, shows the address of the account, with a copy icon to facilitate.

gilles437:
Closes #7209 

Added the copy icon in the sidebar, on the right side of the address.

Screenshot 2022-03-24 at 19 02 28

When the user copies the icon changes to a "check" icon until the user leaves the button area.

### DIFF
--- a/packages/page-accounts/src/Sidebar/AddressSection.tsx
+++ b/packages/page-accounts/src/Sidebar/AddressSection.tsx
@@ -71,7 +71,7 @@ function AddressSection ({ accountIndex, defaultValue, editingName, flags, onCha
               <Button.Group>
                 <Button
                   icon={isCopyShown ? 'check' : 'copy'}
-                  label={t<string>('Copy')}
+                  label={wasCopied ? t<string>('Copy') : t<string>('Copy')}
                   onClick={toggle}
                   onMouseLeave={isCopyShown ? toggle : donothing }
                 />
--- a/packages/page-accounts/src/Sidebar/AddressSection.tsx
+++ b/packages/page-accounts/src/Sidebar/AddressSection.tsx
@@ -21,8 +21,8 @@ interface Props {
 
 function AddressSection ({ accountIndex, defaultValue, editingName, flags, onChange, value }: Props): React.ReactElement<Props> {
   const { t } = useTranslation();
-  const [isCopyShown, toggle] = useToggle();
-  const [wasCopied, donothing] = useToggle(false);
+  const [isCopyShown, toggleIsCopyShown] = useToggle();
+  const [, donothing] = useToggle(false);
 
   return (
     <div className='ui--AddressSection'>
@@ -71,9 +71,9 @@ function AddressSection ({ accountIndex, defaultValue, editingName, flags, onCha
               <Button.Group>
                 <Button
                   icon={isCopyShown ? 'check' : 'copy'}
-                  label={wasCopied ? t<string>('Copy') : t<string>('Copy')}
-                  onClick={toggle}
-                  onMouseLeave={isCopyShown ? toggle : donothing }
+                  label={isCopyShown ? t<string>('Copied') : t<string>('Copy')}
+                  onClick={isCopyShown ? donothing : toggleIsCopyShown }
+                  onMouseLeave={isCopyShown ? toggleIsCopyShown : donothing }
                 />
               </Button.Group>
             </span>
--- a/packages/page-accounts/src/Sidebar/AddressSection.tsx
+++ b/packages/page-accounts/src/Sidebar/AddressSection.tsx
@@ -22,7 +22,7 @@ interface Props {
 function AddressSection ({ accountIndex, defaultValue, editingName, flags, onChange, value }: Props): React.ReactElement<Props> {
   const { t } = useTranslation();
   const [isCopyShown, toggleIsCopyShown] = useToggle();
-  const [, donothing] = useToggle(false);
+  const NOOP = () => undefined;
 
   return (
     <div className='ui--AddressSection'>
@@ -72,8 +72,8 @@ function AddressSection ({ accountIndex, defaultValue, editingName, flags, onCha
                 <Button
                   icon={isCopyShown ? 'check' : 'copy'}
                   label={isCopyShown ? t<string>('Copied') : t<string>('Copy')}
-                  onClick={isCopyShown ? donothing : toggleIsCopyShown }
-                  onMouseLeave={isCopyShown ? toggleIsCopyShown : donothing }
+                  onClick={isCopyShown ? NOOP : toggleIsCopyShown }
+                  onMouseLeave={isCopyShown ? toggleIsCopyShown : NOOP }
                 />
               </Button.Group>
             </span>
--- a/packages/page-accounts/src/Sidebar/AddressSection.tsx
+++ b/packages/page-accounts/src/Sidebar/AddressSection.tsx
@@ -1,12 +1,17 @@
 // Copyright 2017-2022 @polkadot/page-accounts authors & contributors
 // SPDX-License-Identifier: Apache-2.0
 
-import React from 'react';
+//import React from 'react';
+import React, { useState } from 'react';
+
+import { Button } from '@polkadot/react-components';
 
 import { AccountName, IdentityIcon, Input } from '@polkadot/react-components';
 import { AddressFlags } from '@polkadot/react-hooks/types';
 
 import { useTranslation } from '../translate';
+import {CopyToClipboard} from 'react-copy-to-clipboard';
+
 
 interface Props {
   value: string,
@@ -17,10 +22,20 @@ interface Props {
   accountIndex: string | undefined,
 }
 
+
+
 function AddressSection ({ accountIndex, defaultValue, editingName, flags, onChange, value }: Props): React.ReactElement<Props> {
   const { t } = useTranslation();
 
+  // Declare a new state variable, which we'll call "count"
+  const [countCopied, setCountCopied] = useState(0);
+
+
+  
   return (
+
+
+    
     <div className='ui--AddressSection'>
       <IdentityIcon
         size={80}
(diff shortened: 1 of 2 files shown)


jacogr:
@gilles437 I did invite you to the org a while back, just check your invitations.

packages/page-accounts/src/Sidebar/AddressSection.tsx
```diff
@@ -1,12 +1,17 @@
 // Copyright 2017-2022 @polkadot/page-accounts authors & contributors
 // SPDX-License-Identifier: Apache-2.0
 
-import React from 'react';
+//import React from 'react';
```
jacogr, line 4:
```suggestion
```

packages/page-accounts/src/Sidebar/AddressSection.tsx
```diff
@@ -1,12 +1,17 @@
...
+import React, { useState } from 'react';
+
+import { Button } from '@polkadot/react-components';
```
jacogr, line 7: Combine with the import just below this.

packages/page-accounts/src/Sidebar/AddressSection.tsx
```diff
@@ -1,12 +1,17 @@
...
 
 import { AccountName, IdentityIcon, Input } from '@polkadot/react-components';
 import { AddressFlags } from '@polkadot/react-hooks/types';
 
 import { useTranslation } from '../translate';
+import {CopyToClipboard} from 'react-copy-to-clipboard';
```
jacogr, line 13:
```suggestion
import { CopyToClipboard } from 'react-copy-to-clipboard';
```
Additionally it would need to move to just below the `react` usage above.

packages/page-accounts/src/Sidebar/AddressSection.tsx
```diff
@@ -17,10 +22,20 @@ interface Props {
...
 
+
+
 function AddressSection ({ accountIndex, defaultValue, editingName, flags, onChange, value }: Props): React.ReactElement<Props> {
   const { t } = useTranslation();
 
+  // Declare a new state variable, which we'll call "count"
+  const [countCopied, setCountCopied] = useState(0);
+
+
```
jacogr, line 33:
```suggestion
```

Unneeded.

packages/page-accounts/src/Sidebar/AddressSection.tsx
```diff
@@ -17,10 +22,20 @@ interface Props {
...
+  
   return (
+
+
+
```
jacogr, line 38:
```suggestion
```

Unneeded.

packages/page-accounts/src/Sidebar/AddressSection.tsx
```diff
@@ -46,15 +61,36 @@ function AddressSection ({ accountIndex, defaultValue, editingName, flags, onCha
...
           <div className='ui--AddressMenu-index'>
             <label>{t<string>('index')}:</label> {accountIndex}
           </div>
         )}
       </div>
+
+      <div className='ui--AddressSection__CopyColumn'>
+        <div className='ui--AddressMenu-copyaddr'>         
+          <CopyToClipboard text={value}
+          onCopy={() => setCountCopied(countCopied + 1)}
```
jacogr, line 79: Rather use a callback. There is also a `useIncrement` hook that wraps inc/dec already, see https://github.com/polkadot-js/apps/blob/master/packages/react-hooks/src/useIncrement.ts

packages/page-accounts/src/Sidebar/AddressSection.tsx
```diff
@@ -46,15 +61,36 @@ function AddressSection ({ accountIndex, defaultValue, editingName, flags, onCha
...
+          >
+            <span >
+              <Button.Group> 
+                <Button  
+                    icon={countCopied ? 'check' : 'copy'}   
+                    label={t<string>('Copy')} 
+                    onClick={() => setCountCopied(countCopied + 1) }
+                    onMouseLeave={() => setCountCopied(0) }
```
jacogr, line 87: As above, use `useIncrement` if you wish to increment or `useToggle` if you want a true/false (which is probably better suited here)

jacogr (changes requested):
Some comments on inline function creation (which is not advisable, linting should also complain) and possible existing hooks usage.

packages/page-accounts/src/Sidebar/AddressSection.tsx
```diff
@@ -17,9 +21,14 @@ interface Props {
   accountIndex: string | undefined,
 }
 
+
+
```
jacogr, line 25:
```suggestion
```

jacogr (commented):
In addition to the non-closed issues where the functions are re-created each render, please run `yarn lint` locally, CI is failing - https://github.com/polkadot-js/apps/runs/5733108256?check_suite_focus=true#step:3:691

packages/page-accounts/src/Sidebar/AddressSection.tsx
```diff
@@ -46,15 +50,36 @@ function AddressSection ({ accountIndex, defaultValue, editingName, flags, onCha
...
+      <div className='ui--AddressSection__CopyColumn'>
+        <div className='ui--AddressMenu-copyaddr'>
+          <CopyToClipboard
+            text={value}
+          >
+            <span>
+              <Button.Group>
+                <Button
+                  icon={isCopyShown ? 'check' : 'copy'}
+                  label={wasCopied ? t<string>('Copy') : t<string>('Copy')}
```
jsidorenko, line 74: I think it's better to either change the label to "Copied" when a user clicks the button or simply remove the ternary operator: `label={t<string>('Copy')}`

packages/page-accounts/src/Sidebar/AddressSection.tsx
```diff
@@ -46,15 +50,36 @@ function AddressSection ({ accountIndex, defaultValue, editingName, flags, onCha
...
+                  onClick={toggle}
+                  onMouseLeave={isCopyShown ? toggle : donothing }
```
jsidorenko, line 76: What if we write in that way: `onMouseLeave={isCopyShown && toggle}`?
In that case, we won't need to have that hook: `const [wasCopied, donothing] = useToggle(false);`

packages/page-accounts/src/Sidebar/AddressSection.tsx
```diff
@@ -19,6 +21,8 @@ interface Props {
 
 function AddressSection ({ accountIndex, defaultValue, editingName, flags, onChange, value }: Props): React.ReactElement<Props> {
   const { t } = useTranslation();
+  const [isCopyShown, toggle] = useToggle();
```
jsidorenko, line 24: Consider renaming `toggle` to `toggleIsCopyShown`

packages/page-accounts/src/Sidebar/AddressSection.tsx
gilles437: changes the name Copy to Copied when address is copied

packages/page-accounts/src/Sidebar/AddressSection.tsx
gilles437: done

packages/page-accounts/src/Sidebar/AddressSection.tsx
gilles437: removed the wasCopied variable which was useless

jacogr (approved):
Thank you. Sorry about the delay here - have some issues with the teenager having "the virus", so attention was elsewhere in the last day or two.

I'm reasonably happy, will defer to @jsidorenko for all-ok here :) (No pressure, just spreading the load... Happy to see it go.

packages/page-accounts/src/Sidebar/Sidebar.tsx
```diff
@@ -121,11 +130,26 @@ export default React.memo(styled(FullSidebar)`
 
   .ui--AddressMenu-addr {
     word-break: break-all;
-    width: 26ch;
+    width: 24ch;
```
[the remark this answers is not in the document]
gilles437: Ok, thanks @jacogr.
Should I consult Gav about this change or we are ok with it?  
this was needed to have the "Copied" text well aligned after clicking on "Copy".
Also, Displaying the whole address in the Sidebar without an explicit action from the user looks to me unsecured.
Any youtuber recording a tutorial or blogger taking a screenshot will have his address revealed to the world.
So I can "hide" the whole address by shortening the address display to:
<fisrst five characters>...<last five characters>

Like other wallets do (Metamask, Keplr, Binance wallet,...)

What do you think of it?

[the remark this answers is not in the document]
gilles437: format : to "first five letters"..."last five letters" of the address which is 13 characters all in all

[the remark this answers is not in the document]
jsidorenko: Could it be that some browsers restrict copying smth automatically?
I personally feel safer when I see the whole address instead of `"first five letters"..."last five letters"`, otherwise I'm starting to ask why smth is hidden from me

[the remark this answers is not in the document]
gilles437: Although the address is public, if there is a way to logically connect a public address to a real user, then the same user can become a target for a hacker that would like to discover the private key. 

This is the reason why other wallets hide the full address and display only part of it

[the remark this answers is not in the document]
jsidorenko: When you click on the "Copy" button, you'll still get the full address copied, right? So hiding a portion of the address won't make the address' owner less targetable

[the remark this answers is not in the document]
gilles437: The Use Case I am talking about is when a blogger takes a screen shot of his own page and displays it in an article , or when a youtuber records a video showing the sidebar when doing a video about how to use Polkadot, or someone taking a picture of the screen of his neighbor. So in these cases you can associate a real person (blogger, youtuber, or some one that is physically close) to a public address, which makes it more vulnerable. 
A person with bad intentions could know how much money has the account and then possibly decide to target it. 

The address would be copied by clicking on the button but still would not reveal it.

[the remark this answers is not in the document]
jsidorenko: yeah, I see your point

PR #7262, opened 2022-03-29
fix: maybe there's not a conviction of direct voting

JayJay1024:
Maybe there's not a conviction of direct locks?

https://github.com/polkadot-js/api/blob/235a7b1aa952e3717fbd6416102fd821f5011b51/packages/api-derive/src/democracy/locks.ts#L69-L88

### DIFF
--- a/packages/react-components/src/DemocracyLocks.tsx
+++ b/packages/react-components/src/DemocracyLocks.tsx
@@ -51,7 +51,7 @@ function groupLocks (t: TFunction, bestNumber: BN, locks: Partial<DeriveDemocrac
       .reduce((sorted: Entry[], [{ balance, isDelegated, isFinished = false, referendumId, vote }, blocks]): Entry[] => {
         const isCountdown = blocks.gt(BN_ZERO);
         const header = referendumId && vote
-          ? <div>#{referendumId.toString()} {formatBalance(balance, { forceUnit: '-' })} {vote.conviction.toString()}{isDelegated && '/d'}</div>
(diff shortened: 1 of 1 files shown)


packages/react-components/src/DemocracyLocks.tsx
```diff
@@ -51,7 +51,7 @@ function groupLocks (t: TFunction, bestNumber: BN, locks: Partial<DeriveDemocrac
       .reduce((sorted: Entry[], [{ balance, isDelegated, isFinished = false, referendumId, vote }, blocks]): Entry[] => {
         const isCountdown = blocks.gt(BN_ZERO);
         const header = referendumId && vote
-          ? <div>#{referendumId.toString()} {formatBalance(balance, { forceUnit: '-' })} {vote.conviction.toString()}{isDelegated && '/d'}</div>
+          ? <div>#{referendumId.toString()} {formatBalance(balance, { forceUnit: '-' })} {vote?.conviction.toString()}{isDelegated && '/d'}</div>
```
jacogr, line 54: This won't fix the issue - the existence of vote is checked in the preceding line already, so it is defined. However, not sure why a vote won't have conviction? (As per the linked issue)

jacogr, line 54:
```suggestion
          ? <div>#{referendumId.toString()} {formatBalance(balance, { forceUnit: '-' })} {vote.conviction?.toString()}{isDelegated && '/d'}</div>
```

jacogr (approved):

Issue #7251, opened 2022-03-28
Proposal/discussion to update the SS58 format long description tooltip.


chrisdcosta:
In the settings page the the tooltip for the SS58 format describes the format as follows: 

`Prefix indicates the ss58 address format in this network, it is a number between 0 ~ 255 that describes the precise format...`

This refers to the original format for the SS58Prefix being of type `u8`. This has now been changed to `u16` type allowing for more values to be created.

I propose to change the text to read:

`Prefix indicates the ss58 address format in this network, it is a 16-bit integer that describes the precise format...`

However, this impacts many already translated files and so it cannot be globally updated in english without creating inconsistencies with the translated versions.

jacogr:
It sounds spot-on.

That is fine. If untranslated, the English version will show.

PRs really welcome.

chrisdcosta:
Ok I will work on it.

jacogr:
Closed in https://github.com/polkadot-js/apps/pull/7277

PR #7277, opened 2022-03-30
Updates the ss58 prefix long description to align with substrate (u16)

chrisdcosta:
Related to issue https://github.com/polkadot-js/apps/issues/7251

The current long text is not aligned with the SS58 Prefix type used in substrate, cumulus and parachains. 

The text update only affects the English version and translated versions will have to be updated by the community.


jacogr (approved):
Thank you!

PR #7321, opened 2022-04-07
Update automata-network type definitions

RyuH1:


jacogr (changes requested):
Running `yarn lint` locally should fix the quoting issues as per CI - https://github.com/polkadot-js/apps/runs/5865737812?check_suite_focus=true

RyuH1:
> Running `yarn lint` locally should fix the quoting issues as per CI - https://github.com/polkadot-js/apps/runs/5865737812?check_suite_focus=true

Have fixed, thanks for noticing

jacogr (approved):

Issue #7076, opened 2022-03-01
feature request: Support for nomination pools


emostov:
https://github.com/paritytech/substrate/pull/10694 will introduce a MVP for nomination pools.   To facilitate a decent UX with the the initial iteration of pools I think it would help to have the following features:

- _List pools_: list the pools similarly to how the apps UI currently lists validators. This could show the pools stake, the pools admin (linking to any identities of the admin), and maybe have a dropdown that shows its nominations. This could also display warnings for if 1) the pools stake is higher than the min staked validator since this would lead to suboptimal rewards, 2) the pool does not have enough bonded to make it into the next election, 3) the pool has no valid nominations.
- _Join a pool_: if the user is not part of a pool, allow them to join one. This could be an option along side each pool when its listed. See the `join` extrinsic.
- _Claim payout_: if the user has some unclaimed rewards, allow them to claim. See the `claim_payout` extrinsic.
- _Unbond_: If a user is part of a pool, allow them to unbond. See the `unbond_other` extrinsic
- _Withdraw unbonded_: If the unbending duration has elapsed since when the user has unbonded, allow them to withdraw unbounded funds. See the `withdraw_unbonded_other` extrinsic.

Bonus features:

- _Create a pool_: allow a user to to create a pool. See the `create` extrinsic. 
- _Nominate_: allow a pool admin to nominate. The `nominate` extrinsic.
- _Pool performance history_: this would basically be the equivalent of what ever stats can be displayed about a nominator, since the pool has an underlying account nominating.

kianenigma:
We are happy to try and find another contributor for this one to take it off your plate if you are certain that you won't have time for it anytime soon, since it is relatively time-sensitive. But I can imagine the whole process being both faster and smoother if you do it yourself? let me know what you think!

jacogr:
There are a couple of issues here -

1. the branch doesn't build (for me)
2. unsure about how much changes are still going in (assuming it can be run)
3. there are no active external contributors (trying to find a raw body is a recipy for disaster)
4. "time sensitive" - what does that mean? (It is not merged, not in for-merge review)

Happy to take a look if there is a working PR which is not changing daily.

emostov:
Thank you for taking the time to look into this! The feedback about events is very appreciated.

To address your 2nd and 4th point I can just write an update here once I am confident its stable.

PR #7384, opened 2022-04-18
Expansion to pools logic

jacogr:
Closes https://github.com/polkadot-js/apps/issues/7076

- [x] Bond more (free)
- [x] Bond more (rewards)
- [x] Display overall claimable
- [x] Display per-account claimable
- [x] Claiming of rewards
- [x] Integration with normal stash payouts
- [x] Filter join with non-pooled accounts
- [x] Unbond
- [x] Unlocking remaining
- [x] Withdraw unlocked


kianenigma:
Tested all of this today. Some feedback:

1. Pools are designed to have pretty much infinite members. While showing the count is great, I think showing the members without pagination is a dead-end. 
2. Showing the pool roles in the pool list page would be an easy addition. 
3. In the `accounts -> pooled` page, it is a bit hard to distinguish if an account in a pool is the member, or has a special role. For example, if I want to change the nominations of pool 1 below, I have to click the `...` of Ferdie. Knowing the in this list, Ferdie is a `depositor` and Dave and Eve are `Members` is valuable info.
4. Each pool has a bonded account and and a reward account. I see that you already "identify" the bonded account if it pops up somewhere. Is there a way to get these accounts in a simple way? They can probably be added to the pools list page. 
5. Pool management and destruction scenarios are not worked on yet. I assume this is intentional for now. 
6. Keep an eye on https://github.com/paritytech/substrate/pull/11235/files#diff-d19373c4bea562e00466b1163f9d4c7bd4fa9ef8075a09f8bb54a0a4f127ba23R35, hopefully it will make things easier for you.

jacogr:
The RPCs won't be used. It doesn't help at all actually, rather "hinder" since they are non-subscription based.

PR #7422, opened 2022-04-22
Add Bengali translation

itsonal:
Bengali is the fifth biggest spoken language in the world and the national language of Bangladesh. It's also used extensively in India. https://en.wikipedia.org/wiki/Bengali_language


jacogr (approved):
Thank you! To make it active, you also need an entry here - https://github.com/polkadot-js/apps/blob/master/packages/apps-config/src/settings/languages.ts

Yes, it is a tiny bit counter-intuitive...

PR #7426, opened 2022-04-22
Added Shiden Crowdloan 2

akru:
**Description**

This PR introduce empty endpoints for Shiden Network Crowdloan 2 (just add logo and title for parachain id).

### DIFF
--- a/packages/apps-config/src/endpoints/productionRelayKusama.ts
+++ b/packages/apps-config/src/endpoints/productionRelayKusama.ts
@@ -388,6 +388,16 @@ export function createKusama (t: TFunction): EndpointOption {
         }
       },
       {
+        info: 'shiden',
+        homepage: 'https://shiden.plasmnet.io/',
+        paraId: 2120,
+        text: t('rpc.kusama.shiden', 'Shiden Crowdloan 2', { ns: 'apps-config' }),
+        isUnreachable: true,
+        providers: {
+          StakeTechnologies: 'wss://rpc.shiden.astar.network'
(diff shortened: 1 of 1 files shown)


packages/apps-config/src/endpoints/productionRelayKusama.ts
```diff
@@ -387,6 +387,13 @@ export function createKusama (t: TFunction): EndpointOption {
           Dwellir: 'wss://shiden-rpc.dwellir.com'
         }
       },
+      {
+        info: 'shiden',
+        homepage: 'https://shiden.plasmnet.io/',
+        paraId: 2120,
+        text: t('rpc.kusama.shiden2', 'Shiden Crowdloan 2', { ns: 'apps-config' }),
+        providers: {}
```
jacogr, line 395: Lemme just test this one - not sure how it behaves with empty providers. (Normally there is at least 1 + isUnreachable). This is certainly valid, but just concerned that I missed a trick in the filtering.

Will check/approve/comment in the morning.

jacogr (changes requested):
Tested the comment here https://github.com/polkadot-js/apps/pull/7426/files#r856513007

Basically since there are no providers, it doesn't get added. The suggestion would be to add a provider entry but just add the `isUnreachable: true` flag to ensure it never gets used. (Assuming here that the endpoint is not running)

akru:
@jacogr thank you for feedback, fixed.

jacogr (approved):

PR #7507, opened 2022-05-02
calldata hyperlink

gilescope:
Why: there are many locations where you want to tee up a transaction for submission. There are many testing situations where this would make a lot of integrations slicker. Arguably POST could be accepted as well as GET because some call datas could be long but the vast majority are not and ease of use is important (so even if POST was allowed I would still like GET to work).

### DIFF
--- a/packages/page-extrinsics/src/Decoder.tsx
+++ b/packages/page-extrinsics/src/Decoder.tsx
@@ -3,10 +3,11 @@
 
 import type { SubmittableExtrinsic, SubmittableExtrinsicFunction } from '@polkadot/api/types';
 import type { Call } from '@polkadot/types/interfaces';
+import type { HexString } from '@polkadot/util/types';
 import type { DecodedExtrinsic } from './types';
 
 import React, { useCallback, useState } from 'react';
-import { useLocation } from 'react-router-dom';
+import { useParams } from 'react-router-dom';
 import styled from 'styled-components';
 
 import { Button, Call as CallDisplay, Input, InputAddress, InputExtrinsic, MarkError, TxButton } from '@polkadot/react-components';
@@ -18,8 +19,8 @@ import Decoded from './Decoded';
 import { useTranslation } from './translate';
 
 interface Props {
-  calldata?: string;
   className?: string;
+  defaultValue?: HexString | null;
   setLast: (value: DecodedExtrinsic | null) => void;
 }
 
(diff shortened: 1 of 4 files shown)


gilescope:
`location?.search?.replace('?calldata=', '');` isn't quite right either. I wondered if I could use `useSearchParams();` but maybe that is an old way?

jacogr:
This is a step in the right direction. Will walk through what should be the approach. 

The `?calldata=0x...` is not quite right, since it only to the decoder path (unlike `?rpc=...` which applies globally). So it is the same as a `/explorer/query/0x...` block query rather. So couple of things to wire this in -

1. In the `Decoder` component we need the following -

```js
import { useParams } from 'react-router-dom';
...
const { value: encoded } = useParams<{ value: string }>();
const [initialValue] = useState(() => encoded);
...
<div className={className}>
  <Input
     defaultValue={initialValue}
      isError={!extrinsicFn}
...
```

2. In the `index.ts` we need to allow the `decoder/0x...` route as well -

```js
<Route path={[`${basePath}/decode/:value`, `${basePath}/decode`]}>
  <Decoder setLast={setDecoded} />
</Route>
```

3. Additionally in `index.ts` we also need to pass the `hasParam` flag to the header setp, it changes to -

```js
{
  hasParams: true,
  name: 'decode',
  text: t<string>('Decode')
}
```

4. Now it does to populate, but `Input` doesn't fire on initial. This seems like a bug or rather that input was never used this way. However all other `Input*` does always fire initial values, so the component needs adjustment -

```js
const [initialValue] = useState(() => defaultValue);

useEffect((): void => {
  initialValue && onChange && onChange(initialValue);
}, [initialValue, onChange]);
```

As a bonus, we always lose track of the previous decoded value when swapping between Submission & Decoded, so it needs adjustment.

1. In `types.ts` we add the original hex value in -

```js
export interface DecodedExtrinsic {
  call: Call;
  hex: HexString;
  fn: SubmittableExtrinsicFunction<'promise'>;
}
```

2. In `Decoder.ts` -

```js
interface Props {
  className?: string;
  defaultValue?: HexString | null;
  setLast: (value: DecodedExtrinsic | null) => void;
}

function Decoder ({ className, defaultValue, setLast }: Props): React.ReactElement<Props> {
  ...
  const [initialValue] = useState(() => defaultValue || encoded);
  ...
```

3. In `index.ts` we pass the value through of the previous hex, if available -

```js
<Route path={[`${basePath}/decode/:value`, `${basePath}/decode`]}>
  <Decoder
    defaultValue={decoded && decoded.hex}
    setLast={setDecoded}
  />
</Route>
```

gilescope:
Thank you @jacogr for your guidance in this. Much appreciated!

jacogr (approved):
Thank you!

PR #7519, opened 2022-05-04
Teleports to statemint are open

gilescope:
This seems necessary change but not sufficient. There's something more I need to tweak to get the menu showing I think?


jacogr:
In the actual common chain, just also set `teleport: [-1]`

gilescope:
That's the ticket. Seems to appear now.

jacogr (approved):

PR #7535, opened 2022-05-05
Update @substrate/connect to latest version

tomaka:
We really would like the changes we've made to substrate/connect to be deployed

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -3488,7 +3488,7 @@ __metadata:
     "@polkadot/x-fetch": ^9.1.1
     "@polkadot/x-global": ^9.1.1
     "@polkadot/x-ws": ^9.1.1
-    "@substrate/connect": 0.7.2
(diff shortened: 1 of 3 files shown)


yarn.lock
```diff
@@ -3488,7 +3488,7 @@ __metadata:
     "@polkadot/x-fetch": ^9.1.1
     "@polkadot/x-global": ^9.1.1
     "@polkadot/x-ws": ^9.1.1
-    "@substrate/connect": 0.7.2
+    "@substrate/connect": 0.7.3
```
jacogr, line 3491: This is not correct - the specific version refers to a specific version it works with, i.e. 8.3.1 does not ship with 0.7.3

The correct approach would probably to bump the API in the apps UI, which has the latest as part of the beta.

tomaka:
Sorry I have no idea how yarn works

PR #7630, opened 2022-05-23
Add Rococo Snowbridge

alistair-singh:
Resolves SNO-259


packages/apps-config/src/ui/logos/index.ts
```diff
@@ -51,6 +51,7 @@ import chainRococoTrick from './chains/rococo-trick.svg';
 import chainShiden from './chains/shiden.png';
 import chainSkyeKiwi from './chains/skyekiwi.png';
 import chainSnakenet from './chains/snakenet.svg';
+import chainSnowbridge from './chains/snowbridge.jpeg';
```
jacogr, line 54: Ummmm... weird, see https://github.com/polkadot-js/apps/runs/6554470803?check_suite_focus=true#step:3:695



I think this could just be since we have not had jpeg before, so we don't have a loader? Although the loaders array does seem to include it, https://github.com/polkadot-js/apps/blob/master/packages/apps/webpack.base.cjs#L81-L96 - so a bit confused here.

jacogr, line 54: Ummmm... weird, see https://github.com/polkadot-js/apps/runs/6554470803?check_suite_focus=true#step:3:695



I think this could just be since we have not had jpeg before, so we don't have a loader? Although the loaders array does seem to include it, https://github.com/polkadot-js/apps/blob/master/packages/apps/webpack.base.cjs#L81-L96 - so a bit confused here.

packages/apps-config/src/endpoints/testingRelayRococo.ts
```diff
@@ -204,6 +204,14 @@ export const testParasRococo: EndpointOption[] = [
     providers: {
       Zeitggeist: 'wss://roc.zeitgeist.pm'
     }
+  },
+  {
+    info: 'snowbridge',
+    paraId: 3016,
+    text: 'Snowbridge',
```
jacogr, line 211: Just move this up alphabetical, i.e. correct position under "S"

packages/apps-config/src/ui/logos/index.ts
alistair-singh: I have just converted the image to `png`. With `yarn run start` the application builds and works with `jpeg` so it is very strange.

PR #7640, opened 2022-05-26
Adjust litentry parachain config

Kailai-Wang:
This PR adjusts the litentry parachain config.


jacogr (approved):

jacogr:
Unreachable as per https://github.com/polkadot-js/apps/issues/7642 - will disable again.

Kailai-Wang:
> Unreachable as per #7642 - will disable again.

Hi @jacogr 
now the DNS entry for litentry endpoint(rpc.litentry-parachain.litentry.io) should be available, will the `isUnreachable` flag be adjusted automatically, or do I have to submit another PR for it?

jacogr:
The CI checks ignores everything marked `isUnreachable` since it cannot actually help much. There could be a number of reasons -

- it points to the wrong chain (in this case each needs to be checked manually)
- DNS could be unavailable (in this case a check _would_ help)
- it could be permanently playing yo-yo (in this case it means manual PRs in each case)

So the bottom-line is that it is good for detecting misbehaving providers, but it is not good at detecting if providers are available.

Kailai-Wang:
I see, thanks.
I'll submit another manual PR later

PR #7646, opened 2022-05-26
Introduce Subspace Gemini 1 to the list of test networks

nazar-pc:
This is for our upcoming incentivized testnet. Endpoints are not live yet, but will be in the near future.

If there is such option, it would be perfect for this PR to be be reviewed, but only merged when endpoints are live. If this isn't an option, feel free to merge right away.

cc @vedhavyas


jacogr:
Since the endpoints are not available, it doesn't make sense to merge it without a flag. So in this case, rather let me know when they are up and can merge at that point.

vedhavyas:
@jacogr they are live now. We are good to go

jacogr (approved):

PR #7700, opened 2022-06-05
update bifrost name

awesomepan:


packages/apps-config/src/endpoints/productionRelayKusama.ts
```diff
@@ -49,7 +49,7 @@ export const prodParasKusama: EndpointOption[] = [
     info: 'bifrost',
     homepage: 'https://ksm.vtoken.io/?ref=polkadotjs',
     paraId: 2001,
-    text: 'Bifrost (Kusama)',
+    text: 'Bifrost Kusama',
```
jacogr, line 52: Any reason why this is not simply “Bifrost” (this only affects the list display). The reason for asking is that it doesn’t s already under the Kusama tab. 

(Obviously the chain name later on is 100%)

packages/apps-config/src/endpoints/productionRelayPolkadot.ts
```diff
@@ -54,7 +54,7 @@ export const prodParasPolkadot: EndpointOption[] = [
     info: 'bifrost',
     homepage: 'https://crowdloan.bifrost.app',
     paraId: 2030,
-    text: 'Bifrost',
```
jacogr, line 57: Same question as above ;)

awesomepan: updated, thx.

packages/apps-config/src/endpoints/productionRelayKusama.ts
awesomepan: updated thx

jacogr (approved):

PR #7767, opened 2022-06-20
Fix overflow issue in extensions table

SaltyCucumber:
Before:
Screenshot 2022-06-20 at 17 29 54

After:
Screenshot 2022-06-20 at 17 54 11


jacogr (approved):
Thank you!

PR #7771, opened 2022-06-20
Update index.ts

helixstreet:
helixstreet under namedLogos added


helixstreet:
Great! Now it works! Thx indeed. Tom.

PR #7772, opened 2022-06-20
Added Kusama GM + Removed Rococo GM

arrudagates:


packages/apps-config/src/endpoints/productionRelayKusama.ts
```diff
@@ -116,6 +116,15 @@ export const prodParasKusama: EndpointOption[] = [
       Equilibrium: 'wss://node.genshiro.io'
     }
   },
+  {
+    info: 'gm',
```
jacogr, line 120:
```suggestion
    info: 'gm',
    isUnreachable: true,
```

The endpoint doesn't seem reachable. (Or at least doesn't connect for me)

jacogr (changes requested):
Should be marked `isUnreachable: true` as per the comment or actually reachable. (The CI checks will throw this one up as not available if not done, which means more manual PRs to disable and then finally enable)

jacogr (approved):

jacogr:
Thank you.

Issue #7801, opened 2022-06-27
Something went wrong with the query and rendering of this component


StrawberryFlavor:
* **I'm submitting a ...**

  - [x] Bug report
  - [ ] Feature request
  - [ ] Support request (please bear the comments above in mind)
  - [ ] Other

* **What is the current behavior and expected behavior?**

1. Convert encoded call data put into decode, click `Submission`
```
0x3600008100e87648170000000000000000000000010102005d1f0100c0997c4f2b3a83eb07ef74a867cf672a25a2a30cc61abc936dcc994df77ba84a00f2052a01000000
```
<img width="1666" alt="image" src="https://user-images.githubusercontent.com/28721839/175874415-ff9c89bc-9c3c-487f-994e-17239ffac135.png">

2.  Switch currencyId to ForeignAsset
<img width="1646" alt="image" src="https://user-images.githubusercontent.com/28721839/175874707-b10ae54f-0c70-421a-b416-7e915128479a.png">

3. erro
```
Uncaught error. Something went wrong with the query and rendering of this component. Please supply all the details below when logging an issue, it may help in tracing the cause.

Assertion failed

Error: Assertion failed
at n (https://polkadot.js.org/apps/polkadot.02.f3947304.js:1:86864)
at i._initArray (https://polkadot.js.org/apps/polkadot.02.f3947304.js:1:88781)
at i._init (https://polkadot.js.org/apps/polkadot.02.f3947304.js:1:88157)
at new i (https://polkadot.js.org/apps/polkadot.02.f3947304.js:1:87158)
at c (https://polkadot.js.org/apps/polkadot.02.f3947304.js:1:334875)
at https://polkadot.js.org/apps/main.b987ff52.js:2:798309
at Object.useMemo (https://polkadot.js.org/apps/react.01.4b9f06d3.js:2:326385)
at t.useMemo (https://polkadot.js.org/apps/react.01.4b9f06d3.js:2:610137)
at QM (https://polkadot.js.org/apps/main.b987ff52.js:2:798225)
at Ca (https://polkadot.js.org/apps/react.01.4b9f06d3.js:2:320072)
```

* **What is the motivation for changing the behavior?**

* **Please tell us about your environment:**

  - Version:
  - Environment:

    - [ ] Node.js
    - [x] [Browser](https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Fkarura.api.onfinality.io%2Fpublic-ws#/extrinsics/decode)
[...]

PR #7803, opened 2022-06-27
Allow for pre-set Enum changes

jacogr:
Close https://github.com/polkadot-js/apps/issues/7801 

The issue sadly misses all the critical info and is therefore non reproducible, so it is scratching in the dark. However, it _seems_ related to this. Can be re-opened if there is ever reproducible steps and this doesn't address the hunt in the dark (which includes the actual chain)


StrawberryFlavor:
> 

hey, this is 100% reproducible, please check the screenshot I provided and the information I provided

PR #7819, opened 2022-07-03
Remove Plasm

shunsukew:
Plasm is no more active, so we'd like to remove it from Polkadot.js apps in order not to cause confusion.
cc. @hoonsubin


jacogr (approved):
Thank you.

PR #7825, opened 2022-07-04
Add extensions dropdown

SaltyCucumber:
Dropdown has two extensions now, we can easily add additional ones in `const extensionList`
Once connected to an extension it will stay connected and load automatically on page load, no additional action is needed from user
If extensions are not installed a link will be generated depending on user's browser (right now it's chrome and firefox, by default it will go to chrome)

Screenshot 2022-07-04 at 15 24 10


packages/apps-config/src/extensions/index.ts
```diff
@@ -10,22 +14,66 @@ interface Extension {
...
   Object.entries(browsers).forEach(([browser, link]): void => {
     available[browser as 'chrome'].push({ desc, link, name });
   });
 
   return available;
 }, { chrome: [], firefox: [] });
+
+export const supportedExtensionsNames = () => extensionList.map(({ name }) => name);
+
+export const getUserInstalledExtensions = () => {
```
jacogr, line 64:
```suggestion
export function supportedExtensionsNames (): string[] {
  return extensionList.map(({ name }) => name);
}

export function getUserInstalledExtensions (): <....type-to-be-added...> {
```

jacogr: Prefer `function` as opposed to shortcuts.

Also ensure that each function has an explicit return, there are some edge-cases that popup every now and again where lack of return types end up with some "small" head-scratchers.

packages/apps/public/locales/en/apps.json
```diff
@@ -1,5 +1,6 @@
 {
   "Accounts": "Accounts",
+  "Connect": "Connect",
```
jacogr, line 3: It would prefer this under settings

(a) we already have it under accounts (when you have nothing)
(b) it is certainly not something that the user would visit often

packages/apps/src/Menu/ConnectButton.tsx
```diff
@@ -0,0 +1,140 @@
...
+import ConnectExtension from './ConnectExtension';
+
+const SHA_COL = 'rgba(34, 36, 38, 0.12)';
+const SHA_OFF = '5px';
+
+interface ConnectButtonProps {
+  className: string;
+}
+
+const ConnectButton = ({ className }: ConnectButtonProps) => {
```
jacogr, line 20: As per the comment on `function` vs `const`, this signature here should be -

```js
function ConnectButton ({ className }: Props): React.ReactElement<Props> {
...
```

... from here _really_ won't comment on these (as I suggested above already), however since it was the first component, may have been warranted here ...

packages/react-api/src/Api.tsx
```diff
@@ -81,24 +80,64 @@ function getDevTypes (): Record<string, Record<string, string>> {
...
+  const promisedExtensions: Promise<Injected>[] = [];
+  let enabledExtensions: Injected[] = [];
+  const extensionsInfo: InjectedExtensionInfo[] = [];
+  const extensionsWithInfo: InjectedExtension[] = [];
+
+  const supportedExtensions = supportedExtensionsNames().filter((supportedExtension) => !DISALLOW_EXTENSIONS.includes(supportedExtension));
+  const storageData = localStorage.getItem('enabledExtensions');
+  const storedEnabledExtensions: string[] | null = storageData !== null ? JSON.parse(storageData) as string[] : null;
+
+  Object.keys(injectedWindow.injectedWeb3).forEach((extensionName) => {
```
jacogr, line 121: This is really problematic - we are no only allowing 2 extensions to operate, since -

1. the one extension is ours
2. the second extension we seem to be playing favorites to

This means that anything else would be skipped. I don't want to go down the route of playing favorites or playing traffic-warned to which should be allowed. I'm 100% on-board that when teams have added their extension here, it has special treatment -

- allows links to install,
- it shows up properly configured,
- etc.

However if a team has not done that, I really don't want to disallow operation outright.

packages/react-api/tsconfig.build.json
```diff
@@ -3,7 +3,10 @@
   "compilerOptions": {
     "baseUrl": "..",
     "outDir": "./build",
-    "rootDir": "./src"
+    "rootDir": "./src",
+    "lib": [
+      "es2019"
```
jacogr, line 8: Please remove this - this is non-standard for the full project, hence problematic. (We aim for consistency as much as possible, not only within this repo, but also across all in polkadot-js, hence the shared configs)

jacogr (commented):
Some small initial comments.

packages/apps-config/src/extensions/index.ts
jsidorenko: > Prefer `function` as opposed to shortcuts.

May I ask you, what's the reason behind this?

packages/react-api/src/Api.tsx
hamidra: A solution that can be both fair and UX friendly would be to have the other injected extentions that are not added with full info under Others section. The full list of injected extensions are available in windows.injectedWeb3, so for regsitered extensions we can show the full info, for unrgistered extensions you can still get the name and version and show them in the dropdown list as an "Others" section. We still should have the DISALLOW_EXTENSIONS IMO to be able to filter the sketchy/scammy extensions.

packages/apps/public/locales/en/apps.json
SaltyCucumber: Connecting to an extension is basically the first step to use the dapp fully.
Most dapps that I know have a Connect button somewhere in the right top corner.
I didn't put it there cause there are other blocks there now, so decided next to settings would be best.
Most crypto users are very familiar with this pattern and doing it differently here would not be to our benefit.

Simple possible scenario: User enters polkadot-js/apps, looks for Connect, can't find it.
The page that is opened by default is "explorer", which for any new person is just a gibberish of random numbers and letters, quite useless, "I want to see ma balance, sers, not the matrix!", "Tom was right, this polkadot thing is for rocket scientists, I'm out" (I've omitted a lot of swearing from those quotes).

And that is how we lose new people. Since we are not a closed-door invite-only club, on the contrary my team's mission is to onboard the world and not just rocket scientists.
So hiding stuff that is important IMO not the way. Putting a large text in accounts hoping that user actually opens the accounts and reads it is also not the best thing ever I think.

Take into account that with this PR user needs to connect to each extension himself for the first time, this prevents multi-popup issue, this also lets user decide what extension he wants to use/install/connect.

hamidra:
I like the idea of having the user decide which extension they want to connect to and import accounts from, and avoid multi pop-ups. The downside I see with the current implementation is that the connect button is a small tab on top that might get lost between other tabs with no call to action to let the user know they need to connect extensions. So the user might get confused why their account is not showing up in the account page if they don't already know or pay attention that they need to connect the extensions. 
I'd suggest a call to action in the app to let the user know that they need to connect their extnesions. It can be in a form of a  modal window with the list of installed extensions, and popup when the user opens the app while has not connected to any extensions.
Also, on the account page on the top, there can be some info about which extensions are connected so the user would know where the accounts are imported from and be able to connect/disconnect the extensions.

jacogr:
As discussed elsewhere - the whole idea behind the extension interface is that the user certainly doesn't not have to have another "choose this" interface. It doesn't matter where the accounts come from, the interface should be routable to the correct extension.

packages/apps-config/src/extensions/index.ts
jacogr: This is for the sake of consistency - there is one way to write code so there is no overhead in thinking when maintaining it. So doesn't matter which file you touch, it looks and feels the same.

packages/apps/public/locales/en/apps.json
jacogr: Ok, it is not hiding stuff - we are no hiding stuff since we first expect people to install stuff and enable stuff individually, instead of whatever extensions he/she has installed "just working". There is no thinking. When he/she has an extensions, the accounts show up.

If he/she has nothing, they will be prompted on the accounts page (based on the extensions that have actually listed themselves there).

image

The current approach, as imaged here, adds friction - install extension, add account, then randomly connect it. The connection should "just work" when it "is there".

PR #7858, opened 2022-07-12
Add OmniBTC

wbh1328551759:
Hi, we will participate the crowdloan auction .


jacogr (commented):
I understand why the item is added to the Polkadot relay, unsure about the 2 to production and test - these are standalone without providers, so it adds no value.

wbh1328551759:
I understand what you mean, but I'm sorry, I'm not sure what information I need to add is OK, so I'm following someone else's pr to submit the code

packages/apps-config/src/endpoints/productionRelayPolkadot.ts
```diff
@@ -243,6 +243,16 @@ export const prodParasPolkadot: EndpointOption[] = [
...
     }
   },
+  {
+    info: 'omnibtc',
+    isUnreachable: true,
+    homepage: 'https://www.omnibtc.finance',
+    text: 'OmniBTC',
+    paraId: 2053,
+    providers: {
+      OmniBTC: 'wss://omnibtc.io/ws'
```
jacogr, line 253: We have 3 chains here with the exact same endpoint - this is not possible. It is either one of -

- a parachain
- a live network
- a test network

A single RPC cannot be linked to 3 different listings.

I would suggest just removing the entries in test networks and live networks and just keep it under Polkadot. (Which is where the chain lives)

Testing - these are solo test networks, running on their own
Production - these are solo production networks, running on their own

packages/apps-config/src/endpoints/production.ts
```diff
@@ -207,6 +207,13 @@ export const prodChains: EndpointOption[] = [
       OnFinality: 'wss://nodle.api.onfinality.io/public-ws'
     }
   },
+  {
+    info: 'ominibtc',
+    text: 'OminiBTC',
+    providers: {
+      OmniBTC: 'wss://omnibtc.io/ws'
+    }
+  },
```
jacogr, line 216:
```suggestion
```

Remove as per my comment below.

packages/apps-config/src/endpoints/testing.ts
```diff
@@ -369,6 +369,14 @@ export const testChains: EndpointOption[] = [
...
     }
   },
+  {
+    info: 'omnibtc',
+    isDisabled: true, // https://github.com/polkadot-js/apps/issues/6635
+    text: 'OmniBTC',
+    providers: {
+      OmniBTC: 'wss://omnibtc.io/ws'
+    }
+  },
```
jacogr, line 379:
```suggestion
```

Removed as per my comment above.

wbh1328551759:
Okay, thanks for the guidance, I've removed useless content from testing and production

jacogr (approved):
Thank you!

PR #7948, opened 2022-07-31
Update ListNominees.tsx

Nick-1979:
The max value to be replaced is missed in the useTranslation format in the tooltip while a validator is oversubscribed


jacogr (approved):
Thank you.

PR #7998, opened 2022-08-15
Change chain name from "nft" to "unique"

ashkuc:


packages/apps-config/src/api/spec/index.ts
```diff
@@ -226,7 +226,7 @@ const spec: Record<string, OverrideBundleDefinition> = {
   moonshadow: moonbeam,
   'mybank.network Testnet': mybank,
   neatcoin,
-  nft: unique,
+  unique,
```
jacogr, line 229: Would just require a re-order (I think this is where CI is complaining)

ashkuc: Yep, fixed

jacogr (approved):

PR #8003, opened 2022-08-16
A0-1026 update wallet for proper nomination pool support

krzysztofziobro:


jacogr:
This seems like a PR pointed to the wrong repo. (upstream instead of fork)

PR #8050, opened 2022-08-25
Parachain light clients

arrudagates:
This PR defines the URL structure for parachain light clients as `light://substrate-connect/relay/parachain`, adds the functionality in, and also includes 2 parachain light clients!


jacogr (commented):
Thank you for this. I like the idea, I just cannot get over the feeling that 

(a) the const paraOrRelay should be an in-line switch (no need for the extra function there)
(b) it will become very unmanageable, very soon

On the second point I don't have a better idea right now. Pissibly we could rather do something like (untested, messy) -

```js
import * as lightSpecs from './light-client-specs';

...
return paraArg && lightSpecs[paraArg]
  ? new ScProvider(JSON.stringify(lightSpecs[paraArg]), relay)
  : relay

arrudagates:
> Thank you for this. I like the idea, I just cannot get over the feeling that 
> 
> (a) the const paraOrRelay should be an in-line switch (no need for the extra function there)
> (b) it will become very unmanageable, very soon
> 
> On the second point I don't have a better idea right now. Possibly we could rather do something like (untested, messy) -
> 
> ```js
> import * as lightSpecs from './light-client-specs';
> 
> ...
> return paraArg && lightSpecs[paraArg]
>   ? new ScProvider(JSON.stringify(lightSpecs[paraArg]), relay)
>   : relay
> ```
> 
> So basically it only needs to be added as a spec and it will work (instead of adding it as a spec and then also adding it to a switch...)

About the first point, I'm a rust dev trying to write JavaScript 😅
I'll look into both today!

jacogr:
lol, no issues. TS is not quite Rust. Sometimes it will indeed drive you up the wall. It is what it is.

arrudagates:
Your suggestion actually worked perfectly, I just refactored the way the specs are imported and exported to more closely resemble the way things are done in apps-config.

jacogr:
Thank you - one question before I pull it down, have you connected to these chains to see if it does work? (The logic is correct exactly as the connect team suggests, the reason for asking is that I have had reports where parachains do not work via light)

arrudagates:
> Thank you - one question before I pull it down, have you connected to these chains to see if it does work? (The logic is correct exactly as the connect team suggests, the reason for asking is that I have had reports where parachains do not work via light)

Well, there were some issues, but they were all rectified before this PR was opened. This PR fixed the last issue that I had: https://github.com/paritytech/smoldot/pull/2686
And to give a more direct answer to your question, yeah, they work!
image
image

jacogr (approved):
Thank you!

arrudagates:
You better get ready for the masses to follow and start listing their parachain specs too hahaha!

PR #8149, opened 2022-09-20
Kabocha updates for Kusama and Rococo

decentration:
This PR updates Kabocha parachain on Kusama and Rococo.


jacogr (changes requested):
Some CI failures, see https://github.com/polkadot-js/apps/actions/runs/3089206310/jobs/4996552999#step:3:335

(Seems to align with the automated PR content around an unused variable)

packages/apps/public/locales/en/apps-config.json
```diff
@@ -13,5 +13,246 @@
   "rpc.header.test": "Test networks",
   "rpc.header.westend.relay": "Test Westend & parachains",
   "rpc.hosted.via": "via {{host}}",
-  "ss58.default": "Default for the connected node"
+  "rpc.kusama.altair": "Altair",
```
jacogr, line 16: I'm really not sure where these come from - the text for the chain name are not translated (anymore), so it should not build this at all.

jacogr: I would prefer not having these changes in - the reason being that record removal for these is (still) a manual process, so adding it means cleanup for me :)

decentration: Oh perhaps I misunderstood. Yes of course will remove. Though I’m not quite sure what I need to remove because the tests passed.

decentration: have removed the individual name chains that were added in replace for the `"ss58.default": "Default for the connected node"`

jacogr (approved):
Thank you!

PR #8173, opened 2022-09-27
Add hash to judgement call

jacogr:
As introduced in https://github.com/paritytech/substrate/pull/12170

cc @chevdor - please do give it a check as well


chevdor (approved):
If I understand right, you prefill the hash if an identity is set.
This is what I would expect from the UI indeed since the UI is used by humans and they should check the identity (likely not useing the hash). 

Regarding automation, it is fine not having a default value and let the automation profive the hash of what has been verified.

jacogr:
If you use the button for registrar approval, no additional  action needed. It you use the extrinsics tab, you are on your own…

PR #8182, opened 2022-09-29
Link inflation calculations source

jacogr:
... after I spent some time looking and digging, this should make future looks easier


packages/react-hooks/src/useInflation.ts
```diff
@@ -21,14 +21,28 @@ function calcInflation (api: ApiPromise, totalStaked: BN, totalIssuance: BN, num
...
+  console.log({
+    idealInterest,
+    idealStake,
+    inflation,
+    stakedFraction,
+    stakedReturn: stakedFraction
+      ? (inflation / stakedFraction)
+      : 0
+  }, stakeTarget, Math.min(auctionMax, numAuctions.toNumber()), Math.min(auctionMax, numAuctions.toNumber()) * auctionAdjust);
+
```
jacogr, line 45:
```suggestion
```

PR #8198, opened 2022-10-03
Add OnFinality endpoint for Invarch Tinker Network

banananeko:


packages/apps-config/src/endpoints/productionRelayKusama.ts
```diff
@@ -173,7 +173,8 @@ export const prodParasKusama: EndpointOption[] = [
     text: 'InvArch Tinkernet',
     providers: {
       'InvArch Team': 'wss://tinker.invarch.network',
-      'light client': 'light://substrate-connect/kusama/tinkernet'
+      'light client': 'light://substrate-connect/kusama/tinkernet',
+      OnFinality: 'wss://invarch-tinkernet.api.onfinality.io/public-ws'
```
jacogr, line 177: Can we keep light as the last entry? (It is somewhat special, not auto selected)

jacogr (approved):

jacogr:
Thank you, will figure out either auto-sorting for light or add comments on all of them. (It has not been a serious issue since there is only a couple of them, but at the same time have made the same comment twice now, so it just adds friction to PRs)

PR #8253, opened 2022-10-19
Bridgehub Rococo RPC

lazam:
New parachain for Rococo

### DIFF
--- a/packages/apps-config/src/endpoints/testingRelayRococo.ts
+++ b/packages/apps-config/src/endpoints/testingRelayRococo.ts
@@ -391,6 +391,15 @@ export const testParasRococoCommon: EndpointOption[] = [
       'Encointer Association': 'wss://rococo.api.encointer.org'
     },
     teleport: [-1]
+  },
+  {
+    info: 'rococoBridgehub',
+    paraId: 1013,
+    text: 'Bridgehub Rococo',
+    providers: {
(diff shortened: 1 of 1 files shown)


packages/apps-config/src/endpoints/testingRelayRococo.ts
```diff
@@ -391,6 +391,15 @@ export const testParasRococoCommon: EndpointOption[] = [
...
     teleport: [-1]
+  },
+  {
+    info: 'rococoBridgehub',
+    paraId: 1013,
+    text: 'Bridgehub',
+    providers: {
+      Parity: 'wss://rococo-bridge-hub-rpc.polkadot.io'
+    },
+    teleport: [-1]
```
jacogr, line 402: Remove the teleport. These are only for system/common chains.

jacogr (changes requested):

packages/apps-config/src/endpoints/testingRelayRococo.ts
jacogr: Or is it meant to be in that category?

lazam: It is still under discussion if they plan to implement a teleport into bridge hub parachains. I'll temporarily remove it for now, then add it when there's a decision.

jacogr (approved):

PR #8279, opened 2022-10-26
fix: rollback KILT custom type decorations

wischli:
Unfortunately, the KILT Protocol type definitions need to rolled back from version `0.*.*` as any `1.*.*` version (introduced in #8277) currently includes a minor bug and is therefore marked as deprecated.


jacogr (approved):
Thank you.

Issue #7812, opened 2022-06-30
Feature Request: Add `ChargeAssetTxPayment` Signed Extension for Sufficient Assets


joepetrowski:
When transacting from an account with 0 KSM, users should be able to pay fees using sufficient assets by specifying an `AssetId` in the [`ChargeAssetTxPayment` signed extension](https://github.com/paritytech/cumulus/blob/release-parachains-v9230/parachains/runtimes/assets/statemine/src/lib.rs#L599).

Currently the UI does not recognize this:

![image](https://user-images.githubusercontent.com/25483142/176691306-d97d262b-8240-45b0-be67-d9a130831a2d.png)

Request: If an account possesses any sufficient assets, add an option to choose the fee-paying asset (and if they own no KSM/DOT, it should default to one of them).

impelcrypto:
Hi @jacogr,
I'd like to know if you have a plan to solve this issue. I'm also facing the same issue and hope to implement the request above on Polkadot.js.

jacogr:
It is in the queue, so will bubble to the top of the list at some point. 

Interest noted.

fiexer:
This would make the USDT/C coming in from Statemine/Statemint so much more user-friendly to be used in XCM.
Looking forward having this available 🙏 . Thanks for all your work you have been doing for the entire dotsama ecosystem @jacogr

joepetrowski:
Any chance this could get done in the near future? USDT is `sufficient` on Statemint from ~10 Dec and would be nice to have this support.

jacogr:
Will only be able to be done after I can get my hands on sufficient assets (it needs to be tested when done, cannot throw up untested code that could/couldn’t work)

Judging by the queue, unlikely to be looked at by me this year since there are tight 3rd-party deliveries elsewhere.

fiexer:
Feel free to contact me @jacogr if you need assets to test.
Do share your Statemint address so I can send you some USDt.

jacogr:
@fiexer Thanks for the offer. My comment was certainly not "begging for tokens" - I will convert some for testing when the time comes and fund.

marshacb:
[...]

PR #8324, opened 2022-11-07
[Spike] Update `PaymentInfo` to use `Assets` pallet for fee asset if applicable

aidan-starke:
## The problem
 - At `futureverse` we use an asset from our Assets pallet to handle all fees for the network. The number displayed in `PaymentInfo` is correct, but the symbol is not, and the fee error can trigger when it shouldn't

## The solution
 - Add a `feeAssets` map in `@polkadot/apps-config` where any chain can add their fee asset's details
 - Add a hook that attempts to fetch the fee asset balance based on that map
 - If that hook finds anything, use that data for the fee error and `feesForSubmission` section of `PaymentInfo`

### DIFF
--- a/packages/apps-config/src/settings/feeAssets.ts
+++ b/packages/apps-config/src/settings/feeAssets.ts
@@ -1,0 +1,13 @@
+// Copyright 2017-2022 @polkadot/apps-config authors & contributors
+// SPDX-License-Identifier: Apache-2.0
+
+import type { ChainSpecName, FeeAsset } from './types';
+
+// A mapping of chains to their fee asset if different to the native token
+export const feeAssets: Record<ChainSpecName, FeeAsset> = {
+  root: {
+    assetId: 2,
+    decimals: 6,
+    symbol: 'XRP'
+  }
+};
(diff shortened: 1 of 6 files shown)


aidan-starke:
@jacogr keen to get some feedback on this - I've had an idea that we could add in the module/call to the `feeAssets` mapping to keep the `useFeeAssetBalance` hook generic, i.e.

```
root: {
  module: "assets",
  call: "account",
  feeAsset: {
    assetId: 2,
    decimals: 6,
    symbol: 'XRP'
  }  
```

evilrobot-01:
Additional hacky implementation based on Aidan's great work available at https://github.com/evilrobot-01/polkadot-js-apps for anyone interested. It still uses the static assets defined in `feeAssets.ts`, but adds an option to select an asset in the UI, which is then passed through via [SignerOptions](https://github.com/polkadot-js/api/blob/46076c5595ab62e960a1097611a3e150bfa942f2/packages/api-base/src/types/submittable.ts#L19) when sending the signed transaction.

jacogr:
We certainly need UI support, so PRs would be appreciated for "proper" support, see https://github.com/polkadot-js/apps/issues/7812

jacogr (commented):
On first glance, it certainly look above-board, although very localized to a specific instance. But it moves in the right direction, so it is good. 

Lemme take a proper look and pull down/play with it.

ap211unitech:
Closing this PR as it appears to be outdated. Feel free to create a new PR with updated details, if required.

PR #8330, opened 2022-11-10
Astar first wss bootnode

gluneau:
Result of a proof of concept that wss do work with light clients as bootnodes.


gluneau:
Hello @jacogr could you look at this one, when you have a second, Thanks!

jacogr (approved):
All good.

Sorry about the delay here, the notifications *always* get lost when moving from draft -> ready.

PR #8346, opened 2022-11-14
Create dependency-review.yml

lilsunny243:


jacogr:
Thank you, however these won't be merged since it is not tools we can verify or use elsewhere.

PR #8385, opened 2022-11-21
Add subsquare external links for referenda

wliyongfeng:


jacogr (approved):
Will add the extra header column in a follow-up (so display will be a tiny bit scewed atm, but nothing major)

PR #8396, opened 2022-11-22
Minor updates

isunaslabs:
-Added Darwinia's Ethereum Compatibility validation


jacogr:
Not quite minor :) (just due to the fact that this forces all accounts into ETH accounts)

jacogr (approved):

Issue #8174, opened 2022-09-27
Adjust Polkadot inflation with auctions (once enabled)


jacogr:
See https://github.com/paritytech/polkadot/pull/5872

kianenigma:
and how is this number being computed? the comment I see from Gav when he first changed the inflation rate logic says: 

```
// 30% reserved for up to 60 slots.
let auction_proportion = Perquintill::from_rational(auctioned_slots.min(60), 200u64);

// Therefore the ideal amount at stake (as a percentage of total issuance) is 75% less the
// amount that we expect to be taken up with auctions.
let ideal_stake = Perquintill::from_percent(75).saturating_sub(auction_proportion);
```

So with 41 parachains, Kusama's ideal stake should be `75 - ~20 = 55%`.

kianenigma:
okay, I guess the issue is that `auctions::auctionCounter` is mistakenly used (54), and it is wrong. From where is the apps getting the right number of "active parachains" (41, not 54)?

jacogr:
Using auctionCounter is correct.

The discrepancy is most-probably related to this - https://github.com/paritytech/polkadot/blob/816cb64ea16102c6c79f6be2a917d832d98df757/runtime/kusama/src/lib.rs#L526-L527

The [param config is literally 30%](https://github.com/polkadot-js/apps/blob/0701ece72f290ac4bd8bcd24d6e63ddf73547a48/packages/apps-config/src/api/params/inflation.ts#L28) as per the comment as opposed to using `200u6` as the divisor.

Suggestion for the calcs here are always appreciated (probably not this issue, since it related to functionality adjustment, not "it doesn't quite line up") - 

- calcs - https://github.com/polkadot-js/apps/blob/0701ece72f290ac4bd8bcd24d6e63ddf73547a48/packages/react-hooks/src/useInflation.ts#L19-L41
- params - https://github.com/polkadot-js/apps/blob/0701ece72f290ac4bd8bcd24d6e63ddf73547a48/packages/apps-config/src/api/params/inflation.ts#L28

kianenigma:
Any updates on this? it has caused some confusion here and there.

TarikGul:
As a suggestion by Alex Birdo, probably best to just remove the confusion by removing the text for now then, we can add it back in when corrected.

kianenigma:
Yes, happy to see this removed.

PR #8400, opened 2022-11-23
Cater for new generation inflation calcs

jacogr:
- [x] use public paras as input (instead of auctions)
- [ ] re-check calc logic against current generation (i.e. needs 1-to-1 mapping, so don't rely on existing)

Closes https://github.com/polkadot-js/apps/issues/8174


kianenigma:
Looking forward to this -- the status quo is kinda confusing.

PR #8413, opened 2022-11-24
add back OnFinality endpoints for Parallel, Heiko, and Clover

banananeko:


jacogr (approved):

jacogr:
None of these are working - I'm disabling them again on the back of #8421

PR #8425, opened 2022-11-25
Add custom signed extension for the enconter chains and update encointer types.

### LINKED ISSUE #7545: [encointer] can't decode block if it contains a tx payed with a non-native asset.
See: https://github.com/encointer/encointer-parachain/issues/127

The reason here is that `polkadot-js/api` does not generate the signed extensions on the fly; the `AssetId` of the `OnChargeAssetTxPayment` is hardcoded to `Option<Number>`, while we have `Option<CommunityIdentifier>`.

On the client side, I solved this with the following code-block:

```
const api = await ApiPromise.create({
  ...options({
    types: configOverride.types !== null ? configOverride.types : {}
  }),
  signedExtensions: {
    ChargeAssetTxPayment: {
      extrinsic: {
        tip: 'Compact<Balance>',
        assetId: 'Option<CommunityIdentifier>'
      },
      payload: {}
    }
  },
  provider: provider
});
```

As far as I know, I can no longer set custom types in polkadot-js/apps. So how can I get this change there?

clangenb:
* Closes #7545


jacogr (approved):
Thank you.

PR #7807, opened 2022-06-29
Add Amplitude chain to Kusama logos and chain info

gonzamontiel:
The main purpose is to be able to see our chain's name and logo for the Kusama crowdloan, parathread, and parachain pages.

Screenshot from 2022-06-29 17-39-44

PR #8451, opened 2022-11-30
Add Pendulum chain (unreachable) & logo

gonzamontiel:
The main purpose is to be able to see our chain's name and logo for the Polkadot crowdloan, parathread, and parachain pages.
Please that the endpoint is set to Unreachable until we launch.

As reference, similar PR for Kusama: https://github.com/polkadot-js/apps/pull/7807


jacogr (approved):
All good. Thank you - this approach is the best indeed. If there is no RPC endpoint, doesn't need to be active. (If it ever becomes active, and the team wants it reachable, can adjust)

PR #8547, opened 2022-12-15
Enable staking for other chains with different account type

KarishmaBothara:
Enabling staking for chain where account id is not 32 bytes in length ( its 20 bytes in length ethereum id)


jacogr (commented):
Not sure how this would solve anything - `PalletStakingExposure` is provided by the metadata, e.g. for AccountId20 types, it would be indicated as such in the metadata. So the check would still fail since the eq should detect a mismatch on AccountId20 vs AccountId32.

PR #8605, opened 2022-12-23
Fix ChainImg logo to match chain

camfairchild:
This PR fixes the ChainImg component when the `logo` prop is not specified. Previously this would always be an empty logo, using the chain color only, but this change uses the defined chain logo, if it matches, as intended.


camfairchild:
See current (Polkadot):  
image   

Change:  
Screen Shot 2022-12-22 at 9 49 03 PM

jacogr (approved):
Thank you.

PR #8660, opened 2022-12-30
Replace Kusama chain logo with a high quality svg

Nick-1979:
The kusama chain logo was a gif with a very low quality, since replaced with a high quality svg.


jacogr (approved):
As always, thank you.

Issue #8458, opened 2022-12-01
e.derive.balances.account is not a function


sander2:
When we use [Chopsticks](https://github.com/AcalaNetwork/chopsticks) to fork a kintsugi parachain for testing purposes, we get `e.derive.balances.account` when we try to submit transactions. I'm guessing that the overrides are not properly being detected, even though the chainspec name is the same on mainnet. Any ideas?

jacogr:
No idea really. If the specName matches, it should detect all the balance pallets (& derives)

sander2:
Addressed by https://github.com/polkadot-js/apps/pull/8683

PR #8683, opened 2023-01-03
fix: add account derive

sander2:
Without this, we run into [this issue](https://github.com/polkadot-js/apps/issues/8458) when we submit transactions on a chopsticks-ed chain. To be honest I don't understand why this change is required on chopsticks and not on mainnet

### DIFF
--- a/packages/apps-config/src/api/spec/interbtc.ts
+++ b/packages/apps-config/src/api/spec/interbtc.ts
@@ -14,6 +14,7 @@ import { DeriveBalancesAll } from '@polkadot/api-derive/types';
 import { memo } from '@polkadot/api-derive/util';
 import { TypeRegistry, U128 } from '@polkadot/types';
 import { Balance } from '@polkadot/types/interfaces';
+import { FrameSystemAccountInfo } from '@polkadot/types/lookup';
 import { BN, formatBalance } from '@polkadot/util';
 
 function balanceOf (number: number | string): U128 {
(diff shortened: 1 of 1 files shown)


packages/apps-config/src/api/spec/interbtc.ts
```diff
@@ -49,10 +49,11 @@ export function getBalance (
   return memo(
     instanceId,
     (account: string): Observable<DeriveBalancesAll> =>
-      combineLatest<[any]>([api.query.tokens.accounts(account, { Token: nativeToken })]).pipe(
-        map(([data]: [OrmlAccountData]): DeriveBalancesAll => {
+      combineLatest<[any, any]>([api.query.tokens.accounts(account, { Token: nativeToken }), api.query.system.account(account)]).pipe(
+        map(([data, systemAccount]: [OrmlAccountData, any]): DeriveBalancesAll => {
           return {
             ...defaultAccountBalance(),
+            accountNonce: new BN(systemAccount.nonce),
```
jacogr, line 56:
```suggestion
            accountNonce: api.registry.createType('Index', systemAccount.nonce),
```

I guess the (build) issue is that we have `any` here, hence it doesn't know what to do.

The more-correct way is probably to -

```js

import { FrameSystemAccountInfo } from '@polkadot/types/lookup';

...
map(([data, systemAccount]: [OrmlAccountData, FrameSystemAccountInfo]): DeriveBalancesAll => {
```

So the typings are just correct and no faffing around with createType

jacogr (dismissed):

packages/apps-config/src/api/spec/interbtc.ts
sander2: Thanks for the pointer! I updated the PR

jacogr (approved):
Thank you.

PR #8686, opened 2023-01-04
update ipfs endpoint domain

BOBOKaa:


BOBOKaa:
The Crust separates the ipfs upload and download domain.
so it needs to be updated

PR #8696, opened 2023-01-05
chore: replace spec-based with name-based declarations for KILT

ntn-x2:


packages/apps-config/src/api/spec/index.ts
```diff
@@ -208,7 +207,6 @@ const spec: Record<string, OverrideBundleDefinition> = {
   'jur-chain': jur,
   kerria: parallel,
   khala,
-  'kilt-spiritnet': kilt,
```
jacogr, line 211: The changes as applied look fine - do all the chains above have this spec_name as part of the runtime version? If so, it could just be mapped on spec_name, same as what is done with the chain name now.

ntn-x2: And the `kilt-spiritnet` chain spec only identifies our live chain.

jacogr (approved):

PR #8716, opened 2023-01-07
bump @polkadot/* deps

ETeissonniere:
Bump `@polkadot/*` packages deps to take in account the latest changes in polkadot-js/common#1699 which isn't included otherwise. This is needed to allow nodle users to recover their coins hosted in a ledger in the event Ledger does not upgrade their integration on time.


yarn.lock
```diff
@@ -3981,6 +4039,18 @@ __metadata:
...
+  resolution: "@polkadot/x-fetch@npm:10.2.2"
+  dependencies:
+    "@babel/runtime": ^7.20.7
+    "@polkadot/x-global": 10.2.2
+    "@types/node-fetch": ^2.6.2
+    node-fetch: ^3.3.0
+  checksum: 8265dd4ebdb322e883094a5b3d4126b4443bf36a9121a0a466e1a05369366890bcf14185e23da979b61431167ae6f9500c81e4fb85ec15ce47a3c74f750a1bef
+  languageName: node
+  linkType: hard
+
```
jacogr, line 4053: Something is weird where not all deps have gone up, i.e. see the ones just below, still on 10.2.1

jacogr:
Thank you - but due to my comment above, I've reworked it slightly in https://github.com/polkadot-js/apps/pull/8720 and will rather use that one.

PR #8717, opened 2023-01-07
Align staking layout with accounts

jacogr:
This is in playing around mode at this point - will see if it shapes up or not over some time...


jacogr:
Replaced by https://github.com/polkadot-js/apps/pull/8822

PR #8490, opened 2022-12-06
Memoize use* values

jacogr:

PR #8749, opened 2023-01-10
Add & use useMemoValue

jacogr:
Inspiration from https://github.com/polkadot-js/apps/pull/8490 (which is now defunct)

We will probably not merge this, just pull pieces from it like we have done in #8750 (since each and every one needs some extensive testing to ensure it is still on-spec and doesn't create negative effects with any areas)


packages/react-hooks/src/ctx/ApiStats.tsx
```diff
@@ -67,6 +64,16 @@ function getStats (...apis: ApiPromise[]): { stats: ProviderStats, when: number
...
 export const ApiStatsCtx = React.createContext<ApiStats[]>(EMPTY_STATE);
 
+function addStats (prev: ApiStats[], curr: ApiStats): ApiStats[] {
+  if (prev.length === 0) {
+    return [curr];
+  }
+
+  return prev.length === MAX_NUM
+    ? prev.concat(curr).slice(-MAX_NUM)
+    : prev.concat(curr);
```
jacogr, line 74:
```suggestion
  return prev.length === 0
    ? [curr]
    : prev.length === MAX_NUM
      ? prev.concat(curr).slice(-MAX_NUM)
      : prev.concat(curr);
```

packages/react-hooks/src/useAccountInfo.ts
```diff
@@ -247,7 +247,7 @@ function useAccountInfoImpl (value: string | null, isContract = false): UseAccou
 
   const isEditing = useCallback(() => isEditingName || isEditingTags, [isEditingName, isEditingTags]);
 
-  return useMemo(() => ({
+  return {
```
jacogr, line 250:
```suggestion
  return useMemo(() => ({
```

packages/react-hooks/src/useEventTrigger.ts
```diff
@@ -33,6 +33,7 @@ function useEventTriggerImpl (checks: EventCheck[], filter: (record: EventRecord
   const memoChecks = useMemoValue(checks);
   const mountedRef = useIsMountedRef();
   const eventRecords = useCall<Vec<EventRecord>>(api.query.system.events);
+  const memoChecks = useMemoValue(checks);
```
jacogr, line 36:
```suggestion
  const memoChecks = useMemoValue(checks);
```

Added above.

packages/react-hooks/src/useExtrinsicTrigger.ts
```diff
@@ -20,6 +20,7 @@ function useExtrinsicTriggerImpl (checks: ExtrinsicCheck[]): string {
   const mountedRef = useIsMountedRef();
   const memoChecks = useMemoValue(checks);
   const block = useCall<SignedBlockExtended>(api.derive.chain.subscribeNewBlocks);
+  const memoChecks = useMemoValue(checks);
```
jacogr, line 23:
```suggestion
```

Added above.

Issue #8702, opened 2023-01-06
Can not distinguish account type.


AurevoirXavier:
<img width="1474" alt="image" src="https://user-images.githubusercontent.com/16152305/210919011-8168cddc-0669-496e-9432-989ae7b5adf7.png">

It always truncates the `from` account id 32 to account id 20.
I can't input a 32 bytes data.

jacogr:
Yes, it (currently) assumes that all `AccountId` on-chain are the same from an input perspective. 

Which chain is that - since it would be good to have a testcase for any possible 20 + 32 adjustments.

AurevoirXavier:
> Yes, it (currently) assumes that all `AccountId` on-chain are the same from an input perspective, i.e. maps them to the same underlying (single) input.
> 
> Which chain is that - since it would be good to have a testcase for any possible 20 + 32 adjustments.

It's still a testnet.
https://polkadot.js.org/apps/?rpc=ws%3A%2F%2Fg1.dev.darwinia.network%3A20000#/explorer

AurevoirXavier:
@jacogr Hey, is there an estimated time? We are considering renaming the struct to make it work.

jacogr:
This is not easily solvable at all - all the Account dropdowns are backed by the keystore which is set to the default account type for the chain. So effectively to support both types we need something completely non-backed.

(The second type will probably end up as something like we have in the validators graph query, i.e. only allow inputs, no pre-defined selections)

AurevoirXavier:
> (The second type will probably end up as something like we have in the validators graph query, i.e. only allow inputs, no pre-defined selections)

Yes, that's enough for us.

Thanks for your quick response and work.

PR #8752, opened 2023-01-10
Params support for non-native AccountId{20, 32}

jacogr:
Closes https://github.com/polkadot-js/apps/issues/8702

It allows for non-native inputs via a SimpleAddress bridge. There may be dragons lurking since this is not something we actually set out for. (Same with the Ethereum support which is bare-bones)

On the testnet linked in the issue above -

image


AurevoirXavier:
image

But it looks like this component still not work.

jacogr:
With a properly formatted address -

image

PR #8754, opened 2023-01-10
Reformat active balance inputs to all decimals

jacogr:
Additional follow-up tweak after https://github.com/polkadot-js/apps/pull/8753 - basically in this case for active inputs we change the formatting behavior to include all decimals.

As a bonus - all this stuff is hard to follow, so tried to give some sense to it with explaining the what and why of the different steps.


jacogr:
This would make it easier - https://github.com/polkadot-js/common/pull/1703

jacogr:
Marked WIP (although it works), it would make sense to just use the functionality from `formatBalance` directly once we have a release & update.

Issue #8837, opened 2023-01-17
Decoder no longer accepts full blobs


carlosala:
* **I'm submitting a ...**

  - [x] Bug report
  - [ ] Feature request
  - [ ] Support request
  - [ ] Other

* **What is the current behavior and expected behavior?**

Until 9c4cbfcd67311b97a7aa7def48b315b2b7eda2c2, we were able to input in the decoder a full transaction blob, which included both `Call` and also spec version, tx version, etc.
Now, it only accepts the `Call` hex-encoded type.

* **What is the motivation for changing the behavior?**

It'd help us in order to test the Ledger integrations, as we test directly the whole blobs that get to the Ledger to see which call is in it and if we are parsing properly.

* **Please tell us about your environment:**

  - Version:
  - Environment:

    - [x] Node.js
    - [ ] Browser
    - [ ] Other (limited support for other environments)

  - Language:

    - [ ] JavaScript
    - [x] TypeScript (include tsc --version)
    - [ ] Other

carlosala:
[This](https://github.com/polkadot-js/apps/blob/5eec74f1172ac0f94e8e6ce0678c4c771117b1dc/packages/page-extrinsics/src/Decoder.tsx#L77) is the problematic line, as obviously decoded is only using the first part of the blob, and not all the characters.

jacogr:
If you follow the logic there -

1. it first tries to decode as an extrinsic, if that fails
2. try to decode as a call

The line you pointed to is when both fails.

To, here is an actual recent transaction - https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Fpolkadot.public.curie.radiumblock.co%2Fws#/extrinsics/decode/0x05030016991bfe44c8dce7569a74ce7a168fb2d4d176e90e754d4e8a92d26d4870dc600740b149df12

jacogr:
Ahhh, I see what you mean - you want to pass extra characters in. That is quite problematic. (And ended up in the previous position where we ended up decoding invalid stuff without knowing about it... which is why the check was added)

carlosala:
@jacogr for us would be quite interesting to be able to do it. If it makes a lot of trouble, we'd try to make some workaround

jacogr:
We need strict checks. 

If you wish to also supply transaction payloads, we would need an explicit decoding step for that in the above code block. (Also strictly failing like the others if the end result is problematic)

carlosala:
[...]

PR #8840, opened 2023-01-18
Attempt to decode ExtrinsicPayload in decoder

jacogr:
As per https://github.com/polkadot-js/apps/issues/8837

Seems to do the trick, have tested with -

1. Call
2. Extrinsic
3. Un-prefixed payload

Have not done prefixed payloads, but if unprefixed work, it _should_ be sane


carlosala:
It works perfect for me, thanks Jaco!!!

PR #8842, opened 2023-01-18
Motion page to display motion with no preimage

peterwht:
In the next week or two, we will be assisting in the upgrade of the [Watr devnet](https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Frpc.dev.watr.org#/explorer) to include a simple governance. The Mainnet will be launched soon as well. 

This governance only uses a collective, does not have Democracy, and does not rely on preimages. 

As is, the Polkadot.js UI only displays the motion hash for this governance. This PR adds a small change that will display the motion in text if democracy and the preimage does not exist.

Thank you for your help!

### DIFF
--- a/packages/page-democracy/src/Overview/ProposalCell.tsx
+++ b/packages/page-democracy/src/Overview/ProposalCell.tsx
@@ -32,8 +32,8 @@ function ProposalCell ({ className = '', imageHash, proposal }: Props): React.Re
   // while we still have this endpoint, democracy will use it
   const democracy = api.query.democracy;
 
-  const displayProposal = democracy ? 
-    democracy.preimages
+  const displayProposal = democracy
+    ? democracy.preimages
       ? proposal
       : preimage?.proposal
     : proposal;
--- a/packages/page-democracy/src/Overview/ProposalCell.tsx
+++ b/packages/page-democracy/src/Overview/ProposalCell.tsx
@@ -30,9 +30,13 @@ function ProposalCell ({ className = '', imageHash, proposal }: Props): React.Re
   const preimage = usePreimage(imageHash);
 
   // while we still have this endpoint, democracy will use it
-  const displayProposal = api.query.democracy?.preimages
-    ? proposal
-    : preimage?.proposal;
+  const democracy = api.query.democracy;
+
+  const displayProposal = democracy ? 
+    democracy.preimages
+      ? proposal
+      : preimage?.proposal
+    : proposal;
 
   if (!displayProposal) {
     const textHash = imageHash.toString();


packages/page-democracy/src/Overview/ProposalCell.tsx
```diff
@@ -32,8 +32,8 @@ function ProposalCell ({ className = '', imageHash, proposal }: Props): React.Re
   // while we still have this endpoint, democracy will use it
   const democracy = api.query.democracy;
 
-  const displayProposal = democracy ? 
-    democracy.preimages
+  const displayProposal = democracy
+    ? democracy.preimages
```
jacogr, line 36: The `proposal` (which is from an `api.derive`) used the same underlying preimage logic as pre-defined here before the `const democracy = ...` introduction.

- if democracy.preimages, retrieve from there
- if preimage pallet, retrieve from there
- else go empty

That logic would align using simple `?.`. (Adjusted for optional preimage pallet as well)

```js
const displayProposal = api.query.democracy?.preimages
  ? proposal
  : preimage?.proposal;
```

peterwht: Thank you for the response!

The issue is that with using the `democracy?.` only the proposal hash is displayed. But with the code I changed it to, the motion displays properly. So `proposal` is not always undefined. 

I also found this ([here](https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Fpolkadot-rpc.dwellir.com#/council/motions))
Screen Shot 2023-01-18 at 1 34 05 PM

The second row only displays the proposal hash. This is on Polkadot. The code I have does not work for this either. 

The following code does work to fix that issue on Polkadot, and the issue we are having.
```
  const displayProposal = api.query.democracy?.preimages
  ? proposal
  : preimage?.proposal 
    ? preimage.proposal
    : proposal;
```
(could probably be refactored, but first I want to check if that is a reasonable approach).

Here is what Polkadot looks like with that change:
Screen Shot 2023-01-18 at 1 33 56 PM

jacogr:
On Polkadot it should not display. 

The UI is meant to indicate what will happen when it executes. 

So if the chain uses democracy and it has a preimage submitted to the preimage pallet it should not display. (There may be an image on-chain, but it doesn’t mean it is in the correct place)

peterwht:
Ah, okay. That makes sense. 

So, if neither option can be pulled in, do you have a recommended approach I can take to get this resolved? 

As mentioned, I have tested the code changes in this PR. It works for the chain we are working with, as well other chains, such as Polkadot. From what I can tell, it should not have any unintended side-effects for other chains.

Is it a matter of:
- The logic has unintended consequences?
- The code needs to be cleaner / refactored?
- Or something else?

The current code:
```
const displayProposal = api.query.democracy?.preimages // -> undefined -> false
  ? proposal // -> not reached (in our case)
  : preimage?.proposal; // -> undefined
```

follows this flow
- check if democracy exists -> false
- check preimage pallet for proposal -> does not exist
- displayProposal = undefined

The flow we need is:
- check if democracy exists -> false
- set `displayProposal = proposal`

I appreciate your help on this 😊🙏

jacogr:
For your case where does it get the proposal data from to show it?

peterwht:
I have not looked outside the scope of the `ProposalCell` component, but I am just using the `proposal` argument.

```
//.................................................. this one
function ProposalCell ({ className = '', imageHash, proposal } 
```

jacogr:
Yes, proposal is the decode bytes of the preimage. Hence asking how you are storing images on chain.

peterwht:
Ah, I see what you are asking now.

We are storing the proposals directly in `pallet-collective`.

The motions are created via Governance -> Council -> Motions -> Propose Motion. 

Chain state shows that there is nothing in `pallet-preimage` when doing it this way.

jacogr:
So, yes - that is the reason for my question. `proposal` contains the decoded preimage. So it needs to be retrieved from somewhere.

There is a doubling up of logic here -

- `api.derive.democracy.preimages` retrieves the preimage from either democracy or preimages pallet
- `usePreimage` only retrieves from the preimages pallet

If there is no on-chain preimage in either of those locations, the `proposal` would be empty.

peterwht:
🎉 Thank you!! I appreciate the patience you gave us in figuring this out. I just tested, and it works great 😊

PR #8898, opened 2023-01-31
Create dependency-review.yml

lilsunny243:


jacogr:
I'm going to close this one -

1. This is changing the actions (with no pinning to the known-working and checked actions)
2. The PR has no context provided as to why this is a good idea
3. This seems like a fly-by from a new contributor (while new contributions are really appreciated, changing the CI guts without any background is suspicious)

PR #8908, opened 2023-01-31
Add Missing OnFinality Networks

jamesbayly:
@jacogr we have implemented rate limits as requested by network owners that restrict access to these public endpoints to `polkadot.js.org`. We believe that your monitoring that is indicating these endpoints are down due to your monitoring solution being blocked by this rate limit. Can you please let us know what origin or IP you are making these monitoring requests from so we can allow list it

### DIFF
--- a/packages/apps-config/src/endpoints/productionRelayKusama.ts
+++ b/packages/apps-config/src/endpoints/productionRelayKusama.ts
@@ -121,7 +121,8 @@ export const prodParasKusama: EndpointOption[] = [
     paraId: 2105,
     text: 'Darwinia Crab',
     providers: {
-      'Darwinia Network': 'wss://crab-parachain-rpc.darwinia.network/'
+      'Darwinia Network': 'wss://crab-parachain-rpc.darwinia.network/',
+      OnFinality: 'wss://darwinia-crab.api.onfinality.io/public-ws'
     }
   },
   {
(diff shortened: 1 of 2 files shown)


jacogr:
It is automated, running from GitHub CI. Checks run twice daily, once an endpoint shows up unreachable 4 times (48 hours), it gets removed.

jacogr:
So this gives issues, see https://github.com/polkadot-js/apps/issues/8921

jamesbayly:
Yes, your CI runner is certainly getting rate limited https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Fparallel.api.onfinality.io%2Fpublic-ws#/explorer

Is the CI runner GitHub-hosted or self-hosted?

jamesbayly:
And are you able to share the code for the CI runner with me so we can figure out a way to allowlist it without allowing every requests from every runner across GitHub?

jacogr:
GH hosted, so they will spin up an instance on an interval via https://github.com/polkadot-js/apps/blob/master/.github/workflows/chain-endpoints.yml 

The code that does the check is here - https://github.com/polkadot-js/apps/blob/master/packages/apps-config/src/ci/runner.ts

(Which is called from https://github.com/polkadot-js/apps/blob/master/packages/apps-config/src/ci/chainEndpoints.spec.ts via the above workflow)

PR #8947, opened 2023-02-06
fix: bad logo after PR 8941

darkfriend77:
It seems like the Bajun Logo got wrongly changed to the Ajuna Logo. Hopefully, this will fix the change.

Kusama and Rococo are running under the Baju Network (nodesBajunPNG)


jacogr (approved):
Umpf. Sorry about that.

PR #8985, opened 2023-02-09
Add Polkadot Bridgehub

lazam:
New Polkadot Common Good Parachain: Bridgehub


packages/apps-config/src/endpoints/productionRelayPolkadot.ts
```diff
@@ -624,6 +624,17 @@ export const prodParasPolkadotCommon: EndpointOption[] = [
...
+  },
+  {
+    info: 'polkadotBridgeHub',
+    paraId: 1002,
+    providers: {
+      Parity: 'wss://polkadot-bridge-hub-rpc.polkadot.io'
+    },
+    text: 'BridgeHub',
+    ui: {
+      logo: nodesBridgeHubBlackSVG
```
serban300, line 636: We should use the `nodesBridgeHub` logo here. The `nodesBridgeHubBlack` one is for Kusama.

serban300:
> Do we merge as-is or change the logo?
> 
> Sadly I dropped the image initially uploaded when I cleaned up since it was unused - sorry about that. Having said that, it should be simple to add - https://github.com/polkadot-js/apps/tree/master/packages/apps-config#endpoints
> 
> TL;DR
> 
>     1. Add file to apps-config/ui/logos/nodes
> 
>     2. Run `yarn build:images`
> 
>     3. Change import as suggested above

However you prefer. I'm also ok with merging as-is, and then changing the logo in another PR. 

Attaching the image to this message just in case it's needed.

BridgeHub

jacogr (approved):
Merging as-is. Follow-up PRs with an updated logo welcome.

PR #9003, opened 2023-02-10
Change Polkadot BridgeHub logo

serban300:
Follow-up on #8985 

Also change the Westend bridge hub logo since it's my understanding that it will mirror Polkadot bridge hub.


jacogr (approved):
Thank you.

PR #9009, opened 2023-02-13
Update ChainEndpoints Check task

jamesbayly:
Hi @jacogr,

We're not finding an easy way to whitelist just your automated CI for our rate limited endpoints.
We suspect that one of the big issues is that every fork (heres mine doing it https://github.com/jamesbayly/apps/actions/runs/4159421728) will run this automated check task at the same time, so we might have tens if not hundreds of calls from GItHub at the same time (that get rate limited) as each fork executes the check.

I've added a line that only runs the endpoints check in your repo (not on forks) and changes the time it does this to slight later in the 12 hour window since i can't change code in already forked codebases

Let me know if you have any other ideas that we should pursue

James

### DIFF
--- a/.github/workflows/chain-endpoints.yml
+++ b/.github/workflows/chain-endpoints.yml
@@ -5,7 +5,7 @@ on:
 
 jobs:
   types:
-    if: (github.event_name == 'schedule' && github.repository == 'polkadot-js/apps') || (github.event_name != 'schedule')
+    if: github.repository == 'polkadot-js/apps'
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c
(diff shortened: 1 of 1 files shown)


jacogr:
fkit. Ummm... I have a number of cron jobs, didn't think about forks at all on those.

.github/workflows/chain-endpoints.yml
```diff
@@ -1,10 +1,11 @@
 name: Chain endpoints
 on:
   schedule:
-    - cron:  '45 0/12 * * *'
+    - cron:  '50 0/12 * * *'
 
 jobs:
   types:
+    if: (github.event_name == 'schedule' && github.repository == 'polkadot-js/apps') || (github.event_name != 'schedule')
```
jacogr, line 8:
```suggestion
    if: (github.event_name == 'schedule' && github.repository == 'polkadot-js/apps')
```

Would just keep it at schedule. Can be run manually (locally) if need-be.

jacogr (approved):

jamesbayly:
Adding a key is an option, but it would prefer to exhaust other options first to continue just testing the public OnFinality endpoints without the addition of any keys

jacogr:
Indeed.

PR #9012, opened 2023-02-13
Remove Pinknode Westend Endpoints

Shawn-Eng-888:


packages/apps-config/src/endpoints/testingRelayWestend.ts
```diff
@@ -193,7 +193,6 @@ export const testRelayWestend: EndpointOption = {
     'IBP Network': 'wss://rpc.ibp.network/westend',
     OnFinality: 'wss://westend.api.onfinality.io/public-ws',
     Parity: 'wss://westend-rpc.polkadot.io',
-    Pinknode: 'wss://rpc.pinknode.io/westend/explorer',
```
jacogr, line 196: Was this intended?

The reason for asking the same question on all the PRs is that this comes without any background (and I cannot track to a specific team/issue here)

PR #9020, opened 2023-02-14
Fix pangolin rpc

fewensa:
fix pangolin rpc


fewensa:
It's fixed by https://github.com/polkadot-js/apps/pull/9013

PR #9063, opened 2023-02-23
Page area for Supersig... "like multisig but with super powers"

decentration:
This PR is the UI for the Supersig [pallet](https://github.com/kabocha-network/pallet_supersig)and rpc module. Here are the [docs](https://wiki.kabocha.network/docs/Supersig/supersig-tutorial/). Here is a [blog post](https://decentration.medium.com/supersig-like-multisig-but-with-superpowers-86b9ce0412f6). 

In short, supersig is like multisig but with superpowers! 💥 It is also like `pallet_collective`, but instead anyone can create a supersig collective. You can add and remove members, and voting is simpleMajority. Voting is one vote per standard member. And if there are any Master members, then they get 50% of the vote. 

Hopefully this is accepted effortlessly, as it runs and is able to merge prior to pulling this, though apologies if i have missed anything!

All the changes have been in the `page-supersig` section, as well as some addition to the api/specs section adding the chains which have supersig.


README.md
```diff
@@ -36,12 +36,13 @@ You can run a docker container via -
...
 ```
 docker run --rm -it --name polkadot-ui -e WS_URL=ws://someip:9944 -p 80:80 jacogr/polkadot-js-apps:latest
+
 ```
 
 To build a docker container containing local changes -
 
 ```
-docker build -t jacogr/polkadot-js-apps -f docker/Dockerfile .
+docker build -t decentration/apps -f docker/Dockerfile .
```
jacogr, line 45: Please revert.

packages/apps-config/src/api/spec/index.ts
```diff
@@ -267,6 +268,8 @@ const spec: Record<string, OverrideBundleDefinition> = {
   subspace,
   'substrate-contracts-node': substrateContractsNode,
   subzero: zero,
+  supersig,
+  'node-template': supersig,
```
jacogr, line 272: This doesn’t work. Node-template is a standard with the base Substrate types. Please remove.

jacogr (commented):
Would have to keep this on ice until there was a some actual chains using it. (Generally if it is team-specific we don’t add a specific section at all - breakages and ongoing support land on the repo, so there needs to be a real deployment base)

decentration:
hey @jacogr just fixing the linting and changed the readme back to default state. 

Btw Kabocha chain Kusama is currently using this, thereafter hopefully common good chains statemine/mint.

packages/page-addresses/src/modals/Create.tsx
```diff
@@ -102,17 +105,19 @@ function Create ({ onClose, onStatusChange }: Props): React.ReactElement<Props>
           : t<string>('address created');
 
         InputAddress.setLastValue('address', address);
+        
       } catch (error) {
         status.status = 'error';
         status.message = (error as Error).message;
       }
-
+      toggleFavorite(address);
```
jacogr, line 113: This is always meant to a user action, not an auto-action.

packages/page-parachains/src/Overview/Parachain copy.tsx
```diff
@@ -0,0 +1,175 @@
+// Copyright 2017-2022 @polkadot/app-parachains authors & contributors
```
jacogr, line 1: Why this copy?

packages/page-supersig/src/Extrinsics/Submission.tsx
```diff
@@ -0,0 +1,140 @@
...
+    const { method, section } = api.registry.findMetaCall(extrinsicCall.callIndex);
+    const extrinsicFn = api.tx[section][method];
+
+    return { call: extrinsicCall, fn: extrinsicFn, hex};
+  } catch (e) {
+    return null;
+  }
+}
+
+function Selection ({ className, defaultValue }: Props): React.ReactElement<Props> {
```
jacogr, line 67: This is extrinsics duplicated?

Sadly I believe his is a dead-end for this type of functionality in here. It is completely un-integrated in the UI. For instance with current multisig I can submit a transaction from anywhere in the apps UI and I can sign it and create a multisig operation.

Basically whet this means it is a stand-along, non-integrated into the apps UI section. I cannot for instance 

- point the supersig to a proxy - and do do anything, e.g. transfer, I would need to create the full proxy extrinsic myself
- I cannot just change my identities in accounts and get it signed via the normal process - I would need to get the data somehow and fill it in here
- ...

packages/page-supersig/src/Supersig/Dashboard/index.tsx
```diff
@@ -0,0 +1,146 @@
...
+import CreateModal from '../modals/Create';
+import { useTranslation } from '../translate';
+import Address from './Table';
+import Summary from './Summary';
+import { largeNumSum } from '../../util';
+import { encodeAddress } from '@polkadot/util-crypto';
+
+type SortedAddress = { address: string; isFavorite: boolean };
+
+const STORE_FAVS = 'accounts:favorites';
```
jacogr, line 19: This should not be account favorites - they are only managed in the accounts/addresses sections.

packages/page-supersig/src/Supersig/Dashboard/useLoadingDelay.ts
```diff
@@ -0,0 +1,17 @@
+// Copyright 2017-2021 @polkadot/react-hooks authors & contributors
+// SPDX-License-Identifier: Apache-2.0
+import { useEffect, useState } from 'react';
+import { useIsMountedRef } from '../../../../react-hooks/src/useIsMountedRef';
+export function useLoadingDelay (delay = 100): boolean {
```
jacogr, line 5: This is already existing in `react-hooks`?

packages/page-supersig/src/Supersig/useKnownAddresses.ts
```diff
@@ -0,0 +1,18 @@
+// Copyright 2017-2022 @polkadot/app-accounts authors & contributors
```
jacogr, line 1: This is is `react-hooks` already.

packages/page-supersig/src/augment-supersig.ts
```diff
@@ -0,0 +1,26 @@
+import '@polkadot/rpc-core/types/jsonrpc';
```
jacogr, line 1: We cannot bring in additional augmentation files - we only use the defaults from the JS API.

packages/page-supersig/src/modals/IdentityMain.tsx
```diff
@@ -0,0 +1,269 @@
...
+    (value.length >= minLength) &&
+    includes.reduce((hasIncludes: boolean, check) => hasIncludes && value.includes(check), true) &&
+    (!starting.length || starting.some((check) => value.startsWith(check))) &&
+    !excludes.some((check) => value.includes(check)) &&
+    !notStarting.some((check) => value.startsWith(check)) &&
+    !notEnding.some((check) => value.endsWith(check))
+  );
+}
+
+function IdentityMain ({ address, className = '', onClose }: Props): React.ReactElement<Props> {
```
jacogr, line 77: All this stuff seems just duplicated from accounts.

We want integrated pages, this is a stand-alone creation using the apps UI codebase.

We cannot have the same stuff in multiple places, this just means we double the workload. (This is already in need of an update since there is an open PR that extends this specifically)

jacogr (commented):
This doesn't quite work...

We have just duplicated the account UI with no effort on integration. (See the comment around using the existing transactions from any page to trigger this - we don't want to ahev to duplicate the whole UI)

The problem is that here the approach was taken for a stand-alone app, not an app that forms part of the UI and integrates the functionality. 

For a true integration effort -

1. Yes, you can certainly manage the setup for the supersig separately (it is needed)
2. any transaction from anywhere in the apps UI, needs to be able to hook int this - this means no extrinsic duplication, rather just account selection there
3. This means you need integration into the signer since that is where the magic happens based on the account type (of which this is only 1)

jacogr:
I don't particularly care where it is funded from - I care about on-going support of the efforts and proper integration. Since I'm the only polkadot-js dev, I get to spend around 10-15% of my time on the apps UI specifically so don't want to make things worse and explode the support queue (the rest of the time goes to well, the rest... e.g. api, common/ui libs, extension, tools, wasm, and whatever is burning at that point)

The closest community pallet at this point that nears merging is probably the Moonbeam parachain staking - since that is actually used by a large number of chains. so support base is actually seeing people have a need for it and use it.

packages/page-parachains/src/Overview/Parachain copy.tsx
decentration: this can be deleted, was just using this to bring it local to see how it works.

packages/page-supersig/src/Extrinsics/Submission.tsx
decentration: sorry this can also be deleted, again i just brought it locally to see how it worked.

packages/page-supersig/src/Supersig/Dashboard/useLoadingDelay.ts
decentration: i believe you removed useLoadingDelay (may be mistaken) which is why i brought it locally into page-supersig.

packages/page-supersig/src/augment-supersig.ts
decentration: this can be removed will remove.

packages/page-supersig/src/modals/IdentityMain.tsx
decentration: All the duplicates were just brought in to see how they work, they are not actually used. Will clean.

decentration:
Let me close this and come back to you when everything is clean and fixed, and when there is more than one chain using it. 

If i can get it to be used in one of the common good chains, or a larger chain, i think that would be a decent time to come to you, would that be fair to say? 

And i totally appreciate what youre saying, and you want to keep the support requests managable.

PR #9102, opened 2023-03-03
Dali to Picasso

andyjsbell:
Renaming Dali in Rococo to Picasso


jacogr (changes requested):
Fails CI checks.

jacogr (changes requested):
It cannot be merged - there are conflicts. (So ever with an approving review, there is no way to click the button :))

I would suggest merging in master, it is quite possibly due to the latest changes to use with-extension imports everywhere.

jacogr (approved):
Thank you.

PR #9123, opened 2023-03-07
update peaq logo (polkadot)

lavish0000:
This will update the peaq parachain logo in polkadot


lavish0000:
past commits also got included

Issue #9127, opened 2023-03-08
Update teleport functionality to XCM V2/V3


PierreBesson:
* **I'm submitting a ...**

  - [x] Bug report

* **What is the current behavior and expected behavior?**

When using the polkadot.js apps teleport UI, trying to send WND from Westend to either Westmint (para 1000) or Westend-Collective (para 1001) crashes when clicking the Teleport button. We don't see the "submit transaction popup" as expected.

![image](https://user-images.githubusercontent.com/513471/223728550-08883164-6e84-4209-86b0-b02cfa1ac568.png)

Here is the error we have in the browser console:
```
Uncaught Error: createType(Call):: Call: failed decoding xcmPallet.limitedTeleportAssets:: Struct: failed on args: {"dest":"{\"_enum\":{\"V2\":\"XcmV2MultiLocation\",\"V3\":\"XcmV3MultiLocation\"}}","beneficiary":"{\"_enum\":{\"V2\":\"XcmV2MultiLocation\",\"V3\":\"XcmV3MultiLocation\"}}","assets":"{\"_enum\":{\"V2\":\"XcmV2MultiassetMultiAssets\",\"V3\":\"XcmV3MultiassetMultiAssets\"}}","fee_asset_item":"u32","weight_limit":"{\"_enum\":{\"Unlimited\":\"Null\",\"Limited\":\"SpWeightsWeightV2Weight\"}}"}:: Struct: failed on dest: {"_enum":{"V2":"XcmV2MultiLocation","V3":"XcmV3MultiLocation"}}:: Cannot map Enum JSON, unable to find 'V1' in v2, v3
    at h (modu.ff40a6106d8303b6.js:50:68488)
    at j.createTypeUnsafe (modu.ff40a6106d8303b6.js:50:154258)
    at o (modu.ff40a6106d8303b6.js:50:366989)
    at r (modu.ff808c477751f58f.js:1:89034)
    at comm.6d578ff57f493e65.js:1129:2511
    at comm.6d578ff57f493e65.js:115:327
    at Object.je (modu.fc3ee6efdaa06c24.js:2:22463)
    at Ue (modu.fc3ee6efdaa06c24.js:2:22617)
    at modu.fc3ee6efdaa06c24.js:2:42478
    at Ar (modu.fc3ee6efdaa06c24.js:2:42572)
h @ modu.ff40a6106d8303b6.js:50
createTypeUnsafe @ modu.ff40a6106d8303b6.js:50
[...]

PR #9129, opened 2023-03-08
Disable teleports (no detection on versions)

jacogr:
As per https://github.com/polkadot-js/apps/issues/9127


PierreBesson:
@jacogr this seems a bit extreme, especially as the teleport feature still works fine on Polkadot and Kusama but has been hidden from the UI.
I am afraid this could be misinterpreted by some users as XCM transfers not working anymore.

PR #9174, opened 2023-03-14
Fix duration formatting to use proper SI symbols

getsnoopy:
This commit fixes the formatting of durations in the app so that they use the proper SI symbols for the various time units to make it as standardized and international as possible.

### DIFF
--- a/packages/react-hooks/src/useBlockTime.ts
+++ b/packages/react-hooks/src/useBlockTime.ts
@@ -27,24 +27,16 @@ export function calcBlockTime (blockTime: BN, blocks: BN, t: TFunction): Result 
     blockTime.toNumber(),
     `${value < 0 ? '+' : ''}${[
       days
-        ? (days > 1)
-          ? t<string>('{{days}} days', { replace: { days } })
-          : t<string>('1 day')
+        ? t<string>('{{days}} d', { replace: { days } })
         : null,
       hours
-        ? (hours > 1)
-          ? t<string>('{{hours}} hrs', { replace: { hours } })
-          : t<string>('1 hr')
+        ? t<string>('{{hours}} h', { replace: { hours } })
         : null,
       minutes
-        ? (minutes > 1)
-          ? t<string>('{{minutes}} mins', { replace: { minutes } })
-          : t<string>('1 min')
+        ? t<string>('{{minutes}} min', { replace: { minutes } })
         : null,
       seconds
-        ? (seconds > 1)
-          ? t<string>('{{seconds}} s', { replace: { seconds } })
-          : t<string>('1 s')
+        ? t<string>('{{seconds}} s', { replace: { seconds } })
         : null
     ]
       .filter((s): s is string => !!s)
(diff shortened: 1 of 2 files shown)


packages/react-hooks/src/useBlockTime.ts
```diff
@@ -28,17 +28,17 @@ export function calcBlockTime (blockTime: BN, blocks: BN, t: TFunction): Result
     `${value < 0 ? '+' : ''}${[
       days
         ? (days > 1)
-          ? t<string>('{{days}} days', { replace: { days } })
-          : t<string>('1 day')
+          ? t<string>('{{days}} d', { replace: { days } })
+          : t<string>('1 d')
```
jacogr, line 32: We can actually combine these now since we don't have `day` & `days`, but only `d`. GH won't let me comment, but we could simplify -

```
[
  days
    ? t<string>('{{days}} d', { replace: { days } })
    : null,
  hours
    ? t<string>('{{hours}} h', { replace: { hours } })
    : null
...
]
```

jacogr (commented):
One small comment that applies everywhere we had plurals.

I seems to remember we had it this way around initially, but tried to make it more human-ish. Don't have strong feelings either way (since it is indeed standard).

packages/react-query/src/Elapsed.tsx
```diff
@@ -57,7 +57,7 @@ function getDisplayValue (now = 0, value: BN | Date | number = 0): React.ReactNo
     ? formatValue(elapsed, 's', elapsed < 15)
     : (elapsed < 3600)
       ? formatValue(elapsed / 60, 'min')
```
jacogr, line 59: The above we probably want to change to `m` as well.

packages/react-hooks/src/useBlockTime.ts
getsnoopy: I initially simplified it to that, but I reread the condition that it was `days > 1`, so I didn't know if `days` is ever a decimal amount less than 1, where we'd want to display it as **1**, which is why I kept it as is. I can simplify it if that's not the case.

jacogr: It is indeed an integer.

packages/react-query/src/Elapsed.tsx
getsnoopy: The symbol for the *minute* is **min**, so that's why I left it as such.

jacogr: Just feels inconsistent with "d", "h", "s" and then "min". (I guess I care more about consistency that being "technically correct", SI is _nice_ but in this case we never have "month", so cannot overlap)

jacogr (commented):
Needs adjustment to `m` as requested.

Sadly, while I appreciate the effort here, it won't be merged without being consistent single-letter everywhere)

packages/react-query/src/Elapsed.tsx
gilescope: m could be month...

jacogr:
Thanks for the effort. However in this case going to close the PR since the current display is less prone to mistakes and following a consistent (albeit non SI) approach.

PR #9149, opened 2023-03-11
Add IBP system chain support, update naming convention

senseless:
Hello,

This request alters the names of both "Dotters Net" and "IBP Network" to bring them into a naming convention alignment for the IBP program. "IBP Network" will be renamed to "IBP-GeoDNS1" and "Dotters Net" will be renamed to "IBP-GeoDNS2".

Additionally, we are requesting to add our RPC endpoints on IBP-GeoDNS2 for the following system chains: Westmint, Collectives-Westend, Statemine, BridgeHub-Kusama, Encointer-Kusama, Collectives-Polkadot, and Statemint.

Thanks!

PR #9176, opened 2023-03-15
Add IBP endpoints to system parachains, update naming convention

miloskriz:
Dear team, 

This request is to add endpoint support to the system parachains listed below, these endpoints are provided by the Infrastructure Builders' Program:

**Polkadot**:

- Statemint
- Collectives

**Kusama**: 

- Statemine
- Encointer
- BridgeHub

**Westend**:

- Westmint
- Collectives

Additionally, current endpoints which are already active for the relay chains are now renamed (and re-sorted in the list) to show that they are being managed by the Infrastructure Builders' Program.

Please note that this PR replaces #9149 and that one will be now closed.

Many thanks!!.. best regards..

**_Miloš_**


jacogr (approved):
Seems sane.

As an aside, not completely over-the-moon with the `IBP_GeoDNS{1, 2}`, I would have assumed if it is actually balanced it would have a single entry. For now will let it slide since it replaces/renames an existing, but please don't add more :)

PR #9182, opened 2023-03-15
Fix Polymesh types bundle to use correct specNames and update branding

F-OBrien:
- Fix Polymesh types to use correct chain specNames, `polymesh_mainnet` and `polymesh_testnet`.
- Use Polymesh logo and references instead of Polymath logo and references.


F-OBrien:
@jacogr many thanks for updating the Polymesh types bundles today... you got to it before I could get this PR in. Polymesh mainnet and polymesh testnet actually have seprate specNames so this PR fixes them. I've also update the Polymesh logo and removed polymath references.

jacogr:
I was actually clicking through all packages on npm (trying to get all chain deps updated) so saw that the repo is the same, but the package has been renamed (and as I recall that one has a massive jump in increments as well)

jacogr (approved):
Thank you.

Issue #2059, opened 2019-12-16
Support BTreeSet 


kigawas:
#1723

jacogr:
Closed in https://github.com/polkadot-js/apps/pull/9206#event-8805682445

Issue #1723, opened 2019-10-13
Support BTreeMap params


jacogr:

jimmychu0807:
Hit this issue today

Diiaablo95:
Any updates on this issue?

jacogr:
PRs are always welcome. 

It is not something that I use myself, so generally don't bump into it and therefore don't spend time on it. It will eventually get done hence leaving it open, but def. nowhere soon.

ap211unitech:
BTreeMap param component is supported now. Hence, closing this issue.

PR #9204, opened 2023-03-20
Develop update 2023 03

Neopallium:
Removed some unneeded commits.

1. Removed the commit for the "signing data padding" hack that was needed for ITN rewards.
2. Removed `Option<Vec<u8>>` fix, since Polkadot has fixed it upstream.
3. Removed the `polymesh_schema.json` file, not needed anymore as our `@polymeshassociation/polymesh-types` can now be used.
4. Fixed support for `BTreeSet` so we don't need to remap our types using the schema file.
5. Squashed some older commits.


Neopallium:
Sorry, Wrong repo.

jacogr:
No issues. However, if you have upstream fixes, please do make a PR as well.

I see the last commit is "Support BTreeSet", there [is certainly use for that](https://github.com/polkadot-js/apps/issues/2059). (Also for [BTreeMap](https://github.com/polkadot-js/apps/issues/1723))

Neopallium:
Yup.  I have been thinking of upstreaming the BTreeSet feature.  I am not a TS/JS developer and ran into a weird issue with React `useState`.  Maybe if I open a PR for just that someone else would have an idea on how to correctly fix that (just using a hack for now).

Neopallium:
BTreeSet PR: #9206 
I don't know TS/React well enough to try adding BTreeMap support.  Also we don't use it, so I can't test it.

jacogr:
100% - it only makes sense adding them when there is a need and it can be tested.

PR #9214, opened 2023-03-24
Fix EQ&GENS balances 

alex-nax:


jacogr (changes requested):
This is localized to these chains only, so only took a cursory glance and it looks good. 

The build it failing, so that needs resolution.

PR #9227, opened 2023-03-26
Allow to run with isDevelopment on ethereum-like chains

farcaller:
Currently any `isEthereum==true` chain will force the development mode off, making it impossible to add any extra types as needed.


jacogr (commented):
Really can't see why that override exists - it was added alongside the initial Ethereum-like support in this PR - https://github.com/polkadot-js/apps/pull/3961

jacogr (approved):

PR #8227, opened 2022-10-11
Add Stakeworld endpoint for kusama

stakeworld:

PR #9249, opened 2023-03-29
Add stakeworld westend rpc node

stakeworld:
Hi Jaco,

I would like to add our westend rpc node to polkadot.js. I have an [earlier PR](https://github.com/polkadot-js/apps/pull/8227) for a kusama rpc node but I understand you prefer to start with a test net first so i'll close that for now and resubmit later if this one goes well.

Regards
Max | Stakeworld


jacogr (approved):
Thank you.

PR #9293, opened 2023-04-05
feat: update 3dpass logo

fedosov:
3dpass.org network now has a new official logo


jacogr (approved):
Thanks for cleaning up and removing the old one as well :)

PR #9299, opened 2023-04-06
Make disabled input fields selectable

### LINKED ISSUE #8212: Disabled text selection on disabled input fields prevents users to select & copy output
In many places the UI renders outputs in input fields that have text selection disabled (at least in firefox linux) which means copying anything from these outputs is impossible. For example on the extrinsic decode page:

screen_2022-10-06-14-18-58

TorstenStueber:
Closes #8212

This is particularly useful for any fields containing output values, for example when decoding an extrinsic. Currently it is impossible to select and copy and values in Firefox because all pointer events are disabled.


jacogr (approved):
Thank you.

PR #9329, opened 2023-04-14
change aleph testnet name to make it unique

DamianStraszak:
Change the aleph zero testnet name, to make it distinguishable from aleph zero mainnet. Needed for Sirato smart contract verifier.


jacogr (approved):
Just a note: The "info" was used historically to match the colors & logo, which is now inline. The only place where it is still operational is really around Ledger support. (So nothing will change display-wise)

PR #9356, opened 2023-04-20
Add stakeworld kusama rpc node

stakeworld:
Hi @jacogr,

I've been stable on westend for three weeks (#9249) and now would like to add kusama if thats ok with you.

Regards
Max | Stakeworld


jacogr (approved):
Thank you.

PR #9375, opened 2023-04-26
Add blockops endpoint

haroldsphinx:
Now that our westend endpoints has been up and running for a week, This PR will be adding our Kusama and Polkadot endpoints
Screenshot 2023-04-26 at 12 29 51 PM


jacogr (approved):
The guidance is not "running for a week", rather "running for 2 weeks".

The original PR is here - https://github.com/polkadot-js/apps/pull/9326 which according to GH is now merged 2 weeks ago. 

Not sure if it was earlier, this specific PR description here said "for a week", so obviously I left it hang about without a check with the idea to come back to it when I have a gap.

With that said, I have not seen this endpoint pop up in the last _2 weeks_, so it seems to be ok.

PR #9377, opened 2023-04-26
re-activate centrifuge fullnodes

wischli:
This PR re-enables the self-hosted fullnodes for Centrifuge and Altair parachains.

### DIFF
--- a/packages/apps-config/src/endpoints/productionRelayPolkadot.ts
+++ b/packages/apps-config/src/endpoints/productionRelayPolkadot.ts
@@ -127,7 +127,7 @@ export const prodParasPolkadot: Omit<EndpointOption, 'teleport'>[] = [
     info: 'centrifuge',
     paraId: 2031,
     providers: {
-      // Centrifuge: 'wss://fullnode.parachain.centrifuge.io', // https://github.com/polkadot-js/apps/issues/9366
+      Centrifuge: 'wss://fullnode.centrifuge.io',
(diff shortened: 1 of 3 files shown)


packages/apps-config/src/endpoints/productionRelayPolkadot.ts
```diff
@@ -127,7 +127,7 @@ export const prodParasPolkadot: Omit<EndpointOption, 'teleport'>[] = [
     info: 'centrifuge',
     paraId: 2031,
     providers: {
-      // Centrifuge: 'wss://fullnode.parachain.centrifuge.io', // https://github.com/polkadot-js/apps/issues/9366
+      Centrifuge: 'wss://fullnode.centrifuge.io',
```
jacogr, line 130: This cannot be the same as the "production" RPC.

An endpoint is either -

- connected to a relay (i.e. as a parachain)
- operating on it's own (i.e. not a parachain)

The two are mutually exlusive.

jacogr (changes requested):
Tests are failing due to duplicated RPCs used for both parachain and non-parachain operation.

packages/apps-config/src/endpoints/productionRelayPolkadot.ts
wischli: Thanks for the clarification, fixed in c6889fc!

jacogr (approved):

PR #9390, opened 2023-04-30
Add stakeworld polkadot rpc node

stakeworld:
Hi @jacogr,

I've been stable on westend and kusama and also would like to add polkadot, if thats ok. 

Regards
Max | Stakeworld


jacogr (approved):
Thank you.

PR #9402, opened 2023-05-03
Adding LuckyFriday to Polkadot WSS endpoints

paradox-tt:
Dear Devs,

Following our successful implementation of endpoints on Westend, Statemine and Kusama, we are following up with an endpoint for Polkadot.  This endpoint has been in operation for a little over a week and has passed our internal load and up-time checks.

We hope this is satisfactory.

Kind Regards,
Will | Paradox


jacogr (approved):
Thank you.

PR #9419, opened 2023-05-08
Fix default threshold in motion

h4x3rotab:
From the description, the default motion threshold should be "50%+1". However the old implementation was `ceil(50% council)`. When the council has an even member size, it produces 50% council, but not 50%+1. This commit fixes it.


h4x3rotab:
Thanks for the confirmation. Moving to #6924 and I'm happy to provide help.

ap211unitech:
Closing this PR as it appears to be outdated. If the problem persists, Feel free to open a new PR with updated details.

PR #9450, opened 2023-05-15
Add ferrum to list of eth chains

3ierratango:
Changelog:
- Add types for ferrum parachain
- Add ferrum parachain to the list of eth chains


packages/apps-config/src/api/spec/ferrum.ts
```diff
@@ -0,0 +1,32 @@
...
+
+const definitions: OverrideBundleDefinition = {
+  types: [
+    {
+      // on all versions
+      minmax: [0, undefined],
+      types: {
+        AccountId: 'EthereumAccountId',
+        AccountId20: 'EthereumAccountId',
+        AccountId32: 'EthereumAccountId',
```
jacogr, line 17: This is probably not quite correct - `EthereumAccountId` represents a `[u8; 20]` internally, while `AccountId32` represents an `[u8; 32]`. 

So the types are mismatched on this one and anything that uses it (some ETH-compatible chains for instance has a bridge to other Substrate-chains) will fail horribly.

The first 2 is all that is needed to map to ETH-compatible.

jacogr (approved):
Thank you.

PR #9501, opened 2023-05-25
Addressing issue 9480

radiumb:
Addressing https://github.com/polkadot-js/apps/issues/9480

We had switched to using the relay chain as external RPC rather than running a relay chain RPC along with asset chains - to take advantage of a feature in cumulus binary. We are having mixed success with this. Its particularly worse since the relay chain block constantly lags beyond what is usually considered acceptable (>10 blocks) ever since v0.39. We struggle with > v0.39 for validators too. We are seriously considering downgrading back to v0.39 rather than following the convention of using the latest binary.  What makes this hard is that the timeouts due block delays are infrequent and only a service restart seems to fix it. We will revert the external RPC to make sure that choice is not amplifying the issue. We request our statemint RPC to be added back. 

Meanwhile any advice on the matter of binary version is much appreciated.


jacogr:
Sheiße. 

Sadly have no real insight into the binaries and releases (anymore), like you I just follow along and deploy as available.

jacogr (approved):

PR #9529, opened 2023-06-03
fix: issue 9127. teleport reactivated

loanMaster:
Hello Polkadot Apps maintainers

This PR fixes https://github.com/polkadot-js/apps/issues/9127
by using XCM v2 for Kusama and v1 for Polkadot teleports. 

v2 is used as default if not otherwise defined.

### DIFF
--- a/packages/page-parachains/src/Teleport.tsx
+++ b/packages/page-parachains/src/Teleport.tsx
@@ -47,7 +47,7 @@ function Teleport ({ onClose }: Props): React.ReactElement<Props> | null {
   const [recipientId, setRecipientId] = useState<string | null>(null);
   const [senderId, setSenderId] = useState<string | null>(null);
   const [recipientParaId, setParaId] = useState(INVALID_PARAID);
-  const { allowTeleport, destinations, isParaTeleport, oneWay } = useTeleport();
+  const { allowTeleport, destinations, isParaTeleport, oneWay, xcmVersion } = useTeleport();
 
   const call = useMemo(
     (): SubmittableExtrinsicFunction<'promise'> => {
@@ -77,7 +77,7 @@ function Teleport ({ onClose }: Props): React.ReactElement<Props> | null {
   const params = useMemo(
     () => [
       {
-        V1: isParaTeleport
+        [xcmVersion]: isParaTeleport
           ? {
             interior: 'Here',
             parents: 1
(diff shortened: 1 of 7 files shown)


packages/page-parachains/src/Teleport.tsx
```diff
@@ -77,7 +77,7 @@ function Teleport ({ onClose }: Props): React.ReactElement<Props> | null {
   const params = useMemo(
     () => [
       {
-        V1: isParaTeleport
+        [xcmVersion]: isParaTeleport
```
jacogr, line 80: This really is going to be problematic down the line - we are assuming here that all future versions will have the same "form", just because current versions have that form. So it will break at some point.

(We already had the V0 -> V1 swap which was significantly different and certainly not compatible)

I would suggest just supporting the current V2 here - if some chains have not been enabled, they just need to be disabled teleport-wise.

jacogr (changes requested):
Thank you. Some small requests around the versioning maintenance.

packages/page-parachains/src/Teleport.tsx
loanMaster: Thank you for the comment. Yes, this might break with introduction of XCM v3.

The teleport between Polkadot and Statemint is via XCM v1. When trying with XCM v2 I received an error similar to the one shown [here](https://github.com/polkadot-js/apps/issues/9127#issue-1615301770).

If we only allow V2, the most important teleport, the one between Polkadot and Statemint, will not be supported.

I could add a comment and/or try to restrict the type of xcmVersion to "V1" and "V2" in typescript.

loanMaster: How to proceed?

loanMaster: @jacogr I restricted the xcm version via string literal types to V2 and V1. Therefore the build would break if anyone tries using v3. Also I added a comment.

Could you have look?

joepetrowski:
Polkadot and all system chains are now on V3, we really shouldn't use V1 or V2 at all.

jacogr:
Since as per Joe's comments all are (currently) on V3 could just simplify it completely -

- only support V3
- remove all version config

So I guess it could be just as simple as renaming the old key in the message construction to `V3`.

jacogr (commented):
Thank you.

It looks sane, just going through testing through this branch locally (both directions).

PS: I think Rococo is fine to leave as-is. (Since it is a testnet without any funds at risk - since it also evolves quite rapidly, having the functionality available is great, even if it temporaily doesn't quite work as expected).

jacogr (approved):
Thank you! This is great. (And especially for checking/testing on all the affected relays)

The testing on my side went perfectly.

PR #9547, opened 2023-06-06
Peaq rpc add

sfffaaa:
Add RPC calls for the Peaq/Krest/Peaq-dev chain


packages/apps-config/src/api/spec/peaq-type-definitions.d.ts
```diff
@@ -0,0 +1,4 @@
+// Copyright 2017-2023 @polkadot/apps-config authors & contributors
```
jacogr, line 1: Please remove this - rather add explicit overrides inside the actual file used.

(The rest of these were removed yesterday - don't want to go down the rabbit hole of providing types for these packages where it should be done upstream)

sfffaaa: Thanks for the feedback. Already removed this file.

PR #9576, opened 2023-06-13
Fix `jsonrpc` attribute

pmikolajczyk41:
Fixing `jsonrpc` attribute while building full RPC record object.


jacogr (approved):
Thank you!

Really weird, will disect and see where this error snuck in.

PR #9599, opened 2023-06-18
Fix delegation with x0.1 conviction and typo

Tbaut:
Fixes the following
- Delegation with x0.1 conviction (conviction === 0) sets the extrinsic to `null` and the `TxButton` to disabled.
- Typo on delegation

Remarks:
- Also the type of `activityTo` is `VoteResultItem[] | null | undefined` but is set by default to `[]` which is truthy. This result in step 2 being valid even if the selected account has no activity. Is this expected or do you intend to check for `.length`?

### DIFF
--- a/packages/page-referenda/src/Referenda/Delegate/index.tsx
+++ b/packages/page-referenda/src/Referenda/Delegate/index.tsx
@@ -93,7 +93,7 @@ function Delegate ({ className, palletReferenda, palletVote, tracks }: Props): R
   );
 
   const extrinsic = useMemo(
-    () => balance && conviction && toAccount && includeTracks
+    () => balance && conviction >= 0 && toAccount && includeTracks
       ? isAllTracks
         ? api.tx.utility.forceBatch(includeTracks.map((trackId) =>
           api.tx[palletVote as 'convictionVoting'].delegate(trackId, toAccount, conviction, balance)
@@ -103,7 +103,7 @@ function Delegate ({ className, palletReferenda, palletVote, tracks }: Props): R
     [api, balance, conviction, includeTracks, isAllTracks, palletVote, toAccount, trackId]
   );
 
-  const isStep1Valid = !!(accountId && activityFrom && includeTracks && (includeTracks.length > 0));
+  const isStep1Valid = !!(accountId && includeTracks && (includeTracks.length > 0));
   const isStep2Valid = !!(toAccount && activityTo);
 
   return (
--- a/packages/page-referenda/src/Referenda/Delegate/index.tsx
+++ b/packages/page-referenda/src/Referenda/Delegate/index.tsx
@@ -104,7 +104,7 @@ function Delegate ({ className, palletReferenda, palletVote, tracks }: Props): R
   );
 
   const isStep1Valid = !!(accountId && includeTracks && (includeTracks.length > 0));
-  const isStep2Valid = !!(toAccount && activityTo);
+  const isStep2Valid = !!(toAccount && activityTo && (activityTo.length > 0));
 
   return (
     <>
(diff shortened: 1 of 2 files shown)


packages/page-referenda/src/Referenda/Delegate/index.tsx
```diff
@@ -103,7 +103,7 @@ function Delegate ({ className, palletReferenda, palletVote, tracks }: Props): R
     [api, balance, conviction, includeTracks, isAllTracks, palletVote, toAccount, trackId]
   );
 
-  const isStep1Valid = !!(accountId && activityFrom && includeTracks && (includeTracks.length > 0));
+  const isStep1Valid = !!(accountId && includeTracks && (includeTracks.length > 0));
   const isStep2Valid = !!(toAccount && activityTo);
```
Tbaut, line 107: I imagine that this is what was intended. 
``` suggestion
  const isStep2Valid = !!(toAccount && activityTo && !!activityTo.length);
```

packages/page-referenda/src/Referenda/Delegate/index.tsx
```diff
@@ -103,8 +103,8 @@ function Delegate ({ className, palletReferenda, palletVote, tracks }: Props): R
     [api, balance, conviction, includeTracks, isAllTracks, palletVote, toAccount, trackId]
   );
 
-  const isStep1Valid = !!(accountId && activityFrom && includeTracks && (includeTracks.length > 0));
-  const isStep2Valid = !!(toAccount && activityTo);
+  const isStep1Valid = !!(accountId && includeTracks && (includeTracks.length > 0));
+  const isStep2Valid = !!(toAccount && activityTo && (activityTo.length > 0));
```
jacogr, line 107: In these we actually just want to know that we have received _something_ from the hook, not that the actual values are valid.

So the intent was -

- if `null` the hook has not fired, so wait a bit
- if `[...]` (any values), we have a result

So the original actually looks fine.

packages/page-referenda/src/Referenda/Delegate/index.tsx
```diff
@@ -103,8 +103,8 @@ function Delegate ({ className, palletReferenda, palletVote, tracks }: Props): R
     [api, balance, conviction, includeTracks, isAllTracks, palletVote, toAccount, trackId]
   );
 
-  const isStep1Valid = !!(accountId && activityFrom && includeTracks && (includeTracks.length > 0));
-  const isStep2Valid = !!(toAccount && activityTo);
+  const isStep1Valid = !!(accountId && includeTracks && (includeTracks.length > 0));
```
Tbaut, line 106:
```suggestion
  const isStep1Valid = !!(accountId && activityFrom && includeTracks && (includeTracks.length > 0));
```

packages/page-referenda/src/Referenda/Delegate/index.tsx
```diff
@@ -104,7 +104,7 @@ function Delegate ({ className, palletReferenda, palletVote, tracks }: Props): R
   );
 
   const isStep1Valid = !!(accountId && activityFrom && includeTracks && (includeTracks.length > 0));
-  const isStep2Valid = !!(toAccount && activityTo);
+  const isStep2Valid = !!(toAccount && activityTo && (activityTo.length > 0));
```
Tbaut, line 107:
```suggestion
  const isStep2Valid = !!(toAccount && activityTo);
```

packages/page-referenda/src/Referenda/Delegate/index.tsx
Tbaut: Understood thanks. I reverted these changes.

jacogr (approved):
Thank you.

Should actually add a comment on that logic - I had to look again as well... It _does_ look weird at first glance.

PR #9627, opened 2023-06-23
Fix issue with optional rpc parameters.

Neopallium:
This change allows correct handling of `Option<T>` rpc parameter types.  The filter was removing all `null` values, but those are needed for `Option<T>` parameters.  The `isOptional: true` can't be used when mixing optional parameters with required parameters.

### DIFF
--- a/packages/page-rpc/src/Rpc/Selection.tsx
+++ b/packages/page-rpc/src/Rpc/Selection.tsx
@@ -72,7 +72,7 @@ function Selection ({ queueRpc }: Props): React.ReactElement<Props> {
     (): void => queueRpc({
       rpc,
       values: values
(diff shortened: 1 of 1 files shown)


Neopallium:
An example of an RPC endpoint that has this issue see the `asset.canTransfer` endpoint for Polymesh:
https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Fmainnet-rpc.polymesh.network#/rpc

Leave the `from_custodian` and `to_custodian` parameters unset.  You can use `0x` for the `did` fields and for the `ticker`, or just leave them empty.

TarikGul:
Closing since https://github.com/polkadot-js/apps/pull/9911 resolves this.

PR #9686, opened 2023-07-11
please add crust onfinality endpoint, thanks

MinjieTian:
Hi, please add onfinality crust endpoint thanks


jacogr (changes requested):
The URL is listed twice - https://github.com/polkadot-js/apps/actions/runs/5525732800/jobs/10079677166?pr=9686#step:4:532

This is where it currently appears -

https://github.com/polkadot-js/apps/blob/0b166514aadaeeec02728631aac7d9f1d9672c3d/packages/apps-config/src/endpoints/production.ts#L149

Obviously an endpoint can only point to a single network, i.e. a parachain cannot be on two relays or an parachain cannot be a stand-alone network at the same time.

Just adjust the PR to have a single entry with the endpoint at the correct location.

piggydoughnut:
onFinality for crust was added [here](https://github.com/polkadot-js/apps/pull/9775/files)

PR #9667, opened 2023-07-06
Get isEthereum from chain properties

fgamundi:
Giving the growing number of Ethereum based substrate chains, the Tanssi team is [proposing to add](https://github.com/moondance-labs/tanssi/pull/110) `isEthereum` to the chain properties so apps (such as polkadotjs-apps) can easily distinguish if the chain they're connecting to is Ethereum-based or not without having to maintain and update a manual list.

Depends on https://github.com/polkadot-js/api/pull/5686 & https://github.com/polkadot-js/api/pull/5761

PR #9694, opened 2023-07-13
Change ethereum icons based on URL

### LINKED ISSUE #9632: Setting to manually tell Apps to use ethereum-style accounts and signatures
* **I'm submitting a ...**

  - [ ] Bug report
  - [x] Feature request
  - [ ] Support request
  - [ ] Other

* **What is the current behavior and expected behavior?**

Currently, Apps determines whether to use Ethereum style accounts and signatures by comparing the chain's spec name to a hard-coded list in https://github.com/polkadot-js/apps/blob/e55b9fc67b0dc1f36f2123948fb4a8f18cdc7a38/packages/apps-config/src/settings/ethereumChains.ts#L4

When users are developing chains and first try to switch to Ethereum-style signatures, they regularly become stuck at this step because Apps, the standard manual testing tool, can no longer produce valid signatures. I personally got stuck here, and you can see this question is already asked in stack exchange https://substrate.stackexchange.com/questions/9016/troubleshooting-polkadot-js-recognition-of-prefunded-dev-accounts-on-a-network-u

Ideally, there would be some way that a user could manually tell Apps to use the ethereum style accounts and signatures. For example it could be a checkbox, or require entering some manual types like in the old days.

The motivation is that users regularly get confused and stuck when exploring `AccountId20` compatibility, and it would be nice to point them to a simple setting to make the UI work for them.

* **Please tell us about your environment:**

  - Version:
  - Environment:

    - [ ] Node.js
    - [x] Browser
    - [ ] Other (limited support for other environments)

  - Language:

    - [ ] JavaScript
    - [ ] TypeScript (include tsc --version)
    - [ ] Other

wirednkod:
Fixes #9632

### DIFF
--- a/packages/react-api/src/Api.tsx
+++ b/packages/react-api/src/Api.tsx
@@ -132,7 +132,7 @@ async function retrieve (api: ApiPromise, injectedPromise: Promise<InjectedExten
   };
 }
 
-async function loadOnReady (api: ApiPromise, endpoint: LinkOption | null, injectedPromise: Promise<InjectedExtension[]>, store: KeyringStore | undefined, types: Record<string, Record<string, string>>): Promise<ApiState> {
+async function loadOnReady (api: ApiPromise, endpoint: LinkOption | null, injectedPromise: Promise<InjectedExtension[]>, store: KeyringStore | undefined, types: Record<string, Record<string, string>>, urlIsEthereum = false): Promise<ApiState> {
   statics.registry.register(types);
 
   const { injectedAccounts, properties, systemChain, systemChainType, systemName, systemVersion } = await retrieve(api, injectedPromise);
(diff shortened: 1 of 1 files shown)


wirednkod:
There is a better way of doing that (reading the parameters instead of just mapping a string) and could help for customizing the icons based on that new param.. @jacogr let me know if you prefer to extend it that way instead of a single param match

jacogr (approved):
I guess it doesn't hurt having this.

We now have multiple ways -

- as part of the hard-coded list
- as part of the chain properties (which is probably the preferred route)
- this

jacogr:
This PR will add support - https://github.com/polkadot-js/apps/pull/9667 (links through to the API changes, like this PR, also gracefully contributed - so certainly not widespread _yet_)

PR #9707, opened 2023-07-18
Fix the way we read the nft's metadata

jsidorenko:
On the nfts page the metadata disappeared because of the way we extract it. `metadata.data.toString()` was previously converting from u8a to string under the hood, while now it converts u8a to hex and produces the wrong result


jacogr (approved):
THANK YOU.

PR #9714, opened 2023-07-20
Remove now sunset dotscanner

jeluard:
Dotscanner is now sunset and associated links lead to 404.
Remove them.

See more details [here](https://dotscanner.com/)


jacogr (approved):
Thank you.

PR #9739, opened 2023-07-28
Remove Dora Factory

Leemo94:
Removed Dora Factory based on them leaving the ecosystem.

https://twitter.com/DoraFactory/status/1642552723827728385

https://twitter.com/search?q=(Polkadot%20OR%20Kusama%20OR%20Parachain)%20(from%3ADoraFactory)&src=typed_query&f=live


jacogr (changes requested):
Thank you. Just needs some adjustment to the unused imports (GH is now quite clever showing those inline)

packages/apps-config/src/endpoints/productionRelayKusama.ts
```diff
@@ -182,19 +182,6 @@ export const prodParasKusama: Omit<EndpointOption, 'teleport'>[] = [
       logo: nodesIpciSVG
     }
   },
-  {
```
jacogr, line 185: While still registered on-chain, see

image

The complete removal is a bit heavy-handed. All good removing the RPC endpoints, but the link from id -> name should stay in-tact.

jacogr (changes requested):

jacogr:
Ooops, sorry for this - seems I did catch the endpoints as part of https://github.com/polkadot-js/apps/pull/9749

So the RPC is now disabled.

Sorry for stepping on your toes here, just have been a bit behind with the issues logged by CI, so I ended up with a whole bunch to disable when I got around to it...

PR #9772, opened 2023-08-09
Support ipfs://ipfs/ links

jsidorenko:
isIPFS package we use unfortunately can't detect links that start with `ipfs://ipfs/`.
In this PR I've added a new helper method to clean the link and remove the unsupported prefix.

P. S. Shall we switch to the new nfts pallet? It already has more collections created compared to the old one.
Or we can support both pallets if needed.


jacogr:
Generally we want to support old and new - however is the “general support chains” are only n new, having the t to clean it up.

jsidorenko:
So you won't be against it if I switch the code to the new pallet?
There will soon be no UIs supporting the uniques pallet, and I think it will get deprecated at some point.
I would also be happy to add the basic UI to create a new collection (and ideally edit it later)

packages/react-hooks/src/useIpfsFetch.ts
```diff
@@ -68,7 +78,7 @@ export function useIpfsFetch <T> (hashes: string[] | undefined, options?: Option
     }
 
     return hashes
-      .map((hash) => isCid(hash) ? hash : '')
+      .map((hash) => isCid(hash) ? extractCid(hash) : '')
```
jacogr, line 81: Sorry, bit out of it and took me ages to get here:

Doesn't this mean that we lose support for the old format? Or is the new format the only way things are done now?

(Just need to do a search for the usage of this hook as well - it _may_ affect other places, e.g. I think it is used in the files app?)

jsidorenko: I'll double-check the code if affects other places (as I remember, it didn't, but want to be sure).
Also, I think we can write a better regexp here, will check next week.

In general, we won't lose the support of the old format, on the contrary, we failed to extract the cid if it was stored in a format with prefixes `ipfs://` or `ipfs://ipfs/`, this PR fixes that.

jacogr: Perfect.

jsidorenko: Done. I've reworked the whole metadata fetching approach, and now we support both ipfs cid versions + metadata stored on regular http* servers

jsidorenko:
> Hey @jsidorenko if you can fix the merge conflict, I can do a review and we can get this in!

done

TarikGul (approved):

TarikGul:
Thank you, looks great!

PR #9789, opened 2023-08-17
Add typeBundle for Cere Network

shamilkhan:


jacogr (changes requested):
This specific file, typesBundle.ts (as per the comment on line 4) is auto-generated, so any merge would just lose the changes completely.

It is built from the config as set in apps-config/api/spec - unless there are changes made there, trying to change the generated file is futile.

shamilkhan:
> This specific file, typesBundle.ts (as per the comment on line 4) is auto-generated, so any merge would just lose the changes completely.
> 
> It is built from the config as set in apps-config/api/spec - unless there are changes made there, trying to change the generated file is futile.

Thanks! Let me fix it.

shamilkhan:
> This specific file, typesBundle.ts (as per the comment on line 4) is auto-generated, so any merge would just lose the changes completely.
> 
> It is built from the config as set in apps-config/api/spec - unless there are changes made there, trying to change the generated file is futile.

Could you check again, please?

jacogr (approved):
Seems spot-on.

PR #9790, opened 2023-08-17
Add moonbeam apps as external resource for referenda

mmaurello:
Hello, 

in order to enhance user engagement and participation in the Moonbeam Dapp's governance and decision-making processes, we would like to add an external link to the dapp from polkadot js referenda page, for Moonriver, Moonbeam and Moonbase Alpha


jacogr (approved):
Thank you.

PR #9791, opened 2023-08-17
Update productionRelayKusama.ts

radiumb:
This issue was similar to Astar endpoint and related to load and resource balancing. The endpoint has been functional and in use via wallets and polkaassembly etc. We didn't notice the bug in polkadot-js. We request for the endpoint to be re-enable.


jacogr (approved):
I really don't particularly care for other environments - happy if they work as expected, that is great. I really care about the CI runner only.

I have been seeing issues with the endpoint myself in polkadot-js (The RB endpoints are generally my go-to). I can confirm that the actual auto-CI logs matches up with my physical while-using-results, it was doing some really weird stuff, down to dropping some requests.

PR #9870, opened 2023-09-11
Add Enjin & Matrixchain endpoints

leonardocustodio:
This PR adds the endpoint to the network registered at https://github.com/paritytech/ss58-registry/pull/187


SBalaguer:
As commented on [here](https://github.com/paritytech/ss58-registry/pull/187) for the SS58 Registry.

Given the intended nature of the Enjin Blockchain ecosystem's growth, I believe that the best course of action is to rather fork this repository and keep a project going that fully suits your needs, current and future.

leonardocustodio:
Since we are talking in two PRs, just for transparency to everybody. Would you mind explaining how Enjin is different from all "Live Networks" that are here and thus cannot be in this list? 
image

SBalaguer:
It's definitely not that it _can't_ be on that list, that it's far from the point. 

My point is that with the objective of the Enjin Blockchain and the consequent Matrixchain [explained on it's whitepaper](https://enjin.io/whitepaper), the best course of action is to fork PolkadotJS and keep a list of the top-level blockchain with all the subsequent connected chains (like Enjin Matrixchain and whatever others might come in the future). This predicted growth future is what makes the essence of Enjin Blockchain and Enjin Matrixchain different than any other solochain, and that it's why forking will be most probably best to suit your needs, current and future.

jacogr:
Closing. The only relay chains we support is Polkadot & Kusama. (And their testnets). dotapps.io is really just for the specific ecosystem with no appetite to support any relays out of that.

There is a reason any relays don't show up in live chains - these are complete stand-alone chains - unlike the relays which have a top-level position.

PR #9890, opened 2023-09-20
Add subsquare user external link and support vara network

wliyongfeng:
- Add account external link to subsquare user votes page.
- Support vara network.


jacogr (approved):

jacogr:
I guess it would be cool to pass the user address through as wlel - but will play when in and see what it "feels" like.

PR #9900, opened 2023-09-25
A0-3211: Rebase from latest upstream

Marcin-Radecki:
This change incorporates all changes from recent upstream.


Marcin-Radecki:
Sorry, misclick, should have been another fork.

PR #9911, opened 2023-09-27
Support BTreeMap and fix rpc optional

Neopallium:
Closes issue #1723


TarikGul:
@Neopallium I have this on my list to review, and will do it as soon as possible so we can get this in.

packages/react-params/src/Param/BTreeMap.tsx
```diff
@@ -0,0 +1,160 @@
...
+  useEffect((): void => {
+    const output = new Map();
+    let isValid = true;
+
+    for (const entry of values) {
+      const [key, value] = entry.value as RawParam[];
+
+      if (output.has(key)) {
+        isValid = false;
+        console.error('BTreeMap: Duplicate key', key);
```
TarikGul, line 104:
```suggestion
        console.error('BTreeMap: Duplicate key ', key);
```

packages/react-params/src/Param/Vector.tsx
```diff
@@ -61,9 +60,9 @@ function Vector ({ className = '', defaultValue, isDisabled = false, label, onCh
   useEffect((): void => {
     inputParams.length &&
       setParams((prev) =>
-        getParams(inputParams, prev, isDisabled ? (defaultValue.value as RawParam[] || []).length : count)
+        getParams(inputParams, prev, isDisabled ? values.length : count)
```
TarikGul, line 63: Ahh nice catch here 👍!

packages/page-rpc/src/Rpc/Selection.tsx
```diff
@@ -72,7 +72,7 @@ function Selection ({ queueRpc }: Props): React.ReactElement<Props> {
     (): void => queueRpc({
       rpc,
       values: values
-        .filter(({ value }) => !isNull(value))
+        .filter(({ value }, idx) => !rpc.params[idx].isOptional || !isNull(value))
```
TarikGul, line 75: Apologies but can you give a brief explanation as to what this is fixing? Thanks again.

TarikGul (changes requested):
Overall, looks super solid Thanks so much for the PR! The BTreeMap Component is spot on, and the vector fix is awesome.

Just a small nit with the console, and a quick question about the rpc optional fix. Other than that it should be ready to go.

packages/page-rpc/src/Rpc/Selection.tsx
Neopallium: The `isOptional` setting only really works for last parameters, not for a mix of Option and non-Option parameters.  So the fix keeps null values for `isOptional=false` parameters.

We are using this fix here: https://mainnet-app.polymesh.network/#/rpc
Open the `asset.canTransferGranular` endpoint and just click the submit button (don't need to fill in values).

To see the error try the same endpoint using the Polkadot.js.org app without the fix: https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Fmainnet-rpc.polymesh.network#/rpc

The error is:
```
1: asset.canTransferGranular: CanTransferGranularReturn
Expected 7 parameters (1 optional), 4 found instead
```

TarikGul:
Just needs a `yarn lint`

packages/page-rpc/src/Rpc/Selection.tsx
TarikGul: Ahh okay, thanks so much for the explanation. Makes sense

Neopallium:
`lint` applied.

TarikGul (approved):

PR #9915, opened 2023-09-28
Add Metaquity Network Testnet endpoints

Ekaanth:
We would like to add [Metaquity network](https://metaquity.xyz) to the polkadot-js testnet.


jacogr (changes requested):
This a a bit of a pain, however the image seems to be detected as slightly large, see https://github.com/polkadot-js/apps/actions/runs/6339744805/job/17243525191?pr=9915#step:4:511

Since we only use small versions in the UI we try to keep these sizes down since if each network has 1MB images (we had those in the past) it actaully slows down the load for all users since they all need to retrieve it.

Ekaanth:
Hi @jacogr after reviewing your feedback, I have updated the logo to a more suitable version. Please let me know if this is acceptable or if I should obtain a new compressed logo.

jacogr (approved):

Issue #9901, opened 2023-09-25
Error when open "Developer -> Contracts" tab with polkadot-v1.0.0 based chain


vampik33:
* **I'm submitting a ...**

  - [x] Bug report
  - [ ] Feature request
  - [ ] Support request
  - [ ] Other

* **What is the current behavior and expected behavior?**

Currently, when I tried to open "Developer -> Contracts" tab I have this error:

```
Uncaught error. Something went wrong with the query and rendering of this component. Please supply all the details below when logging an issue, it may help in tracing the cause.

Cannot read properties of undefined (reading 'keys')

TypeError: Cannot read properties of undefined (reading 'keys')
at https://polkadot.js.org/apps/page.52e72cfdfbfee15a.js:739:5722
at os (https://polkadot.js.org/apps/modu.7e3c157f7ba748aa.js:2:106322)
at Bu (https://polkadot.js.org/apps/modu.7e3c157f7ba748aa.js:2:126444)
at uu (https://polkadot.js.org/apps/modu.7e3c157f7ba748aa.js:2:118745)
at $o (https://polkadot.js.org/apps/modu.7e3c157f7ba748aa.js:2:59735)
at https://polkadot.js.org/apps/modu.7e3c157f7ba748aa.js:2:125246
at Cu (https://polkadot.js.org/apps/modu.7e3c157f7ba748aa.js:2:125251)
at au (https://polkadot.js.org/apps/modu.7e3c157f7ba748aa.js:2:117748)
at _ (https://polkadot.js.org/apps/modu.5d78a093107b232f.js:2:63097)
at MessagePort.I (https://polkadot.js.org/apps/modu.5d78a093107b232f.js:2:63629)
```

* **Please tell us about your environment:**

  - Version: n/a
  - Environment: Docker

    - [ ] Node.js
    - [x] Browser
    - [ ] Other (limited support for other environments)

  - Language:

    - [ ] JavaScript
    - [ ] TypeScript (include tsc --version)
    - [x] Other

* **Additional comments:**

The error happens due to this line:
https://github.com/polkadot-js/apps/blob/6ca0731c49d1bb24be79505339860586368904f9/packages/page-contracts/src/Contracts/Summary.tsx#L33

`CodeStorage` was deleted from `contracts` in this commit:
https://github.com/paritytech/substrate/commit/ffbc02fecc5507b6a21415eeaffcdd80da5054da

* **Possible solution(?):**
Replace `codeStorage` with `pristineCode`

[...]

PR #9951, opened 2023-10-15
#9901 - Fix issue with Contracts tab for polkadot-v1.0.0 based chain

vampik33:
#9901 - Fix issue with Contracts tab for polkadot-v1.0.0 based chain

https://github.com/polkadot-js/apps/issues/9901

### DIFF
--- a/packages/page-contracts/src/Codes/Code.tsx
+++ b/packages/page-contracts/src/Codes/Code.tsx
@@ -24,7 +24,7 @@ interface Props {
 function Code ({ className, code, onShowDeploy }: Props): React.ReactElement<Props> {
   const { t } = useTranslation();
   const { api } = useApi();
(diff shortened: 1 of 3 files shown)


jacogr (changes requested):
Thank you.

There are some tweaks required - with this we now break compat with older chains, so we need both. The approach I would suggest is the following -

Instead of -

```js
const optCode = useCall<Option<Codec>>(api.query.contracts.pristineCode, [code.json.codeHash]);
```

Rather use -

```js
const optCode = useCall<Option<Codec>>(api.query.contracts.pristineCode || api.query.contracts.codeStorage, [code.json.codeHash]);
```

(In all cases where the swap is made)

vampik33:
Thank you for review. 
Add additional checks you mentioned.

Also tested "Contracts" tab for both pre and post `polkadot-v1.0.0` branch chains

jacogr (approved):
Thank you.

PR #10052, opened 2023-11-24
Add innovatorchain in evm chains

masterdubs:
Add innovatorchain as evm type chain for correct address formatting.


jacogr (approved):
Thank you.

masterdubs:
Thanks for pushing the PR, seems it is not enough and I still have something to do to make it works with the chain ( wss://rpc.innovatorchain.com not adding it yet to the list ). This chain have Ethereum like account format, however now it seems activated but I don't see any in the account tab from extension ( all is allowed and so on ). If I put "frontier-template" in runtime it works. Could you give me a bit of direction what I need to push on the explorer ? 
Thank you again @jacogr  .

PR #10073, opened 2023-12-04
OnFinality adds websocket support for polkadex network

CrazySurfBoy:


jacogr (changes requested):
Looks fine overall, just need to make CI linting happy with key ordering. (So it can be merged)

Issue #9869, opened 2023-09-11
ink! contract upload and deploy failure


benoitdevos:
* **I'm submitting a ...**

  - [x] Bug report
  - [ ] Feature request
  - [ ] Support request
  - [ ] Other

* **What is the current behavior and expected behavior?**

**Current**:
When uploading an ink! contract (Developers > Contracts > Upload & deploy code), 
the UI freezes endless at this step.
```
contracts.instantiateWithCode
broadcast
```
and the contract never gets uploaded nor deployed. It does not seem to be an issue with the contract code or the connected blockchain, because it perfectly deploys and runs as expected when using contracts-ui.

The following error appears in the console:

```
TypeError: cyclic object value
    I https://cloudflare-ipfs.com/ipns/dotapps.io/comm.12610700b267d6ed.js:4084
    d https://cloudflare-ipfs.com/ipns/dotapps.io/modu.3b4705abeb23323f.js:1
    a https://cloudflare-ipfs.com/ipns/dotapps.io/modu.b8aa2d4a2c75f52e.js:50
    promise callback*17965/a/< https://cloudflare-ipfs.com/ipns/dotapps.io/modu.b8aa2d4a2c75f52e.js:50
    setTimeout handler*a https://cloudflare-ipfs.com/ipns/dotapps.io/modu.b8aa2d4a2c75f52e.js:50
    d https://cloudflare-ipfs.com/ipns/dotapps.io/modu.3b4705abeb23323f.js:1
    next https://cloudflare-ipfs.com/ipns/dotapps.io/modu.5d78a093107b232f.js:2
    _next https://cloudflare-ipfs.com/ipns/dotapps.io/modu.5d78a093107b232f.js:2
    next https://cloudflare-ipfs.com/ipns/dotapps.io/modu.5d78a093107b232f.js:2
    s https://cloudflare-ipfs.com/ipns/dotapps.io/modu.5d78a093107b232f.js:2
```

**Expected**
The contract is deployed, as it's the case when using [https://contracts-ui.substrate.io/](https://contracts-ui.substrate.io/).

* **What is the motivation for changing the behavior?**

* **Please tell us about your environment:**

  - Chain: Tested on Astar Shibuya testnet (rpc: shibuya-rpc.dwellir.com)
  - Version: 
  - Environment:

    - [ ] Node.js
    - [x] Browser
    - [ ] Other (limited support for other environments)

  - Language:

    - [x] JavaScript
[...]

PR #10078, opened 2023-12-05
Fix json stringify circular object error

sweatpotato13:
# Overview

Fix json stringify circular object error

This PR can resolve https://github.com/polkadot-js/apps/issues/9869

# Changes

```typescript
export function handleTxResults (handler: 'send' | 'signAndSend', queueSetTxStatus: QueueTxMessageSetStatus, { id, txFailedCb = NOOP, txSuccessCb = NOOP, txUpdateCb = NOOP }: QueueTx, unsubscribe: () => void): (result: SubmittableResult) => void {
  return (result: SubmittableResult): void => {
    if (!result?.status) {
      return;
    }

    const status = result.status.type.toLowerCase() as QueueTxStatus;

    console.log(`${handler}: status :: ${stringify(result)}`);

    queueSetTxStatus(id, status, result);
    txUpdateCb(result);
```

`JSON.stringify`  function It causes an error when deploying the ink! contract.

So, the process after that is not executed and the spinner continues.

And even if the block is actually finalized, the deployed contract will not appear on the UI.

I saw that a circular object error occurred when run `JSON.stringify(result)`  and resolved this using the circular JSON parser.

success

### DIFF
--- a/packages/react-signer/package.json
+++ b/packages/react-signer/package.json
@@ -17,7 +17,8 @@
   "version": "0.133.2-3-x",
   "dependencies": {
     "@polkadot/hw-ledger": "^12.6.1",
-    "@polkadot/react-components": "^0.133.2-3-x"
+    "@polkadot/react-components": "^0.133.2-3-x",
+    "flatted": "3.2.9"
   },
   "peerDependencies": {
     "react": "*",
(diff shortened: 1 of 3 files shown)


jacogr:
Really don't want to pull in additional packages on these, this is additional longer-term maintenance. (Generally where we use `stringify` it is from `@polkadot/util`)

PR #10137, opened 2024-01-02
Add Polimec Network 

vstam1:


vstam1:
Hi @jacogr, I hope it's alright I directly message you. It would be amazing if you could look at this PR. We are in the middle of the process of going live, and currently have to tell all our users that they have to add the RPC node manually.

Nexus2k (approved):

jacogr (approved):

PR #10151, opened 2024-01-08
Vara Network staking params

EugenWay:


wirednkod (approved):
lgtm

jacogr (changes requested):
Not 100% sure where the updated `yarn.lock` comes from since there were no `package.json` changes. (The rest are fine, the specific lock file creates noise and makes it difficult to review to see if funnies snuck in)

PR #10164, opened 2024-01-09
Set up Vara inflation pararms

EugenWay:


EugenWay:
@jacogr sorry. i just closed the last pr with wrong yarn.lock. Here is only right files and all checks pased

jacogr (approved):

PR #10180, opened 2024-01-15
Ternoa is a live network

soufiane-capsule:


soufiane-capsule:
Dear @kdembler 

Ternoa.network is actually a live network and a production one, I would like to understand how do you proceed with monitoring?

I have highly doubts that you generate some traffic which can be considered as malicious traffic and getting blocked

Thank you for your time and consideration. I look forward to your response.

mohsinriaz17:
Hi @jacogr 

we tried running the `ci:chainEndpoints` many times, Ternoa did not end up in the error list in any of those attempts.
we suspect that the IP form which the test is being run ended up in the banned IP list on cloudflare proxy.

if possible for you to provide us the IP, we can look into banned list to make sure we are on the right track to finding why the test pass for us but failed for you

If the issue is confirmed to be banned IP related, we can working on making the cloudflare policy more lenient

Thank you

soufiane-capsule:
> I'm not sure why you've tagged me here but I have no idea what this is about and have very little to do Polkadot apps

Sorry I made a typo 🙏 appologies

jacogr (approved):

jacogr:
There is no "single IP" - the [check job is run on Github CI](https://github.com/polkadot-js/apps/blob/master/.github/workflows/chain-endpoints.yml), so whichever IP the assigned runner has is what is used for the specific run.

PR #10181, opened 2024-01-15
Fix Files(IPFS) module status link and download link

XueMoMo:
1. The old download link is no longer available, so we should use cf-ipfs.com
2. Fix the status link to link to specific file information

### DIFF
--- a/packages/page-files/src/CrustFiles.tsx
+++ b/packages/page-files/src/CrustFiles.tsx
@@ -43,7 +43,7 @@ const shortStr = (name: string, count = 6): string => {
 };
 
 function createUrl (f: SaveFile) {
-  const endpoint = f.UpEndpoint || 'https://ipfs.io';
+  const endpoint = 'https://cf-ipfs.com';
 
   return `${endpoint}/ipfs/${f.Hash}?filename=${f.Name}`;
 }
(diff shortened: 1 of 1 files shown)


itsbirdo:
@TarikGul are you able to look at this please sir ?

packages/page-files/src/CrustFiles.tsx
```diff
@@ -43,7 +43,7 @@ const shortStr = (name: string, count = 6): string => {
 };
 
 function createUrl (f: SaveFile) {
-  const endpoint = f.UpEndpoint || 'https://ipfs.io';
+  const endpoint = 'https://cf-ipfs.com';
```
TarikGul, line 46: Why do we get rid of `f.UpEdnpoint`?

TarikGul (changes requested):
Just one question. Otherwise looks good.

packages/page-files/src/CrustFiles.tsx
XueMoMo: Because currently UpEndpoint is only used for uploading files and cannot be used to access files.

TarikGul (approved):
Apologies on the delay. LGTM

PR #10189, opened 2024-01-17
Allow pool metadata to display UTF-8 emojis

ArthurHoeke:
Currently Polkadot-js allows for UTF-8 emojis to be displayed on both the validators page and address identities. However, the pools page currently strips any non-standard characters from pool names.

For consistency this pull request adjusts the `transformName` regex to allow for UTF-8 emojis, as is also the case on staking.polkadot.network/#/pools.

### DIFF
--- a/packages/page-staking2/src/Pools/usePoolInfo.ts
+++ b/packages/page-staking2/src/Pools/usePoolInfo.ts
@@ -36,7 +36,7 @@ const OPT_MULTI = {
 };
 
 function transformName(input: string): string {
-  return input.replace(/[^\x20-\x7E\uD83C-\uDBFF\uDC00-\uDFFF]/g, '');
+  return input.replace(/[^\x20-\x7E\uD83C[\uDF00-\uDFFF]]/g, '');
 }
 
 function usePoolInfoImpl (poolId: BN): PoolInfo | null | undefined {
--- a/packages/page-staking2/src/Pools/usePoolInfo.ts
+++ b/packages/page-staking2/src/Pools/usePoolInfo.ts
@@ -35,8 +35,8 @@ const OPT_MULTI = {
       : null
 };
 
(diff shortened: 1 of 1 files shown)


jacogr:
Identities -

- don't show emojis when the identity has not been approved
- do show emojis when the identity is deemed at least reasonable by a registrar

aka not by default until it is deemed "non-spam" and viable.

There is no built-in spam/frustration filter on these specific user inputs, hence the filter.

ArthurHoeke:
> Identities -
> 
> * don't show emojis when the identity has not been approved
> * do show emojis when the identity is deemed at least reasonable by a registrar
> 
> aka not by default until it is deemed "non-spam" and viable.
> 
> There is no built-in spam/frustration filter on these specific user inputs, hence the filter applied here.

How about allowing say up to 4 emoji's to prevent spam and filtering out any unwanted ones? As of right now a bad actor could already put anything they'd want in the pool metadata, don't think there aren't that many potentially bad emojis out there.

`
function transformName(input: string): string {
  const naughtyEmojis = ['🔫','💩','☠️','🍆'];

  let sanitizedInput = input.replace(new RegExp(naughtyEmojis.join('|'), 'g'), '');

  const emojiRegex = /([\uD800-\uDBFF][\uDC00-\uDFFF]|[^\x20-\x7E\uD83C-\uDBFF\uDC00-\uDFFF])/g;
  const emojis = sanitizedInput.match(emojiRegex) || [];
  const limitedEmojis = emojis.slice(0, 4);

  emojis.forEach((emoji, index) => {
    sanitizedInput = sanitizedInput.replace(emoji, index < 4 ? limitedEmojis[index] : '');
  });

  return sanitizedInput;
}
`

ap211unitech:
Closing this PR as it appears to be outdated. Feel free to open a new PR if required.

PR #10196, opened 2024-01-19
chore: proofreading CONTRIBUTORS

sqltrigger:
Came across a few typos while reading about contributors


jacogr (commented):
This is automatically generated based on the PR commits - so once CI runs, this will be undone and it will be reverted to the same descriptions, with the typos.

TL;DR You are fighting against automation based on git history here, unless the git history itself it is rewritten to change the specific commits (which we won't do, the history is what is is) it will always stay as as based on the last commit from the author

PR #10201, opened 2024-01-19
Dwellir: update Polkadex parachain, add Polkadex mainnet

jakobilobi:
Hello,

We'd like to update our Polkadex endpoints.

Cheers


jakobilobi:
Note: `polkadex-mainnet-rpc.dwellir.com` is still syncing so might be best to wait a little before merging.

jacogr (approved):

PR #10209, opened 2024-01-22
chore: update parallel rpc

0x8f701:
we switched to use dwellir


jacogr (changes requested):
It probably makes sense in this case to uncomment the Dwellir entry and comment the Parallel entry (since it shows the origin, especially it multiples does get added again)

PR #10240, opened 2024-02-02
Typo in condition check for a curve type

rzadp:
This seems to be a typo, luckily there are probably no issues because of this since there are no other types of curves.


IkerAlus (approved):
Good catch!

TarikGul:
Before this goes in @rzadp do you mind pulling in master so the CI will run the tests again. Thanks again for the patience.

rzadp:
> Before this goes in @rzadp do you mind pulling in master so the CI will run the tests again. Thanks again for the patience.

Sure, done.

TarikGul (approved):

PR #10296, opened 2024-02-26
fix: updating the democracy proposal modal

Dudleyneedham:
Fix: The democracy page doesn't include a field to add the proposal length.

I have based the fix on the `propose` in the referendum and altered some variables.

The issue is that the proposal length is set to 0 with no option to change this and no warning. This will result in an invalid proposal.


TarikGul:
Hey @Dudleyneedham, I plan on reviewing this tomorrow. Thanks again for the patience

TarikGul (approved):
LGTM, seems sensible. Thanks 🚀

PR #10302, opened 2024-02-27
Adding RockX endpoint

AgoL5866:
Bring online RockX endpoint


TarikGul (changes requested):
Just need to run the linter

TarikGul (approved):

PR #10306, opened 2024-02-28
Enable LuckyFriday Bridgehub

paradox-tt:
Hello,

We've noticed that this entry was removed from Polkadot.JS; we've since retested the endpoint and it appears to be in working order.  Kindly confirm and re-add if there aren't any issues.

Regards,
Will

### DIFF
--- a/packages/apps-config/src/endpoints/productionRelayKusama.ts
+++ b/packages/apps-config/src/endpoints/productionRelayKusama.ts
@@ -902,7 +902,7 @@ export const prodParasKusamaCommon: EndpointOption[] = [
       'Dwellir Tunisia': 'wss://kusama-bridge-hub-rpc-tn.dwellir.com',
       'IBP-GeoDNS1': 'wss://sys.ibp.network/bridgehub-kusama',
       'IBP-GeoDNS2': 'wss://sys.dotters.network/bridgehub-kusama',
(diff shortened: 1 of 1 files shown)


packages/apps-config/src/endpoints/productionRelayKusama.ts
```diff
@@ -902,7 +902,7 @@ export const prodParasKusamaCommon: EndpointOption[] = [
       'Dwellir Tunisia': 'wss://kusama-bridge-hub-rpc-tn.dwellir.com',
       'IBP-GeoDNS1': 'wss://sys.ibp.network/bridgehub-kusama',
       'IBP-GeoDNS2': 'wss://sys.dotters.network/bridgehub-kusama',
-      // LuckyFriday: 'wss://rpc-bridge-hub-kusama.luckyfriday.io', // https://github.com/polkadot-js/apps/issues/9947
+      LuckyFriday: 'wss://rpc-bridge-hub-kusama.luckyfriday.io', // https://github.com/polkadot-js/apps/issues/9947
```
TarikGul, line 905: Lets remove the comment from the previous issue, and then were good to go.

paradox-tt: Thanks for the quick feedback, it's removed.

TarikGul (approved):

Issue #10305, opened 2024-02-28
Browser fork option is persist accross site visits


peetzweg:
While I was playing with the new "Fork Locally" feature I stumbled upon a few ux issues which broke my flow working with the pjs console.

Usually I have multiple tabs open each on different pages of the console for the same chain. Using the "Fork Locally" feature each of my browsers tabs of the same network have their own fork. So if I fund "Demo Account" in Tab A, it's not funded in Tab B.

https://github.com/polkadot-js/apps/assets/839848/49cc7600-01cc-4a23-b4d5-a4248a4352f7

The other hiccup I had with this feature is that the console persists the option to "Fork Locally" even when I visit a completely different chain. So when I was on POlkadot and did "Fork Locally", and than open a link pointing to PJs console using Kusama it will do the local fork thing. Which is most likely not what the user wants.

https://github.com/polkadot-js/apps/assets/839848/fa77f5ab-7771-40d7-96cd-7bf412f12606

All in all I think this "Fork Locally" mode should not be persisted between console visits.

adamsteeber:
I'm going to dog pile on this issue by mentioning the "Fork Locally" button **should not** be where the "Switch" button used to be. I keep clicking it out of muscle memory and it bricks my computer because Chopsticks is very resource intensive. The "Fork Locally" button should be less conspicuous and moved somewhere else.

qiweiii:
> Using the "Fork Locally" feature each of my browsers tabs of the same network have their own fork. So if I fund "Demo Account" in Tab A, it's not funded in Tab B.

@peetzweg right now we use a web worker for chopsticks instance in browser, so it only works in one window. Maybe we can experiment with shared worker instead so it will work across tabs under the same origin.

> persists the option to "Fork Locally" even when I visit a completely different chain

the local fork state is save in browser localstorage, so it persists after you refresh. To stop using local fork, you can just click on the `Switch` button again.

For switch chains, I didn't think abt user doing it by changing url directly, if you switch chain using the UI, you can use `Switch` button to quit local fork mode, but this case can be fixed though, we can make it go back to normal mode when the rpc endpoint is changed

qiweiii:
@adamsteeber 
How about switch the `Switch` and `Fork Locally` buttons, will that be better? so `Switch` button will go back to its old position. 
Or maybe add a secondary dropdown next to switch button and put fork button inside.

Any other suggestion is welcomed. @TarikGul

peetzweg:
[...]

PR #10308, opened 2024-02-29
fix: switch rpc in url should not auto use local fork

qiweiii:
As mentioned in #10305, this PR includes
- Fix when user change rpc endpoint to another chain in url and load the page, it should not automatically enter local fork mode (although refreshing the page will still be in local fork mode)
- Refactor `Sidebar` buttons slightly and change the order of `Switch` and `Fork` buttons, so `Switch` remains in the old position that users are used to.


TarikGul (approved):
Tested this out locally, thanks for the fix!

Issue #10288, opened 2024-02-22
PJS browser app is not listing all on-chain Governance->Preimages 


acatangiu:
* **I'm submitting a ...**

  - [x] Bug report
  - [ ] Feature request
  - [ ] Support request
  - [ ] Other

* **What is the current behavior and expected behavior?**

For some chains (saw this on PolkadotCollectives and Kusama, but possibly others), PJS is not showing all `Preimages` that are still on-chain.

For example, this preimage was submitted to Collectives: https://collectives.statescan.io/#/extrinsics/3201877-2
and on-chain preimage `0x953d2a661349cbbf2f7f14af56bad37393a54d1878b593fc4ee9e893bd7eb1ba` is not showing up in `Governance -> Preimages`.

Doing a bit of debugging, it seems that PJS is:

only listing the items found using deprecated [pallet_preimage::StatusFor](https://github.com/paritytech/polkadot-sdk/blob/9bf1a5e23884921498b381728bfddaae93f83744/substrate/frame/preimage/src/lib.rs#L173) storage

![Screenshot 2024-02-22 at 17 01 05](https://github.com/polkadot-js/apps/assets/31917973/0572808d-293d-479e-b72b-1455c263df3c)

and not listing items found using new [pallet_preimage::RequestStatusFor](https://github.com/paritytech/polkadot-sdk/blob/9bf1a5e23884921498b381728bfddaae93f83744/substrate/frame/preimage/src/lib.rs#L178C18-L178C34) storage item.

![Screenshot 2024-02-22 at 17 01 22](https://github.com/polkadot-js/apps/assets/31917973/4c5b825d-ae7d-4c4c-a8d5-1e22e68ceeaa)

* **What is the motivation for changing the behavior?**

Preimages should show up in the PJS GUI for a number of reasons:
- it allows easy listing
- it allows easy decoding and inspection of preimages
- it allows easy access to un-note and refund deposit once it is no longer needed
- governance proposals using preimages will correctly render the inner call (when PJS can find and decode the preimage)

* **Please tell us about your environment:**

  - Version: `apps v0.133.2-87`
  - Environment:

    - [ ] Node.js
    - [x] Browser
    - [ ] Other (limited support for other environments)

TarikGul:
Fixed via: https://github.com/polkadot-js/apps/pull/10310

Issue #10141, opened 2024-01-04
Preimages now use RequestStatusFor


dnjscksdn98:
* **I'm submitting a ...**

  - [ ] Bug report
  - [ ] Feature request
  - [x] Support request
  - [ ] Other

* **What is the current behavior and expected behavior?**
Due to the major update on `pallet-preimage` ([ref.](https://github.com/paritytech/polkadot-sdk/pull/1363)), it seems that the original storage value `StatusFor` has been deprecated and now `RequestStatusFor` is used instead.
But it seems that the preimage page ([ref.](https://github.com/polkadot-js/apps/blob/9c911354f421fca2c1a6b159bda5934772064656/packages/apps-routing/src/preimages.ts#L14)) are only referencing to the `StatusFor` field.
So for chains that had upgraded their runtime higher than `v1.2.0` (that contains the upper-mentioned PR), will have some incompatibility among preimages. I guess the stored preimage data is fine, but it's just the data display issue on the page.
In my opinion, both `StatusFor` and `RequestStatusFor` should be supported. The reason is because some chains might not yet upgraded their runtime, or either, the lazy storage migration might not have been done.
<img width="1761" alt="스크린샷 2024-01-04 오후 1 42 57" src="https://github.com/polkadot-js/apps/assets/59077132/ec797517-adc3-4fff-a870-bfcd1483b0a3">

* **What is the motivation for changing the behavior?**

* **Please tell us about your environment:**

  - Version:
  - Environment:

    - [ ] Node.js
    - [ ] Browser
    - [ ] Other (limited support for other environments)

  - Language:

    - [ ] JavaScript
    - [ ] TypeScript (include tsc --version)
    - [ ] Other

dnjscksdn98:
Also mentioned on this [issue](https://github.com/paritytech/polkadot-sdk/issues/2869), every action related to preimages won't work such as governance using `pallet-democracy`. The issues that I've encountered was that even though the proposal has been passed, `pallet-scheduler` wasn't able to call the proposal emitting a `CallUnavailable` event.

albertov19:
Hey, @dnjscksdn98, I noticed the same issue as new preimages are no longer shown in the "Preimages" tab of governance. I guess this is expected because of what you mentioned above. 

I do see also a bunch of errors in the console

andresilva:
cc @TarikGul could you look into this? I just found this issue while submitting a fellowship proposal. Came here to report it but the issue already exists.

[...]

PR #10310, opened 2024-02-29
Support requestStatusFor for usePreimage

TarikGul:
closes: https://github.com/polkadot-js/apps/issues/10288, https://github.com/polkadot-js/apps/issues/10141

Due to the deprecation of `StatusFor` given https://github.com/paritytech/polkadot-sdk/pull/1363 the `usePreimage` hook needs to support both `StatusFor` and `RequestStatusFor` calls when gathering data for preimage hashes.

### DIFF
--- a/packages/apps-routing/src/preimages.ts
+++ b/packages/apps-routing/src/preimages.ts
@@ -11,6 +11,7 @@ export default function create (t: TFunction): Route {
     display: {
       needsAccounts: true,
       needsApi: [
+        'query.preimage.requestStatusFor',
         'query.preimage.statusFor',
         'tx.preimage.notePreimage'
       ]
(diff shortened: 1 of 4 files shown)


TarikGul:
Okay so I did some testing with kusama, and this should fix in all the areas that use `usePreimage`. The only page that is a little unknown at the moment and is not related to this specifically is the whitelist page which will be investigated seperately

wischli:
@TarikGul Unfortunately this breaks backwards compatibility with chains which still run on Polkadot v0.9.43. Centrifuge is one of these. We cannot access the `Preimages` page anymore, which basically block users from being able to preimages and therefore referenda as well as council motions. Could you please re-add support for the old API as well?

TarikGul:
> @TarikGul Unfortunately this breaks backwards compatibility with chains which still run on Polkadot v0.9.43. Centrifuge is one of these. We cannot access the `Preimages` page anymore, which basically block users from being able to preimages and therefore referenda as well as council motions. Could you please re-add support for the old API as well?

Interesting, I didn't remove any of the old calls, or old logic. Let me see the issue.

TarikGul:
@wischli Okay I found the issue, PR to resolve this will be up soon.

TarikGul:
@wischli Should be fixed.

PR #10315, opened 2024-03-01
fix: make local fork non persistent across sessions

qiweiii:
issue: #10305

This pr removes the perisitent local fork mode


TarikGul (approved):
Thanks so much for resolving the issue. LGTM 🚀

PR #10317, opened 2024-03-01
Fix break in backwards compatibility with preimages

TarikGul:
https://github.com/polkadot-js/apps/pull/10310 broke backwards compatibility for chains that don't have `requestStatusFor`. This PR fixes that issue.


TarikGul:
This just removes `requestStatusFor` for the `needsApi` routing for the display so it does not break forward compatibility either.

PR #10342, opened 2024-03-08
feat: local fork allow fake accounts

### LINKED ISSUE #10333: Allow signing from any account once in "Local Fork More" (Chopsticks pt2)
What would be very useful is to also provide an easier way to fake any account once you click into that button. I am looking into the code to see if the CS instance is executed with `mock-signature-host` or not, but even if yes, atm I need to hack my way around. Ideally, the UI would facilitate this and allow you to:

1. Add any account you wish merely by the public key
2. (fake) Sign with all of your accounts.  

Without too much manual work needed.

_Originally posted by @kianenigma in https://github.com/polkadot-js/apps/issues/9903#issuecomment-1978555648_

qiweiii:
resolve #10333

image

image

image

image


kianenigma:
Ideally, when we are in "fork-mode" there is a button that allows me to add any account by mere public key.

kianenigma:
In a shorter term, is there a place in the code where I can add some a few lines and make the App "think" it has some extra accounts? although I don't think this is enough, we also need to tweak the tx-submission code path to just fill the signature with `0x00000..` or something fake, knowing it will pass the runtime.

qiweiii:
@kianenigma You can add some accounts in setStorage [here](https://github.com/polkadot-js/apps/blob/master/packages/react-api/src/Api.tsx#L244-L248), but you are right, still need to fake signature. I am planning to do something like [this](https://github.com/AcalaNetwork/chopsticks/blob/509c6241767cc8c701c959c13613ac64420e2eef/packages/core/src/blockchain/block-builder.ts#L365-L375) for fork mode, when user want to sign using fake accounts

For now, depends on what you want to test, I think you could use chopsticks cli, for example, using [dry-run](https://github.com/AcalaNetwork/chopsticks/blob/558c3e82f384681c5e29585fe3dcbcc4e6b22f68/README.md#L73-L77)

ermalkaleci:
Screenshot 2024-03-14 at 2 27 31 PM
The best way to implement this is by adding a button here 'FakeSignSend' which will request an address. Do fake sign and sent tx. This will be available only on forks and chopsticks need to be started with --mock-signature-host

ermalkaleci:
or this window will be better
Screenshot 2024-03-14 at 2 31 31 PM

qiweiii:
Yes, it's definitely easier for user if can add these in the `sign and submit` popup, and I think we should also put the added fake account in context or somehow, so it's synced to accounts page and can be selected for any future sign and submit

packages/page-accounts/src/modals/LocalAdd.tsx
```diff
@@ -0,0 +1,130 @@
...
+export async function createLocalAccount (chain: Blockchain, address: string, { balance = 1000, name }: CreateOptions): Promise<ActionStatus> {
+  const status = { action: 'create' } as ActionStatus;
+  const existing = store.get('localAccounts') as { address: string, name: string, balance: number }[] | undefined;
+
+  try {
+    await setStorage(chain, {
+      System: {
+        Account: [
+          ...(existing || []).map((account) => {
+            return [[account.address], { data: { free: account.balance * 1e12 }, providers: 1 }];
```
ermalkaleci, line 39: I think we should not change user state. This account should remain as is. If founds are needed then user can use Alice account to fund the address

qiweiii:
@TarikGul could you try this have a review?

TarikGul:
@qiweiii Thanks so much for the PR, I am going to have this reviewed today!

TarikGul (changes requested):
Conflict just needs to be resolved. But overall, it looks great. I am just going to run it locally as well! Thanks again

TarikGul:
Ran it locally as well, and everything was smooth Great Job, and thanks so much for the PR.

qiweiii:
Thanks for review @TarikGul. I have fixed the conflict🫡

TarikGul (approved):
Looks great, and thanks again :)

packages/page-accounts/src/modals/LocalAdd.tsx
```diff
@@ -0,0 +1,132 @@
...
+export async function createLocalAccount (chain: Blockchain, address: string, { balance, name }: CreateOptions): Promise<ActionStatus> {
+  const status = { action: 'create' } as ActionStatus;
+
+  try {
+    keyring.addExternal(address, { isLocal: true, name, tags: ['local'] });
+
+    await setStorage(chain, {
+      System: {
+        Account: [
+          [[address], { data: { free: balance * 1e12 }, providers: 1 }]
```
ermalkaleci, line 38: it will be wise to not do this for mocked account because this will reset the state. Account has additional data that might be important. I suggest we don't modify it. If funds needed they can come from Alice account which has plenty of funds or the dev can call `setStorage` PRC and modify this.

qiweiii: I see, will have PR this remove this

kianenigma:
Amazing, thanks for this :)

PR #10350, opened 2024-03-12
Add XCM encoder utility

serban300:
I am part of the bridges & XCM team

We plan to change some extrinsics from the `XCM` pallet to receive an encoded `VersionedXcm` instead of a `VersionedXcm`. More details here: https://github.com/paritytech/polkadot-sdk/pull/1798 . 

This would make it harder to call these extrinsics from PJS and in order to improve the user experience we would like to add a new utility UI for crafting encoded `VersionedXcm` messages.


TarikGul (approved):
Overall, its really clean. Nice work and lgtm.

The conflicts just need to be resolved.

Edit: Also tested it on your branch and it looks great.

serban300:
@TarikGul thank you for the review ! I resolved the conflicts. Could you please take another look ?

PR #10365, opened 2024-03-18
dry run transactions prior to execution

moliholy:
This PR adds dry run check prior to extrinsic submission and prints an alert if the extrinsic would fail in its execution. Useful to warn the user before needlessly spending network fees.

Screenshots:

Screenshot 2024-03-18 at 08 47 48

### DIFF
--- a/packages/react-signer/src/PaymentInfo.tsx
+++ b/packages/react-signer/src/PaymentInfo.tsx
@@ -80,7 +80,7 @@ function PaymentInfo ({ accountId, className = '', extrinsic, isHeader }: Props)
         <MarkWarning content={t('The account does not have enough free funds (excluding locked/bonded/reserved) available to cover the transaction fees without dropping the balance below the account existential amount.')} />
       )}
       {isDryRunError && (
-        <MarkError content={t('The transaction would not be successfully executed')} />
+        <MarkError content={t('The transaction did not pass a dry run and would likely not be successfully executed')} />
       )}
     </>
   );


Moliholy:
@TarikGul any feedback for the PR? do you think it's a good idea?

packages/react-signer/src/PaymentInfo.tsx
```diff
@@ -67,6 +79,9 @@ function PaymentInfo ({ accountId, className = '', extrinsic, isHeader }: Props)
       {isFeeError && (
         <MarkWarning content={t('The account does not have enough free funds (excluding locked/bonded/reserved) available to cover the transaction fees without dropping the balance below the account existential amount.')} />
       )}
+      {isDryRunError && (
+        <MarkError content={t('The transaction would not be successfully executed')} />
```
TarikGul, line 83: I would ensure we also bring up that it failed via the dry run.

So something like: `The transaction did not successfully pass a dry run, something something`

TarikGul:
Apologies on the delay, I haven't gotten around to truly running this yet. I do think it is a good idea though, i just need to test it locally as well!

packages/react-signer/src/PaymentInfo.tsx
Moliholy: I just changed it to `The transaction did not pass a dry run and would likely not be successfully executed`. Is that OK?

Moliholy:
No worries! I just updated the text and rebased to the latest `master` branch.

TarikGul:
I haven't forgotten about this PR :), I'll have a response soon but there was some internal discussion on our team on some potential problems that this can introduce. 

It's definitely a wanted feature so I would love to see this get in, but we just need to ensure the edge cases are resolved (to be written out soon)

Moliholy:
Thanks for the update. This is actually a small part of a bigger task that I had in mind. In a real-world production scenario a user would not have the private key accessible in the browser and instead would likely be using wallets like Talisman. In such case the account will very likely be locked and hence the feature implemented here would be useless, as it would not prevent real funds from being spent.

For that case I thought to implement a `Dry run` button near the `Sign and submit` one that would require the user to unlock the account and execute the dry run operation on demand. But before that I wanted to see how was the reception of this PR, as I supposed it could bring some friction.

IkerAlus:
Important to note that this feature implies calling an unsafe RPC method and most likely won't be available in any public provider.

Moliholy:
Which is, since we're on it, a huge disadvantage IMO comparing to Ethereum. See [here](https://web3js.readthedocs.io/en/v1.2.11/web3-eth.html?highlight=call#call):

> Executes a message call transaction, which is directly executed in the VM of the node, but never mined into the blockchain.

If you can query the chain state like in Ethereum while spending computational resources, why can't you dry-run a transaction also spending computational resources?

Actually, I hoped implementing this feature (and others that might be coming) could make some pressure to the community so that **by default** there exists a way for any client to perform dry-runs of their transactions.

IkerAlus:
@Moliholy Good points, it may be worth it to revive [this old issue in Polkadot-SDK](https://github.com/paritytech/polkadot-sdk/issues/68).

xlc:
Note that the dry run RPC isn't that useful. See this for more details https://github.com/paritytech/polkadot-sdk/issues/341
Once we have the new dry run runtime API, we should use it instead https://github.com/paritytech/polkadot-sdk/issues/4432
Otherwise should use Chopsticks to dry run locally. However it can take up to a few minutes so can't be enabled by default.

alastair0cook01-source:
This is a valid and practical observation about real-world security scenarios. In production environments, private keys are typically managed through secure wallets like Talisman, keeping accounts locked for safety. That means additional browser-based safeguards may have limited impact on actual funds. Tools like [saper pro](https://grasaper.pl/) can help streamline processes, but strong wallet-level protection remains essential for genuine transaction security and user trust overall.

Issue #10285, opened 2024-02-21
Update polkadot.js UI for coretime


seadanda:
* **I'm submitting a ...**
Feature request

* **What is the current behavior and expected behavior?**
The current UI does not have dedicated views for coretime. As a change in network architecture this should be reflected in the UI.
More information below motivation.

* **What is the motivation for changing the behavior?**
Coretime has been deployed on Rococo and Westend, and will soon be deployed on Kusama with Polkadot following in a few months. This is a permanent change to the network architecture and the way parachains obtain blockspace. The crowdloan/auction mechanism will no longer be used.
Currently the dedicated tabs for Parachains do not reflect the new scheme and the only way to see parachains on cores or the coretime sales status is to poke around in the chain state on the coretime and relay chains.

## More information
Currently in the Network->Parachains tab there are four tabs: Overview, Parathreads, Auctions, Crowdloans.

I would suggest as a starting point that the auctions and crowdloans tabs be removed, and the parathreads tab be changed to a parachains tab. (parathreads is no longer a relevant term).

A rough structure could then be:
### Overview
- Top level summary figures of current timeslice, region, sale and human-readable timelines
- Workload for current region
- Workplan for next region
- On-demand usage
- ...

Parachains
Registered parachains showing:
- parachain id
- whether they're on core or not, and whether that is via reservation, bulk or on-demand, and in the short term whether they are on a legacy lease
- if they are on bulk core whether they qualify to renew their core
- if they are on a legacy lease, the last region of that lease and when they can renew
- when they're next scheduled to be on core
- their last included block and when that was
- ...

Bulk sales
More details specifically about bulk coretime sale
- timeline including important points in time e.g. sale start, interlude, lead-in period, fixed price period, advance notice etc.
- list of regions (and owners?)
- details about current sale from `sale_info` e.g. number of cores for sale, sold and the prices/last price
- pricing information
- price projections for the next sale based on the number of cores sold and current last price
- ...

On-demand
It's a bit harder to think of what to show here that isn't already on the overview
- number of cores in the pool
- price of on-demand (and price history?)
- usage of on-demand as a percentage
- ...

### Summary
This is just a starting point and by no means the best layout or content that we could have.

vyto2021:
It does make sense

chungquantin:
Hi, is this issue open for outside contributors? I am really interested in contributing to this issue.

seadanda:
[...]

PR #10383, opened 2024-03-22
feat(coretime): add Overview page

bee344:
### Description
Adds the `Overview` page to the `Coretime` tab. This first step aims to add a `Coretime` tab to integrate its functionalities with PJS/APPS, with an `Overview` page that provides some relevant Summary information in a user-friendly manner. 

#### Tasks for Coretime Chain
- [x] Add Summary.
- [x] Add Workload.
- [x] Add Workplan as a separate table.
- [x] Update Workload to improve readability.
- [x] Add filters to Workplan and Workload to filter by core and timeslice.
- [x] Modify appearance.

#### Tasks for Relay Chain
- [x] Add Summary including Spot Traffic, core usage.
- [x] Add Core Descriptors as table.
- [x] Add Core Schedules as table

##### Optional:
- [ ] Modify appearance adding a timeline.

Regarding the appearance, I'm thinking on something simmilar to the `Referenda` page, since the amount of information for Workload and Workplan is substantial.
E.G. Image:

image

This is only as to reference the functionality of collapsing a table and adding a filter.

Part 1 of #10285


bee344:
A small sidenot: `onDemandCoretimeAssigner` wasn't included in the `Overview` since it isn't in use currently, and although we could make assumptions on how it would look liker, as evidenced with the key for `coreDescriptors`, sometimes stuff doesn't show up as it should. So I think it's better to wait until we have something to work with and then including it.

TarikGul:
After some discussion we decided to switch the logic into 2 pages. 

pages-broker
pages-coretime

TarikGul:
I think the core idea is there but there is some major cosmetic issues we need to fix before we release it:

1. Usage should have a percentage attached to it.  Next index is blank for me as well.
Screenshot 2024-06-10 at 7 16 47 PM

2. The left and right divs here are should have the same vertical width. Also the border of where the 100% task is denoted has no bottom border which is should.
Screenshot 2024-06-10 at 7 17 39 PM

3. When the string increases in task the div should not widen as below.
Screenshot 2024-06-10 at 7 19 04 PM

Currently I am not getting any info in the dropdowns.

TarikGul (changes requested):
Few cosmetic things to change.

bee344:
> 1. Usage should have a percentage attached to it. Next index is blank for me as well.

The info not showing is because of a missmatch between the pallet versions used by Rococo and Westend and the one used by Kusama (v1 vs v0), and that info not appearing in v0. I used Westend and Rococo for testing so I didn't catch that, my bad.

The `usage` percentage on the other hand, when you hover over the graph you can see what's the percentage assigned to that "assignment" (`idle`, `task` or `pool`). Since there can be 3 states at the same time, how do we choose which percentage we show? And if we show all three it gets messy. If you check on Rococo or Westend, you can see the graph has a different color assigned to each "assignment", and hovering over them you can see the percentage of cores dedicated to that "assignment". It's the way I found to have it look cleaner, but let me know if you still think we should change it.

TarikGul:
We also need to ensure we add the `.skip-npm` and `.skip-build` files.

piggydoughnut:
Visual updates :)

## Coretime Chain - Network - Coretime Broker Overview Tab

Screenshot 2024-09-04 at 21 28 11

## Some data 

Screenshot 2024-09-04 at 21 28 37

PR #10423, opened 2024-04-03
feat: add asset filter

monkeyWie:
Currently there are more and more assets on polkadot, and we need to spend a lot of time on the page to find the assets, so I submitted this pr to try to improve that, both asset ID and name filtering are supported.

### Preview

pr2

### DIFF
--- a/packages/page-assets/src/Balances/index.tsx
+++ b/packages/page-assets/src/Balances/index.tsx
@@ -1,9 +1,10 @@
 // Copyright 2017-2024 @polkadot/app-assets authors & contributors
 // SPDX-License-Identifier: Apache-2.0
 
+import type { DropdownItemProps } from 'semantic-ui-react';
 import type { AssetInfo, AssetInfoComplete } from '../types.js';
 
-import React, { useEffect, useMemo, useRef, useState } from 'react';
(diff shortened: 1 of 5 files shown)


packages/page-assets/src/Balances/index.tsx
```diff
@@ -39,9 +40,9 @@ function Balances ({ className, infos = [] }: Props): React.ReactElement<Props>
   );
 
   const assetOptions = useMemo(
-    () => completeInfos.map(({ id, metadata }, index) => ({
+    () => completeInfos.map(({ id, metadata }) => ({
       text: `${metadata.name.toUtf8()} (${formatNumber(id)})`,
-      value: index
+      value: id.toNumber()
```
TarikGul, line 45: Why do we replace `index` here?

monkeyWie: Because if use index then we can't exact match by asset ID.

TarikGul: Makes total sense. thanks!

TarikGul (approved):

TarikGul:
Thanks so much for the PR.

PR #10425, opened 2024-04-03
Remove Automata 1RPC from Kusama

laboon:
The Automata 1RPC was previously removed from Polkadot RPC endpoint list, due to constant re-connects (see https://github.com/polkadot-js/apps/pull/10366 ). The same issue also occurs on their Kusama endpoint and I recommend to remove this as well.


TarikGul (approved):

TarikGul:
Thank you for the PR 🙏

PR #10431, opened 2024-04-04
Nodle testnet migrated from Rococo to Paseo

simonsso:


simonsso:
When nodle is migrated https://github.com/paseo-network/support/issues/36

TarikGul (approved):

PR #10440, opened 2024-04-05
[paseo] LAOS Omega added

asiniscalchi:
Add LAOS Omega parachain to Paseo. The network is already up and producing blocks


TarikGul (changes requested):
Why is there changes to the yarn.lock?

asiniscalchi:
> Why is there changes to the yarn.lock?

fixed

TarikGul (approved):

PR #10457, opened 2024-04-10
use runtime APIs to query block time

Moliholy:
Follow up from https://github.com/polkadot-js/apps/pull/10456#discussion_r1559407903

### DIFF
--- a/packages/react-hooks/src/useBlockInterval.ts
+++ b/packages/react-hooks/src/useBlockInterval.ts
@@ -6,11 +6,11 @@ import type { BabeGenesisConfiguration } from '@polkadot/types/interfaces/babe';
 
 import { useMemo } from 'react';
 
-import { A_DAY } from '@polkadot/react-hooks/useBlocksPerDays';
 import { BN, BN_THOUSAND, BN_TWO, bnMin } from '@polkadot/util';
 
 import { createNamedHook } from './createNamedHook.js';
 import { useApi } from './useApi.js';
+import { A_DAY } from './useBlocksPerDays.js';
 import { useCall } from './useCall.js';
 
 // Some chains incorrectly use these, i.e. it is set to values such as 0 or even 2
--- a/packages/react-hooks/src/useBlockInterval.ts
+++ b/packages/react-hooks/src/useBlockInterval.ts
@@ -2,48 +2,29 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import type { ApiPromise } from '@polkadot/api';
+import type { BabeGenesisConfiguration } from '@polkadot/types/interfaces/babe';
 
 import { useMemo } from 'react';
 
-import { BN, BN_THOUSAND, BN_TWO, bnMin } from '@polkadot/util';
+import { BN } from '@polkadot/util';
 
 import { createNamedHook } from './createNamedHook.js';
 import { useApi } from './useApi.js';
-import { A_DAY } from './useBlocksPerDays.js';
-
-// Some chains incorrectly use these, i.e. it is set to values such as 0 or even 2
-// Use a low minimum validity threshold to check these against
-const THRESHOLD = BN_THOUSAND.div(BN_TWO);
-const DEFAULT_TIME = new BN(6_000);
-
-function calcInterval (api: ApiPromise): BN {
-  return bnMin(A_DAY, (
-    // Babe, e.g. Relay chains (Substrate defaults)
-    api.consts.babe?.expectedBlockTime ||
-    // POW, eg. Kulupu
-    api.consts.difficulty?.targetBlockTime ||
-    // Subspace
-    api.consts.subspace?.expectedBlockTime || (
-      // Check against threshold to determine value validity
-      api.consts.timestamp?.minimumPeriod.gte(THRESHOLD)
-        // Default minimum period config
-        ? api.consts.timestamp.minimumPeriod.mul(BN_TWO)
-        : api.query.parachainSystem
-          // default guess for a parachain
-          ? api.consts.aura?.slotDuration ?? DEFAULT_TIME.mul(BN_TWO)
-          // default guess for others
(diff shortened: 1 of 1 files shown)


packages/react-hooks/src/useBlockInterval.ts
```diff
@@ -2,48 +2,29 @@
...
-          // default guess for others
-          : DEFAULT_TIME
-    )
-  ));
-}
+import { useCall } from './useCall.js';
 
 function useBlockIntervalImpl (apiOverride?: ApiPromise | null): BN {
-  const { api } = useApi();
+  let { api } = useApi();
```
TarikGul, line 16: We shouldn't reassign here, instead have a `const { api } = useApi()`, and then rename the api below to be `curApi` or something of the sort, so: `const curApi = apiOverride || api;`

That way there is less confusion.

Moliholy: Done in [0ed9b77](https://github.com/polkadot-js/apps/pull/10457/commits/0ed9b779e3855f390409812b25075c1b0dec67a7)

packages/react-hooks/src/useBlockInterval.ts
```diff
@@ -2,48 +2,29 @@
...
+  const blockTimeAura = useCall(currApi.call.auraApi?.slotDuration && currApi.call.auraApi.slotDuration, []);
+  const blockTimeBabe = useCall(currApi.call.babeApi?.configuration && currApi.call.babeApi.configuration, [], {
+    transform: (data: BabeGenesisConfiguration | undefined) => data?.slotDuration
+  });
+
   return useMemo(
-    () => calcInterval(apiOverride || api),
-    [api, apiOverride]
-  );
+    () => (blockTimeAura || blockTimeBabe) ?? new BN(6_000),
```
TarikGul, line 25: I think it would also make more sense here to have it be like `(blockTimeAura || blockTimeBabe) ?? calcInterval(curApi), [api, apiOverride]`.

My initial thought is, not all chains configured to apps are required to use either babe, or aura AFAIU. For example this might end of breaking for `Subspace`, and `Kulupu`, as seen in `calcInterval`

TarikGul (changes requested):
Just a discussion point above on the affect of removing `calcInterval`.

packages/react-hooks/src/useBlockInterval.ts
Moliholy: Done in [c3383e9](https://github.com/polkadot-js/apps/pull/10457/commits/c3383e9710d6a6783b758290dbec11df36e9cef3).

packages/react-hooks/src/useBlockInterval.ts
```diff
@@ -2,14 +2,16 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import type { ApiPromise } from '@polkadot/api';
+import type { BabeGenesisConfiguration } from '@polkadot/types/interfaces/babe';
 
 import { useMemo } from 'react';
 
+import { A_DAY } from '@polkadot/react-hooks/useBlocksPerDays';
```
TarikGul, line 9: The import here should be `import { A_DAY } from './useBlocksPerDays.js';`

TarikGul (changes requested):
Looks like that last import nit should be the last blocker for this.

packages/react-hooks/src/useBlockInterval.ts
Moliholy: Fixed in [af37d65](https://github.com/polkadot-js/apps/pull/10457/commits/af37d6593521611490e1193a54ac7b48c57562bd).

Moliholy:
> Looks like that last import nit should be the last blocker for this.

Ready for the final review then.

packages/react-hooks/src/useBlockInterval.ts
```diff
@@ -40,10 +42,16 @@ function calcInterval (api: ApiPromise): BN {
 function useBlockIntervalImpl (apiOverride?: ApiPromise | null): BN {
   const { api } = useApi();
 
+  const currApi = apiOverride || api;
+  const blockTimeAura = useCall(currApi.call.auraApi?.slotDuration && currApi.call.auraApi.slotDuration, []);
```
TarikGul, line 46:
```suggestion
  const blockTimeAura = useCall<BN>(currApi.call.auraApi?.slotDuration && currApi.call.auraApi.slotDuration, []);
```

packages/react-hooks/src/useBlockInterval.ts
```diff
@@ -40,10 +42,16 @@ function calcInterval (api: ApiPromise): BN {
...
+    transform: (data: BabeGenesisConfiguration | undefined) => data?.slotDuration
+  });
+
   return useMemo(
-    () => calcInterval(apiOverride || api),
-    [api, apiOverride]
-  );
+    () => (blockTimeAura || blockTimeBabe) ?? calcInterval(currApi),
+    [blockTimeAura, blockTimeBabe, currApi]
+  ) as BN;
```
TarikGul, line 54: Then with the above you can remove the `as BN`.

Thanks so much for the patience. I just realized we can remove the typecasting here by using the Generic input above.

Moliholy: Thanks to you for the feedback!

Addressed in [0e593af](https://github.com/polkadot-js/apps/pull/10457/commits/0e593afd5d5c90d6b1d2cd36eee770dc647a9138).

packages/react-hooks/src/useBlockInterval.ts
Moliholy: Addressed in [0e593af](https://github.com/polkadot-js/apps/pull/10457/commits/0e593afd5d5c90d6b1d2cd36eee770dc647a9138).

TarikGul (approved):
Thanks for the patience, and the PR.

PR #10461, opened 2024-04-11
fix: remove local account storage override

qiweiii:
to fix this [problem](https://github.com/polkadot-js/apps/pull/10342/files/01f677603bcd7c27ef03224897a2db0aa155d98f#r1560596487), so I removed account storage override when add a local account in `local fork` mode


TarikGul (changes requested):
Needs the linter

TarikGul (approved):

PR #10480, opened 2024-04-18
Update polkadot/* deps

TarikGul:
Updates: 

`@polkadot/ui` -> 3.6.6
`@polkadot/tools` -> 0.56.7
`@polkadot/phishing` -> 0.22.5


TarikGul:
The extension needs to be updated before this goes in.

PR #10491, opened 2024-04-22
Identity fix. Username has been introduced in identity pallet. 

niklabh:
identity.identityOf call now returning a tuple with username. Based on this [commit](https://github.com/paritytech/polkadot-sdk/commit/d1f678c0ec44e9734cf242247dab90c2678774a9#diff-4a95a3948151a729b1fa4e91da9c72cdc6cd2d1611842fca6f177efeed34fd1aR221)

### DIFF
--- a/packages/react-components/src/AccountSidebar/RegistrarJudgement.tsx
+++ b/packages/react-components/src/AccountSidebar/RegistrarJudgement.tsx
@@ -1,7 +1,7 @@
 // Copyright 2017-2024 @polkadot/react-query authors & contributors
 // SPDX-License-Identifier: Apache-2.0
 
-import type { Option } from '@polkadot/types';
+import type { Bytes, Option } from '@polkadot/types';
 import type { PalletIdentityRegistration } from '@polkadot/types/lookup';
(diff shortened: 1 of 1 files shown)


packages/react-components/src/AccountSidebar/RegistrarJudgement.tsx
```diff
@@ -34,9 +35,9 @@ const JUDGEMENT_ENUM = [
 ];
 
 const OPT_ID = {
-  transform: (optId: Option<PalletIdentityRegistration>): HexString | null =>
+  transform: (optId: Option<ITuple<[PalletIdentityRegistration]>>): HexString | null =>
```
TarikGul, line 38: I believe the accurate typing here would be: 

`Option<ITuple<[PalletIdentityRegistration, Option<Bytes>]>>`

Ref: https://github.com/polkadot-js/api/blob/master/packages/api-augment/src/polkadot/query.ts#L745

TarikGul (changes requested):
Just one small nit.

TarikGul (approved):

TarikGul:
Thank you so much for the fix

Issue #10482, opened 2024-04-18
Recent metadata update introduced several UI problems on Kusama


CoinStudioDOT:
After a recent metadata update, there are several UI issues on Kusama:

1. Network - Staking - Overview: Other stake field empty
2. Network - Account: Staking accounts showing all nominations as invalid and all nominated validators as Inactive (confirmed they are in Active set)
3. Network - Payout - Own stashes: Empty - no account displayed

TarikGul:
rel: https://github.com/polkadot-js/apps/issues/10396 https://github.com/polkadot-js/apps/issues/10004 https://github.com/polkadot-js/apps/issues/10388

CoinStudioDOT:
Thanks, appreciate your work on this understanding there is a lot in line. The challenge here is operational limitation to payout staking rewards from stash or have full control of nominating accounts.

Issue #10396, opened 2024-03-25
Staking upcoming todos


TarikGul:
Staking is in need of some TLC. With the upcoming changes for v1.4 in Kusama and Polkadot there is a list of tasks to focus on:

- [x] Compatibility with `legacyClaimedRewards` and `claimedRewards` for in `apps-routing` for staking. ([Resolved](https://github.com/polkadot-js/apps/pull/10394)) Rel: https://github.com/polkadot-js/apps/issues/10388
- [ ] Ensure apps-routing staking has support proper logic for the `erasStakers` check.
- [X] Ensure compatibility for upstream changes to eraExposure derive (ref: https://github.com/polkadot-js/api/pull/5815) 10.12.4 of the api which had the following changes were merged [here](https://github.com/polkadot-js/apps/pull/10355)
- [ ] Finish support in `page-staking` has full comaptibility  with the changes explained in https://github.com/polkadot-js/apps/issues/10004

Great Ref PR: https://github.com/polkadot-js/api/issues/5771

PR #10498, opened 2024-04-24
Fix Staking with latest breaking changes!

TarikGul:
closes: https://github.com/polkadot-js/apps/issues/10396
closes: https://github.com/polkadot-js/apps/issues/10482

With the most recent breaking changes in the `api` https://github.com/polkadot-js/api/pull/5860  https://github.com/polkadot-js/api/pull/5862 due to changes in pallet-staking, this PR hopes to fix a lot of those issues.

Just to preface, it is very possible that this PR does not fix every issue that might be present but it is a step in the right direction. What I mean by this is there is a lot changes and logic that were covered in the above PR's and it is difficult to track every single one down in the UI, and how that data might be nested to x amount of Components.

I also updated all the following deps:

`@polkadot/api` -> `11.0.1`
`@polkadot/extension` -> `0.47.2`
`@polkadot/phishing` -> `0.22.6`
`@polkadot/tools` -> `0.56.8`

**Disclaimer!!** This does not touch the page-staking2 components (As noted by Jaco it seems to have been a staking test app). When the time comes if needed I can update that as well.


TarikGul:
By removing `withClaimedRewardsEras` it ended up working as is. Not sure how it effects claimedRewards yet though. I am also willing to just merge this in and fix that regression seperately.

Issue #10493, opened 2024-04-22
Regression / Identity: Judgements no longer possible


chevdor:
- [x] Bug report
  - [ ] Feature request
  - [ ] Support request
  - [ ] Other

* **What is the current behavior and expected behavior?**

Setting the judgement via the "Add identity judgement" link on an account's view no longer works.
The UI hangs.

<img width="716" alt="image" src="https://github.com/polkadot-js/apps/assets/738724/0e92951b-530a-4c21-a6d9-d70d4cb323a4">

with the error:
```
comm.88d5168f95c4b7c1.js:3892 Error: useCall(identity.identityOf):: Cannot read properties of undefined (reading 'hash'):: TypeError: Cannot read properties of undefined (reading 'hash')
    at transform (comm.88d5168f95c4b7c1.js:1724:919)
    at comm.88d5168f95c4b7c1.js:3892:41322
    at modu.c5e227c622fddc08.js:1:111325
    at modu.b05dd42ef8c37bcb.js:2:429403
    at u (comm.88d5168f95c4b7c1.js:3892:40118)
    at comm.88d5168f95c4b7c1.js:3892:41347
    at modu.c5e227c622fddc08.js:1:111325
    at modu.b05dd42ef8c37bcb.js:2:429403
```

* **What is the motivation for changing the behavior?**

I would rather have the behavior NOT change :)

* **Please tell us about your environment:**
  - Version:
  - Environment:

    - [ ] Node.js
    - [x] Browser
    - [ ] Other (limited support for other environments)

  - Language:

    - [x] JavaScript
    - [ ] TypeScript (include tsc --version)
    - [ ] Other

chevdor:
The issue can be seen on Polkadot and Kusama.
The selected RPC endpoint does not matter.

TarikGul:
@chevdor Let me know if this fixes your issue. Happy to close once confirmed.

chevdor:
It does fix the issue, thx.

niklabh:
Sorry, this has broken polkadex and older chains now.

TarikGul:
@niklabh I guess our best route is to check the type to see if its a tuple or not that either do the legacy option or the new one.

PR #10570, opened 2024-05-13
Backwards compatibility for identityOf

TarikGul:
rel: https://github.com/polkadot-js/apps/issues/10493

https://github.com/polkadot-js/apps/pull/10491 broke backwards compatibility for chains that did not update their identity pallets. This ensures that there is both forward and backwards compatibility

### DIFF
--- a/packages/react-components/src/AccountSidebar/RegistrarJudgement.tsx
+++ b/packages/react-components/src/AccountSidebar/RegistrarJudgement.tsx
@@ -44,7 +44,7 @@ const OPT_ID = {
     return !id
       ? null
       : Array.isArray(id)
-        ? optId.unwrap()[0].info.hash.toHex()
+        ? id[0].info.hash.toHex()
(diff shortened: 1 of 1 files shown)


packages/react-components/src/AccountSidebar/RegistrarJudgement.tsx
```diff
@@ -35,10 +35,16 @@ const JUDGEMENT_ENUM = [
...
 const OPT_ID = {
-  transform: (optId: Option<ITuple<[PalletIdentityRegistration, Option<Bytes>]>>): HexString | null =>
-    optId.isSome
+  transform: (optId: Option<ITuple<[PalletIdentityRegistration, Option<Bytes>]>>): HexString | null => {
+    const id = optId.isSome
+      ? optId.unwrap()
+      : null;
+
+    // Backwards compatibility - https://github.com/polkadot-js/apps/issues/10493
+    return Array.isArray(id)
```
TarikGul, line 44: Need to return null if Id is null

Issue #10573, opened 2024-05-14
Uncaught error in balances tab 


nuel77:
Hey,
We have noticed an issue in the balances page (network -> assets -> balances), The page throws the following error:

<img width="1439" alt="image" src="https://github.com/polkadot-js/apps/assets/50878033/efda7536-9116-465c-9d4a-a5899b6b7cc2">

Can we get more insight into this ? if this is an issue with our node runtime or something wrong in the apps-ui itself?

Thanks for your support!

TarikGul:
Yes this looks to be an issue for assetHub on polkadot, and kusama as well. Will look into it today!

TarikGul:
For `Polkadex` this error seems to be the source of the issue: `Number can only safely store up to 53 bits`

But for assetHub I still need to find the root of the problem

TarikGul:
Ahh okay, balances works fine on assetHub its just for polkadex in this instance. That was my bad I just needed to link some accounts.

PR #10576, opened 2024-05-14
Fix polkadex assets balances page

TarikGul:
closes: https://github.com/polkadot-js/apps/issues/10573

@nuel77 Do you mind testing this out and confirming that it fixes your issue?


nuel77:
yep! Thanks for the prompt actions @TarikGul 👍

Issue #10580, opened 2024-05-15
Set Identity fails on accounts that already have identities set


TarikGul:
Replaces: https://github.com/polkadot-js/apps/issues/10540 (See the following issue for more information).

If an account has an identity set already if you try to run the `Set Identity` functionality under `...` in the `/accounts` page it will fail. 

The following error will occur: 
```
Uncaught error. Something went wrong with the query and rendering of this component. Please supply all the details below when logging an issue, it may help in tracing the cause.

Cannot read properties of undefined (reading 'display')

TypeError: Cannot read properties of undefined (reading 'display')
at https://polkadot.js.org/apps/page.e65471d6c0991232.js:43:12598
at os (https://polkadot.js.org/apps/modu.a73333ada4998f0e.js:2:107879)
at Bu (https://polkadot.js.org/apps/modu.a73333ada4998f0e.js:2:128001)
at https://polkadot.js.org/apps/modu.a73333ada4998f0e.js:2:124605
at E (https://polkadot.js.org/apps/modu.7c3b262d6289bfdf.js:2:63843)
at MessagePort.I (https://polkadot.js.org/apps/modu.7c3b262d6289bfdf.js:2:64375)
```

More information can be found in [this forum post](https://forum.polkadot.network/t/people-chain-launch-and-identity-migration-plan/5930) which discusses the migration process of the identity pallet to the people chain.

PR #10581, opened 2024-05-15
Fix set identity on accounts

TarikGul:
closes: https://github.com/polkadot-js/apps/issues/10580

### DIFF
--- a/packages/page-accounts/src/modals/IdentityMain.tsx
+++ b/packages/page-accounts/src/modals/IdentityMain.tsx
@@ -117,43 +117,47 @@ function IdentityMain ({ address, className = '', onClose }: Props): React.React
   const [valDiscord, setValDiscord] = useState('');
   const [valWeb, setValWeb] = useState('');
   const [gotPreviousIdentity, setGotPreviousIdentity] = useState(false);
-  // HACK since the types are different in the peoples chains, we need to add this check to ensure we deal with legacy systems.
-  const [isLegacyIdentity, setIsLegacyIdentity] = useState(!specName.includes('people-'));
+  // HACK This ensures we can deal with legacy indentity information for chains while
+  // also giving support for the new identity logic inside of the Peoples chain.
+  // The new logic is specific to people system parachains.
+  const [isLegacyIdentity, setIsLegacyIdentity] = useState<boolean>(!specName.includes('people-'));
 
   useEffect((): void => {
     if (identityOpt && identityOpt.isSome) {
       const identity = identityOpt.unwrap();
       const foundInfo = Array.isArray(identity) ? identity[0].info : (identity as Registration).info;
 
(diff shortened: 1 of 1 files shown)


packages/page-accounts/src/modals/IdentityMain.tsx
```diff
@@ -100,8 +103,8 @@ function IdentityMain ({ address, className = '', onClose }: Props): React.React
 
       setData(info.web, setHasWeb, setValWeb);
 
-      [info.display, info.email, info.legal, info.riot, info.twitter, infoDiscord, info.web].some((info: Data) => {
-        if (info.isRaw) {
+      [info.display, info.email, info.legal, info.riot, info.twitter, infoDiscord, info.web].some((info: Data | null) => {
```
joepetrowski, line 106: See new `IdentityInfo`: https://github.com/polkadot-fellows/runtimes/blob/v1.2.1/system-parachains/people/people-kusama/src/people.rs#L96-L143

- `riot` is now `matrix`
- `github` and `discord` are first class members
- `additional` is no more

```suggestion
      [info.display, info.email, info.legal, info.matrix, info.twitter, info.discord, info.github, info.web].some((info: Data | null) => {
```

joepetrowski: Anyone with `additional` fields of `github` or `discord` had those fields migrated into the new struct.

joepetrowski:
Looks like this fixes setting Identity, but there also needs to be a PR to change Identity rendering? As in, when I'm on the Relay Chain, accounts with Identity on the People Chain should still render with that Identity.

packages/page-accounts/src/modals/IdentityMain.tsx
TarikGul: Ahh thanks for linking that, I will add those additions - Also need to keep the old structures though to make sure other chains don't have this break in the process.

TarikGul:
> Looks like this fixes setting Identity, but there also needs to be a PR to change Identity rendering? As in, when I'm on the Relay Chain, accounts with Identity on the People Chain should still render with that Identity.

Yea this only fixes setting the identity. The actual retrieving of the information is going to be a way larger overhaul. I don't know exactly how I am going to do it yet, but I will have a plan figured out today. Same as the above, but basically we need to keep the old while bringing in the new, so that we don't break everything for every chain that wont be using this identity structure.

joepetrowski:
> The actual retrieving of the information is going to be a way larger overhaul. I don't know exactly how I am going to do it yet

Hmm, I would not expect that, as the Identity on system chains rendered from the Relay Chain. I would expect that it just means always making a wss connection with People Chain and changing the provider of Identity info?

TarikGul:
Tested on both Polkadot, and westend-people and the logic works as it should.

PR #10598, opened 2024-05-20
Use apiSystemPeople for identities on relay chains

TarikGul:
[WIP]

This adds the functionality for getting identities from the peoples chain when connected to the relay chain.

### DIFF
--- a/packages/apps-config/src/endpoints/productionRelayKusama.ts
+++ b/packages/apps-config/src/endpoints/productionRelayKusama.ts
@@ -995,6 +995,7 @@ export const prodRelayKusama: EndpointOption = {
   dnslink: 'kusama',
   genesisHash: KUSAMA_GENESIS,
   info: 'kusama',
+  isRelay: true,
   linked: [
     ...prodParasKusamaCommon,
     ...prodParasKusama
--- a/packages/apps-config/src/endpoints/productionRelayPolkadot.ts
+++ b/packages/apps-config/src/endpoints/productionRelayPolkadot.ts
@@ -898,6 +898,7 @@ export const prodRelayPolkadot: EndpointOption = {
   dnslink: 'polkadot',
   genesisHash: POLKADOT_GENESIS,
   info: 'polkadot',
+  isRelay: true,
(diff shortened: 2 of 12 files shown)


TarikGul:
Last i need to ensure a safe way to disable all tx.identity calls for any chain but the people chain. While also keeping support for all parachains and other chains...

Also need to ensure the endpoint urls are populated for systemParachains

TarikGul:
Ready!

pandres95:
@TarikGul does this change allows for other parachains to register people as their identity provider?

TarikGul:
@pandres95 Yes! For sure, there is one small edge case I would need to fix, but I designed it in a way where any chain can opt in.

TarikGul:
@pandres95 For now though its just system chains, and relay chains that have support. With a follow up PR I will clean up one small piece of logic, then it should only take adding the following to apps config for the parachain you want to opt into the people chain for identities:

```
isPeopleForIdentity?: boolean;
relayName?: string;
```

pandres95:
> @pandres95 Yes! For sure, there is one small edge case I would need to fix, but I designed it in a way where any chain can opt in. 

Awesome @TarikGul! Is there a way I can propose that some parachains register themselves as compatible with the relay chain addresses?

We have this case where Kreivo (Kusama/2281) uses the same prefix as the relay, is using people chain as provider for identity, and for practical purposes, we expect users not to have to change their accounts in polkadot.js wallet to run on _"Any Chain"_ for their accounts being discoverable in Kreivo. I wanted to propose the concept of relay-compatible. And half of that proposal is included in this PR.

TarikGul:
> > @pandres95 Yes! For sure, there is one small edge case I would need to fix, but I designed it in a way where any chain can opt in.
> 
> Awesome @TarikGul! Is there a way I can propose that some parachains register themselves as compatible with the relay chain addresses?

As I added in an edit above, I actually don't know if a parachain would even be able to opt in to using the People chain as a source of identity. That being said I am definitely not the best person to give a great answer. cc: @joepetrowski

PR #10602, opened 2024-05-21
Remove Rococo Trappist RPC

PierreBesson:
The Trappist testnet (https://github.com/paritytech/trappist) is unfortunately no longer maintained and the network has been shut down by Parity.


TarikGul (changes requested):
I believe the logo in `/apps-config/src/ui/logos/nodes/trappist-rococo.png` needs to be removed as well

TarikGul:
Awesome thanks so much! Just as a heads up this will go in once the github actions status is back to stable

TarikGul (approved):

PR #10603, opened 2024-05-21
Adds null check before accessing multisigs

gupnik:


TarikGul (approved):
Sounds reasonable, was there a coinciding issue that was present, just so I can be aware?

gupnik:
> Sounds reasonable, was there a coinciding issue that was present, just so I can be aware?

Thanks for the quick review! 

Yeah, the omni node POC [here](https://github.com/kianenigma/pba-omni-node) doesn't work with the accounts section. Not sure why its the only one that returns it as `undefined` while the other templates continue to work though.

TarikGul:
Ahh okay makes sense, thanks so much for the PR. Actions are a bit degraded today so it will take a second before this can go in.

PR #10609, opened 2024-05-22
Set Kusama People as identity provider for Kreivo

pandres95:
Following #10598, we're registering People as the identity provider for Kreivo in polkadot.js apps

### DIFF
--- a/packages/apps-config/src/endpoints/productionRelayKusama.ts
+++ b/packages/apps-config/src/endpoints/productionRelayKusama.ts
@@ -438,8 +438,8 @@ export const prodParasKusama: Omit<EndpointOption, 'teleport'>[] = [
       Kippu: 'wss://kreivo.kippu.rocks/',
       Virto: 'wss://kreivo.io/'
     },
-    text: 'Kreivo - By Virto',
     relayName: 'kusama',
+    text: 'Kreivo - By Virto',
     ui: {
       color: '#294940',
       identityIcon: 'polkadot',
--- a/packages/apps-config/src/endpoints/productionRelayKusama.ts
+++ b/packages/apps-config/src/endpoints/productionRelayKusama.ts
@@ -432,12 +432,14 @@ export const prodParasKusama: Omit<EndpointOption, 'teleport'>[] = [
   {
     homepage: 'https://virto.network/',
     info: 'kreivo',
+    isPeopleForIdentity: true,
     paraId: 2281,
     providers: {
       Kippu: 'wss://kreivo.kippu.rocks/',
       Virto: 'wss://kreivo.io/'
     },
     text: 'Kreivo - By Virto',
+    relayName: 'kusama',
     ui: {
       color: '#294940',
       identityIcon: 'polkadot',


pandres95:
@TarikGul, do we have any updates on the support for this feature?

TarikGul:
Yes, apologies on the delay. I have had some pressing work to take care of. 

Have you confirmed if parachains can also leverage identities on the Peoples Chain?

pandres95:
Yes! We've checked and confirmed that's possible to manage via mechanisms like XCM (both for querying and setting).

[Here's an example](https://kusama.subsquare.io/user/Ftuq9bHvQb5NiU5JA7q79fxYn9FVBeRjNBHL3RH5raN9qck/posted/referenda of an account that recently set its identity on People Chain from a Parachain (note: this is a keyless account and only could have setup its identity this way) via XCM.

TarikGul:
> Yes! We've checked and confirmed that's possible to manage via mechanisms like XCM (both for querying and setting).
> 
> [Here's an example](https://kusama.subsquare.io/user/Ftuq9bHvQb5NiU5JA7q79fxYn9FVBeRjNBHL3RH5raN9qck/posted/referenda) of an account that recently set its identity on People Chain from a Parachain (note: this is a keyless account and only could have setup its identity this way) via XCM.

So if I understand this correctly given what you said, both querying and setting of the identity requires XCM calls?

The reason this is important is because currently the way both querying and setting of accounts identity works is by creating a seperate ws connection to the people chain, and making general storage calls, and or tx calls. This wont work for parachains if it requires XCM. There will need to be a significant addition to the code to make it work.

pandres95:
> > Yes! We've checked and confirmed that's possible to manage via mechanisms like XCM (both for querying and setting).
> > 
> > [Here's an example](https://kusama.subsquare.io/user/Ftuq9bHvQb5NiU5JA7q79fxYn9FVBeRjNBHL3RH5raN9qck/posted/referenda) of an account that recently set its identity on People Chain from a Parachain (note: this is a keyless account and only could have setup its identity this way) via XCM.
> 
> So if I understand this correctly given what you said, both querying and setting of the identity requires XCM calls?
> 
> The reason this is important is because currently the way both querying and setting of accounts identity works is by creating a seperate ws connection to the people chain, and making general storage calls, and or tx calls. This wont work for parachains if it requires XCM. There will need to be a significant addition to the code to make it work. 

Oh! Guess I didn't made myself understand correctly. 😅😅

What I meant with my last comment is that it is possible for parachains to use the services of People Chain via XCM.

But that was merely a comment to illustrate the level of integration of a Parachain with People Chain, but it's definitely not the case we're discussing here.

Not sure if I understand you correctly, but it seems you're talking about leveraging on People Chain as being somehow tightly  integrated to People Chain by some means (a.k.a. direct connection via Runtime), which, AFAIK, is not the case for any parachain (not even system ones), but is a service dApps tend to consume via a client.

In this sense, for example, system parachains don't have an `identity` pallet, because they don't need it. And instead what dApps need to know about identities can be found exactly by doing what you just mentioned.

That should apply for any parachain, not just system chains.

In conclusion, what has already been achieved by #10598 is perfectly fine for allowing end users to get the identity for their accounts in any parachain, no extra steps needed on the parachain's end.

The only thing that would be necessary IMHO is creating that separate WS connection whenever a parachain indicates in the config file (like we're doing in this PR) to use the people chain as _"identity provider"_.

This would be a HUGE improvement in terms of UX. 😊

TarikGul:
This has been on the backburner a bit too long, going to tackle this again, and get this merged.

TarikGul (approved):

PR #10624, opened 2024-05-29
[Crust] Add crust parachain 2 for crowdloan

wuhaixian1984:
Add a new endpoint for Crust Parachain, named 'Crust Parachain 2', which is used in Crowd Loan for the new paraId (3375).

This change will let our brand shows on the Crowd Loan page, so that contributors recognize us as official and wouldn't treat our Crowd Loan activity as SCAM.


TarikGul (changes requested):
The test is failing because apps-config tests do not allow multiple configs to have the same `ws` endpoint.

wuhaixian1984:
> The test is failing because apps-config tests do not allow multiple configs to have the same `ws` endpoint.

Another commit has been submitted to fix the failing test.

ap211unitech:
Closing this PR as it appears to be outdated. Feel free to open a new PR if required.

PR #10643, opened 2024-06-05
[endpoints] Adjusted some Dwellir endpoints.

awnns:
Added some active Dwellir endpoints that had fallen out at various times.


TarikGul (changes requested):
Ordering of keys needs to be adjusted.

awnns:
> Ordering of keys needs to be adjusted.

Thanks for letting me know. Sorry about that. It should be adjusted now.

TarikGul (approved):

PR #10651, opened 2024-06-07
fix(paseo): relay for identity

al3mart:
I have noticed several changes regarding the identity pallet, its types and how it is displayed in the app.

@validorange has reported that setting an identity was not available in the accounts tab of pjs/apps.

I believe this change should re-enable that and and use legacy identities instead of People chain ids. Please, point me in the right direction if that is not the case @TarikGul

### DIFF
--- a/packages/apps-config/src/endpoints/testingRelayPaseo.ts
+++ b/packages/apps-config/src/endpoints/testingRelayPaseo.ts
@@ -174,7 +174,7 @@ export const testRelayPaseo: EndpointOption = {
   dnslink: 'paseo',
   genesisHash: PASEO_GENESIS,
   info: 'paseo',
-  isPeopleForIdentity: true,
+  isPeopleForIdentity: false,
   isRelay: true,
   linked: [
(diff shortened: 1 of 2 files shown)


TarikGul:
Yea I think that should do it, thanks for the fix!

TarikGul (approved):

TarikGul:
Ahh actually, there needs to be one more addition, since its a relay chain: 

https://github.com/polkadot-js/apps/blob/master/packages/react-api/src/Api.tsx#L302

Needs to have `paseo` added to it as well. Unfortunately it's hard coded for now since everything is going to be transferred to the identity chain for all relays.

TarikGul (changes requested):
Changes the status: My bad for approving earlier

al3mart:
Thanks!

I have added a check for `paseo` following how the check for `polkadot` has been made.
Saw this at first, but was not sure if my first change was enough

TarikGul (approved):
Nice job, thanks for updating it!

PR #10669, opened 2024-06-14
HydraDX -> Hydration rebrand

jak-pan:


jak-pan:
I have similar question here as in common. Is the key in the map 'hydradx': taken from rpc.chain or something else? Should I add another config with hydration key to support the transition when we change this?

Cryptopod13 (approved):

Cryptopod13 (approved):

Cryptopod13 (approved):

Cryptopod13 (approved):
1

Cryptopod13 (commented):
1

Cryptopod13 (approved):
1

Cryptopod13 (approved):
3

TarikGul:
> I have similar question here as in common. Is the key in the map 'hydradx': taken from rpc.chain or something else? Should I add another config with hydration key to support the transition when we change this?

Just answered in Common :)

jak-pan:
Done @TarikGul

TarikGul (approved):

PR #10675, opened 2024-06-16
Increase Alice Balance - Chopsticks

DrW3RK:
1000 KSM is not enough for testing parachain registration on Kusama. Bumping the balance of Alice to 100K KSM. This applies to other networks as well.

### DIFF
--- a/packages/react-api/src/Api.tsx
+++ b/packages/react-api/src/Api.tsx
@@ -243,7 +243,7 @@ async function createApi (apiUrl: string, signer: ApiSigner, isLocalFork: boolea
       await setStorage(chopsticksFork, {
         System: {
           Account: [
-            [['5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY'], { data: { free: 1000 * 1e12 }, providers: 1 }]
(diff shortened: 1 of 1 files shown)


packages/react-api/src/Api.tsx
```diff
@@ -243,7 +243,7 @@ async function createApi (apiUrl: string, signer: ApiSigner, isLocalFork: boolea
       await setStorage(chopsticksFork, {
         System: {
           Account: [
-            [['5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY'], { data: { free: 1000 * 1e12 }, providers: 1 }]
+            [['5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY'], { data: { free: 100000 * 1e12 }, providers: 1 }]
```
TarikGul, line 246: This will not work because it will exceed the the allowed max integer: `Number needs to be an integer <= Number.MAX_SAFE_INTEGER, i.e. 2 ^ 53 - 1`. In this case best if you change the value to a string, i.e - '100000000000000000'

TarikGul (changes requested):

packages/react-api/src/Api.tsx
DrW3RK: Ah, okay! I guess 5000 is still within the bounds and should be good for almost all key things to test on Kusama. I guess this suffices for now.

TarikGul (approved):

PR #10680, opened 2024-06-18
Fix SubQuery endpoint for Kusama

deverka:


TarikGul (changes requested):
Linting is failing. Ordering of the keys needs to be adjusted.

deverka (commented):
Fixed

TarikGul (approved):

Issue #9963, opened 2023-10-19
UI becomes unresponsive for development chain


rajdeep7Singh:
* **I'm submitting a ...**

  - [x] Bug report
  - [ ] Feature request
  - [ ] Support request
  - [ ] Other

* **What is the current behavior and expected behavior?**

**Current Behavior**: UI becomes unresponsive. In my case happens when a referenda is about to leave decision period (seconds left), i have modified trackInfo locally. Attaching screenshots for the bug.

<img width="1509" alt="Screenshot 2023-10-19 at 11 32 59 PM" src="https://github.com/polkadot-js/apps/assets/42423438/0e3fe5d6-7a08-4536-bc03-610a19ffdca8">

<img width="756" alt="Screenshot 2023-10-19 at 11 33 24 PM" src="https://github.com/polkadot-js/apps/assets/42423438/f26735d4-8d93-4338-a67d-42a8c3badff5">

**expected Behavior**:  It should be responsive.
<!---

rzadp:
The same still happens - both Firefox and Chrome, also on most recent `polkadot-sdk` release built locally.

hemantwasthere:
Hey, I would like to work on this issue

TarikGul:
I need to confirm this by reproducing it locally, but @hemantwasthere If you are able to reproduce it, and have the bandwidth, all the help is appreciated :)

hemantwasthere:
@TarikGul Okay I'll try then

rzadp:
Hey @hemantwasthere, did you have any luck with this?

rzadp:
Hey, I poked around myself, consistently reproduced the issue, and proposed a fix: https://github.com/polkadot-js/apps/pull/10690

cc @TarikGul

PR #10690, opened 2024-06-20
Higher precision of referenda graphs

rzadp:
Closes https://github.com/polkadot-js/apps/issues/9963

Currently, the UI is not prepared to display graphs of referenda with short timeframes (which exist on development chains).

For example [here](https://github.com/polkadot-js/apps/blob/6b8e76edbd2024ee10671ed1a80300e9e8e8da8d/packages/page-referenda/src/util.ts#L230), if the decision period is short enough, the division will bring the `step` to `0` and no graph will work.

It causes the graphs to be botched, and even freezes the UI in an infinite loop in some cases (when the confirmation period overruns the usual decision period).

In the PR, I'm bringing the precision up by multiplying by `100` first, and dividing back by `100` after relevant calculations.

It fixes the graphs on local development chains.

I also compared side-by-side the changed UI with un-changed version, looking at current Polkadot referenda - all seems good to me.

### DIFF
--- a/packages/page-referenda/src/Referenda/Referendum.tsx
+++ b/packages/page-referenda/src/Referenda/Referendum.tsx
@@ -143,14 +143,14 @@ function getChartResult (totalEligible: BN, isConvictionVote: boolean, info: Pal
 
       // Bringing it to a higher precision.
       // Otherwise, graphs with short periods (on dev chains) are invalid.
-      const stepWithPrecision = x[x.length - 1].sub(x[0]).muln(100).divn(x.length)
+      const stepWithPrecision = x[x.length - 1].sub(x[0]).muln(100).divn(x.length);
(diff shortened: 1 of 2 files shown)


packages/page-referenda/src/Referenda/Referendum.tsx
```diff
@@ -141,13 +141,16 @@ function getChartResult (totalEligible: BN, isConvictionVote: boolean, info: Pal
         supx = (supn || supx !== -1) ? supx : i;
       }
 
-      const step = x[1].sub(x[0]);
```
rzadp, line 144: Can't really do that, because on short enough timeframes, `x[0] == x[1]` and `step` becomes zero.

TarikGul (approved):
Awesome work! Thank you so much :)

PR #10696, opened 2024-06-23
Add RegionX Cocos

Szegoo:
This PR is a request to add the RegionX test parachain called 'Cocos' to Polkadot.js.


Szegoo:
@TarikGul I updated the order

TarikGul (changes requested):
Looks like another ordering issues, It is kind of silly but it does 1 level at a time, so maybe run `yarn test` locally after fixing it or try to see where the it should be manually in the order.

Szegoo:
@TarikGul Checked with `yarn test` locally, should be good now.

TarikGul (approved):

PR #10744, opened 2024-07-10
Ledger support for PolkadotGenericApp

TarikGul:
# Part 1 (Current PR)

This PR revives ledger support in PJS apps. I kept the User interface exactly the same in order to not bring any confusion. Internally it is using the new `LedgerGeneric` wrapper from Common.

This is specific to the new ledger PolkadotGenericApp, where all addresses are kept under one slip44 in the derivation path...

This PR does not provide a migration App or support for the old Ledger App. To do any migrations here is a guide to [migrate old accounts on talisman](https://support.polkadot.network/support/solutions/articles/65000188075-talisman-migrate-to-the-new-ledger-universal-app).


TarikGul:
Currently working perfectly for polkadot. On Kusama I am getting `CLA not supported`...

PR #10747, opened 2024-07-11
Not show identify in account page for failed set identity txs.

kaichaosun:
Failed transactions, https://polkadot.subscan.io/extrinsic?module=identity&call=set_identity

UI after the fix:
image


TarikGul:
Thank you!

Issue #10313, opened 2024-03-01
Preimage tab has some preimages showing `Unable to decode preimage bytes into a valid Call`


TarikGul:
Currently in the Preimages tab under Governance some Preimages are coming back as `Unable to decode preimage bytes into a valid Call`, this might be an upstream issue coming from the `/polkadot-js/api` (most likely), but his requires some investigation.

![Screenshot 2024-02-29 at 8 52 22 PM](https://github.com/polkadot-js/apps/assets/47201679/d6eb79a4-cc5a-480c-9f38-2e04e07cc526)
![Screenshot 2024-02-29 at 8 53 24 PM](https://github.com/polkadot-js/apps/assets/47201679/db6e6748-3373-49e7-8577-d9351fb03838)

Apps version v0.133.2-105

ryanleecode:
taking this one @TarikGul

ggwpez:
Are you sure that it is a valid call?  
There is a case that when a pallet is removed from the runtime or someone notes a non-call, then it wont decode for sure.

KarimJedda:
Hitting the same problem today :) 

Both these extrinsics (Westend) decode in the extrinsics tab:

`0x630004000100a50f04082f000006030300286bee42420f00510249020428001049044a0fde15a111eda94aba678170fd212dbe70f26c3b38cd89d67ad327aacc171949054a0fde15a111eda94aba678170fd212dbe70f26c3b38cd89d67ad327aacc1719010049054a0fde15a111eda94aba678170fd212dbe70f26c3b38cd89d67ad327aacc1719020049054a0fde15a111eda94aba678170fd212dbe70f26c3b38cd89d67ad327aacc17190300`

`0xbd0204630004000100a50f04082f000006030300286bee42420f00510249020428001049044a0fde15a111eda94aba678170fd212dbe70f26c3b38cd89d67ad327aacc171949054a0fde15a111eda94aba678170fd212dbe70f26c3b38cd89d67ad327aacc1719010049054a0fde15a111eda94aba678170fd212dbe70f26c3b38cd89d67ad327aacc1719020049054a0fde15a111eda94aba678170fd212dbe70f26c3b38cd89d67ad327aacc17190300`

But the preimage of the second one doesn't decode in the PreImage view. 

![Screenshot from 2024-06-13 10-58-33](https://github.com/polkadot-js/apps/assets/5029053/4bbb35b4-60d8-4f89-8f87-c7590329e09b)

[...]

PR #10755, opened 2024-07-15
Try decoding a preimage as a transaction

rzadp:
- Closes https://github.com/polkadot-js/apps/issues/10313

First, this includes some refactoring.

A `registry` doesn't really make sense in the preimage type - replaced with `api` as a parameter to `createResult`, which gives access to `registry` needed in this function.

Next, replaced the way of returning the final result with pre-defined values with direct return of a result.

- It allows us to return earlier - reducing nesting.
- Works better with multiple attempts at decoding.
- we know exactly what's being returned at a given line.

Finally, added a new attempt at decoding - with `api.tx(callData.toString())`

It doesn't seem to help _very much_, but fixes at least two preimages that I've seen.

Besides the preimage [described in the issue](https://github.com/polkadot-js/apps/issues/10313#issuecomment-2165153954) (it doesn't exist anymore on Westend so I tested locally), it helps with one more preimage on Rococo.

### Before

local-westend

### After

local-westend-after

### Before

rococo

### After

rococo-after


TarikGul (approved):
Really amazing job!

PR #10757, opened 2024-07-16
Add muse-paseo

Moliholy:
This PR adds Muse testnet to Paseo.

### DIFF
--- a/packages/apps-config/src/endpoints/testingRelayPaseo.ts
+++ b/packages/apps-config/src/endpoints/testingRelayPaseo.ts
@@ -4,7 +4,7 @@
 import type { EndpointOption } from './types.js';
 
 import { chainsFrequencyPaseoSVG, chainsLaosOmegaPNG, chainsPaseoPNG, chainsPopNetworkSVG } from '@polkadot/apps-config/ui/logos/chains';
-import { nodesAjunaPNG, nodesAssetHubSVG, nodesBajunPNG, nodesBifrostSVG, nodesDarwiniaKoiSVG, nodesHyperbridgeSVG, nodesIntegriteeSVG, nodesKiltPNG, nodesNodleSVG } from '@polkadot/apps-config/ui/logos/nodes';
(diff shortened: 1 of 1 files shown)


TarikGul (approved):

ShankarWarang:
@Moliholy By any chance do you know the json rpc url and chain id of this newer Muse testnet? eth.chainId rpc call doesn't seem to exist for it.
Looking to add support on the [Universal Faucet](https://substratefaucet.xyz/).

Moliholy:
> eth.chainId rpc call doesn't seem to exist for it.

This is not an Ethereum-compatible blockchain and hence does not enable any Ethereum endpoint. It simply uses Ethereum format for addresses, but that's all.

Available data is all substrate-based:
- Endpoint: wss://paseo-muse-rpc.polkadot.io
- Parachain ID: 3369

ShankarWarang:
> This is not an Ethereum-compatible blockchain and hence does not enable any Ethereum endpoint. It simply uses Ethereum format for addresses, but that's all.
Are the ethereum format addresses directly mapped to ss58 corresponding substrate addresses in 29972 prefix format or is there any custom implementation? @Moliholy  

Looks like we might need a new chain type in polkadot-js @TarikGul ?
Something like 'substrateHex' or 'hybrid' instead of the 'ethereum' type to enable support for substrate extensions/wallets on such chains ?

Moliholy:
> Are the ethereum format addresses directly mapped to ss58 corresponding substrate addresses in 29972 prefix format or is there any custom implementation?

No addresses are directly derived from the public key using Ethereum method with keccak256, and the ss58 prefix is totally irrelevant.

> Looks like we might need a new chain type in polkadot-js @TarikGul ?
Something like 'substrateHex' or 'hybrid' instead of the 'ethereum' type to enable support for substrate extensions/wallets on such chains ?

There is indeed a very significant lack of support and tooling throughout the whole ecosystem for this use case, so any contribution in this regard would be greatly appreciated.

PR #10777, opened 2024-07-22
Add support for migration ledger app

TarikGul:
This PR works to give support for the Ledger `Migration App`. 

In the settings you can now toggle which app you would like to use. The options will be the `Generic` or the `Migration`.

TODO:

- [x] When signing with a ledger account - ensure there is a message saying which app the user is using.
- [x] Display the same as above but for when connecting an account as well.


IkerAlus (approved):
LGTM

bee344 (approved):

PR #10780, opened 2024-07-23
feat: enable uploading `rawMetadata` to the extension

bee344:
Companion to [#1408](https://github.com/polkadot-js/extension/pull/1408).

This enables Apps to upload the `rawMetadata` to the extension. The appearance is unchanged, the only difference is the button to upload the metadata is disabled until the `rawMetadata` is retrieved from the endpoint.

### DIFF
--- a/packages/page-settings/src/Metadata/Extensions.tsx
+++ b/packages/page-settings/src/Metadata/Extensions.tsx
@@ -2,6 +2,8 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import type { ChainInfo } from '../types.js';
+import type { HexString } from '@polkadot/util/types';
+import type { RawMetadataDef } from '@polkadot/extension-inject/types';
 
 import React, { useCallback, useMemo, useRef, useState } from 'react';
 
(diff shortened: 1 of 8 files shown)


packages/page-settings/src/Metadata/Extensions.tsx
```diff
@@ -17,13 +20,22 @@ import iconOption from './iconOption.js';
...
+function Extensions ({ chainInfo, className, rawMetadata }: Props): React.ReactElement<Props> {
+  const isMetadataReady = rawMetadata !== null;
+
   const { t } = useTranslation();
   const { extensions } = useExtensions();
   const [selectedIndex, setSelectedIndex] = useState(0);
-  const [isBusy, toggleBusy] = useToggle();
+  const [isBusy, toggleBusy] = useToggle(true);
+
+  useEffect((): void => {
```
TarikGul, line 34: Does this also affect users and dApps who don't care to use ledger accounts?

TarikGul: Ahh ignore this comment - I was thinking of the wrong thing.

packages/page-settings/src/Metadata/Extensions.tsx
```diff
@@ -37,14 +49,16 @@ function Extensions ({ chainInfo, className }: Props): React.ReactElement<Props>
       if (chainInfo && extensions?.[selectedIndex]) {
         toggleBusy();
 
+        const rawDef: MetadataDef = objectSpread<MetadataDef>({}, { ...chainInfo, rawMetadata: rawMetadata || '0x00' });
```
TarikGul, line 52: I think it can just be `rawMetadata` can just be `rawMetadata || undefined`

Since we now assert in the extension.

TarikGul (approved):
LGTM, one nit

PR #10808, opened 2024-08-06
Added OnFinality CommuneAI endpoint

mikespook:
Added OnFinality CommuneAI endpoint

`wss://commune.api.onfinality.io/public-ws`


TarikGul (changes requested):
Object keys need to be in correct order - linter is complaining

TarikGul (approved):

Issue #10823, opened 2024-08-12
Cannot import Astar address with ledger astar app and got CLA not support error


EV6999:
Tried to import Astar address on polkadot-js extension on chrome with Ledger Astar app and got CLA not support error.

TarikGul:
Yes, this is known, and I am actually in the process of getting this sorted in both the extension and apps. It should be ready and in soon.

PR #10826, opened 2024-08-13
Enable "Chain Specific App" for Ledger

TarikGul:
closes: https://github.com/polkadot-js/apps/issues/10823

This enables another setting called `Chain Specific App` which allows users to use Ledger apps that are not included in the Generic App.


packages/page-accounts/src/Accounts/Account.tsx
```diff
@@ -136,10 +141,16 @@ function createClearReferendaTx (api: ApiPromise, address: string, ids: [BN, BN]
...
+  const currApp = settings.get().ledgerApp;
   const ledger = getLedger();
 
-  await ledger.getAddress(ss58Prefix, true, meta.accountOffset || 0, meta.addressOffset || 0);
+  if (currApp === 'migration' || currApp === 'generic') {
+    await (ledger as LedgerGeneric).getAddress(ss58Prefix, true, meta.accountOffset || 0, meta.addressOffset || 0);
+  } else {
+    // This will always be the `chainSpecific` setting if the above condition is not met
+    await (ledger as Ledger).getAddress(true, meta.accountOffset || 0, meta.addressOffset || 0);
+  }
```
bee344, line 153: Why is `confirm = true` in these two cases while on other uses of `getAddress` is set as `false`?

bee344 (approved):
Just one question but the rest works well

packages/page-accounts/src/Accounts/Account.tsx
TarikGul: It's just what was originally there: https://github.com/polkadot-js/apps/commit/afa3efb43fbabaaf3e251b5d922a9500df672d33#diff-5b91fe9ddf75c614a7078acb7623af5dce21fe6f8fe35af6b8e4c381e0752be0L142

PR #10827, opened 2024-08-13
feat: enable metadata update to the extension on initialization

bee344:
Now the metadata gets injected into the extensions when the app is initialized, in order to streamline the process. The extensions still need to approve the metadata update request, but the step of going to `settings` and selecting the extension to then update the metadata has been removed, along with that component.


packages/page-settings/src/index.tsx
```diff
@@ -30,7 +29,6 @@ function SettingsApp ({ basePath, onStatusChange }: Props): React.ReactElement<P
       text: t('General')
     },
     {
-      count: numExtensions,
```
TarikGul, line 33: Why do we remove this?

bee344: It was removed when the popup was added, has been reinstated now

TarikGul:
Closing.

Issue #9988, opened 2023-10-27
Transferable balance is incorrect


michalisFr:
* **I'm submitting a ...**

  - [X] Bug report
  - [ ] Feature request
  - [ ] Support request
  - [ ] Other

* **What is the current behavior and expected behavior?**

Transferable balance seems to be calculated incorrectly. After the merging of [this PR](https://github.com/paritytech/substrate/pull/12951), `reducible` is calculated as indicated [here](https://github.com/paritytech/polkadot-sdk/blob/3069b0af398e5a0374802c31201637e97f4a530a/substrate/frame/balances/src/impl_fungible.rs#L44).

It seems transferable is calculated with the deprecated:
`transferable = free - reserved - max(feeFrozen, miscFrozen)`

* **What is the motivation for changing the behavior?**

Display transferable balance properly. 

Perhaps `free` balance should also be displayed because it can be used for holds (aka reserves) even if transferable is insufficient.

* **Please tell us about your environment:**

  - Version: apps v0.132.2-134
  - Environment:

    - [ ] Node.js
    - [X] Browser
    - [ ] Other (limited support for other environments)

  - Language:

    - [ ] JavaScript
    - [ ] TypeScript (include tsc --version)
    - [ ] Other

Albertpolkadot:
Hello @IkerAlus & @TarikGul ,

Today, I encountered a critical issue while attempting to submit a proposal. After creating a pre-image, I noticed that my transferrable balance was showing as 0 DOT.

In an attempt to rectify this, I initiated a transaction of 20 DOT from another wallet to my identified wallet. However, even after completing the transaction, my transferrable balance remains unchanged. Subscan indicates that the transaction was successfully processed, as seen [here](https://polkadot.subscan.io/extrinsic/20409079-2).

![Captura de pantalla 2024-04-19 122056](https://github.com/polkadot-js/apps/assets/144892919/d5fd60f1-a897-4f99-b3fe-4528df407a40)

This issue is particularly urgent as I urgently need to submit two proposals on Polkassembly and create and sign child bounties within the next 10 days. Unfortunately, my account is currently frozen, preventing me from proceeding with these crucial tasks.

Thanks!

Albertpolkadot:
Hi @IkerAlus,

Yes, this is the problem, I sent 20 Dots to my wallet, but I can't see these 20 DOTS that they should be free...

[...]

PR #10842, opened 2024-08-18
Fix transferable balance

TarikGul:
closes: https://github.com/polkadot-js/apps/issues/9988

Before this can go in - https://github.com/polkadot-js/api/pull/5955/files must be merged and released in the api, and bumped inside of apps.

This also has backwards compatibility for chains that don't have support for the `FrameSystemAccountData` type be fallbacking on the old `availableBalance` value.


TarikGul:
That Accounts page errors with `Cannot read properties of undefined (reading 'negative')`

PR #10850, opened 2024-08-21
Update Kilt type definitions

rflechtner:
Bumps `@kiltprotocol/type-definitions` to ^v1.


packages/apps-config/package.json
```diff
@@ -28,7 +28,7 @@
     "@fragnova/api-augment": "0.1.0-spec-1.0.4-mainnet",
     "@frequency-chain/api-augment": "1.11.1",
     "@interlay/interbtc-types": "1.13.0",
-    "@kiltprotocol/type-definitions": "0.35.1",
+    "@kiltprotocol/type-definitions": "^1.11200.0",
```
TarikGul, line 31: Out of curiosity what is with the versioning here?

rflechtner: The package version scheme has so far been disconnected from the runtime versioning scheme, and we've seen the major version bump (which is mainly motivated by a move from chain-name runtime matching to spec-name runtime matching) as an opportunity to change that. The minor version now represents the runtime version at which an update to the type definitions had to be introduced.

E.g., when we upgrade spiritnet to runtime version `11400` (which should happen soonish) we'll introduce version `1.11400.0`.

We've chosen minor version bumps as this is a non-breaking, feature-update style upgrade. It'll support the new runtime, but nothing else will change.

TarikGul: Okay that makes sense, thank you for the explanation.

packages/apps-config/src/api/spec/index.ts
```diff
@@ -221,6 +222,7 @@ const spec: Record<string, OverrideBundleDefinition> = {
   'jur-node': jur,
   kerria: parallel,
   khala,
+  ...kilt,
```
TarikGul, line 225: I think this might be a mistake since we want the key to be `kilt`, unless the object passed from `kilt.js` is `{ kilt: {} }`

TarikGul (changes requested):
Small nit and question

packages/apps-config/src/api/spec/index.ts
rflechtner: I've based this on acala's approach. The kilt object contains a key per runtime - we've got spiritnet, the production network, and several test networks with separate runtime/spec names. I tested this locally and it seems to work as expected.

TarikGul: Ahh okay thanks for the explanation, everything looks good then!

TarikGul (approved):

Issue #10837, opened 2024-08-15
App does not allow voting with reserved balance


Ank4n:
With the changes in pool to allow open gov participation, the staked balance will migrate from locked (frozen) to reserved (held). The conviction-voting pallet allows [total balance](https://github.com/paritytech/polkadot-sdk/blob/master/substrate/frame/conviction-voting/src/lib.rs#L397) including reserved balance to vote.

But the UI does not allow to vote with reserved balance.
![image](https://github.com/user-attachments/assets/02b8ee8d-7f78-4e1f-afa5-5aa4dff1eab5)

![image](https://github.com/user-attachments/assets/a2ce8376-82fb-4003-9a56-1c35ea82759b)

The above scenario works if I call the extrinsic directly.
![image](https://github.com/user-attachments/assets/37163e57-9edb-49bf-ac9f-2d5331e8cac0)
https://westend.subscan.io/extrinsic/22143408-2

This bug is probably due to the old democracy pallet that only allowed locked (frozen) funds for vote.

TarikGul:
@Ank4n Is there anything specific in the runtime that would say this logic is valid? Like a call, storage entry, etc.?

Basically I want to do a check like: 'If this is available on chain then voting balance should include reserves'

Ank4n:
> @Ank4n Is there anything specific in the runtime that would say this logic is valid? Like a call, storage entry, etc.?
> 
> Basically I want to do a check like: 'If this is available on chain then voting balance should include reserves'

@TarikGul If `pallet-conviction-voting` is used for voting, you can use reserved balance.

TarikGul:
@Ank4n So the pallet just needs to be active in the runtime?

Ank4n:
> @Ank4n So the pallet just needs to be active in the runtime?

Yeah the above pallet supports reserved balance since [this commit](https://github.com/paritytech/polkadot-sdk/commit/5d81f23f8f8734c5d5fbb845e06ef972e952c339) which was well over a year ago. I guess it is okay to assume if this pallet is used, that it supports reserved balance? (not sure if any chain is using the soft deprecated `pallet-democracy` which does not allow reserved balance for voting).

Ank4n:
How does the voting UI work? I am guessing it is hardcoded to support only these two pallets?

PR #10858, opened 2024-08-22
Allow reserved in voting balance

TarikGul:
closes: https://github.com/polkadot-js/apps/issues/10837

TODO:

- [x] Ensure there is a check I can do that shows this is allowed on chain. Like some kind of call, or query that would only be available if this logic is available in the runtime. This is to avoid breaking all other chains in the UI.

### DIFF
--- a/packages/react-components/src/VoteValue.tsx
+++ b/packages/react-components/src/VoteValue.tsx
@@ -2,8 +2,8 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import type { DeriveBalancesAll } from '@polkadot/api-derive/types';
-import type { BN } from '@polkadot/util';
 
+import BN from 'bn.js';
 import React, { useCallback, useEffect, useState } from 'react';
 
 import { useApi, useCall } from '@polkadot/react-hooks';
@@ -17,6 +17,7 @@ interface Props {
   accountId?: string | null;
   autoFocus?: boolean;
   isCouncil?: boolean;
+  isReferenda?: boolean;
   label?: string;
   noDefault?: boolean;
   onChange: (value: BN) => void;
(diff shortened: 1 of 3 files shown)


packages/react-components/src/VoteValue.tsx
```diff
@@ -55,7 +57,7 @@ function getValues (selectedId: string | null | undefined, noDefault: boolean |
     })
     .map(({ amount }) => amount);
 
-  const maxValue = allBalances.votingBalance;
+  const maxValue = isReferenda && api.query.convictionVoting ? allBalances.votingBalance.add(allBalances ? allBalances.reservedBalance : new BN(0)) : allBalances?.votingBalance;
```
bee344, line 60:
```suggestion
  const maxValue = isReferenda && api.query.convictionVoting ? allBalances.votingBalance.add(allBalances.reservedBalance) : allBalances?.votingBalance;
```
Why are we checking for `allBalances` inside `allBalances.votingBalance.add`?

packages/react-query/src/BalanceVoting.tsx
```diff
@@ -13,19 +14,20 @@ import FormatBalance from './FormatBalance.js';
...
+function BalanceVoting ({ children, className = '', isReferenda, label, params }: Props): React.ReactElement<Props> {
   const { api } = useApi();
   const allBalances = useCall<DeriveBalancesAll>(api.derive.balances?.all, [params]);
 
   return (
     <FormatBalance
       className={className}
       label={label}
-      value={allBalances?.votingBalance}
+      value={isReferenda && api.query.convictionVoting ? allBalances?.votingBalance.add(allBalances ? allBalances.reservedBalance : new BN(0)) : allBalances?.votingBalance}
```
bee344, line 30:
```suggestion
      value={isReferenda && api.query.convictionVoting ? allBalances?.votingBalance.add(allBalances.reservedBalance) : allBalances.votingBalance}
```
Simmilar to above

packages/react-components/src/VoteValue.tsx
TarikGul: Since the useCall returns `<T> | undefined`

packages/react-query/src/BalanceVoting.tsx
TarikGul: Since the useCall returns `<T> | undefined`

packages/react-components/src/VoteValue.tsx
TarikGul: This one actually doesnt need it because allBalances is not going to be undefined since its checked in the component. Fixed this one.

PR #10862, opened 2024-08-23
fix(bug): user can input invalid lifetime for tx

### LINKED ISSUE #10737: GUI allows to specify invalid era period
* **I'm submitting a ...**

  - [X] Bug report
  - [ ] Feature request
  - [ ] Support request
  - [ ] Other

* **What is the current behavior and expected behavior?**

Currently the GUI allows to specify any arbitrary number for the era when signing an extrinsic. However, the maximum period is defined in the [runtime](https://github.com/paritytech/polkadot-sdk/blob/master/substrate/frame/system/src/lib.rs#L544-L547), and it is usually 4096 blocks.

Screenshot 2024-07-08 at 16 16 18

This can lead users to sign extrinsics with an era that will yield an `InValidSignature` error, which is not intuitive. At least a warning should be shown to notify the user of this possible behaviour.

bee344:
Closes #10737

I'm currently split between limitting the lifetime to c`onst.system.blockHashCount`, so that if they input a larger value, it gets converted to that max value automatically and displaying a warning or something like that. I'm leaning towards the first case since screen realestate is limited.

### DIFF
--- a/packages/react-components/src/InputNumber.tsx
+++ b/packages/react-components/src/InputNumber.tsx
@@ -24,6 +24,7 @@ interface Props {
   isDisabled?: boolean;
   isError?: boolean;
   isFull?: boolean;
+  isLifetime?: boolean;
   isLoading?: boolean;
   isSi?: boolean;
   isDecimal?: boolean;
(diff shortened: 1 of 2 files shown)


packages/react-components/src/InputNumber.tsx
```diff
@@ -163,23 +191,27 @@ function InputNumber ({ autoFocus, bitLength = DEFAULT_BITLENGTH, children, clas
       : null
   );
   const [[value, valueBn, isValid], setValues] = useState<[string, BN, boolean]>(() =>
-    getValues(api, propsValue || defaultValue, si, bitLength, isSigned, isZeroable, maxValue, siDecimals)
+    getValues(api, propsValue || defaultValue, si, bitLength, isLifetime || false, isSigned, isZeroable, maxValue, siDecimals)
```
TarikGul, line 194:
```suggestion
    getValues(api, propsValue || defaultValue, si, bitLength, !!isLifetime, isSigned, isZeroable, maxValue, siDecimals)
```

packages/react-components/src/InputNumber.tsx
```diff
@@ -163,23 +191,27 @@ function InputNumber ({ autoFocus, bitLength = DEFAULT_BITLENGTH, children, clas
...
   }, [isValid, onChange, valueBn]);
 
   const _onChangeWithSi = useCallback(
-    (input: string, si: SiDef | null) => setValues(
-      getValuesFromString(api, input, si, bitLength, isSigned, isZeroable, maxValue, siDecimals)
-    ),
-    [api, bitLength, isSigned, isZeroable, maxValue, siDecimals]
+    (input: string, si: SiDef | null) =>
+      setValues(
+        getValuesFromString(api, input, si, bitLength, isLifetime || false, isSigned, isZeroable, maxValue, siDecimals)
```
TarikGul, line 206:
```suggestion
        getValuesFromString(api, input, si, bitLength, !!isLifetime, isSigned, isZeroable, maxValue, siDecimals)
```

TarikGul (approved):

TarikGul:
Great job lgtm!

PR #10923, opened 2024-09-17
Fix: Estimated Bulk Core Time price

piggydoughnut:
Showing `endPrice` as Estimated Core Time price in the Overview UI tab on Coretime chain.

`endPrice` 
-  a price at which Coretime starts selling during the leadin period after the interlude period.
- pallet_broker::pallet::SalesInfo - [sdk reference](https://paritytech.github.io/polkadot-sdk/master/pallet_broker/type.SaleInfoRecordOf.html#)

Screenshot 2024-09-17 at 17 44 06
[Image Source](https://wiki.polkadot.network/docs/learn-agile-coretime)


packages/react-hooks/src/utils/dataProcessing.ts
```diff
@@ -23,3 +25,17 @@ export function processHexMask (mask: PalletBrokerScheduleItem['mask'] | undefin
 
   return buffArr;
 }
+
+export function stringToBN (value: string | undefined): BN {
```
TarikGul, line 29: We could totally add this utility function to `@polkadot/util` in the future. 🚀

TarikGul (approved):
Looks great to me, nice improvement!

PR #10926, opened 2024-09-18
Fix: updated parsing for brokerConfig and brokerSalesInfo

piggydoughnut:


packages/react-hooks/src/useBrokerConfig.ts
```diff
@@ -11,17 +11,17 @@ import { createNamedHook, useApi, useCall } from '@polkadot/react-hooks';
 import { stringToBN } from './utils/dataProcessing.js';
 
 function parseConfig (config: PalletBrokerConfigRecord): SimplifiedPalletBrokerConfigRecord {
-  const c = config.toJSON();
+  const c = config?.toJSON();
```
TarikGul, line 14: This should actually be like this:

```typescript
function parseConfig (config: PalletBrokerConfigRecord): SimplifiedPalletBrokerConfigRecord {
  return {
    advanceNotice: config.advanceNotice.toNumber(),
    contributionTimeout: config.contributionTimeout.toNumber(),
    idealBulkProportion: stringToBN(config.idealBulkProportion?.toString()),
    interludeLength: config.interludeLength.toNumber(),
    leadinLength: config.leadinLength.toNumber(),
    limitCoresOffered: config.limitCoresOffered.isSome ? config.limitCoresOffered.unwrap().toNumber() : 0,
    regionLength: config.regionLength.toNumber(),
    renewalBump: stringToBN(config.renewalBump?.toString())
  };
}
```

TarikGul: And if config can be undefined, then the param type needs to be updated.

packages/react-hooks/src/useBrokerSalesInfo.ts
```diff
@@ -10,22 +10,20 @@ import { createNamedHook, useApi, useCall } from '@polkadot/react-hooks';
 
 import { stringToBN } from './utils/dataProcessing.js';
 
-function parsePalletBrokerSaleInfoRecord (
-  record: PalletBrokerSaleInfoRecord
-): SimplifiedPalletBrokerSaleInfoRecord {
-  const rec = record.toJSON();
+function parsePalletBrokerSaleInfoRecord (record: PalletBrokerSaleInfoRecord): SimplifiedPalletBrokerSaleInfoRecord {
```
TarikGul, line 13: Same with this as above.

TarikGul (changes requested):
It's always better to use the built in type structures, instead of calling `toJSON` in order to get all the values.

packages/react-hooks/src/useBrokerSalesInfo.ts
```diff
@@ -0,0 +1,43 @@
...
+    coresOffered: record.coresOffered.toNumber(),
+    coresSold: record.coresSold.toNumber(),
+    endPrice: record.endPrice,
+    firstCore: record.firstCore.toNumber(),
+    idealCoresSold: record.idealCoresSold.toNumber(),
+    leadinLength: record.leadinLength.toNumber(),
+    regionBegin: record.regionBegin.toNumber(),
+    regionEnd: record.regionEnd.toNumber(),
+    saleStart: record.saleStart.toNumber(),
+    selloutPrice: record.selloutPrice.isSome ? record.selloutPrice.unwrap() : 0
```
TarikGul, line 22: Small nit, but I think you can set the `0` to `new BN(0)` if you want the values to be inline. That being said its a small nit.

TarikGul (approved):

packages/react-hooks/src/useBrokerSalesInfo.ts
piggydoughnut: good nit 🤩

PR #10933, opened 2024-09-19
Fix: UI Coretime type fix

piggydoughnut:


TarikGul (approved):
Nice, catch yea I confirmed as well, the result that is returned from the call is indeed wrapped in an option.

packages/react-hooks/src/useBrokerConfig.ts
```diff
@@ -1,37 +1,40 @@
...
 import type { PalletBrokerConfigRecord } from '@polkadot/types/lookup';
 import type { PalletBrokerConfigRecord as SimplifiedPalletBrokerConfigRecord } from './types.js';
 
 import { useEffect, useState } from 'react';
 
 import { createNamedHook, useApi, useCall } from '@polkadot/react-hooks';
 
-function parseConfig (config: PalletBrokerConfigRecord): SimplifiedPalletBrokerConfigRecord {
+function extractInfo (config: Option<PalletBrokerConfigRecord>): SimplifiedPalletBrokerConfigRecord {
+  const c = config.unwrap();
```
TarikGul, line 13: I think there should be a check here though to ensure that `c.isSome`

packages/react-hooks/src/useBrokerSalesInfo.ts
```diff
@@ -9,32 +10,34 @@ import { useEffect, useState } from 'react';
 import { createNamedHook, useApi, useCall } from '@polkadot/react-hooks';
 import { BN } from '@polkadot/util';
 
-function parsePalletBrokerSaleInfoRecord (record: PalletBrokerSaleInfoRecord): SimplifiedPalletBrokerSaleInfoRecord {
+function extractInfo (record: Option<PalletBrokerSaleInfoRecord>): SimplifiedPalletBrokerSaleInfoRecord {
+  const v = record.unwrap();
```
TarikGul, line 14: I think there should be a check here though to ensure that `record.isSome`

PR #10947, opened 2024-09-24
Coretime Broker on Coretime Chain - Update after Coretime Polkadot release

piggydoughnut:
Coretime has been released to the Polkadot Relay chain, but since the data was relying on the `Workload` pallet call, there is not data shown in the `Network` -> `Coretime broker(Experimental)` Tab currently.

In the initial sales period (sale 0), the Coretime chain isn't actively driving the relay chain; it's essentially in the same state as it was prior to the transition. During the bootstrapping phase, things operate a bit differently, but after the 28-day period, the workload will align with the current state of the relay chain moving forward.

This PR changes data organization, and takes the `coreCount` from `broker.status` call as the truth to how many cores are there, instead of relying on `Workload` to provide information about the cores.

Other changes:
- added `cores sold/offered` to the summary above the table
- fixed the tasks/pools/idles graph which had pools and idles swapped
- improved calculation of start and end dates for Bulk Coretime tasks
- double checked the calculations for time estimations and all seems to be correct, I added information on calculations in the code and also in my notes.

[My Coretime notes](https://hackmd.io/2Jnps35mRCOepqCpCfGQ5w)


packages/page-broker/src/Overview/Filters.tsx
```diff
@@ -16,31 +17,31 @@ const StyledDiv = styled.div`
...
-  onFilter: (data: CoreWorkloadType[]) => void
+  data: CoreInfo[];
+  onFilter: (data: CoreInfo[]) => void
 }
 
-const filterLoad = (parachainId: string, load: CoreWorkloadType[] | CoreWorkplanType[], workloadCoreSelected: number): CoreWorkloadType[] | CoreWorkplanType[] => {
+const filterLoad = (parachainId: string, data: CoreInfo[], workloadCoreSelected: number): CoreInfo[] => {
   if (parachainId) {
-    return load?.filter(({ info }) => info.task === parachainId);
+    return data.filter(({ workload }) => !!workload?.filter(({ info }) => info.task === parachainId).length);
```
TarikGul, line 26: Wouldn't `!!workload` here just return a boolean?

packages/page-broker/src/Overview/Filters.tsx
```diff
@@ -16,31 +17,31 @@ const StyledDiv = styled.div`
...
-function Filters ({ onFilter, workLoad }: Props): React.ReactElement<Props> {
+function Filters ({ data, onFilter }: Props): React.ReactElement<Props> {
   const [workloadCoreSelected, setWorkloadCoreSelected] = useState(-1);
   const [_parachainId, setParachainId] = useState<string>('');
 
   const coreArr: number[] = useMemo(() =>
-    workLoad?.length
-      ? Array.from({ length: workLoad?.length || 0 }, (_, index) => index)
+    data?.length
+      ? Array.from({ length: data?.length || 0 }, (_, index) => index)
```
TarikGul, line 42:
```suggestion
      ? Array.from({ length: data.length || 0 }, (_, index) => index)
```
since you are already checking the ternary with `data?.length`

packages/page-broker/src/Overview/Workload.tsx
```diff
@@ -3,76 +3,105 @@
...
+
+  const status = useCall<Option<PalletBrokerStatusRecord>>(isApiReady && api.query.broker?.status);
   const [isExpanded, toggleIsExpanded] = useToggle(false);
-  const [tableData, setTableData] = useState<InfoRow>();
-  const [currentRegion, setCurrentRegion] = useState<RegionInfo | undefined>();
+  const [workloadData, setWorkloadData] = useState<InfoRow[]>();
+  const [workplanData, setWorkplanData] = useState<InfoRow[]>();
+
+  const currentTimeSlice = useMemo(() => {
+    if (!!status && !!status?.toHuman()) {
```
TarikGul, line 36:
```suggestion
    if (status?.isSome && status.unwrap().toHuman()) {
```

packages/page-broker/src/Overview/Workload.tsx
```diff
@@ -3,76 +3,105 @@
...
+    return 0;
+  }, [status]);
+
   const regionInfo = useRegions(api);
+  const region: RegionInfo | undefined = useMemo(() => regionInfo?.find((v) => v.core === core && v.start <= currentTimeSlice && v.end > currentTimeSlice), [regionInfo, core, currentTimeSlice]);
 
   useEffect(() => {
-    if (info) {
-      const region: RegionInfo | undefined = regionInfo?.find((v) => v.core === core && v.start <= timeslice && v.end > timeslice);
+    if (!!workload?.length && !!salesInfo) {
```
TarikGul, line 47:
```suggestion
    if (workload?.length && salesInfo) {
```

I think that would work just fine. Im guessing here you just want to ensure that if length is greater than 0?

packages/page-broker/src/Overview/Workload.tsx
```diff
@@ -3,76 +3,105 @@
...
+      setWorkloadData(formatRowInfo(workload, core, region, currentTimeSlice, salesInfo));
+    } else {
+      return setWorkloadData([{ core }]);
+    }
+  }, [workload, region, currentTimeSlice, core, salesInfo]);
 
-      setTableData(formatRowInfo(info, core, region, timeslice, type, lastBlock));
-      setCurrentRegion(region);
+  useEffect(() => {
+    if (!!workplan?.length && !!salesInfo) {
```
TarikGul, line 55: Same as the above.

packages/page-broker/src/Overview/Workload.tsx
```diff
@@ -3,76 +3,105 @@
...
+  }
+  , [workplan, region, currentTimeSlice, core, salesInfo]);
 
-  const hasWorkplan = !!workplan?.length;
+  const hasWorkplan = workplan?.length;
 
   return (
     <>
-      {tableData &&
+      {!!workloadData &&
```
TarikGul, line 65: This can just be `workloadData`. Also if this value has the possibility of being false, it should probably be noted in the types in its `useState`.

packages/page-broker/src/Overview/Workload.tsx
```diff
@@ -3,76 +3,105 @@
...
-              (
-                <ExpandButton
-                  expanded={isExpanded}
-                  onClick={toggleIsExpanded}
-                />
-              )
-            }
-            {!hasWorkplan && 'none'}
-          </td>
+          {workloadData?.map((one) => {
```
TarikGul, line 70:
```suggestion
          {workloadData.map((one) => {
```

packages/page-broker/src/Overview/index.tsx
```diff
@@ -1,102 +1,98 @@
...
-  useEffect(() =>
-    setWorkPlan(formatLoad(workplanInfos, leaseMap, reservationMap))
-  , [workplanInfos, leaseMap, reservationMap]);
+  const leaseMap: LeaseMapType = useMemo(() => leases ? createTaskMap(leases) : [], [leases]);
+  const reservationMap: ReservationMapType = useMemo(() => reservations ? createTaskMap(reservations) : [], [reservations]);
 
   useEffect(() => {
-    setWorkLoad(formatLoad(workloadInfos, leaseMap, reservationMap));
-  }, [workloadInfos, leaseMap, reservationMap]);
+    !!workplanInfos && !!workloadInfos && !!coreCount && !!leaseMap && !!reservationMap &&
```
TarikGul, line 70: I dont think we need the double bangs on any of these values?

packages/page-broker/src/Overview/index.tsx
```diff
@@ -1,102 +1,98 @@
...
   return (
     <div className={className}>
       <Summary
         apiEndpoint={apiEndpoint}
-        workloadInfos={workLoad}
+        coreCount={coreCount}
+        workloadInfos={workloadInfos}
       ></Summary>
-      {!!workPlan?.length &&
+      {!!data?.length &&
```
TarikGul, line 81:
```suggestion
      {data?.length &&
```

TarikGul (changes requested):
Overally super nice! Great job.

Just some nits, and I think generally we don't all the double bangs, but if you prefer them for readability (explicit booleans for checks) then I am okay with it.

packages/page-broker/src/Overview/Filters.tsx
piggydoughnut: Yes, that is exactly what I am going for :) Coz I need to see if the is  a workload with parachainId that I am looking for :) Then I can filter out objects of CoreInfo 

```
export interface CoreInfo {
  core: number,
  workload: CoreWorkloadType[] | undefined,
  workplan: CoreWorkplanType[] | undefined
}
```

packages/page-broker/src/Overview/index.tsx
piggydoughnut: Typescript begs to differ (for some of them) 😊 I might have thrown in some extras just in case 😅

piggydoughnut:
I love my double bangs 😆 They leave no other room for other interpretation 😊

PR #10953, opened 2024-09-25
chore: bump KILT dep

ntn-x2:


ntn-x2:
CC @TarikGul. Should be rather quick to merge and deploy. We have a new feature which relies on these runtime APIs, thanks!

TarikGul:
Just a heads up the API now decorates runtime api calls off of the metadata when possible (if the chain supports v15). It was just added to the UI last night.

TarikGul (approved):

PR #10959, opened 2024-09-27
Reflecting changes to Identity pallet 

piggydoughnut:
Identity object used to hold `discord` information under `others`, now it is part of the object on a higher level.

Reference for the changes after mirgration

https://github.com/paritytech/polkadot-sdk/pull/2931/files

`cumulus/scripts/migrate_storage_to_genesis/identityMigration.js`

```
// Migrate `IdentityInfo` data to the new format:
// - remove `additional` field
// - add `discord` field
// - add `github` field`
```

PR #10962, opened 2024-09-27
Renamed riot to matrix while displaying identity

niklabh:
As riot is renamed to matrix changing this while displaying.


piggydoughnut:
Hey 👋🏽  Thanks for this PR, we try to keep the old fields around for backward compatibility in case someone has not updated to the latest version of the pallet.
Changes you require will be up and running soon :) https://github.com/polkadot-js/apps/pull/10959

Issue #10954, opened 2024-09-25
Error on connection to `LAOS` networks


asiniscalchi:
all `laos` network are experiencing the following error on connection:
```
Uncaught error. Something went wrong with the query and rendering of this component. Please supply all the details below when logging an issue, it may help in tracing the cause.

createType(AccountId):: Invalid AccountId provided, expected 32 bytes, found 20

Error: createType(AccountId):: Invalid AccountId provided, expected 32 bytes, found 20
at o (https://polkadot.js.org/apps/modu.6724c4c4ec0fc7a7.js:1:87040)
at J.createType (https://polkadot.js.org/apps/modu.d704469d4359a33d.js:2:48759)
at Array.<anonymous> (https://polkadot.js.org/apps/comm.bf3973086a259e8c.js:693:1230)
at D (https://polkadot.js.org/apps/comm.bf3973086a259e8c.js:693:1820)
at P (https://polkadot.js.org/apps/comm.bf3973086a259e8c.js:693:2176)
at https://polkadot.js.org/apps/comm.bf3973086a259e8c.js:743:195
at Object.zi [as useState] (https://polkadot.js.org/apps/modu.bd8f3266470441f0.js:2:80363)
at https://polkadot.js.org/apps/modu.bd8f3266470441f0.js:2:223782
at https://polkadot.js.org/apps/comm.bf3973086a259e8c.js:743:189
at _i (https://polkadot.js.org/apps/modu.bd8f3266470441f0.js:2:77446)
```

https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Flaos-rpc.dwellir.com#/accounts
![image](https://github.com/user-attachments/assets/5250ca7b-9605-4e00-a763-4abd298bc862)

https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Frpc.laossigma.laosfoundation.io#/accounts
![image](https://github.com/user-attachments/assets/411f23b9-2223-4855-be59-4318c2b560ec)

asiniscalchi:
it seems it's the **api** the problem:
I run locally polkadot.js and it works with the following:
![image](https://github.com/user-attachments/assets/113d5072-20ba-460b-b77d-652687e63b30)

but the one published is not working with:
![image](https://github.com/user-attachments/assets/1cc47333-f36a-48d7-85db-3ed8e51f8aef)

---
So it seems 
v10.12.6 working.
**v13.2.1 not working**.

asiniscalchi:
[...]

PR #10967, opened 2024-09-30
Patch for crashing explorer

piggydoughnut:
Helps with https://github.com/polkadot-js/apps/issues/10954 but does not solve not parsing the `AccountId`


Imod7 (approved):
Tested it and it works. It loads the Laos apps page correctly and logs the error in the console.

TarikGul (approved):
Looks good for now, looks like it just needs AccountId20 as the type, but it will need a special case.

PR #10969, opened 2024-10-01
H160 address parsing issues on Address book and AccountName component

piggydoughnut:
Choosing to use `AccountId20` type if the address is of 20 bytes 😊 

This should solve issues with parsing H160 addresses for Laos Network https://github.com/polkadot-js/apps/issues/10954

### DIFF
--- a/packages/react-components/src/AccountName.tsx
+++ b/packages/react-components/src/AccountName.tsx
@@ -10,7 +10,8 @@ import React, { useCallback, useContext, useEffect, useState } from 'react';
 import { statics } from '@polkadot/react-api/statics';
 import { useApi, useDeriveAccountInfo } from '@polkadot/react-hooks';
 import { AccountSidebarCtx } from '@polkadot/react-hooks/ctx/AccountSidebar';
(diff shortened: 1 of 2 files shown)


packages/react-components/src/AccountName.tsx
```diff
@@ -48,9 +49,9 @@ function createNumMatcher (prefix: string, name: string, add?: string): AddrMatc
 
   return (addr: unknown): string | null => {
     try {
-      const u8a = isCodec(addr)
-        ? addr.toU8a()
-        : statics.registry.createType('AccountId', addr as string).toU8a();
+      const decoded = isU8a(addr) ? addr : isCodec(addr) ? addr.toU8a() : decodeAddress(addr?.toString() || '');
+      const type = decoded.length === 20 ? 'AccountId20' : 'AccountId';
```
TarikGul, line 53: I wonder if using `encodedLength` here makes sense?

TarikGul (approved):
Looks good to me :)

packages/react-components/src/AccountName.tsx
TarikGul: This only makes sense for Codec classes, resolving!

PR #11006, opened 2024-10-14
Remove Rococo relay and parachains

PierreBesson:
Rococo Shutdown is shutting down today October 14th 2024, see https://forum.polkadot.network/t/rococo-to-be-deprecated-in-october/8702/9


TarikGul:
Thanks, LGTM if there are any additional remnants I will remove them in a follow up PR, but overall this looks great!

Issue #11028, opened 2024-10-22
Coretime: Relay chain Tab Bugs 


piggydoughnut:
Changes for the recently deployed Coretime UI Tab on Relay chain:

- [x] The chain should have its status as  "renewed" if we found its future workplan on the same core and it is of type Bulk Coretime
- [x] "Eligible" for renewal should only be shown during the renewal phase
- [x] Show correct end date if the chain did not extend their core lease
- [x] Some chains can be on multiple cores
- [x] Kusama does not have system parachains and their reservations displayed
- [x] better mobile view

PR #11043, opened 2024-10-28
Relay Chain Coretime UI bugs

piggydoughnut:
Addressing some of the issues in #11028

- Fixed end dates for the chains that did not renew during the renewal period (interlude).
- Showing "renewed" only if there is workplan for the same chain on the same core.
- Taking taskIds from `api.query.coretimeAssignmentProvider.coreDescriptors` instead of `api.query.paras.parachains`
- Showing renewable eligibility and price of renewal only if the current phase is within the interlude period of the current cycle/sale.
- Added expandable rows to the table in order to show if chain owns multiple cores.

Screenshot 2024-10-29 at 14 43 10

### DIFF
--- a/packages/page-broker/src/utils.ts
+++ b/packages/page-broker/src/utils.ts
@@ -1,17 +1,12 @@
 // Copyright 2017-2024 @polkadot/app-broker authors & contributors
 // SPDX-License-Identifier: Apache-2.0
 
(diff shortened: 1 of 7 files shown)


TarikGul (approved):

TarikGul:
Amazing job!

Issue #11046, opened 2024-10-29
Treasury page does not check treasury pallet


sesh92:
https://github.com/polkadot-js/apps/issues/11031 

> @TarikGul
> 
> This change breaks chains that do not have treasury pallet.
> 
> https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Fexplorer-rpc-ws.mainnet.stages.humanode.io#/treasury
> 
> <img alt="image" width="1434" src="https://private-user-images.githubusercontent.com/1701852/381233806-d5032efa-df3b-47ae-a4ba-92fcadc9ff18.png?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MzAyMjE2NDksIm5iZiI6MTczMDIyMTM0OSwicGF0aCI6Ii8xNzAxODUyLzM4MTIzMzgwNi1kNTAzMmVmYS1kZjNiLTQ3YWUtYTRiYS05MmZjYWRjOWZmMTgucG5nP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI0MTAyOSUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNDEwMjlUMTcwMjI5WiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9MTg3NjQ1YmRiYjQ2YTNmYTUxZTViMTJiMjI0ZDEzZjliMjgyMjcyZTdlYTYwYmNjYjEwNzhlYTliNzUwOTg0ZCZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QifQ.25P2jAljk8Dq4gUYRcuBDNkK4IlhaUdBII9OA6ZedEA">

Issue #11031, opened 2024-10-23
Treasury Page No Longer Available


TarikGul:
Currently with the most recent calls that have been removed from runtime 1003003 (ex: `tx.treasury.proposeSend`), it has caused the treasury page from no longer being available since app routing requires the above example tx. 

In addition to that, there is a weird issue currently where when I get the page up and running again, it fails to persist the data it originally gets. For example, it will get all the proposals and render them correctly, but on the next subscribed call it makes it gets back no data, and empties the page of all proposals. I can't narrow down the issue exactly just yet, but it feels like to me it somehow is hitting: https://github.com/polkadot-js/api/blob/master/packages/api-derive/src/treasury/proposals.ts#L97-L101

That being said, `const totalProposals = useCall<BN>(api.query.treasury.proposalCount);` is also returning zero which is incorrect.

sesh92:
@TarikGul 

This change breaks chains that do not have treasury pallet.

https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Fexplorer-rpc-ws.mainnet.stages.humanode.io#/treasury

<img width="1434" alt="image" src="https://github.com/user-attachments/assets/d5032efa-df3b-47ae-a4ba-92fcadc9ff18">

PR #11047, opened 2024-10-29
API check for treasury page

sesh92:
Fixes: https://github.com/polkadot-js/apps/issues/11046
Ref: https://github.com/polkadot-js/apps/issues/11031#issuecomment-2444860196


TarikGul (approved):
Thanks - definitely a needed addition.

PR #11079, opened 2024-11-12
Fix staking payouts

TarikGul:
This should be the end of this staking payouts bug saga! (Finally)

Once the api deps are updated and these PR's are available:

https://github.com/polkadot-js/api/pull/6027
https://github.com/polkadot-js/api/pull/6026

What this fixes:

- Ensures the right eras are listed in the staking page.
- Ensures the right extrinsics are created when submitting a payout


TarikGul:
Once https://github.com/polkadot-js/apps/pull/11082 goes in this PR just needs to be up to date with master then can go in

PR #11081, opened 2024-11-12
Allow identity on Shibuya

Dinonard:
Attempt to make Shibuya have identity option in the UI.


TarikGul (changes requested):
Needs to be linted - can you confirm this solves your issue with identities on Shibuya

Dinonard:
Confirm it solves the issue. Thanks for the help!

TarikGul (approved):

PR #11135, opened 2024-12-11
Add Autonomys Taurus EVM testnet endpoints

jfrank-summit:
Add Autonomys Taurus EVM testnet to test networks.


TarikGul (changes requested):
Via the tests: `Autonomys Taurus EVM needs to be before Autonomys Taurus Testnet`

TarikGul (approved):

PR #11161, opened 2024-12-26
Disable unreachable endpoints

piggydoughnut:


TarikGul (approved):
Conflicts just need to be resolved. But thanks for taking care of this - we can close all the issues once this goes.

PR #11171, opened 2025-01-02
Add westend penpal to list of networks

franciscoaguirre:
We recently launched a new test network in Westend: Penpal. It's a non-system chain so we can test reserve asset transfers among other things. I added it to the list of networks for ease of use.


TarikGul (changes requested):
The ordering needs to be changed in order to satisfy the linter. 

I will also have the headers updated by tomorrow. Once it goes in you can pull it in, and that error wont appear anymore.

TarikGul:
Headers are updated, it would just need to be pulled in from master.

Then the ordering just needs to be updated

franciscoaguirre:
@TarikGul I ran the linter locally and I didn't get any error about the order. Is there a way to see the linter output in the CI? How should it be?

TarikGul:
You can run `yarn test ./packages/apps-config/src/endpoints/index.spec.ts` to test it out

franciscoaguirre:
Done, please check again

TarikGul (approved):

PR #11180, opened 2025-01-06
Fix test that writes to typesbundle

TarikGul:
Fixes the incorrect header from the test that writes to file.


FlorianFranzen (approved):
Fun fix for an interesting linter. LGTM!

PR #11190, opened 2025-01-09
update polkadot api

Rihyx:


TarikGul:
Thank you for the effort in trying to update these, but I am going to take care of this task (as I am doing it part of a release today) because its a bit more complicated than just bumping up the deps for the api. Thank you again for the effort though

PR #11205, opened 2025-01-15
Fix Block Weight title

ap211unitech:
## 📝 Description

This PR aims to fix the `block weight` typo, which is now replaced with `ref time`. This PR also adds all the required translations for supported languages.

Image


TarikGul (approved):
LGTM, thanks!

PR #11213, opened 2025-01-19
fix: get next asset id issue

ap211unitech:
## 📝 Description 

When creating a new asset, the Asset ID field should default to the next possible asset_id which is basically `last_created_asset_id + 1`

Screenshot from 2024-10-04 12-46-08

## 🤔 Previous behavior

The current UI makes it seem that any asset id would work, but after a runtime upgrade it is required that newly created asset id's follow a specific rule: the newly created asset's id should be the next possible id of the last created asset. 

In other terms, next_asset_id = previous_asset_id + 1

Without this change, users will face this error:

Screenshot from 2024-10-04 12-48-33

## 🚀 Current behavior

image

## ✅ Tasks

- [x] Fix `findOpenId` method to calculate default asset id
- [x] For Asset Hub, Next asset id is `prev asset id + 1`

### DIFF
--- a/packages/apps-config/src/api/constants.ts
+++ b/packages/apps-config/src/api/constants.ts
@@ -6,7 +6,7 @@ import type { HexString } from '@polkadot/util/types';
 import { knownGenesis } from '@polkadot/networks/defaults';
 import { assert, BN } from '@polkadot/util';
 
-function getGenesis (name: string): HexString {
+export function getGenesis (name: string): HexString {
   const network = Object.entries(knownGenesis).find(([network]) => network === name);
 
   assert(network?.[1][0], `Unable to find genesisHash for ${name}`);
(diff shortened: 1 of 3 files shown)


packages/apps-config/src/api/constants.ts
```diff
@@ -6,7 +6,7 @@ import type { HexString } from '@polkadot/util/types';
 import { knownGenesis } from '@polkadot/networks/defaults';
 import { assert, BN } from '@polkadot/util';
 
-function getGenesis (name: string): HexString {
+export function getGenesis (name: string): HexString {
```
TarikGul, line 9: Why are we exporting?

TarikGul: Nvm, i see it :)

TarikGul (approved):
👍 Looks great, really good job!

PR #11216, opened 2025-01-20
Disable unreachable endpoints

ap211unitech:
## 📝 Description

There are some endpoints which are unreachable, so we are commenting it out.


TarikGul (approved):
👍

PR #11222, opened 2025-01-21
fix: default asset selection

ap211unitech:
## 📝 Description

On visiting the balances page for assets, a continuous loading screen is displayed. However, searching for assets manually from the dropdown works fine. But this can be confusing and difficult for some users to notice.

Image

## ✅ Current Behavior

https://github.com/user-attachments/assets/acbe85e8-e8f5-4cc0-add2-3592f082df49


TarikGul (approved):
👍  THanks :)

PR #11229, opened 2025-01-23
AssetHub: Fee payment with any sufficient asset

ap211unitech:
## 📝 Description

This PR will add the feature to select the asset for paying fee. 

## ✅ Tasks

- [x] Create `PayWithAsset` component which allow users to choose asset to pay fee
- [ ] Integrate it with Transfer Modal (for now)
- [ ] End to End testng for it (just for AssetHub)
- [ ] Allow this feature to other chains which can accept fee in other assets

### DIFF
--- a/packages/react-components/src/PayWithAsset/index.tsx
+++ b/packages/react-components/src/PayWithAsset/index.tsx
@@ -1,0 +1,79 @@
+// Copyright 2017-2025 @polkadot/react-components authors & contributors
+// SPDX-License-Identifier: Apache-2.0
+
+import type { DropdownItemProps } from 'semantic-ui-react';
+import type { AssetInfoComplete } from '@polkadot/react-hooks/types';
+
+import React, { useCallback, useEffect, useMemo, useState } from 'react';
+
+import { getGenesis } from '@polkadot/apps-config';
+import { Dropdown } from '@polkadot/react-components';
+import { useApi, useAssetIds, useAssetInfos } from '@polkadot/react-hooks';
+import { formatNumber } from '@polkadot/util';
+
+import { useTranslation } from '../translate.js';
+
+const ALLOWED_CHAINS = [getGenesis('statemint')];
+
+const PayWithAsset = () => {
+  const { t } = useTranslation();
+  const { api } = useApi();
+  const ids = useAssetIds();
+  const assetInfos = useAssetInfos(ids);
+  const [infoIndex, setInfoIndex] = useState('0');
+
+  const nativeAsset = useMemo(
+    () => api.registry.chainTokens[0],
+    [api]
+  );
+
+  const completeInfos = useMemo(
+    () => (assetInfos
+      ?.filter((i): i is AssetInfoComplete =>
+        !!(i.details && i.metadata) && !i.details.supply.isZero() && !!i.details?.toJSON().isSufficient)
+      .sort((a, b) => a.id.cmp(b.id))) || [],
+    [assetInfos]
+  );
+
+  const assetOptions = useMemo(
+    () => [
+      { text: `${nativeAsset} (Native)`, value: nativeAsset },
+      ...completeInfos.map(({ id, metadata }) => ({
+        text: `${metadata.name.toUtf8()} (${formatNumber(id)})`,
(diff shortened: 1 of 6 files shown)


packages/react-hooks/src/constants.ts
```diff
@@ -18,3 +20,6 @@ export const ChainRenewalStatus = {
 
 // block time on coretime chain is 2 x slower than on relay chain
 export const BlockTimeCoretimeToRelayConstant = 2;
+
+// list of chains which support sufficient non-native assets to pay fee
+export const CHAINS_WITH_FEE_ASSET = [getGenesis('statemint')];
```
TarikGul, line 25: We need to add support for statemine as well

ap211unitech: Done

packages/react-hooks/src/constants.ts
```diff
@@ -18,3 +20,6 @@ export const ChainRenewalStatus = {
 
 // block time on coretime chain is 2 x slower than on relay chain
 export const BlockTimeCoretimeToRelayConstant = 2;
+
+// list of chains which support sufficient non-native assets to pay fee
+export const CHAINS_WITH_FEE_ASSET = [getGenesis('statemint'), getGenesis('statemine')];
```
TarikGul, line 25: Is there a world where we can make this generic - where we count on what is available in the metadata - therefore this can extend to all chains that support this functionality and not just assethubs?

For now I am okay with this since it will give support to the Asset Hubs, but generally would like to see a TODO comment here saying we can improve on this, and how.

packages/react-hooks/src/utils/getFeeAssetLocation.ts
```diff
@@ -0,0 +1,31 @@
...
+import { CHAINS_WITH_FEE_ASSET } from '../constants.js';
+
+export const getFeeAssetLocation = (api: ApiPromise, feeAsset: AssetInfoComplete | null): AnyNumber | object | undefined => {
+  const genesis = api.genesisHash.toHex();
+
+  if (!CHAINS_WITH_FEE_ASSET.includes(genesis) || !feeAsset?.id) {
+    return undefined;
+  }
+
+  switch (genesis) {
```
TarikGul, line 17: This switch case seems a bit unnecessary since there is only one case.

Similar to the constants I think we need to better optimize for multi chains. Here we are only accessing the first value in the `CHAINS_WITH_FEE_ASSET` so it seems the content of the constant is a bit unorganized.

TarikGul (changes requested):
I think we can better optimize the way we handle chains that support using assetId's. Even if the support only extends to assset-hub (statemine, statemint) we can better improve our workflow here.

The best approach IMO is to see if the necessary assets pallet exists (there might be some additional checks we can make to ensure we can pay with assets for fees) - check what the palletIndex is (for inputting in the multilocation) - and then from there offer support for paying with assets for fees.

packages/react-hooks/src/constants.ts
ap211unitech: I think, It can be made generic. If `SignedExtensions` (in metadata) contains `ChargeAssetTxPayment`, then it should accept non-native sufficient assets as fee.

Also, It should implement `assetConversion` Pallet.

```ts
const isEnabled = useMemo(() =>
    api.registry.metadata.extrinsic.signedExtensions.some(
      (a) => a.identifier.toString() === 'ChargeAssetTxPayment'
    ) &&
    api.tx.assetConversion && completeInfos.length > 0,
  [api.registry.metadata.extrinsic.signedExtensions, api.tx.assetConversion, completeInfos.length]
  );
```

Source - https://paritytech.github.io/polkadot-sdk/master/pallet_asset_tx_payment/struct.ChargeAssetTxPayment.html

packages/react-hooks/src/utils/getFeeAssetLocation.ts
ap211unitech: Removed `switch` statement and use of constant variable & Fetched pallet Index from metadata itself

TarikGul (approved):
Amazing job! 👍

PR #11240, opened 2025-01-29
fix: 🐛 Update `moonbeam-types-bundle` to latest

timbrinded:
Updates the Moonbeam chains to use the latest types-bundle for custom RPCs and types. 

This will users to access the ethapis: `debug` `trace` `txpool` (and fixed `moon`) via the PJS app without having to rely on a raw JSON RPC call.


TarikGul (approved):
👍 Nice!

PR #11243, opened 2025-01-30
fix: make use of `spendLocal` in treasury

ap211unitech:
## 📝 Description

This PR aims to make use of `spendLocal` instead of `proposeSpend` as this has been deprecated.

### DIFF
--- a/packages/page-treasury/src/Overview/ProposalCreate.tsx
+++ b/packages/page-treasury/src/Overview/ProposalCreate.tsx
@@ -111,7 +111,7 @@ function Propose ({ className }: Props): React.ReactElement<Props> | null {
               label={t('Submit proposal')}
               onStart={toggleOpen}
               params={[value, beneficiary]}
-              tx={api.tx.treasury.spendLocal}
+              tx={api.tx.treasury.spendLocal ?? api.tx.treasury.proposeSpend}
             />
           </Modal.Actions>
         </Modal>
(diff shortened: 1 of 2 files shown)


TarikGul (changes requested):
Nice work, so technically speaking this is great. But it doesn't take into account backwards compatibility for chains that use `proposeSend`. So there needs to be support for both depending on what is exposed at runtime.

ap211unitech:
> Nice work, so technically speaking this is great. But it doesn't take into account backwards compatibility for chains that use `proposeSend`. So there needs to be support for both depending on what is exposed at runtime.

To fix this, We provide `proposeSpend` as option if `spendLocal` doesn't exists like this `api.tx.treasury.spendLocal ?? api.tx.treasury.proposeSpend`.

If `spendLocal` and `proposeSpend` both does not exists, then `Submit Proposal` button would be disabled.

TarikGul:
Is the call in `apps-routing` correct as well? Will it work for proposeSend?

ap211unitech:
> Is the call in `apps-routing` correct as well? Will it work for proposeSend?

Yes it is. In `apps-routing`, it need this only. 

```ts
needsApi: [
        'query.treasury.proposals'
      ]
```

TarikGul (approved):
LGTM, nice job!

albertov19:
Hey just a small comment. `spendLocal` will be deprecated in favour of `treasury.spend` in the near future.

I did not check, but just FYI as if we only support `spendLocal` the flow will be broken either way later

TarikGul:
> Hey just a small comment. `spendLocal` will be deprecated in favour of `treasury.spend` in the near future.
> 
> I did not check, but just FYI as if we only support `spendLocal` the flow will be broken either way later

Thanks for the heads up - we'll put in a guard for now that disables the send button if proposeSend and or spendLocal doesn't exist.

We'll work on getting support for spend as well.

PR #11247, opened 2025-01-31
Ensure safeguard against treasury spendLocal functions

TarikGul:
This ensures that for certain chains that don't have `proposeSend` and or `spendLocal` we can safeguard from the active send button


TarikGul:
rel: https://github.com/polkadot-js/apps/pull/11243

PR #11256, opened 2025-02-03
fix: showing what core the chain is renewed on

piggydoughnut:
- Fixed not showing "renewed"
- Added what core the chain was renewed on.

Screenshot 2025-02-03 at 7 29 05 PM

### DIFF
--- a/packages/react-hooks/src/useCoretimeInformation.ts
+++ b/packages/react-hooks/src/useCoretimeInformation.ts
@@ -144,14 +144,8 @@ function useCoretimeInformationImpl (api: ApiPromise, ready: boolean): CoretimeI
           renewalStatus = `${ChainRenewalStatus.Renewed} on ${workplan[0].core}`;
         }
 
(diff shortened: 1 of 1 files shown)


TarikGul:
Question what is the something `renewed on <something>` mean?

piggydoughnut:
@TarikGul 
Renewed on X, is indeed on core X. Chains do not always get the same core when they renew, they can be allocated anything. So the will "migrate" to another core.

I m not sure what you mean by overlap. Do you mean that it says "renewed on 13", but there is a chain on 13 right now?
In that case maybe we could put something like. "Next cycle on core 6"

TarikGul:
> @TarikGul Renewed on X, is indeed on core X. Chains do not always get the same core when they renew, they can be allocated anything. So the will "migrate" to another core.
> 
> I m not sure what you mean by overlap. Do you mean that it says "renewed on 13", but there is a chain on 13 right now? In that case maybe we could put something like. "Next cycle on core 6"

Yea That is exactly what I meant by overlap. I think something referring to it being the next cycle would help clear confusion, since I was confused when reading it initially.

TarikGul (approved):
🚀

PR #11270, opened 2025-02-06
Deleted OnFinality 2 endpoints

FiveKG:
- Khala
`wss://khala.api.onfinality.io/public-ws`

- Ajuna
`wss://ajuna.api.onfinality.io/public-ws`


FiveKG:
> Deleting or just disabling? Because they looked to be just commented out.

@TarikGul  
Onfinality no longer supports these two networks, but they might be supported again in the future, so I commented them out.

TarikGul (approved):
👍

PR #11272, opened 2025-02-06
chore: update @polkadot/types dep

ap211unitech:
This PR is a follow-up PR to https://github.com/polkadot-js/apps/pull/11229. 

We need to upgrade `@polkadot/types` dependency.


TarikGul:
Generally we need to update all the @polkadot/* deps at once. I'll have the extension released today - then we can go ahead and upgrade everything.

TarikGul:
Since this is just bumping the patch we can update it for now actually, and I'll follow up with a full version bump!

TarikGul (approved):

PR #11332, opened 2025-02-25
Maximum transfer value while teleport

ap211unitech:
## 📝 Description

Currently, users can input any value when teleporting assets, which can lead to unexpected errors if the amount exceeds the allowable limit or if it doesn't leave existential deposit behind. This PR addresses the issue by fetching the maximum transferable value for Teleport and enforcing this limit to ensure users cannot exceed it.

This update improves reliability and provides a smoother user experience by preventing invalid transfers and unexpected errors.

Image

Image


TarikGul (approved):
👍

PR #11338, opened 2025-02-27
Disable Unreachable Endpoints

ap211unitech:
## 📝 Description

This PR aims to disable the unreachable RPC endpoints. They can be restored if required.


helixstreet:
Apologies for the hassle. Would you mind checking one more time? If it looks good, can we let the CI test run overnight? I think I found a solution.

ap211unitech:
> Apologies for the hassle. Would you mind checking one more time? If it looks good, can we let the CI test run overnight? I think I found a solution.

Working fine now. Restoring the endpoints.

helixstreet:
Thank you very much indeed. Your understanding and assistance in this matter are highly appreciated. I now understand the cause of the issues with the CI and my endpoint, and they have been resolved.

TarikGul (approved):

Issue #11339, opened 2025-02-27
[ajuna] runtime can't verify transaction anymore: wasm trapped


clangenb:
We did not change anything on our end, but we suddenly can't send transactions anymore with polkadot-js/apps:

See error:

![Image](https://github.com/user-attachments/assets/2dac8251-ba64-47f0-8a5d-3685c0adf473)

![Image](https://github.com/user-attachments/assets/5f012e43-9a96-48a3-80d3-6ba158852db9)

I am confident that this has to do with the latest release as the commits have something to do with asset fee payment, and we also use asset payment.

Our runtime does not do anything fancy regarding asset-tx-payment. It should be the same as asset hub essentially: https://github.com/ajuna-network/ajuna-parachain/blob/00b58102a2410ac3b4c822ba5dcecc70b53770e6/runtime/ajuna/src/lib.rs#L150

TarikGul:
cc: @ap211unitech

ap211unitech:
Fixing with https://github.com/polkadot-js/apps/pull/11341

clangenb:
Awesome, thanks for the quick action!

PR #11341, opened 2025-02-28
Enhancement: Improve Fee Token Verification & Asset ID Handling

ap211unitech:
## 📝 Description

This PR aims to fix - 

- **Initial `selectedAssetValue` Update:** Now initializing with `-1` instead of `0`, as non-native assets can have an asset ID of `0`, which may cause issues like [this](https://github.com/polkadot-js/apps/issues/11339).
- **Enhanced Fee Token Verification:** Improved checks to ensure a chain implements sufficient fee token mechanism.

https://github.com/user-attachments/assets/d26dfefc-1ca4-4846-a165-ca3138a30cc2


darkfriend77:
I tested the transfer on ajuna and it works again with the fix.

image

ap211unitech:
Great! It will be merged soon.

TarikGul (approved):
Thanks!

PR #11344, opened 2025-02-28
Fix: Voting on referenda breaks Apps

ap211unitech:
## 📝 Description

After v1.4.0 runtime upgrade, Relay chains i.e. Kusama and Polkadot, or other parachains chains, replaced `voting` method with `votingFor`. So, We will be adding a safety check here so that app doesn't break.


TarikGul (approved):
Great job, thank you!

PR #11350, opened 2025-03-03
Improve Address Book Search

ap211unitech:
## 📝 Description

In current version of polkadot.js apps, Search feature for Address book doesn't work properly. For example, it doesn't filter records based on account address, on-chain identity etc.

In this PR, we are aiming to fix these issues which will eventually improve the UX. 

https://github.com/user-attachments/assets/f91f6507-e4a4-4b24-845b-1e38da07d09b


TarikGul (approved):
🚀

PR #11355, opened 2025-03-04
Replace `isEmpty` checks with utility function

ap211unitech:
## 📝 Description

This PR aims to create a custom utility function for checking `isEmpty` state.  The reason is, For any address that is not a validator, the `staking.validators` method returns - 

```ts
{
    "commission": 0,
    "blocked": false
}
```

We have to check absolute emptyness for this Struct.

image

image


TarikGul (approved):
Great job 🚀

PR #11378, opened 2025-03-11
Fix dark mode display hash display result in Fork 

Amity808:
Users can read the fork hash display in Explorer/fork when the result is not finalized in darkmode. Fix from this 
This was the original dark mode followed by light mode 
Screenshot 2025-03-11 at 03 42 28

Screenshot 2025-03-11 at 03 42 57

The images below are taken after I fixed the dark mode, which is also compatible with the light mode.
Screenshot 2025-03-11 at 03 47 31
Screenshot 2025-03-11 at 03 52 26


ap211unitech (approved):

TarikGul (approved):
LGTM, thanks!

PR #11381, opened 2025-03-11
Handle the case when pallet slots is not being used

tatulea:
The `slots` module does not have to be implemented by all relay chains. Currently, the frontend will crash if the implementation is not available, so making `slots` optional was required.

### DIFF
--- a/packages/page-parachains/src/Overview/useParaInfo.ts
+++ b/packages/page-parachains/src/Overview/useParaInfo.ts
@@ -71,7 +71,7 @@ function useParaInfoImpl (id: ParaId): Result {
     [(api.query.parasHrmp || api.query.paraHrmp || api.query.hrmp)?.hrmpWatermarks, id],
     [(api.query.parasInclusion || api.query.paraInclusion || api.query.inclusion)?.pendingAvailability, id],
     [api.query.registrar.paras, id],
-    [api.query.slots.leases, id]
+    [api.query.slots?.leases, id]
   ], MULTI_OPTS);
 }
 
--- a/packages/page-parachains/src/Parathreads/useParaMap.ts
+++ b/packages/page-parachains/src/Parathreads/useParaMap.ts
@@ -63,7 +63,7 @@ function useParaMapImpl (ids?: ParaId[]): Result | undefined {
     [hasLinksMap]
   );
 
-  return useCall<Result>(ids && api.query.slots.leases.multi, [ids], {
+  return useCall<Result>(ids && api.query.slots?.leases.multi, [ids], {
     transform,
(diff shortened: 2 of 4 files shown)


ap211unitech:
One question, Does it break anything currently?

tatulea:
> One question, Does it break anything currently?

not for the manual testing that i was conducting. if the slots pallet is not available, then the data won't be displayed. without this change, the frontend crashes because it expects slots to be available

ap211unitech:
@tatulea For now, the slots pallet is always present in the Polkadot runtime, but its existence in the long term depends on the network's evolution. If there are any governance decisions or updates that modify slot allocation, changes may be required accordingly. So these changes are not necessary at this time. 
However, if future updates remove or replace the `slots` pallet, we can make the necessary adjustments accordingly. In my opinion, this PR can be closed for now.

tatulea:
The slots pallet is not required to be present in every ralay chain. There are chains where it is not present, even tho Polkadot has it

TarikGul:
@tatulea  If you can give a reproducible step by step on how you're getting the UI to crash that would be great.

If anything we can just add `api.query.slots` to the apps-routing of page-parachains, and it wont load those assets.

tatulea:
> @tatulea If you can give a reproducible step by step on how you're getting the UI to crash that would be great.
> 
> If anything we can just add `api.query.slots` to the apps-routing of page-parachains, and it wont load those assets.

One example is `https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Fstagelight.tanssi-dev.network#/parachains`

TarikGul:
> > @tatulea If you can give a reproducible step by step on how you're getting the UI to crash that would be great.
> > If anything we can just add `api.query.slots` to the apps-routing of page-parachains, and it wont load those assets.
> 
> One example is `https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Fstagelight.tanssi-dev.network#/parachains`

Okay so it's crashing because the chain you provided has `query.paras.parachains` available, but you don't have the slots pallet. 

The PR I added ensures that the parachain page wont be available if the slots pallet is not there.

TarikGul:
> > The PR I added ensures that the parachain page wont be available if the slots pallet is not there.
> 
> But can it be that the relay chain does not require to use slots, but can still have parachains?

The logic around the UI for the parachains page was built with slots so no, unless you have the slots pallet active on the chain it wont work. That being said, if I understand correctly, once all the leases are done we can remove all the leases info since they won't be continued.

ap211unitech:
Appears to be an outdated PR. Feel free to create another PR if issue exists.

PR #11400, opened 2025-03-18
Remove the dev notification for 1 month

### LINKED ISSUE #11398: Add ability to permanantly hide dotapps.io message
Please add the ability to permanently hide this warning message.

It is one of the most annoying things known to mankind.

Image

wirednkod:
With this implementation, once the X is click on the "dev" notification - then a localstorage is added in order to keep it hidden
An extra check is added - if the notification is hidden more than 1 month - then it will reappear and user needs to re-deactivate it.

Fixes #11398


TarikGul (approved):
👍  nice thank you!

PR #11403, opened 2025-03-19
Improvement: Multisig Call Data Field

ap211unitech:
## 📝 Description

This PR addresses layout inconsistencies in the **Multisig Pending Approvals Modal**, specifically when switching between multiple pending transaction hashes in the dropdown.  

## 🤔 Issue Details  

- When toggling between multiple pending hashes, the **call data input field** occasionally disappears.  
- If a user enters call data and then switches to another transaction, the previously entered call data becomes invisible.  
- Although this issue is **not consistently reproducible**, performing these actions **three to four times** typically triggers the bug.  

## 🚀 Fix & Improvements  

- Ensures that the call data input field remains visible when switching between transactions.
- Adjust the width of the **"Multisig message with call (for final approval)"** toggle to match its content. Currently, clicking on the empty space to the left also triggers the toggle action.
- Improves the layout to prevent unintended UI glitches.

### DIFF
--- a/packages/page-accounts/src/modals/MultisigApprove.tsx
+++ b/packages/page-accounts/src/modals/MultisigApprove.tsx
@@ -79,6 +79,7 @@ function MultisigApprove ({ className = '', onClose, ongoing, threshold = 0, who
       isMultiCall: !!multisig && (multisig.approvals.length + 1) >= threshold,
       multisig
     });
+    setCallHex('');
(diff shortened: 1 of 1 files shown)


valentinfernandez1 (approved):
Great Work!!

TarikGul (commented):
🚀

TarikGul (approved):

PR #11412, opened 2025-03-21
Fix: Multisig calls linked to a proxy account

ap211unitech:
## 📝 Description

The current polkadot-js explorer has a bug where MultiSig calls fail to execute properly under specific conditions.

## 🤔 Steps to Reproduce

1. Use a multisig account that also has an associated proxy account.
2. Attempt to perform a transaction **without** using the proxy (i.e., select "Don't use a proxy call for this call").
3. The UI gets stuck on the loading icon over the "Sign and Submit" button, making it unresponsive.
4. The only workaround previously was to remove the proxy account from the explorer.

## ✅ Fix Implemented

- Identified and resolved the root cause of the issue, ensuring proper handling of multisig transactions regardless of proxy usage.
- Updated transaction submission logic to prevent UI from freezing.
- Ensured that the "Sign and Submit" button remains functional in all cases.

## 🔥 Testing
- Conducted thorough testing covering all possible cases to prevent recurrence of this issue.
- Verified successful execution of MultiSig transactions with and without an associated proxy account.
- Ensured the UI remains responsive and functional under all conditions.
- This fix enhances the reliability of MultiSig transactions within polkadot-js explorer.
- Users are no longer required to remove the proxy account as a workaround.

## Proxy exists but not using proxy for signing

https://github.com/user-attachments/assets/1acb6689-725e-436b-8283-3143bf28648f

## Using Proxy for signing

https://github.com/user-attachments/assets/88a62247-6882-4c42-9556-011e5c8ab4cb

### DIFF
--- a/packages/react-signer/src/Address.tsx
+++ b/packages/react-signer/src/Address.tsx
@@ -142,13 +142,15 @@ function filterProxies (allAccounts: string[], tx: Call | SubmittableExtrinsic<'
     .map(([address]) => address);
 }
 
-async function queryForMultisig (api: ApiPromise, requestAddress: string | null, proxyAddress: string | null, tx: SubmittableExtrinsic<'promise'>): Promise<MultiState | null> {
+// This function shouldn't depend on proxyAddress, it can fetch multisig info based on source multisig account.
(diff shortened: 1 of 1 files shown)


TarikGul (approved):
🚀

PR #11426, opened 2025-03-26
FIX: Unwanted Errors on Sudo Page When Submitting Calls

ap211unitech:
## 📝 Description

The Sudo page sometimes returns unexpected errors when submitting a call. However, the same action works correctly when performed from the Extrinsics page.  

To resolve this issue, we are aligning the Sudo page's submission behavior with the Extrinsics page, which functions correctly. This involves analyzing the working implementation on the Extrinsics page and applying the necessary adjustments to the Sudo page.  

This fix ensures a more consistent and error-free experience when submitting calls via the Sudo page.

## 🤔 Previous Behaviour

https://github.com/user-attachments/assets/e6335184-f137-4fae-83a8-b5c5288a102b

## ✅ Current Behaviour

https://github.com/user-attachments/assets/02a75839-0dfc-4b57-bfda-e50e4fdedac1

### DIFF
--- a/packages/page-sudo/src/Sudo.tsx
+++ b/packages/page-sudo/src/Sudo.tsx
@@ -2,14 +2,13 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import type { SubmittableExtrinsic } from '@polkadot/api/types';
-import type { BN } from '@polkadot/util';
 
 import React, { useCallback, useState } from 'react';
 
-import { Button, Icon, InputNumber, styled, Toggle, TxButton } from '@polkadot/react-components';
+import { Button, Icon, styled, Toggle, TxButton } from '@polkadot/react-components';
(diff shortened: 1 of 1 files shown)


TarikGul (approved):
Clean and simple fix. 🚀

valentinfernandez1 (approved):

packages/page-sudo/src/Sudo.tsx
```diff
@@ -21,67 +20,46 @@ interface Props {
...
+    },
     []
   );
 
   return isMine
     ? (
       <StyledSection className={className}>
         <Extrinsic
-          defaultValue={apiDefaultTxSudo}
+          defaultValue={withWeight ? api.tx.sudo.sudoUncheckedWeight : api.tx.sudo.sudo}
```
MOZGIII, line 38: This causes state loss when the weight override is toggled. Please fix.

PR #11427, opened 2025-03-26
Require slots pallet for parachain page

TarikGul:
closes: https://github.com/polkadot-js/apps/pull/11381

Ensures slots pallet is also required for the parachain page.


ap211unitech:
Surely a much cleaner approach :) 

However, we also have to make sure `api.consts.slots` is not undefined or null. Please add this check as well.

ap211unitech:
@TarikGul Can you please add the required changes here? So, we can merge this PR and closed the related ones.

PR #11435, opened 2025-03-28
Fix multisig when the proxy pallet is not available

### LINKED ISSUE #11434: Unable to submit a multisig transaction on a chain without proxy pallet
~~#11412 didn't take into account that a chain might not have a proxy pallet enabled. Please fix.~~ UPD: might be something else, as I've found the tx submission not on a sudo page is also broken.

CC @ap211unitech @TarikGul 

Image

Image

You can try it with Humanode chain.

MOZGIII:
Fixes #11434.

### DIFF
--- a/packages/react-signer/src/Address.tsx
+++ b/packages/react-signer/src/Address.tsx
@@ -149,7 +149,7 @@ async function queryForMultisig (api: ApiPromise, requestAddress: string | null,
   if (isFunction(api.query[multiModule]?.multisigs)) {
     const address = requestAddress;
     const { threshold, who } = extractExternal(address);
-    const isProxyPalletAvailable = api.tx.proxy !== undefined;
+    const isProxyPalletAvailable = isFunction(api.tx.proxy?.proxy);
     const hash = (address && isProxyPalletAvailable ? api.tx.proxy.proxy(requestAddress || '', null, tx) : tx).method.hash;
 
     const optMulti = await api.query[multiModule].multisigs<Option<Multisig>>(address, hash);
--- a/packages/react-signer/src/Address.tsx
+++ b/packages/react-signer/src/Address.tsx
@@ -149,7 +149,8 @@ async function queryForMultisig (api: ApiPromise, requestAddress: string | null,
   if (isFunction(api.query[multiModule]?.multisigs)) {
     const address = requestAddress;
     const { threshold, who } = extractExternal(address);
-    const hash = (address ? api.tx.proxy.proxy(requestAddress || '', null, tx) : tx).method.hash;
+    const isProxyPalletAvailable = api.tx.proxy !== undefined;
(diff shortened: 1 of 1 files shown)


ap211unitech:
I would encourage using `isFunction(api.tx.proxy)` :)

ap211unitech:
that's fine :)

TarikGul:
> Note though I had to change the check from api.tx.proxy to api.tx.proxy?.proxy - and that .? is a non-trivial thing cause you have to know to add it, as it doesn't follow from the type-system automatically.

> I suggest adding | undefied to api.tx.proxy type to make the types actually true to life.

Not really possible atm. The static type generation is for substrate, polkadot, and kusama, having static types for each chain would be too much overhead. This is one of the trade offs with apps supporting any substrate based chain.

MOZGIII:
> The static type generation is for substrate, polkadot, and kusama, having static types for each chain would be too much overhead.

I'm not suggesting that though - just making the existing generated types properly describe the world. Supporting any chain means you actually *must* check for those pallets to conform to the expected API, to a degree at least, before attempting to use them. In fact, the apps *already do it*.
So, yes, polkadot and kusama do have the proxy pallet always. But you still have to write the code as if they might not have to support all the other chains. I think making the places where the checks are needed visible as the type level would be a simple fix.

So, you only need one set of types really, or maybe two - but still they'd come not from the chains you want to support, but from the particular subsets of APIs you require to implement a particular functionality, i.e. what you actually depend on.

That would be a dream. Yet, in practice, you'd likely still want to codegen types based on some well-known chains capabilities - like Polkadot and Kusama - simply to avoid manually authoring the API declarations (which can actually be automated with a new and different codegen - i.e. say we have a command that would generate the types only for a particular pallet - but alas...). Even *with all this said*! Even when working with Polkadot chain and Polkadot-specific type codegen. A pallet can still be missing - i.e. be gone from the runtime in the next version - but the codegen just doesn't allow for that possibility. You have to actually check for the pallet to be available at every time.

This seems more like the polkadot-api issue than the apps issue really. It would be great if we had something like this:

```ts
const api: Api<..., never> = new Api(...);

const _ = api.tx.proxy.proxy(...); // causes a type error at `api.tx.proxy`: invoking method an `undefined`

const hasProxy: boolean = api.check({ pallet: "proxy" });

const verifiedApi: Api<..., "proxy"> = api.assert({ pallet: "proxy" }); // `<Pallet>(what: { pallet: Pallet }) => asserts this is Api<..., Pallet>` - throw is the `pallet` is not available, is it is correct the `api` type to mark `proxy` as available

const _ = verifiedApi.tx.proxy.proxy(...); // works
```

Would also be nice to adjust the runtime error at the `api.tx.<pallet>.<method>` position when the `pallet` is `undefined` to day something more explicit about the situation. This is completely irrelevant to the types - but just something to improve the DX and UX for errors. Should also be done at the API level.

MOZGIII:
Also, merge? Any more work to do here?

ap211unitech (approved):

PR #11448, opened 2025-04-01
Add zkVerify Volta inflation parameters

drgora:
This PR adds the inflation parameters for the zkVerify Volta testnet (https://github.com/zkVerify/zkVerify).

It also updates the inflation calculation to avoid dividing by zero when `idealStake == 0`. This is useful, for example, when we want to hide the "ideal staked" field from the staking tab in the web-ui, when the ideal stake is not accounted for when computing the era rewards.


drgora:
> Needs `yarn lint`

Oops, sorry about that, should be fixed now, thank you!

ap211unitech (approved):
Looks fine 🚀

PR #11499, opened 2025-04-22
Add Deregister ParaId component and integrate into Actions

ap211unitech:
## 📝 Description

This PR aims to add the component for **deregister a reserved parachain ID (paraID)**.

https://github.com/user-attachments/assets/7c33ee54-c207-41af-b684-886e23b516c4

## ✨ What's New

- Reserved paraIDs (from `registrar.reserve`) are now **visible** in the UI.
- Users can **deregister** unregistered paraIDs directly from the interface.

## 🎯 Why

Users previously had no clear way to cancel a reserved paraID, leading to confusion and frustration over "locked" DOT/KSM deposits. This improves visibility and control during the parathread onboarding process.


valentinfernandez1:
Based on the implications of this action and the fact that once the button is clicked the UI is the same as any other transaction signing action, I’d advise adding some kind of extra confirmation before sending.

Something like:

    "You are about to deregister a paraId which will ...(describe consequences). Are you sure? Yes / No"

ap211unitech:
> Based on the implications of this action and the fact that once the button is clicked the UI is the same as any other transaction signing action, I’d advise adding some kind of extra confirmation before sending.
> 
> Something like:
> 
> ```
> "You are about to deregister a paraId which will ...(describe consequences). Are you sure? Yes / No"
> ```

### Added a warning message here to clearly explain what deregistering a paraID does and its consequences.

image

valentinfernandez1 (approved):
LGTM 🔥

TarikGul (approved):
Nice job, 👍

PR #11552, opened 2025-05-14
fix: multisig data disappears while signing multiple multisig tx's from same account

### LINKED ISSUE #11546: Multisig data disappears when signing multiple multisig tx's on same account
A similar bug to this seems to have come back: https://github.com/polkadot-js/apps/issues/11395 

When I have multiple multisig tx's on the same account, when signing the second, after posting the call data, the call data disappears.

The workaround seems to be refresh the page between multisig signatures.

rajk93:
fixes #11546

## Problem

The MultisigApprove modal was experiencing a UI glitch where the "call data for final approval" input field was constantly being reset during user interaction. This occurred because:

1. Any blockchain event (even those unrelated to multisig) would trigger the useMultisigApprovals hook
2. This caused unnecessary re-renders of the MultisigApprove component 
3. Each re-render would reset the input field value (`callHex`), making it impossible for users to input or edit call data

## Solution

The fix optimizes the useMultisigApprovals hook to prevent unnecessary triggers and re-renders:

1. Added a `useRef` to track previously processed multisig events
2. Implemented a filtering mechanism that only processes events that are:
   - From the multisig section
   - Related to the current address
   - Different from previously processed events
3. Created a comparison system using JSON.stringify to generate a hash of relevant events and only trigger updates when this hash changes

## Previous behaviour
https://github.com/user-attachments/assets/64e57944-bb25-4395-a4f8-7cf61595f3fe

## Current behaviour (after fix)
https://github.com/user-attachments/assets/a5e8dc42-9d53-49fb-b1de-0d0a22562f0d

### DIFF
--- a/packages/page-accounts/src/Accounts/useMultisigApprovals.ts
+++ b/packages/page-accounts/src/Accounts/useMultisigApprovals.ts
@@ -31,6 +31,7 @@ function useMultisigApprovalsImpl (address: string): [H256, Multisig][] | undefi
 
     if (multisigEvents.length) {
       const eventsHash = JSON.stringify(multisigEvents.map((e) => e.key));
+
       if (eventsHash !== prevEventsRef.current) {
         prevEventsRef.current = eventsHash;
         incTrigger();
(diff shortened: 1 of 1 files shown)


ap211unitech:
Needs `yarn lint`

rajk93:
> Needs `yarn lint`

resolved in [ef241a3](https://github.com/polkadot-js/apps/pull/11552/commits/ef241a3e4bc2afb8d0adbf64b5d2d22b88a3f819)

ap211unitech (approved):
Great job 👌

PR #11579, opened 2025-05-26
Auto suggestion for proxy based on transaction type

### LINKED ISSUE #9201: inaccurate suggestion to use the proxy
I have a non-transfer proxy set on one of my accounts. Whenever I want to do any operation on this account, the app always auto-suggests to me to use the proxy. I can unselect it with a toggle.

This is very handy, however when I want to do a transfer, it is impossible to use the proxy. So, the proxy auto-suggestion should be smarter and not actually suggest the proxy when it's a transfer. For any operation, it should anticipate whether proxy would work or not, and make the suggestion accordingly.

rajk93:
fixes #9201 

## Problem

Currently, when a user has a proxy account set up (e.g., a 'NonTransfer' proxy), the UI may auto-suggest this proxy for all transactions, even for actions the proxy isn’t allowed to do (e.g., suggesting a 'NonTransfer' proxy for a balance transfer).

## Solution

This PR refines the proxy filtering logic in [Address.tsx](https://github.com/polkadot-js/apps/blob/master/packages/react-signer/src/Address.tsx) to enable more intelligent auto-suggestion of proxies:

1. The global `BYPASS_PROXY_CHECK` constant has been removed.
2. The [filterProxies](https://github.com/polkadot-js/apps/blob/master/packages/react-signer/src/Address.tsx) function now accepts an optional `bypassProxyTypeCheck` boolean parameter (defaulting to false).
  - When `bypassProxyTypeCheck` is false (the default), `filterProxies` will strictly filter proxies based on their type and the nature of the transaction. This is intended for use by the auto-suggestion logic.
  - When `bypassProxyTypeCheck` is true, the type-specific filtering is skipped, allowing a list of all structurally valid proxies to be returned (e.g., for a UI element where the user explicitly wants to see all proxies and make their own choice).

## Testing Summary:
Tested with an account having two proxies: `PROXY1` (NonTransfer type) and `PROXY2` (Any type).
* For balance transfers, `PROXY2` (Any) was correctly auto-suggested, while` PROXY1` (NonTransfer) was not.
* For non-transfer operations (e.g. remarks), both proxies were identified as suitable, with the UI auto-suggesting one of them. The `filterProxies` logic behaved as expected in both scenarios.

### DIFF
--- a/packages/react-signer/src/Address.tsx
+++ b/packages/react-signer/src/Address.tsx
@@ -49,10 +49,6 @@ interface ProxyState {
   proxiesFilter: string[];
 }
 
-// true if we don't try to filter the list of proxies by type and
-// instead leave it up to the user
-const BYPASS_PROXY_CHECK = true;
-
 function findCall (tx: Call | SubmittableExtrinsic<'promise'>): { method: string; section: string } {
   try {
     const { method, section } = tx.registry.findMetaCall(tx.callIndex);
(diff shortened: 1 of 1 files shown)


ap211unitech (approved):
👍

TarikGul (approved):
🚀

PR #11635, opened 2025-06-18
Polkadex updated end point

charanks030:
We are moving on from onFinality to our own node deployed for our user. We have updated the changed endpoint in file production.ts

### DIFF
--- a/packages/apps-config/src/endpoints/production.ts
+++ b/packages/apps-config/src/endpoints/production.ts
@@ -456,7 +456,7 @@ export const prodChains: Omit<EndpointOption, 'teleport'>[] = [
   {
     info: 'polkadex',
     providers: {
-      OnFinality: 'wss://polkadex.api.onfinality.io/public-ws',
+      PolkadexSup: 'wss://so.polkadex.ee',
       RadiumBlock: 'wss://polkadex.public.curie.radiumblock.co/ws'
     },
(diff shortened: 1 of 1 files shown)


ap211unitech:
image

<hr />

The endpoint is occasionally slow—at times, I’ve had to wait nearly a minute for a response. Could you please take a look at the performance and see what's causing the delay?

charanks030:
Hi Polkadot-Js/Apps,

I checked, I don't see that it's slow, we run this with other Polkadot
validators, please check again its very quick in response.

Regards,
Charan

charanks030:
Also looks like you are trying to connect to "wss://polkadex-mainnet-rpc.dwellir.com" thats why its taking time. We are exactly changing that since we are in contract with dwellier anymore. 
Screenshot 2025-06-20 at 10 25 07 AM
Screenshot 2025-06-20 at 10 16 08 AM

ap211unitech:
> Also looks like you are trying to connect to "wss://polkadex-mainnet-rpc.dwellir.com" thats why its taking time. We are exactly changing that since we are in contract with dwellier anymore.  

Ofcourse, I am not trying that. I just tried `wss://so.polkadex.ee`. Right now, it is performing better. So, I'm merging it. However, if it gets slow and cron detects it, then it may be temporarily disabled in the future.

PR #11669, opened 2025-07-02
Handle Import/Export of In-browser accounts gracefully

ap211unitech:
## 📝 Description

Since the Accounts page already supports importing and exporting accounts, this PR focuses on addressing existing discrepancies and ensuring the feature works as expected.

In the app, the following account types are supported: in-browser, chopsticks, hardware, injected, multisig, proxied, QR, and testing. These represent the possible sections an account can fall under. Ultimately, every account fits into one of these categories. Among them, only in-browser, multisig, and QR accounts can be exported.

https://github.com/user-attachments/assets/2da55daf-5456-4452-9422-c1c0261cafe1

### DIFF
--- a/packages/page-accounts/src/modals/ExportAll.tsx
+++ b/packages/page-accounts/src/modals/ExportAll.tsx
@@ -198,7 +198,12 @@ const BrowserAccountsDiv = styled.div`
   margin-top: 1.5rem;
   padding-inline: 2rem;
 
-  div:nth-child(2) {width: 100%;}
+  div:nth-child(2) { 
+    width: 100%;
+    label { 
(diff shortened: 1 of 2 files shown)


valentinfernandez1:
Why are "testing" accounts not allowed?

ap211unitech:
> Why are "testing" accounts not allowed?

The only reason is that they are not specific to a single user. Their mnemonic or private key can be easily found on the web—for example, for accounts like Alice or Bob. They don't carry real significance for any individual user IMO.

rajk93 (approved):
LGTM!
Awesome work 🚀

TarikGul (approved):
🚀

Issue #11667, opened 2025-07-01
Staking Async Integration Part 2


kianenigma:
Hey all. First off, thanks for the rapid work on #11401. 

I didn't look too close of a look into what was merged there. Now that I have, I will compile my list of issues + follow-ups here: 

### `Command Center` 

1. > events: No events available

- [x] Please add a tooltip to this saying that there are no _relevant_  events, naming which events are being filtered. Otherwise it might imply that this block had no events at all, which is not true

2. > Active

I don't really know what this is about

3. Performance

I have not used this page much yet, but I do worry it will become too soon too hard to render. Perhaps we should render a new row on each side, only if one of the important data being displayed has changed

### `Staking` tab

- [x] is still present, but most of the data is not populated. I suggest removing it as a whole. We will soon remove the pallet from the runtime as well. 

## Westend Asset Hub 

<img width="1667" alt="Image" src="https://github.com/user-attachments/assets/a2069ccf-c0d7-4a08-a865-0e6f2f5e2772" />

The above picture has at least two issues: 

* [x] The validators shown are the asset hub collators, not the relay chain validators
* [x] The number of active nominators is miscalculated. What you should do to calculate this is as follows: 

```typescript
const era = (await api.query.staking.currentEra()).unwrap();

// you don't need this one, just for demonstration of how the storage works.
const overviews = (await api.query.staking.erasStakersOverview.entries(era)).map(([key, value]) => {
	const stash = key.args[1].toHuman();
	const metadata = value.unwrap();
	return { stash, metadata }
});
console.log(`overviews/exposed validators: ${overviews.length}`);

const exposedNominators = (await api.query.staking.erasStakersPaged.entries(era)).map(([key, value]) => {
	// @ts-ignore
	return value.unwrap().others.map((x) => x.who.toString());
}).flat();
console.log(`exposed backers: ${exposedNominators.length}`);

// this is what you want to show.
const uniq = [...new Set(exposedNominators)];
console.log(`unique exposures: ${uniq.length}`);

// This is equal to `exposedNominators.length`
const sumNominators = overviews.map(({ metadata }) => metadata.nominatorCount.toNumber()).reduce((a, b) => a + b, 0);
console.log(`sumNominators: ${sumNominators}`);
```

Moreover: 

* [x] We are lacking an overview here. A tooltip with the above message would be great: 

[...]

PR #11681, opened 2025-07-08
Staking Async Feedback

ap211unitech:
## 📝 Description

This PR addresses the feedback outlined in [this issue](https://github.com/polkadot-js/apps/issues/11667). Below is an overview of the updated changes.

image

image


kianenigma:
I think you are still parsing the snapshot range wrong. It should have a value like: 
- None if `Phase::None`
- `31 -> 15` (when we are for example in `Phase::Snapshot(14)`), 
- and `31 -> 0` in all other phases

This should always be the first and last item in this list: https://github.com/paritytech/polkadot-scripts/blob/51a99829460322be11379ea642f0642940604b1a/src/handlers.ts#L253C25-L254C15

ap211unitech:
Hi @kianenigma , I'm fetching it like this:

```ts
const snapshotRange = (
      await ahApi.query.multiBlockElection.pagedVoterSnapshotHash.entries()
    )
      .map(([k]) => k.args[0])
      .sort();

console.log(snapshotRange.map((a) => a.toHuman()));
```

and I am already rendering first and last item in list:

```ts
const uniquePages = Array.from(new Set(snapshotRange.map((p) => Number(p.toString())))).sort((a, b) => a - b);
const [minPage, maxPage] = [uniquePages[0], uniquePages[uniquePages.length - 1]];
```

The script you provided seems to fetch it the same way. Could you please clarify what I might be doing wrong?

kianenigma:
I think it is likely because `pagedVoterSnapshotHash` is now a double map of `(u32, u32)` (`(round, page_index)`), and we are interested in `page_index`, so you should use `k.args[1]`.

ap211unitech:
> I think it is likely because `pagedVoterSnapshotHash` is now a double map of `(u32, u32)` (`(round, page_index)`), and we are interested in `page_index`, so you should use `k.args[1]`.

Updated it now (and attached image as well)

ap211unitech:
Hi @kianenigma, When you have a moment, could you please provide your go-ahead so we can proceed with the review?

kianenigma (approved):

ap211unitech:
@valentinfernandez1 @TarikGul Hi guys, I think you can start reviewing this PR now—Kian just gave it the green light.

valentinfernandez1 (approved):

PR #11728, opened 2025-07-29
Filter events by user accounts

ap211unitech:
## 📝 Description

Currently, the explorer displays all chain events, which can make it difficult to identify events relevant to the user—especially on active networks with high event volume.

This PR introduces an **optional toggle** that allows users to filter and view **only the events involving their registered accounts** (as detected from the wallet extensions). This enhancement improves the developer experience by:

- Reducing visual noise in the event feed
- Making it easier to track specific extrinsics or state changes related to the user
- Improving usability for testers and developers on active networks

The toggle is off by default to preserve the existing behavior, and when enabled, it filters the events based on account involvement—either directly in the event data or as an block author/signer.

https://github.com/user-attachments/assets/eb49a638-e4f9-4336-a86d-a2cd4dc553f1

### DIFF
--- a/packages/apps/public/locales/en/app-explorer.json
+++ b/packages/apps/public/locales/en/app-explorer.json
@@ -10,6 +10,7 @@
   "No justifications available": "No justifications available",
   "No logs available": "No logs available",
   "Node info": "Node info",
+  "Show my events": "Show my events",
   "Unable to decode the block events. {{error}}": "Unable to decode the block events. {{error}}",
   "Unable to retrieve the specified block details. {{error}}": "Unable to retrieve the specified block details. {{error}}",
   "Waiting for next block...": "Waiting for next block...",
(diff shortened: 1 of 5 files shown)


TarikGul (approved):
Awesome feature!

rajk93 (approved):
Awesome feature
LGTM! 🚀

valentinfernandez1:
Awesome to have, I would honestly also like to see this toggle as part of the specific block page, so that the user doesn't need to scroll down until finding its event on a specific block. This will be useful specifically for chains that have a lot of transactions per block.

ap211unitech:
> Awesome to have, I would honestly also like to see this toggle as part of the specific block page, so that the user doesn't need to scroll down until finding its event on a specific block. This will be useful specifically for chains that have a lot of transactions per block.

Certainly! That would be a nice to have. I'll look into soon as I get some free time.

ap211unitech:
@valentinfernandez1 This feature has been added to the block details page now. Feel free to review it again.

PR #11742, opened 2025-08-01
Simplified Chain Endpoint Selection

ap211unitech:
## 📝 Description

Selecting an RPC endpoint can be confusing for new users, who may not understand what it is or which one to use. To improve usability, we’re proposing automatic selection of a rotating default endpoint from a pool of reliable public nodes. This avoids overloading a single provider and helps prevent centralization. Advanced users will still be able to manually choose their preferred endpoint via a dropdown.

Also, UI improvements has been done for Favorite chains header.

https://github.com/user-attachments/assets/7ff18a94-0a56-4c82-b1d3-262a90c16d02


valentinfernandez1:
This PR also includes the functionality for adding "Favorite chains" @ap211unitech please update the PR description accordingly.

PR #11748, opened 2025-08-04
Accumulate proof size for block details page

ap211unitech:
## 📝 Description

This PR fixes the calculation of `proofSize` on the block details page.

Previously, the implementation only captured the `proofSize` from the final `system` event in the block, which often led to inaccurate or underestimated values. This update changes the logic to fetch the actual block weight from storage, ensuring accurate reporting of `proofSize` and `refTime`.

image


valentinfernandez1 (approved):
LGTM 🚀

valentinfernandez1:
I am just wondering why would the system event display a value different that the actually stored one? I will do some investigation to make sure that we are not just misunderstanding something or if this is an issue that should be addresses by the core team.

Let's hold back on merging this for a bit.

PR #11764, opened 2025-08-12
Prioritize Selected Chain RPC Provider for System Chains

ap211unitech:
## 📝 Description

Currently, when a user selects a specific RPC provider for a relay chain (e.g., Kusama), Polkadot.js randomly selects RPC providers for the required system chains (e.g., people, coretime, etc.).  

This behavior can result in an inconsistent and sometimes unreliable user experience, as the randomly chosen system chain providers may not be as stable or performant as the user's chosen relay chain provider.

## 💡Motivation

The current randomization strategy for system chain RPC endpoints does not respect the user’s choice of relay chain provider, leading to inconsistent experiences. If a user has explicitly selected a provider for their relay chain, it is reasonable to assume they trust that provider’s infrastructure and would prefer to use it for associated system chains as well.

By aligning system chain RPC selection with the relay chain’s chosen provider (when available), we:

- Preserve user intent.
- Increase predictability and reliability in RPC connections.
- Leverage potentially better latency and reliability from the same provider’s network.

This approach ensures a more coherent and stable multi-chain browsing experience.

## 🖋️ Notes

This change only affects URL ordering. No functional changes to RPC connection logic or endpoint filtering are introduced. This behavior will only apply to chains that require connections with system chains.

<hr></hr>

https://github.com/user-attachments/assets/db66915e-5126-40a6-ad99-73f337e978e0

### DIFF
--- a/packages/react-hooks/src/useApiUrl.ts
+++ b/packages/react-hooks/src/useApiUrl.ts
@@ -30,10 +30,10 @@ function useApiUrlImpl (url?: null | string | string[]): ApiPromise | null {
     const map = new Map<string, string>();
 
     for (const endpoint of endpoints) {
-      const val = endpoint.textBy;
+      const rpcProvider = endpoint.textBy;
 
-      if (typeof val === 'string' && val.length > 0 && val !== 'Placeholder') {
-        map.set(endpoint.value, val);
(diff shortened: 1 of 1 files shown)


helixstreet:
Thanks indeed for this PR!

PR #11876, opened 2025-10-03
Improve block time handling for chains with elastic scaling

ap211unitech:
## 📝 Description

This PR fixes the Explorer’s `recent blocks` column, related metrics, and latency chart for chains with **elastic scaling**.  
The changes are backward-compatible and do not affect chains without elastic scaling (i.e., the original UI behavior remains unchanged).

## 🚀 Changes

- **Group blocks per slot** in the `recent blocks` column.  
  Blocks produced in the same slot share the exact same on-chain timestamp, so grouping them visually makes them easier to distinguish.

- **Improve the `last block` timer and target display** for better accuracy.

- **Fix block time chart** to correctly represent blocks produced in the same slot.  
  Since multiple blocks in a slot have the same on-chain timestamp, the chart now:
  - Computes the average block time for a slot as 

$$
\text{AvgBlockTime} = \frac{\text{blockTime}(\text{previousSlot}) - \text{blockTime}(\text{currentSlot})}{\text{blocksInSlot}}
$$

  - Plots values based on this calculation.  

  This approach ensures consistency, as no alternative on-chain metrics exist for distinguishing blocks within the same slot.  
  While off-chain timestamps could be used, they risk introducing inconsistencies and incorrect values.

<hr></hr>

image

image

### DIFF
--- a/packages/page-explorer/src/BlockHeader.tsx
+++ b/packages/page-explorer/src/BlockHeader.tsx
@@ -10,7 +10,6 @@ import { AddressSmall, styled } from '@polkadot/react-components';
 import { formatNumber } from '@polkadot/util';
 
 interface Props {
-  isLast: boolean
   value: AugmentedBlockHeader;
 }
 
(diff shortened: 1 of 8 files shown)


TarikGul (approved):
NOTE: I just looked at the general code, and did not look deep into the calculations, but overall looks good

PR #11877, opened 2025-10-04
fix: filter-out non eth accounts from eth chains

### LINKED ISSUE #11875: Regression: Extension's account list shows sr25519 accounts on chains that support only ethereum accounts
In #11592 @valentinfernandez1 implemented the support account type in `InjectedAccountExt`. The side effect of this implementation is that the non ethereum account from all the extensions are not filtered. Previously the non ethereum account was filtered in `keyring.loadAll` because try to add these account generate an exception but now that we're providing _an explicit account `type`_ the `addFromAddress` doesn't raise any exception for the standard polkadot accounts.

The practical issue is that you can find some accounts listed in _"My Accounts/extension"_ that seam legit and connected but you cannot use. The user can thinks that are valid address and transfert some funds on these accounts that doesn't have any know private key. I'm really scared by this regression because the founds moved on these accounts are lost.

 I've experimented this behavior on several chains (Moonbeam, Moonriver, VFlow,...) with several extensions connected (Polkadot.js, Talisman and Subwallet). 

I wasn't able to find a fix and I cannot understand in depth if the filtering due the inject extension is just a lucky side effect or something on that the code relay on.

This is the commit that generate the issue
- 'Support for Extension Ethereum Signing (#11592)' `4ebe82b8defb3d8d81ab99db9fd384130db5cd83`
- [Video example that show the issue](https://github.com/user-attachments/assets/f0f5c584-66a3-404d-891c-9895cc539468)

Console on previous version, in the last version these errors disappear.
```text
Api.tsx_+_3_modules:236 Keyring: Unable to inject {"address":"5D9txxK9DTvgCznTjJo7q1cxAgmWa83CzHvcz8zhBtLgaLBV","meta":{"name":"ZkVerify Sudo Testnet (talisman)","source":"talisman","whenCreated":0}}
eval @ bundle.js_+_11_modules:656
loadAll @ bundle.js_+_11_modules:650
loadOnReady @ Api.tsx_+_3_modules:236
await in loadOnReady
eval @ Api.tsx_+_3_modules:394
emit @ index.js:202
emit @ Api.js_+_127_modules:5647
#onProviderConnect @ Api.js_+_127_modules:6859
await in #onProviderConnect
eval @ Api.js_+_127_modules:6522
emit @ index.js:201
#emit @ index.js_+_1_modules:443
#onSocketOpen @ index.js_+_1_modules:559Understand this warning
Api.tsx_+_3_modules:236 Keyring: Unable to inject {"address":"5EhREncHsntgJaax9YQphk1xN3LxPu2Rzbz4A3g7Ut8cRXWq","meta":{"name":"Unknow Zkverify Testnet (talisman)","source":"talisman","whenCreated":1}}
eval @ bundle.js_+_11_modules:656
loadAll @ bundle.js_+_11_modules:650
loadOnReady @ Api.tsx_+_3_modules:236
await in loadOnReady
eval @ Api.tsx_+_3_modules:394
emit @ index.js:202
emit @ Api.js_+_127_modules:5647
#onProviderConnect @ Api.js_+_127_modules:6859
await in #onProviderConnect
eval @ Api.js_+_127_modules:6522
emit @ index.js:201
#emit @ index.js_+_1_modules:443
#onSocketOpen @ index.js_+_1_modules:559Understand this warning
...
```

Let mu know if I can help, Thanks

josepot:
Closes #11875 . I already provided an explanation of the cause in the issue.

### DIFF
--- a/packages/react-api/src/Api.tsx
+++ b/packages/react-api/src/Api.tsx
@@ -174,8 +174,18 @@ async function loadOnReady (api: ApiPromise, endpoint: LinkOption | null, fork: 
     isDevelopment,
     ss58Format,
     store,
-    type: isEthereum ? 'ethereum' : 'ed25519'
-  }, injectedAccounts);
+    type: 'ed25519'
+  },
+  isEthereum
+    ? injectedAccounts.map((account) => {
+      const copy = { ...account };
+
+      copy.type = 'ethereum';
+
+      return copy;
+    })
+    : injectedAccounts
+  );
 
   const defaultSection = Object.keys(api.tx)[0];
   const defaultMethod = Object.keys(api.tx[defaultSection])[0];
--- a/packages/react-api/src/Api.tsx
+++ b/packages/react-api/src/Api.tsx
@@ -174,7 +174,7 @@ async function loadOnReady (api: ApiPromise, endpoint: LinkOption | null, fork: 
     isDevelopment,
     ss58Format,
     store,
-    type: 'ed25519'
+    type: isEthereum ? 'ethereum' : 'ed25519'
   },
   isEthereum
     ? injectedAccounts.map((account) => {


TarikGul (approved):
Nice fix 👍 Thanks for the explanation.

> Leaky abstractions: UI context, ui-keyring, and keyring each make assumptions about type and “what the chain supports,” but there’s no single contract that all layers share. The behavior depends on who supplies type and when exceptions happen to bubble up.

This bit you wrote really resonates with me.

josepot:
> Nice fix 👍

I've put an extra commit to leave the ctx type as it was... Just in case. I'm pretty sure that it's only used as a fallback mechanism, when the type of the account is not provided, but who  knows, right? So, anyways, just in case, I left that part untouched. 

> This bit you wrote really resonates with me.

I swear that when I first saw the commit the OP pointed as the culprit of the issue I thought: "yeah, no, this can't be it... there must be some other commit that he skipped or something", boy, was I wrong...

ap211unitech:
Thanks @josepot for quick fix

josepot:
> Thanks @josepot for quick fix

No worries! Thanks for the quick merge!

Since this is a rather serious problem, may I request for a prompt hot-fix release? 🙏

ap211unitech:
Certainly! I’ll look into where we can apply the fix to get this flow working correctly.

josepot:
> Certainly! I’ll look into where we can apply the fix to get this flow working correctly.

What's wrong with the hot-fix of this PR?

Again, it's not meant to fix the architectural issues, it's just meant to get us out of this pickle. Is there a reason why you think that this PR doesn't do the trick? Could you please elaborate?

TarikGul:
If I remember correctly the fix should be live in https://polkadot.js.org/apps/#/explorer based on how the CI does releases per PR (since its considered the dev env). The only thing that needs to get released next is the IPFS version which requires an actual manual release of `apps`

ap211unitech:
> > Certainly! I’ll look into where we can apply the fix to get this flow working correctly.
> 
> What's wrong with the hot-fix of this PR?
> 
> Again, it's not meant to fix the architectural issues, it's just meant to get us out of this pickle. Is there a reason why you think that this PR doesn't do the trick? Could you please elaborate?


Apologies for the confusion. I assumed you were referring to fixing the architectural issues. Also, its already live now on https://polkadot.js.org/apps/#/explorer

PR #11592, opened 2025-05-30
Support for Extension Ethereum Signing

TarikGul:

PR #11884, opened 2025-10-06
Avoid to add Relay Chain Genesis Hash for Ethereum chain even if they are system parachain

la10736:
I know that polkadot's parachains are all not Ethereum parachain, but zkVerify has implement also a relaychain protocol and its system parachain Volta is an Ethereum chain.

Now if we don't filter the ethereum chains we're sharing the same keystore between a substrate chain (zkVerify Relay chain) and a ethereum chain (VFlow system parachain). This behaviour leed to show some silly and missleading accounts as legit like what's happened in #11592 for the injected accounts fixed in #11877 by @josepot .

I know that's not a general purpose case, and add `!isEthereum` seams a little bit hardcoded but if you have some other ideas of how to solve it I'm open to any discussion.


la10736:
@ap211unitech There're some chances that you can take a look to this or it's completely out of scope?

ap211unitech:
Thanks for your patience. Last few days, I have been quite busy honestly, didn't got the chance to look into it. Although, its a simple change, I would still have to check, it does not have any side effect. I will give it a closer look really soon.

ap211unitech:
LGTM

la10736:
@TarikGul Are there some chances that you can take a look on this? I know that is not an issue on a polkadot ecosystem, but for us is quite dangerous and the fix is really conservative.

TarikGul:
@ap211unitech review is more than good enough to merge :) I don't have the bandwidth to do any testing, but if he gave a check it's good to merge

PR #11930, opened 2025-10-19
Better staking overview

kianenigma:
Half backed approach. Still not quite good as: 

1. We are probably loading the entire nominator list under the hood, while we should not, which slows down the page. We only need to read `ErasStakersOverview` here
2. This data that comes only has the first page of nominators, which is why if you open this on westend, all validators have nominator count 64 (incorrect.)


kianenigma:
@ap211unitech any status update on this?

ap211unitech:
> @ap211unitech any status update on this?

Hey! I've gone through this earlier, but progress on it has been slow. I'll definitely take this up as soon as I have some bandwidth.

kianenigma:
@ap211unitech @TarikGul how about we merge this as-is, and do the improvements later? 

I am finding it ever more useful to know the accurate self+total stake of validators in the staking overview page.

ap211unitech:
@kianenigma I'll check this again today (or tomorrow at the latest) and merge it as soon as possible.

Issue #11859, opened 2025-09-25
Correctly compute block time for parachains


sandreim:
Currently block time seems to be computed from the inherent timestamps. For parachains using elastic scaling and building multiple blocks per 6s slot, the reported block time remains still 6s, because the timestamp in this blocks identifies the slot and not the block. We need to change and properly show the block time based on the amount of blocks produced per slot.

TarikGul:
cc: @ap211unitech

sandreim:
This chain is running with multiple cores and makes multiple blocks per 6s: https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Fkusama-yap-3392.parity-chains.parity.io#/explorer

josepot:
> We need to change and properly show the block time based on the amount of blocks produced per slot.

WDYT about the way that the PAPI console [handles this](https://dev.papi.how/explorer#networkId=localhost&endpoint=wss%3A%2F%2Fkusama-yap-3392.parity-chains.parity.io)?

🤔 

How do you think it should work, instead?

I mean, the reality is that often what happens is that it takes almost 6 secs to produce some blocks, and then when they come, then there are many in a row... I kinda like the way that the PAPI console displays that. However, I would like to know if you think that there is a better way to do it? cc @sandreim 

EDIT: maybe we could also display the average block-time?

sandreim:
This is where the block time needs to be reflected and is not:
<img width="1725" height="457" alt="Image" src="https://github.com/user-attachments/assets/f3471793-247b-407e-bbfa-470e04905e83" />

ap211unitech:
It's true that not every slot will contain the same number of blocks always, so updating the chart here would be a better idea rather than displaying it on the explorer page. On the explorer page, the target time should be the slot time only, in my opinion.

sandreim:
> It's true that not every slot will contain the same number of blocks always, so updating the chart here would be a better idea rather than displaying it on the explorer page. On the explorer page, the target time should be the slot time only, in my opinion.

Yes, we should keep the slot time (in the future we are planning to increase the slot time to 24s). But we should not call it the target block time in this case. 

<img width="185" height="81" alt="Image" src="https://github.com/user-attachments/assets/626441f4-afcb-4180-9fb9-538f5892c7fc" />

But what is missing right now is the effect of collators building multiple blocks per slot. We need a solution to show the block times on the `explorer` page and in the latency page. For `explorer`, something like the block times of past few blocks.

[...]

PR #11999, opened 2025-11-13
Displayed average block time for chains in explorer page

ap211unitech:
## 📝 Description

Rather than showing the slot duration, we will aim to display the moving average of the last few blocks (maybe 50), and we’ll call it the average block time (This PR aims to do so), as 'slot time' might be confusing and unfamiliar to regular users.

Even though the slot time was initially decided to be kept there after this discussion: https://github.com/polkadot-js/apps/issues/11859 but the requirement has since changed.

---

image

image


ap211unitech:
> It would be nice for a tooltip next to "average block time" that explains that it spans an average over x amount of blocks.

Added now.

PR #12023, opened 2025-11-24
Keep Sidebar Buttons outside the scrollable content

peetzweg:
uses flex column layout with dedicated scrollable area just for chains and keeping the header visible at the top for easy access of the actual controls available in the sidebar.

Before:

https://github.com/user-attachments/assets/ffc7d98e-abad-467d-9f03-f12aa744f306

After:

https://github.com/user-attachments/assets/e6b7c884-4cfb-4331-a035-a4d71a0d0fdc


ap211unitech:
Thanks for your contribution, @peetzweg! This will definitely improve the UX.

PR #12053, opened 2025-12-09
Derive proxy type name from `addProxy` call arguments

ap211unitech:
## 📝 Description

Previously, the UI attempted to detect all available proxy types using the logic implemented [here](https://github.com/polkadot-js/apps/blob/master/packages/page-accounts/src/modals/ProxyOverview.tsx#L152). However, on **Bittensor**, the proxy type ends with `RuntimeCommonProxyType` instead of `RuntimeProxyType`, causing the UI to miss some proxy types.

Additionally, not all proxy types were being populated for **Kusama** and **Polkadot Relay** networks.

This PR updates the detection logic to correctly infer available proxy types across different chains and prevents these inconsistencies going forward.

---

https://github.com/user-attachments/assets/2f9d05d0-c249-46f2-83b3-5350c0cea3d7

### DIFF
--- a/packages/page-accounts/src/modals/ProxyOverview.tsx
+++ b/packages/page-accounts/src/modals/ProxyOverview.tsx
@@ -148,8 +148,12 @@ function NewProxy ({ index, onChangeAccount, onChangeType, onRemove, proxiedAcco
 }
 
 function getProxyTypeInstance (api: ApiPromise, index = 0): KitchensinkRuntimeProxyType {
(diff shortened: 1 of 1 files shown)


rajk93 (approved):
LGTM!

cxpgo:
Thank you very much for your hard work. May I ask approximately how long it will take for the official website to be updated?

ap211unitech:
As soon as we merge this PR, it will be live then. Later today or tomorrow.

PR #12054, opened 2025-12-09
remove dwellir endpoint for acala

gflow33:


gflow33:
There was some miscommunication, closing the PR

PR #12065, opened 2025-12-12
Fix handling of v1 weights for historical blocks

F-OBrien:
Visiting a histroical block e.g. https://polkadot.js.org/apps/#/explorer/query/1 was throwing an error due to the legacy v1 weight. `Cannot read properties of undefined (reading 'toBn')`. This fix ensure both v1 and v2 weights are correctly handled.


ap211unitech (approved):
LGTM

PR #12100, opened 2026-01-06
Updated uniform naming schema for stakeworld endpoints

stakeworld:
Updating endpoints to more uniform naming schema for better manageability. Old endpoint url's will keep working for a while.


stakeworld:
Also disabling some endpoints since there is no funding at this moment. Hopefully again in the future.

@ap211unitech let me know if there are any questions.

bkchr (approved):

Issue #12155, opened 2026-02-07
Kusama governance curves don't appear


laboon:
The graphical curves showing approval and support curves do not appear for referenda on Kusama, only on Polkadot. They should appear on both.

Kusama:

<img width="1469" height="402" alt="Image" src="https://github.com/user-attachments/assets/de822412-7268-48b5-b417-a31d9d23a544" />

Polkadot:

<img width="1441" height="512" alt="Image" src="https://github.com/user-attachments/assets/602fe951-3faf-45b6-9cc1-aa8392f209c7" />

josepot:
> They should appear on both.

If you say so... why don't you open a PR?

KoxyG:
Hi @josepot  I would love to take up this issue and make the graphical curves appear for referenda on Kusama.

josepot:
> Hi [@josepot](https://github.com/josepot) I would love to take up this issue and make the graphical curves appear for referenda on Kusama.

go for it! Also, while you are at it, could you please do the same for Paseo and Westend? 🙏 

You are aware that these config values differ from chain to chain and that, unfortunately, AFAIK there is no runtime-api, storage-query, or view-function that makes these values easily accessible via runtime interactions, right? Do you know where to obtain them from?

KoxyG:
> Paseo and Westend?

Yes sure, I will do the sane for Paseo and Westend.

I don't yet know where to obtain the config value from. Pls can you point me to the right direction?

josepot:
> > Paseo and Westend?
> 
> Yes sure, I will do the sane for Paseo and Westend.
> 
> I don't yet know where to obtain the config value from. Pls can you point me to the right direction?

Apologies, the most relevant info is available via the `Referenda.Tracks` runtime constant. What's not easily available via runtime interactions are the expendable limits of each track.

KoxyG:
> > > Paseo and Westend?
> > 
> > Yes sure, I will do the sane for Paseo and Westend.
> > 
> > I don't yet know where to obtain the config value from. Pls can you point me to the right direction?
> 
> Apologies, the most relevant info is available via the `Referenda.Tracks` runtime constant. What's not easily available via runtime interactions are the expendable limits of each track.

Noted. Thank you!.

KoxyG:
@josepot   I made a pr here https://github.com/polkadot-js/apps/pull/12168

PR #12168, opened 2026-02-15
Fix: Show referenda approval/support curves on Kusama, Paseo & Westend (no more blank charts)

KoxyG:
### Summary
Approval and support curves were not rendered for referenda on Kusama (and similarly on Paseo and Westend), while they appeared correctly on Polkadot. This PR fixes that by gating curve calculation on the fields that are actually used for the curves instead of decisionDeposit.
Fixes #https://github.com/polkadot-js/apps/issues/12155

### Root cause
Curve data is computed in expandTracks() in useTracks.ts. The condition for calling calcCurves(info) was:
```
info.decisionDeposit && info.minApproval && info.minSupport
```

- calcCurves() (in util.ts) only uses decisionPeriod, minApproval, and minSupport from the track. It does not use decisionDeposit.

- On chains like Kusama, some tracks have decisionDeposit set to 0. In JavaScript, 0 is falsy, so the condition failed and the code fell back to zeroGraph, which produces no visible curves.

- Polkadot tracks often have a non-zero decisionDeposit, so the condition passed there and curves were shown.

So the bug was: we required a field that isn’t used for the curves (decisionDeposit) and that can be zero on some chains, which hid the curves on those chains.

### Change
packages/page-referenda/src/useTracks.ts
**Before:**
```
graph: info.decisionDeposit && info.minApproval && info.minSupport ? calcCurves(info) : zeroGraph,
```

**After:**
```
graph: info.decisionPeriod && info.minApproval && info.minSupport ? calcCurves(info) : zeroGraph,
```
Screenshot 2026-02-15 at 10 25 06 PM

We now require exactly the three fields that calcCurves() uses and that come from the Referenda.Tracks runtime constant:

- decisionPeriod – used for the time axis and curve math.

- minApproval – approval curve

- minSupport – support curve

We no longer require decisionDeposit, so tracks with zero decision deposit (e.g. on Kusama) still get curves when they have valid decisionPeriod, minApproval, and minSupport.

### DIFF
--- a/packages/page-referenda/src/useTracks.ts
+++ b/packages/page-referenda/src/useTracks.ts
@@ -16,7 +16,7 @@ const zeroGraph = { approval: [BN_ZERO], support: [BN_ZERO], x: [BN_ZERO] };
 
 function expandTracks (tracks: [BN, PalletReferendaTrackDetails][]): TrackDescription[] {
   return tracks.map(([id, info]) => ({
-    graph: info.decisionDeposit && info.minApproval && info.minSupport ? calcCurves(info) : zeroGraph,
+    graph: info.decisionPeriod && info.minApproval && info.minSupport ? calcCurves(info) : zeroGraph,
     id,
     info
   }));


KoxyG:
@josepot  Attaching the issue to the pr

PR #12222, opened 2026-03-31
DNM-yet: Add VerifyMultiSignature custom signed extension for people-paseo-next

bkontur:
Relates to: https://github.com/paritytech/polkadot-sdk/blob/129a48aa9f17a3d5ccbcb8cb4167dd505165d35f/substrate/frame/verify-signature/src/extension.rs#L43

cc: @peetzweg just PoC not sure if I match the runtime spec_name= people-paseo-next, you should be able to run my branch locally and see if it is working for you, then we can finish this PR,

but I think this TX is general and not related to the people, so it should go directly here: https://github.com/polkadot-js/api/blob/master/packages/types/src/extrinsic/signedExtensions/substrate.ts (when can do it afterwards, when we verify that it is working)

### DIFF
--- a/packages/apps-config/src/api/spec/index.ts
+++ b/packages/apps-config/src/api/spec/index.ts
@@ -96,6 +96,7 @@ import opportunity from './opportunity.js';
 import parallel from './parallel.js';
 import parami from './parami.js';
 import peaq from './peaq.js';
+import people from './people.js';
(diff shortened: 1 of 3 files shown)


bkontur:
closing in favor of https://github.com/polkadot-js/api/pull/6253

PR #12274, opened 2026-05-07
Move Bulletin Chain to Polkadot system parachains (paraId 1010)

mudigal:
https://github.com/paritytech/polkadot-bulletin-chain/issues/343

## Summary
  - Move Bulletin Chain from standalone production chains to Polkadot system parachains (`prodParasPolkadotCommon`) with paraId 1010
  - Add Spectrum (SimplyStaking) as a new RPC provider 
  - Remove the standalone Bulletin entry from `production.ts` to avoid duplication

  ## Changes
  - `productionRelayPolkadot.ts`: Added Bulletin Chain entry to `prodParasPolkadotCommon` with `chainsBulletinPNG` logo import
  - `production.ts`: Removed standalone Bulletin entry and unused logo import
  
  @bkontur - PTAL


packages/apps-config/src/endpoints/productionRelayPolkadot.ts
```diff
@@ -881,6 +881,20 @@ export const prodParasPolkadotCommon: EndpointOption[] = [
       logo: nodesAssetHubSVG
     }
   },
+  {
+    info: 'polkadotBulletin',
+    isPeopleForIdentity: true,
```
bkontur: What is this flag? Why true fór Bulletin?

packages/apps-config/src/endpoints/productionRelayPolkadot.ts
```diff
@@ -881,6 +881,20 @@ export const prodParasPolkadotCommon: EndpointOption[] = [
...
+    paraId: 1010,
+    providers: {
+      Spectrum: 'wss://spectrum-03.simplystaking.xyz/cG9sa2Fkb3QtMDMtOTFkMmYwZGYtcG9sa2Fkb3Q/9QbAeudedsupNA/polkadotbulletin/mainnet/'
```
bkontur: Is this correct? 

We have also: https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Fbulletin.amperfix.de#/rpc

packages/apps-config/src/endpoints/productionRelayPolkadot.ts
mudigal: My understanding - isPeopleForIdentity means the chain delegates identity storage to the People chain rather than managing identities itself.

packages/apps-config/src/endpoints/productionRelayPolkadot.ts
mudigal: Yes, me and Karol had a deep debate and figured out that this is supposed to be RPC nodes and not Collator Nodes. We currently have only one RPC node. Once we get Archive node end point, we will also put that.

bkontur: Amperfix is not rpc node?

franciscoaguirre: This is only for RPC nodes. If we only have one that's fine

franciscoaguirre (approved):

packages/apps-config/src/endpoints/productionRelayPolkadot.ts
mudigal: No Amperfix is not RPC node. They are Collator Node.

packages/apps-config/src/endpoints/productionRelayPolkadot.ts
```diff
@@ -881,6 +881,20 @@ export const prodParasPolkadotCommon: EndpointOption[] = [
...
+    },
+    relayName: 'polkadot',
+    text: 'Bulletin',
```
bkontur:
```suggestion
    teleport: [-1, 1000],
    text: 'Bulletin',
```

bkontur (approved):
I think we should add also `teleport: [-1, 1000],`

packages/apps-config/src/endpoints/productionRelayPolkadot.ts
mudigal: Do we support XCM teleports to Asset Hub?

bkontur: Yes, it Is systém para

karolk91:
> I think we should add also `teleport: [-1, 1000],`

lets maybe avoid adding teleports config for now? (assuming that lack of this config will not crash PJS)

There are some calls that are not feeless now - but the question is if that is correct in the first place? We are slowly migrating to feeless logic on every call

bkontur:
> > I think we should add also `teleport: [-1, 1000],`
> 
> lets maybe avoid adding teleports config for now? (assuming that lack of this config will not crash PJS)
> 
> There are some calls that are not feeless now - but the question is if that is correct in the first place? We are slowly migrating to feeless logic on every call

But there are other standard pallets which requires dot - collator session Keys or something

karolk91:
> > > I think we should add also `teleport: [-1, 1000],`
> > 
> > 
> > lets maybe avoid adding teleports config for now? (assuming that lack of this config will not crash PJS)
> > There are some calls that are not feeless now - but the question is if that is correct in the first place? We are slowly migrating to feeless logic on every call
> 
> But there are other standard pallets which requires dot - collator session Keys or something

true, but this is system chain - we are using invulnerables, and adding new ones seems to require to call `set_keys` by the collator signed origin - but I guess this is something to be sorted with collators directly before referendum to add invulnerables. There should not be normal user use-cases requiring balance on Bulletin

bkontur:
> > > > I think we should add also `teleport: [-1, 1000],`
> > > 
> > > 
> > > lets maybe avoid adding teleports config for now? (assuming that lack of this config will not crash PJS)
> > > There are some calls that are not feeless now - but the question is if that is correct in the first place? We are slowly migrating to feeless logic on every call
> > 
> > 
> > But there are other standard pallets which requires dot - collator session Keys or something
> 
> true, but this is system chain - we are using invulnerables, and adding new ones seems to require to call `set_keys` by the collator signed origin - but I guess this is something to be sorted with collators directly before referendum to add invulnerables. There should not be normal user use-cases requiring balance on Bulletin

Shell runtime supports teleports, why to restrict PJS? 

Standard system para should support teleports. Bulletin business should be orthogonal and our feesless stuff should be guarded by authorizations and tx extension

karolk91:
> > > > > I think we should add also `teleport: [-1, 1000],`
> > > > 
> > > > 
> > > > lets maybe avoid adding teleports config for now? (assuming that lack of this config will not crash PJS)
> > > > There are some calls that are not feeless now - but the question is if that is correct in the first place? We are slowly migrating to feeless logic on every call
> > > 
> > > 
> > > But there are other standard pallets which requires dot - collator session Keys or something
> > 
> > 
> > true, but this is system chain - we are using invulnerables, and adding new ones seems to require to call `set_keys` by the collator signed origin - but I guess this is something to be sorted with collators directly before referendum to add invulnerables. There should not be normal user use-cases requiring balance on Bulletin
> 
> Shell runtime supports teleports, why to restrict PJS?
> 
> Standard system para should support teleports. Bulletin business should be orthogonal and our feesless stuff should be guarded by authorizations and tx extension

I agree that Bulletin should support teleports because its a system chain in general - the only reason I was thinking about skipping this on the PJS level is exactly because Bulletin business is not related to DOT so why to expose this to users. It wont stop anyone to just manually do it from the "Extrinsic" tab - choosing manually the xcm pallet to send DOT to Bulletin. I'm also not super strong against adding this config and making it visible in the PJS "Teleport" tab 

if we want to add it we need to do it on both sides in this PR:
1. AssetHub config -> add 1010 there to the list 
2. Bulletin -> add [-1, 1000]

mudigal:
> > > > > > I think we should add also `teleport: [-1, 1000],`
> > > > > 
> > > > > 
> > > > > lets maybe avoid adding teleports config for now? (assuming that lack of this config will not crash PJS)
> > > > > There are some calls that are not feeless now - but the question is if that is correct in the first place? We are slowly migrating to feeless logic on every call
> > > > 
> > > > 
> > > > But there are other standard pallets which requires dot - collator session Keys or something
> > > 
> > > 
> > > true, but this is system chain - we are using invulnerables, and adding new ones seems to require to call `set_keys` by the collator signed origin - but I guess this is something to be sorted with collators directly before referendum to add invulnerables. There should not be normal user use-cases requiring balance on Bulletin
> > 
> > 
> > Shell runtime supports teleports, why to restrict PJS?
> > Standard system para should support teleports. Bulletin business should be orthogonal and our feesless stuff should be guarded by authorizations and tx extension
> 
> I agree that Bulletin should support teleports because its a system chain in general - the only reason I was thinking about skipping this on the PJS level is exactly because Bulletin business is not related to DOT so why to expose this to users. It wont stop anyone to just manually do it from the "Extrinsic" tab - choosing manually the xcm pallet to send DOT to Bulletin. I'm also not super strong against adding this config and making it visible in the PJS "Teleport" tab
> 
> if we want to add it we need to do it on both sides in this PR:
> 
> 1. AssetHub config -> add 1010 there to the list
> 2. Bulletin -> add [-1, 1000]

If there are no consequences of adding the teleport, I am fine with adding. And agree if so, we need to add on both sides. My question is should it be only with Asset Hub or also with other system chains? 

I went ahead and added the teleport changes.

bkontur (approved):

bkontur:
@TarikGul could you please take a look and merge this?

TarikGul (approved):

PR #12373, opened 2026-07-16
Add SIK and TunaStaking RPC providers to Bulletin (Paseo)

alexbcasas:
## Summary

Adds two more RPC providers to the Bulletin system chain entry on Paseo: SIK and TunaStaking.


alexbcasas:
@TarikGul another one super small 🙏

TarikGul (approved):

PR #12391, opened 2026-07-30
fix: validators reported as idle in staking accounts

### LINKED ISSUE #12388: Validators shown as idle in Staking Accounts on Polkadot Asset Hub
### Summary

On Polkadot Asset Hub, **Staking Async → Accounts** reports own validator stashes as idle: a lone `Nominate` button instead of `Stop`, absent from the **Validators** filter (shown under **Inactive**), no commission or min-validator-bond warning, and in **Payouts** the "Own validators" option is greyed out so that view is unreachable. **All Validators** lists the same stashes correctly - it uses a membership path
that never reads preferences.

### Cause
`isStashValidating` (`react-hooks/src/useOwnStashInfos.ts:41`) tests whether the stash's `ValidatorPrefs` are *empty*, via `isEmpty` (`react-hooks/src/utils/isEmpty.ts`). But `Validators` is a `ValueQuery` map (`polkadot-sdk substrate/frame/staking-async/src/pallet/mod.rs:662`), so an account that never validated returns `ValidatorPrefs::default()` - `{ commission: 0, blocked: false }` - byte-identical to a real validator at 0% commission accepting nominations.

The intention to validate *is* registered on chain. The UI reads that intention's contents instead of its existence, and those contents are indistinguishable from absence.

### Why this affects most validators, and cannot be worked around

[Referendum 1909](https://polkadot.subsquare.io/referenda/1909) set validator commission to 0% and capped it there: `staking.maxCommission` reads **0** on Polkadot Asset Hub. Commission is therefore 0 for everyone, `blocked` is the only field that can make a record non-blank, and "set a commission" is not an available workaround. **405 of the 600** active validators have `blocked: false` and are misreported - `blocked: true` is the only way to be classified correctly, so any validator that wants delegations is affected. (The absolute count drifts as validators are chilled below the minimum bond; the proportion holds near two thirds.)

Observed both ways on two validators differing only in that flag: `blocked: true` → `Stop`, listed under **Validators**; `blocked: false` → lone `Nominate`, absent from it.

The distinction the check needs exists on chain, just not on the path it queries (era 2246):

| | Raw storage (`state_getStorage`) | `staking.validators.multi()` |
|---|---|---|
| Real validator, `blocked: false` | the encoded default | `{"commission":"0.00%","blocked":false}` |
| Never validated | **absent (`null`)** | `{"commission":"0.00%","blocked":false}` |

### Reproduce

1. Open apps on Polkadot Asset Hub with an own validator stash whose preferences are `{ commission: 0, blocked: false }`: 405 of 600 qualify, so almost any elected validator will do (list `staking.validators.entries()` and take one whose value equals the default).
2. **All Validators**: the stash is listed. ✅
3. **Accounts**: lone `Nominate` instead of `Stop`; absent under the **Validators** filter, present under
   **Inactive**. ❌
4. **Payouts**: "Own validators" greyed out, view falls back to "Own stashes". ❌

`Validate` never appears on Asset Hub: it requires session keys, which live on the Relay Chain (`page-staking-async/src/Actions/Account/index.tsx:246-285`). To test without the stash key, add the address via **+ Proxied**; address-book contacts do not work, since `useOwnStashes` reads `useAccounts().allAccounts`.

### Affected code

`page-staking-async` renders on Asset Hub; `page-staking` is hidden there (`apps-routing/src/staking.ts:14-19`) but shares the hook and supplies the Payouts component.

- **Accounts filter**: `page-staking-async/src/Actions/index.tsx:93-97`
- **Row detail** (commission, min-bond warning): `page-staking-async/src/Actions/Account/index.tsx:210-224`
- **Payouts**: `ownValidators` (`page-staking-async/src/System/index.tsx:105-108`) empties, disabling the
  "Own validators" toggle (`page-staking/src/Payouts/index.tsx:165,244`); the view is unreachable, not just
  empty
- **`react-hooks/src/useAccountInfo.ts:63`**: same check for the `isValidator` badge (`:70` `isNominator`
  is correct)

### Fix

Test **presence** of the `Validators` entry, not emptiness of its value — raw storage returns `null` when
absent and the encoded default when present.

- Own stashes are few: one batched `queryStorageAt` over their `Validators` keys is exact.
- Or fetch `staking.validators.keys()` once and test membership (`counterForValidators` is available, since
  it is a `CountedStorageMap`).
- Keep `.multi()` for reading the preferences themselves (commission display); it simply cannot answer the
  presence question.
- Same change at `useAccountInfo.ts:63`.

Separately, `Payouts/index.tsx:165-166` latches `hasOwnValidators` and `myStashesIndex` in `useState`
initializers, so "Own validators" never re-enables if `ownValidators` resolves after mount.

### Environment

- Polkadot-JS Apps (`polkadot-js/apps`) `master` @ `65dd9cba45`
- Polkadot-JS API (`@polkadot/api`) 16.5.x
- Polkadot Asset Hub, `statemint` 2003002, era 2246
- runtime `pallet-staking-async`.

Closest existing reports, not duplicates: #997 (stale Nominate/Validate button state), #4798 (blocked
validators shown as nomination targets).

Imod7:
Closes #12388

### Description

- Validator detection checked whether an account's `ValidatorPrefs` value was empty.
- `staking.validators` is a `ValueQuery` map, so an account that never validated returns the default `{ commission: 0, blocked: false }`, which is the same value as a real validator on 0% commission.
- Referendum 1909 capped commission at 0 on Polkadot Asset Hub, so most validators returned that default and were reported as idle.
- Now tests whether the storage entry exists, instead of whether its value is empty.

### Changes

`react-hooks/src/useOwnStashInfos.ts`

- `isStashValidating` is now `!validateInfo.isNone`, read from one single-key `state_subscribeStorage` per stash instead of `staking.validators.multi`.
- `validatorPrefs` still comes from the derive, so the commission display is unchanged.

`react-hooks/src/useAccountInfo.ts`

- `isValidator` uses the same presence check, fixing the validator badge.
- `isNominator` now uses `!nominator.isNone`. It also used `isEmpty`, which always returns `true` for an `Option`, so it was always `false`.

`page-staking/src/Payouts/index.tsx` and `page-staking-legacy/src/Payouts/index.tsx`

- `hasOwnValidators` was latched in a `useState` initializer, so "Own validators" never became selectable when `ownValidators` resolved after mount. It is now derived each render.

`react-hooks/src/utils/isEmpty.ts`

- Deleted, no remaining users.

One key per request is deliberate: `rpc-core` only uses its per-key cache when a request has more than one key, and that cache is shared with the typed `staking.validators` reads. A multi-key request can return a `ValidatorPrefs`, which has no `isNone`, making a non-validator look like a validator.

### Testing

- `yarn lint` and `yarn test` pass, 3279 tests.
- Verified on Polkadot Asset Hub, `statemint` 2003002: 0% commission validators are now detected; `blocked: true` validators, nominators and accounts with no entry are unchanged.


packages/react-hooks/src/useOwnStashInfos.ts
```diff
@@ -4,22 +4,23 @@
...
 import { createNamedHook } from './createNamedHook.js';
 import { useAccounts } from './useAccounts.js';
 import { useApi } from './useApi.js';
 import { useIsMountedRef } from './useIsMountedRef.js';
 import { useOwnStashes } from './useOwnStashes.js';
 
-type ValidatorInfo = ITuple<[ValidatorPrefs, Codec]> | ValidatorPrefs;
+// The raw `staking.validators` storage value for a stash, as returned by
+// state_subscribeStorage - `isNone` when the stash has no entry, i.e. is not validating
+type ValidatorInfo = Option<StorageData>;
```
bee344:
```suggestion
type ValidatorInfo = Option<Raw>;
```

packages/page-staking-legacy/src/Payouts/index.tsx
```diff
@@ -159,8 +159,11 @@ function getOptions (blockTime: BN, eraLength: BN | undefined, historyDepth: BN
 function Payouts ({ className = '', historyDepth, isInElection, ownPools, ownValidators }: Props): React.ReactElement<Props> {
   const { t } = useTranslation();
   const { api } = useApi();
-  const [hasOwnValidators] = useState(() => ownValidators.length !== 0);
-  const [myStashesIndex, setMyStashesIndex] = useState(() => hasOwnValidators ? 0 : 1);
+  const [selectedIndex, setMyStashesIndex] = useState<number | undefined>();
```
bee344:
```suggestion
  const [selectedIndex, setSelectedIndex] = useState<number | undefined>();
```

packages/page-staking/src/Payouts/index.tsx
```diff
@@ -162,8 +162,11 @@ function getOptions (blockTime: BN, eraLength: BN | undefined, historyDepth: BN
 function Payouts ({ className = '', historyDepth, isInElection, ownPools, ownValidators }: Props): React.ReactElement<Props> {
   const { t } = useTranslation();
   const { api } = useApi();
-  const [hasOwnValidators] = useState(() => ownValidators.length !== 0);
-  const [myStashesIndex, setMyStashesIndex] = useState(() => hasOwnValidators ? 0 : 1);
+  const [selectedIndex, setMyStashesIndex] = useState<number | undefined>();
```
bee344:
```suggestion
  const [selectedIndex, setSelectedIndex] = useState<number | undefined>();
```

packages/react-hooks/src/useAccountInfo.ts
```diff
@@ -60,14 +70,15 @@ function useAccountInfoImpl (value: string | null, isContract = false): UseAccou
   useEffect((): void => {
     validator && setFlags((flags) => ({
       ...flags,
-      isValidator: !isEmpty(validator)
+      isValidator: !validator[0].isNone
```
bee344:
```suggestion
      isValidator: !validator.isStorageFallback
```

packages/react-hooks/src/useAccountInfo.ts
```diff
@@ -45,8 +45,18 @@ function useAccountInfoImpl (value: string | null, isContract = false): UseAccou
...
+    () => value && api.query.staking?.validators
+      ? api.query.staking.validators.key(value)
+      : null,
+    [api, value]
+  );
+  // `staking.validators` is a ValueQuery map, so the value of an entry cannot tell a
+  // validator apart from an account that never validated - test for the entry itself.
+  // One key per request: a request with more than one key falls back to a cache shared
+  // with the typed reads of these keys, which yields a value that has no `isNone`
+  const validator = useCall<[Option<StorageData>]>(!!validatorKey && api.rpc.state.subscribeStorage, [[validatorKey]]);
```
bee344:
```suggestion
  const nominator = useCall<Option<Nominations>>(api.query.staking?.nominators, [value]);
  const validator = useCall<ValidatorPrefs>(api.query.staking?.validators, [value]);
```
We could simplify this whole thing, since `isStorageFallback` is only set  only when the entry is not present.

bee344 (changes requested):

packages/react-hooks/src/useOwnStashInfos.ts
Imod7: Good catch! Reverted to `staking.validators.multi` with `!isStorageFallback`, so the raw read, the hand-built key and the `Option<Raw>` annotation are all gone. Verified on Polkadot Asset Hub: the flag separates a 0% validator from an account with no entry, and it still does when the staking accounts derive has already read the same keys. Addressed in [`2753334`](https://github.com/polkadot-js/apps/pull/12391/commits/2753334a9cc59df3fd3b0a316caa1679f2b84132)

packages/page-staking-legacy/src/Payouts/index.tsx
Imod7: Renamed in both Payouts pages - Commit [`54a5f86`](https://github.com/polkadot-js/apps/pull/12391/commits/54a5f86cb50b76c66d44f2151c72832025429a71).

packages/page-staking/src/Payouts/index.tsx
Imod7: Renamed in both Payouts pages - Commit [`54a5f86`](https://github.com/polkadot-js/apps/pull/12391/commits/54a5f86cb50b76c66d44f2151c72832025429a71).

packages/react-hooks/src/useAccountInfo.ts
Imod7: Back to the plain `staking.validators` read with `isValidator: !validator.isStorageFallback`, so the memoised key and the raw subscription are gone. Verified on Polkadot Asset Hub: a 0% validator gets the validator badge, an account that only nominates gets the nominator badge and not the validator one. Addressed in commit [`de1b7d3`](https://github.com/polkadot-js/apps/pull/12391/commits/de1b7d3438c46bd3fb6d7e993bc5f4989fa3b065)

packages/react-hooks/src/useAccountInfo.ts
Imod7: Agreed, and it simplifies both hooks. `useAccountInfo` is back to the plain `staking.validators` read with `!validator.isStorageFallback` ([`de1b7d3`](https://github.com/polkadot-js/apps/pull/12391/commits/de1b7d3438c46bd3fb6d7e993bc5f4989fa3b065)), `useOwnStashInfos` back to `staking.validators.multi` ([`2753334`](https://github.com/polkadot-js/apps/pull/12391/commits/2753334a9cc59df3fd3b0a316caa1679f2b84132)) - no raw reads, no hand-built keys. Verified on Polkadot Asset Hub: a 0% validator shows as validating, an account that only nominates does not.

packages/page-staking-legacy/src/Payouts/index.tsx
```diff
@@ -159,8 +159,11 @@ function getOptions (blockTime: BN, eraLength: BN | undefined, historyDepth: BN
 function Payouts ({ className = '', historyDepth, isInElection, ownPools, ownValidators }: Props): React.ReactElement<Props> {
   const { t } = useTranslation();
   const { api } = useApi();
-  const [hasOwnValidators] = useState(() => ownValidators.length !== 0);
-  const [myStashesIndex, setMyStashesIndex] = useState(() => hasOwnValidators ? 0 : 1);
+  const [selectedIndex, setSelectedIndex] = useState<number | undefined>();
+  const hasOwnValidators = ownValidators.length !== 0;
+  // `ownValidators` can resolve after mount, so the default has to follow it rather than
+  // latch in a `useState` initializer - otherwise "Own validators" never becomes selectable
+  const myStashesIndex = selectedIndex ?? (hasOwnValidators ? 0 : 1);
```
bee344:
```suggestion
  const myStashesIndex = hasOwnValidators
    ? (selectedIndex ?? 0)
    : 1;
```
If the user selects "Own Validators"  and that eventually becomes empty, `selectedIndex` stays 0 while option 0 becomes disabled, leaving the view on an empty validator set

packages/page-staking/src/Payouts/index.tsx
```diff
@@ -162,8 +162,11 @@ function getOptions (blockTime: BN, eraLength: BN | undefined, historyDepth: BN
 function Payouts ({ className = '', historyDepth, isInElection, ownPools, ownValidators }: Props): React.ReactElement<Props> {
   const { t } = useTranslation();
   const { api } = useApi();
-  const [hasOwnValidators] = useState(() => ownValidators.length !== 0);
-  const [myStashesIndex, setMyStashesIndex] = useState(() => hasOwnValidators ? 0 : 1);
+  const [selectedIndex, setSelectedIndex] = useState<number | undefined>();
+  const hasOwnValidators = ownValidators.length !== 0;
+  // `ownValidators` can resolve after mount, so the default has to follow it rather than
+  // latch in a `useState` initializer - otherwise "Own validators" never becomes selectable
+  const myStashesIndex = selectedIndex ?? (hasOwnValidators ? 0 : 1);
```
bee344: Same as above.

bee344 (changes requested):

packages/page-staking-legacy/src/Payouts/index.tsx
Imod7: Great catch! Fixed in [9228721](9228721e24aa0e235014455d9cd38e1e4e34546f)

packages/page-staking/src/Payouts/index.tsx
Imod7: Fixed in same commit as above, [9228721](9228721e24aa0e235014455d9cd38e1e4e34546f)

bee344 (approved):

